PR #5, opened 2026-04-28
Remove Roo Code Cloud and Router remnants

taltas:
## Summary
- remove remaining Roo Code Cloud and Roo Code Router extension/webview/provider surfaces
- migrate legacy `roo` provider state safely for existing users during upgrade
- clean up CLI and evals management references while leaving `apps/web-roo-code` untouched

## Validation
- repo lint and typecheck completed via pre-push hooks
- targeted `src`, `webview-ui`, `apps/cli`, and `apps/web-evals` vitest coverage was run during implementation


src/activate/handleUri.ts
```diff
@@ -30,17 +28,7 @@ export const handleUri = async (uri: vscode.Uri) => {
...
-			const organizationId = query.get("organizationId")
-			const providerModel = query.get("provider_model")
-
-			await CloudService.instance.handleAuthCallback(
-				code,
-				state,
-				organizationId === "null" ? null : organizationId,
-				providerModel,
-			)
+			// Legacy Roo Cloud auth callbacks are ignored during compatibility mode.
```
edelauna: Instead of no-opting, should we call `showInformationMessage` toast informing users until we rip out this function completely.

src/core/webview/webviewMessageHandler.ts
```diff
@@ -103,6 +104,14 @@ export const webviewMessageHandler = async (
 		return provider.getCurrentTask()?.cwd || provider.cwd
 	}
 
+	const isCloudServiceAvailable = () => CloudService.hasInstance()
+
+	const showCloudUnavailableMessage = () => {
+		vscode.window.showInformationMessage(
+			"Roo Code Cloud sign-in is currently unavailable. Configure another provider to continue.",
```
edelauna: We should use i10n strings

edelauna: I wonder if this logic should just be moved to router removal class

edelauna (commented):
Nice! Makes sense, just had a small comment regarding no-opt on the auth callback and then adding i10n strings to the `routerRemoval` class

src/core/config/routerRemoval.ts
```diff
@@ -0,0 +1,54 @@
...
+	return translated === key ? defaultValue : translated
+}
+
+export const getRouterRemovalMessage = () =>
+	getLocalizedMessage(ROUTER_REMOVAL_I18N_KEY, ROUTER_REMOVAL_DEFAULT_MESSAGE)
+
+export const getRouterUnavailableSignInMessage = () =>
+	getLocalizedMessage(ROUTER_SIGN_IN_UNAVAILABLE_I18N_KEY, ROUTER_SIGN_IN_UNAVAILABLE_DEFAULT_MESSAGE)
+
+export const ROUTER_REMOVAL_MESSAGE = getRouterRemovalMessage()
```
edelauna: Since `ROUTER_REMOVAL_MESSAGE` is just a module-level constant computed via t(...) at import time.
    * If locale can change at runtime, a function call (getRouterRemovalMessage()) everywhere is safer than a precomputed constant.

edelauna (dismissed):
Nice! Had one follow up nit, but looks good.

navedmerchant (approved):
lgtm, verified cloud is gone, and changes work in vscode extension development host 👍

PR #11, opened 2026-05-01
feat: bundle marketplace configs locally

taltas:
## Summary
- bundle marketplace MCP and mode catalog YAML files into the extension assets
- load marketplace configuration from local bundled assets instead of remote marketplace endpoints
- update marketplace packaging and tests to cover bundled asset availability and loader behavior

## Summary by CodeRabbit

* **Refactor**
  * Marketplace now loads items from bundled local assets; organization-scoped customization and MCP filtering removed.

* **Tests**
  * Added and updated tests to validate bundled marketplace assets, combined loading behavior, caching, and error handling using filesystem-backed fixtures.

* **Chores**
  * Build and ignore configs updated so bundled marketplace assets are included in distribution and test environments provide a concrete extension path.


src/services/marketplace/ConfigLoader.ts
edelauna, line 1: should this just become `ConfigLoader` ?

src/services/marketplace/RemoteConfigLoader.ts
```diff
@@ -19,24 +19,17 @@ const mcpMarketplaceResponse = z.object({
 })
 
 export class RemoteConfigLoader {
-	private apiBaseUrl: string
+	private readonly marketplacePath: string
 	private cache: Map<string, { data: MarketplaceItem[]; timestamp: number }> = new Map()
 	private cacheDuration = 5 * 60 * 1000 // 5 minutes
```
edelauna: we could probably get rid of the cache now since it'll be bundled.

src/services/marketplace/MarketplaceManager.ts
```diff
@@ -30,52 +29,17 @@ export class MarketplaceManager {
...
-					marketplaceItems = allMarketplaceItems.filter(
-						(item) => item.type !== "mcp" || !hiddenMcpIds.has(item.id),
-					)
-				}
-			}
+			const marketplaceItems = await this.configLoader.loadAllItems()
 
 			return {
-				organizationMcps,
+				organizationMcps: [],
```
edelauna, line 41: what are `organizationMcps` now? Should we just remove this?

src/assets/marketplace/mcps.yml
edelauna, line 1: worth raising as a governance topic - regarding how best to maintain these .yml files

Should we maintain an openclaw approach, or a more curated style?

edelauna (dismissed):
I think this is a good start. Had some comments regarding follow up tickets.

Tested it by adding a custom MCP and it showed up in the marketplace and installed successfully. 

Image

src/assets/marketplace/mcps.yml
taltas: I think we should keep it curated, but people can submit PRs. We could introduce security issues by letting anything just come in.

doctarock:
I addressed the pending marketplace loader/package review comments in follow-up PR #37: https://github.com/Zoo-Code-Org/Zoo-Code/pull/37\n\nThis renames RemoteConfigLoader to ConfigLoader, removes the obsolete cache path, keeps the organizationMcps response shape because the webview still consumes it, and copies marketplace YAML into dist/assets/marketplace for the dist asset test. I could not push directly to this branch with the local account, so #37 targets local-marketplace directly.

src/services/marketplace/MarketplaceManager.ts
edelauna: Filed issue: https://github.com/Zoo-Code-Org/Zoo-Code/issues/62 to follow up eventually.

edelauna (approved):
🚢

PR #13, opened 2026-05-02
feat: migrate Roo config handling to Zoo config

taltas:
## Summary
- make Zoo config files the preferred source while preserving compatibility with legacy Roo config files
- migrate handling across config, ignore, custom modes, marketplace, MCP, skills, commands, prompts, and related webview flows
- add migration-focused test coverage and include the migration PRD

## Migration behavior
- prefer `.zoo`, `.zoomodes`, and `.zooignore` when present
- continue reading legacy `.roo`, `.roomodes`, and `.rooignore` inputs for backwards compatibility
- preserve user workflows during transition by recognizing both naming schemes where required

## Safety guarantees
- Zoo-first resolution avoids ambiguous writes to legacy paths
- backwards-compatible fallbacks reduce upgrade risk for existing workspaces
- added tests cover migration paths and compatibility behavior

## Validation
- existing implementation was already validated before release workflow
- pre-push hooks completed lint and typecheck successfully during push


taltas:
Stale PR

Issue #8, opened 2026-04-30
[ENHANCEMENT] Code refactor on monolith files


doctarock:
### Problem (one or two sentences)

I understand there are historical attempts at this, I think we can handle it in small pieces, so I am submitting it as an issue I can pull against.

Task.ts (line 162) is ~4,700 lines, ClineProvider.ts (line 107) is  3,556 lines.

### Context (who is affected and when)

This is a code debt issue, I am flagging it for a course correction.

### Desired behavior (conceptual, not technical)

Make Task.ts or ClineProvider.ts thinner while preserving method names and call flow where possible.

Turn Task and ClineProvider into orchestrators. They should coordinate collaborators, not personally implement every detail.

### Constraints / preferences (optional)

We should preserve call order, async timing, event emission, persistence timing, and side effects unless we are intentionally fixing a bug.

Before extracting anything, decide who owns the state. Good extraction candidates have a clear owner, like pending edit operations, API history shaping, task history persistence, or checkpoint coordination. Bad candidates need half the class fields passed in.

Task.ts is full of order-sensitive behavior: streaming, aborts, saved assistant messages, tool result flushing, retries, delegation, and webview updates. Any extraction touching that must be conservative.

Pure helpers are safer. Stateful services need focused tests. For risky areas, we should add or extend tests before big moves.

We should avoid exposing internals just to make extraction easy. If a helper needs ten private fields, that is a sign to extract a smaller piece first.

Task and ClineProvider already know about each other. New modules need careful placement so we do not create circular dependency traps.

A file like apiConversationHistory.ts is useful because it says what it owns. Vague names like TaskUtils.ts just create a junk drawer.

Aim for cohesive modules around 50-300 lines when possible.

Most extractions should not change runtime behavior. But we need to watch allocations in hot paths, especially streaming and message processing.

### Request checklist

- [x] I've searched existing Issues and Discussions for duplicates
- [x] This describes a specific problem with clear context and impact

### Proposed approach (optional)

Our strategy is incremental extraction, not “rewrite the mountain.”

1. Pick Real Clusters
We only extract code that already behaves like a unit: state stores, message preparation, persistence helpers, webview HTML setup, task stack/delegation, telemetry, etc. 

2. Keep Public Behavior Stable
Each pass should make Task.ts or ClineProvider.ts thinner while preserving method names and call flow where possible. The monolith keeps orchestrating; extracted modules take over focused logic.

3. Verify Every Cut
After each extraction:

[...]

PR #27, opened 2026-05-06
Refactor core monolith helpers

doctarock:
### Related GitHub Issue

Refs #8

### Description

Extract two low-coupling pieces from the core monolith files while preserving the existing call flow:

- Move API conversation message shaping from `Task.ts` into `src/core/task/apiConversationHistory.ts`
- Move pending edit operation timeout/state management from `ClineProvider.ts` into `src/core/webview/PendingEditOperationStore.ts`

`Task` and `ClineProvider` keep orchestrating the same operations, but the focused logic now lives in named collaborators.

### Test Procedure

- `pnpm --dir src exec vitest run core/task/__tests__/reasoning-preservation.test.ts core/task/__tests__/grounding-sources.test.ts core/webview/__tests__/checkpointRestoreHandler.spec.ts`

### Notes

This is intentionally scoped to the two extraction candidates called out in #8.

## Summary by CodeRabbit

* **Refactor**
  * Centralized API conversation message preparation for consistent assistant/user messages.
  * Replaced in-provider pending-edit handling with a dedicated timed pending-edit store; provider updated to use the store.

* **Bug Fixes**
  * More reliable validation and rewriting of tool-result IDs and conditional conversion of tool-result content to plain text to preserve context.

* **Tests**
  * Added tests for conversation message preparation and pending-edit store behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/27?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/core/task/apiConversationHistory.ts
+++ b/src/core/task/apiConversationHistory.ts
@@ -77,6 +77,7 @@ function prepareAssistantMessage(
 		const reasoningBlock = {
 			type: "reasoning",
 			text: reasoning,
+			summary: [] as any[],
 		}
 
 		prependContentBlock(messageWithTs, reasoningBlock)
--- a/src/core/task/apiConversationHistory.ts
+++ b/src/core/task/apiConversationHistory.ts
@@ -1,0 +1,151 @@
+import { Anthropic } from "@anthropic-ai/sdk"
+
+import { type ProviderSettings, getApiProtocol, getModelId, isRetiredProvider } from "@roo-code/types"
+
+import type { ApiHandler } from "../../api"
+import { getEffectiveApiHistory } from "../condense"
+import type { ApiMessage } from "../task-persistence"
+import { validateAndFixToolResultIds } from "./validateToolResultIds"
+
+type ApiHistoryHandler = ApiHandler & {
+	getResponseId?: () => string | undefined
+	getEncryptedContent?: () => { encrypted_content: string; id?: string } | undefined
+	getThoughtSignature?: () => string | undefined
+	getSummary?: () => any[] | undefined
+	getReasoningDetails?: () => any[] | undefined
+}
+
+interface PrepareApiConversationMessageOptions {
+	message: Anthropic.MessageParam
+	reasoning?: string
+	api: ApiHandler
+	apiConfiguration: ProviderSettings
+	apiConversationHistory: ApiMessage[]
+}
+
+export function prepareApiConversationMessage({
+	message,
+	reasoning,
+	api,
+	apiConfiguration,
+	apiConversationHistory,
+}: PrepareApiConversationMessageOptions): ApiMessage {
+	if (message.role === "assistant") {
+		return prepareAssistantMessage(message, reasoning, api as ApiHistoryHandler, apiConfiguration)
+	}
+
+	return prepareUserMessage(message, apiConversationHistory)
+}
+
+function prepareAssistantMessage(
+	message: Anthropic.MessageParam,
+	reasoning: string | undefined,
+	handler: ApiHistoryHandler,
+	apiConfiguration: ProviderSettings,
+): ApiMessage {
+	const responseId = handler.getResponseId?.()
+	const reasoningData = handler.getEncryptedContent?.()
+	const thoughtSignature = handler.getThoughtSignature?.()
+	const reasoningSummary = handler.getSummary?.()
+	const reasoningDetails = handler.getReasoningDetails?.()
+
+	const modelId = getModelId(apiConfiguration)
+	const apiProvider = apiConfiguration.apiProvider
+	const apiProtocol = getApiProtocol(
+		apiProvider && !isRetiredProvider(apiProvider) ? apiProvider : undefined,
+		modelId,
+	)
+	const isAnthropicProtocol = apiProtocol === "anthropic"
+
+	const messageWithTs: any = {
+		...message,
+		...(responseId ? { id: responseId } : {}),
+		ts: Date.now(),
+	}
+
+	if (reasoningDetails) {
+		messageWithTs.reasoning_details = reasoningDetails
+	}
+
+	if (isAnthropicProtocol && reasoning && thoughtSignature && !reasoningDetails) {
+		const thinkingBlock = {
+			type: "thinking",
+			thinking: reasoning,
+			signature: thoughtSignature,
+		}
+
+		prependContentBlock(messageWithTs, thinkingBlock)
+	} else if (reasoning && !reasoningDetails) {
+		const reasoningBlock = {
+			type: "reasoning",
+			text: reasoning,
+			summary: reasoningSummary ?? ([] as any[]),
+		}
+
+		prependContentBlock(messageWithTs, reasoningBlock)
+	} else if (reasoningData?.encrypted_content) {
+		const reasoningBlock = {
+			type: "reasoning",
+			summary: [] as any[],
+			encrypted_content: reasoningData.encrypted_content,
+			...(reasoningData.id ? { id: reasoningData.id } : {}),
+		}
+
+		prependContentBlock(messageWithTs, reasoningBlock)
+	}
+
+	if (thoughtSignature && !isAnthropicProtocol) {
+		const thoughtSignatureBlock = {
+			type: "thoughtSignature",
+			thoughtSignature,
+		}
+
+		appendContentBlock(messageWithTs, thoughtSignatureBlock)
+	}
+
+	return messageWithTs
+}
+
+function prepareUserMessage(message: Anthropic.MessageParam, apiConversationHistory: ApiMessage[]): ApiMessage {
+	const effectiveHistoryForValidation = getEffectiveApiHistory(apiConversationHistory)
+	const lastEffective = effectiveHistoryForValidation[effectiveHistoryForValidation.length - 1]
+	const historyForValidation = lastEffective?.role === "assistant" ? effectiveHistoryForValidation : []
+
+	let messageToAdd = message
+	if (lastEffective?.role !== "assistant" && Array.isArray(message.content)) {
+		messageToAdd = {
+			...message,
+			content: message.content.map((block) =>
+				block.type === "tool_result"
+					? {
+							type: "text" as const,
+							text: `Tool result:\n${typeof block.content === "string" ? block.content : JSON.stringify(block.content)}`,
+						}
+					: block,
+			),
+		}
+	}
+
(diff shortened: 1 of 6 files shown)


src/core/webview/PendingEditOperationStore.ts
```diff
@@ -0,0 +1,62 @@
+export interface PendingEditOperation {
+	messageTs: number
+	editedContent: string
+	images?: string[]
+	messageIndex: number
+	apiConversationHistoryIndex: number
+	timeoutId: NodeJS.Timeout
```
edelauna, line 7: Since `get()` is the only public read path and callers (e.g. `ClineProvider.createTaskWithHistoryItem`) only consume the data fields, consider not exposing `timeoutId` and `createdAt` to consumers. One option:

```ts
export type PendingEditOperationView = Omit<PendingEditOperation, "timeoutId" | "createdAt">

get(operationId: string): PendingEditOperationView | undefined { ... }
```

Keeps the timeout handle as a true implementation detail of the store.

src/core/webview/PendingEditOperationStore.ts
```diff
@@ -0,0 +1,62 @@
...
+		private readonly log: (message: string) => void,
+	) {}
+
+	set(operationId: string, editData: PendingEditOperationInput): void {
+		this.clear(operationId)
+
+		const timeoutId = setTimeout(() => {
+			this.clear(operationId)
+			this.log(`[setPendingEditOperation] Automatically cleared stale pending operation: ${operationId}`)
+		}, this.timeoutMs)
```
edelauna, line 28: The bracket tags here (and at lines 35, 50, 60) — `[setPendingEditOperation]`, `[clearPendingEditOperation]`, `[clearAllPendingEditOperations]` — name the old `ClineProvider` wrapper methods that no longer exist. Now that the logs are emitted from the store, worth renaming to match (e.g. `[PendingEditOperationStore.set]`) so they actually point at the call site.

src/core/task/apiConversationHistory.ts
```diff
@@ -0,0 +1,151 @@
...
+import type { ApiHandler } from "../../api"
+import { getEffectiveApiHistory } from "../condense"
+import type { ApiMessage } from "../task-persistence"
+import { validateAndFixToolResultIds } from "./validateToolResultIds"
+
+type ApiHistoryHandler = ApiHandler & {
+	getResponseId?: () => string | undefined
+	getEncryptedContent?: () => { encrypted_content: string; id?: string } | undefined
+	getThoughtSignature?: () => string | undefined
+	getSummary?: () => any[] | undefined
```
edelauna: `getSummary` isn't implemented by any provider in `src/api/providers/`, so `reasoningSummary` is always `undefined` and the `summary: reasoningSummary ?? []` on line 82 always resolves to `[]`. Pre-existing dead branch from the original `Task.ts`, but extracting it is a natural moment to drop both `getSummary` from the duck-type and the `?? []` fallback.

src/core/webview/PendingEditOperationStore.ts
```diff
@@ -0,0 +1,62 @@
...
+	editedContent: string
+	images?: string[]
+	messageIndex: number
+	apiConversationHistoryIndex: number
+	timeoutId: NodeJS.Timeout
+	createdAt: number
+}
+
+export type PendingEditOperationInput = Omit<PendingEditOperation, "timeoutId" | "createdAt">
+
```
edelauna, line 13: Now that this is a standalone unit, a `PendingEditOperationStore.spec.ts` would be cheap and worth it — particularly to cover the `setTimeout` auto-clear path, which isn't exercised by any existing integration test. Suggested cases:
- `set` clears a prior op with the same id
- `clear` returns `false` on miss / `true` on hit
- auto-clear fires after `timeoutMs` and logs
- `clearAll` clears timers without leaks (use fake timers)

src/core/task/apiConversationHistory.ts
```diff
@@ -0,0 +1,151 @@
...
+
+interface PrepareApiConversationMessageOptions {
+	message: Anthropic.MessageParam
+	reasoning?: string
+	api: ApiHandler
+	apiConfiguration: ProviderSettings
+	apiConversationHistory: ApiMessage[]
+}
+
+export function prepareApiConversationMessage({
```
edelauna, line 25: Same suggestion for an `apiConversationHistory.spec.ts` — `prepareApiConversationMessage` now has a clean signature and is straightforward to unit test. The current coverage via `reasoning-preservation.test.ts` goes through `Task`, which makes the protocol-branching logic (Anthropic vs. non-Anthropic, encrypted-content path, thoughtSignature append) hard to exercise in isolation.

edelauna (commented):
Nice - thanks for taking this on - couple comments based on some opportunities we could take with this refactor.

src/core/webview/PendingEditOperationStore.ts
doctarock: Added in the following PR

src/core/webview/PendingEditOperationStore.ts
doctarock: Added in the following PR [9da86b1](https://github.com/Zoo-Code-Org/Zoo-Code/pull/27/commits/9da86b1b4281341b268aacb50cbf2d63705578f3)

src/core/task/apiConversationHistory.ts
doctarock: Added in the following PR [9da86b1](https://github.com/Zoo-Code-Org/Zoo-Code/pull/27/commits/9da86b1b4281341b268aacb50cbf2d63705578f3)

src/core/webview/PendingEditOperationStore.ts
doctarock: Added in the following PR [9da86b1](https://github.com/Zoo-Code-Org/Zoo-Code/pull/27/commits/9da86b1b4281341b268aacb50cbf2d63705578f3)

src/core/task/apiConversationHistory.ts
doctarock: Added in the following PR [9da86b1](https://github.com/Zoo-Code-Org/Zoo-Code/pull/27/commits/9da86b1b4281341b268aacb50cbf2d63705578f3)

src/core/webview/PendingEditOperationStore.ts
```diff
@@ -0,0 +1,74 @@
...
+		const timeoutId = setTimeout(() => {
+			this.clear(operationId)
+			this.log(`[PendingEditOperationStore.set] Automatically cleared stale pending operation: ${operationId}`)
+		}, this.timeoutMs)
+
+		this.operations.set(operationId, {
+			...editData,
+			timeoutId,
+			createdAt: Date.now(),
```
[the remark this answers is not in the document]
doctarock: Pushed commit 9a824017e to doctarock/Zoo-Code:issue-8-monolith-refactor.

What changed:

Clones images on set().
Clones images on get().
Added a regression test covering mutation of both the input array and returned array.

src/core/task/apiConversationHistory.ts
```diff
@@ -0,0 +1,148 @@
...
+			thinking: reasoning,
+			signature: thoughtSignature,
+		}
+
+		prependContentBlock(messageWithTs, thinkingBlock)
+	} else if (reasoning && !reasoningDetails) {
+		const reasoningBlock = {
+			type: "reasoning",
+			text: reasoning,
+		}
```
edelauna, line 83: Subtle behavioral change worth calling out: the original `Task.ts` stored `summary: reasoningSummary ?? []` on plain-text reasoning blocks (always `[]` since `getSummary` was never implemented). The refactored block here omits `summary` entirely.

For models with `preserveReasoning: true` (deepseek-reasoner, ZAI, Fireworks), these blocks aren't dropped by `buildCleanConversationHistory` — they're round-tripped back to the API at `Task.ts:4412-4436`. So the wire payload for those providers changes from `{ type: "reasoning", text, summary: [] }` to `{ type: "reasoning", text }`.

Almost certainly safe (the field was always empty and these APIs don't appear to require it), but a refactor PR should either preserve the wire shape exactly or note the intentional change in the description. Could you either:
1. Add `summary: []` back to the plain-text reasoning block to keep this strictly behavior-preserving, or
2. Update the PR description noting that `summary: []` is being dropped from plain-text reasoning blocks for these three providers (and ideally a one-liner on why it's safe)?

doctarock:
Addressed the reasoning payload-shape comment in `3ecf73d02` by choosing the behavior-preserving path: plain-text reasoning blocks now include `summary: []` again, matching the old `Task.ts` wire shape.

Validation: `corepack pnpm --dir src exec vitest run core/task/__tests__/apiConversationHistory.spec.ts` passed.

edelauna (approved):
🚀 this is a good start, thanks for taking this on - ci should pass once you merge in the updates

edelauna:
@doctarock if you can get this merged - then I dont have to manually approve workflow runs for your PRs since you'll no longer be a first time contributor.

edelauna (changes requested):
Merge https://github.com/Zoo-Code-Org/Zoo-Code/pull/68 into this one.

src/core/task/apiConversationHistory.ts
```diff
@@ -0,0 +1,149 @@
...
+import type { ApiHandler } from "../../api"
+import { getEffectiveApiHistory } from "../condense"
+import type { ApiMessage } from "../task-persistence"
+import { validateAndFixToolResultIds } from "./validateToolResultIds"
+
+type ApiHistoryHandler = ApiHandler & {
+	getResponseId?: () => string | undefined
+	getEncryptedContent?: () => { encrypted_content: string; id?: string } | undefined
+	getThoughtSignature?: () => string | undefined
+	getReasoningDetails?: () => any[] | undefined
```
edelauna, line 14: The original `Task.ts` code called `handler.getSummary?.()` and fell back to `[]` — was that method intentionally dropped here because no handler ever returns a non-empty summary? If so, might be worth a brief comment so a future reader knows it was deliberate rather than an accidental omission.

edelauna (commented):
Thanks for adding the commit - on re-review had one followup comment.

src/core/task/apiConversationHistory.ts
doctarock: getSummary was intentionally omitted because no provider implements it and the preserved wire shape is the explicit summary: [] fallback, it does deserve a comment though

edelauna (approved):
Thanks for doing this, and also documenting it out via issues.

PR #32, opened 2026-05-07
feat: added log requests to backend for observability

JamesRobert20:
### Summary

This PR adds Zoo Code observability auth + telemetry and folds in the follow-up fixes from review.

- verify persisted Zoo Code tokens on startup before reporting the user as connected, and clear stale cached profile state when verification fails
- keep telemetry truly fire-and-forget with a timeout boundary
- hydrate Zoo Code auth/profile state through one async webview path so the badge has consistent user data
- add focused regression coverage for `/auth-callback`, startup auth validation, revoke/clear flows, and telemetry behavior
- clean up the auth badge toward Tailwind / VS Code token classes

### Validation

- `pnpm --dir src exec vitest run activate/__tests__/handleUri.spec.ts services/__tests__/zoo-code-auth.test.ts services/__tests__/zoo-telemetry.test.ts`
- `pnpm --dir webview-ui check-types`

### Notes

- One reviewer report about avatar/telemetry behavior is still open for manual verification in a real VS Code + zoocode.dev flow. The extension-side fixes are on this branch, but that end-to-end path is not runnable from the sandbox browser surface.
- No browser proof is attached for the same reason: the exposed sandbox browser target is `apps/web-roo-code`, while this change lives in the VS Code extension/webview auth flow.


JamesRobert20:
@coderabbitai review

JamesRobert20:
@coderabbitai review

doctarock:
I addressed the pending review comments in follow-up PR #34: https://github.com/Zoo-Code-Org/Zoo-Code/pull/34\n\nThis covers the callback URI, token verification before persistence, fire-and-forget telemetry, resolved task mode, and async auth state loading. I could not push directly to this branch with the local account, so #34 targets eat/observability directly.

src/services/zoo-code-auth.ts
```diff
@@ -0,0 +1,283 @@
...
+	// Check subscription status after successful auth
+	await checkSubscriptionStatus().catch(() => {})
+
+	vscode.window.showInformationMessage(
+		"Zoo Code: Successfully connected! You can now use Zoo Code as your AI provider.",
+	)
+	return true
+}
+
+export async function verifyZooCodeToken(): Promise<boolean> {
```
edelauna: same thing with this function, I'm not seeing any callers?

src/services/zoo-code-auth.ts
```diff
@@ -0,0 +1,283 @@
...
+	_lastSubscriptionCheck = 0
+	await clearZooCodeUserInfo()
+}
+
+export function getZooCodeBaseUrl(): string {
+	const config = vscode.workspace.getConfiguration("zoo-code")
+	return config.get<string>("baseUrl") || process.env.ZOO_CODE_BASE_URL || "https://www.zoocode.dev"
+}
+
+export async function startZooCodeAuth(): Promise<void> {
```
edelauna: what is calling this function? I'm not seeing any references anywhere.

webview-ui/src/oauth/urls.ts
```diff
@@ -11,3 +11,14 @@ export function getOpenRouterAuthUrl(uriScheme?: string) {
 export function getRequestyAuthUrl(uriScheme?: string) {
 	return `https://app.requesty.ai/oauth/authorize?callback_url=${getCallbackUrl("requesty", uriScheme)}`
 }
+
+const ZOO_CODE_DEFAULT_BASE_URL = "https://www.zoocode.dev"
+
+export function getZooCodeAuthUrl(uriScheme?: string, baseUrl?: string) {
+	const resolvedBaseUrl = baseUrl || ZOO_CODE_DEFAULT_BASE_URL
+	const callbackUri = getCallbackUrl("auth-callback", uriScheme)
+	const deviceName = encodeURIComponent("VS Code")
```
edelauna: how come both are `VS Code`?

edelauna (commented):
We should probably also update `PRIVACY.md` as part of this PR.

The current policy explicitly states:

    Telemetry is powered by PostHog and includes your VS Code machine ID, feature usage patterns, and exception reports. This telemetry does not collect personally identifiable information…

This PR introduces a separate, non-anonymous pipeline to zoocode.dev that sends task IDs, provider, model, token counts, and cost — linked to an authenticated user account whose email is stored in SecretStorage. That contradicts the policy as written.

Would suggest updating the `PRIVACY.md` file.

webview-ui/src/components/chat/ZooCodeAuthBadge.tsx
```diff
@@ -0,0 +1,214 @@
...
+					fontWeight: 600,
+					background: zooCodeIsAuthenticated ? (showImage ? "transparent" : avatarColor) : undefined,
+					color:
+						zooCodeIsAuthenticated && !showImage ? "var(--vscode-button-foreground, #ffffff)" : undefined,
+				}}
+				title={zooCodeIsAuthenticated ? `Zoo Code: ${zooCodeUserEmail || "Connected"}` : "Sign in to Zoo Code"}>
+				{zooCodeIsAuthenticated ? (
+					showImage ? (
+						<img
+							src={zooCodeUserImage}
```
taltas, line 99: I tried this out on my local, my image did not show when I signed up with github. None of the telemetry is showing up either, not sure what's going on there with the telemetry.

Image

One other workflow issue I had when signing up with github for the first time, the server never sent me back to vscode after I **signed up**. To reporuce

1. Clicked on 'sign in' -> send me to server
2. Click on sign in with github, asks for permission, authorized
3. Land back on server's app dashboard.
4. Zoo code doesn't see me as loged in.

I then click on sign in via zoo code again, it sends me to the server sign in page, I sign in with github, THEN the browser asks for permission to go to VScode etc.

I wouldn't be surprised if this is a general problem with sign up and not just github

src/activate/handleUri.ts
```diff
@@ -32,6 +31,33 @@ export const handleUri = async (uri: vscode.Uri) => {
 			vscode.window.showInformationMessage(getRouterUnavailableSignInMessage())
 			break
 		}
+		case "/auth-callback": {
```
taltas, line 34: We need to add a test for this end point in [src/activate/tests/handleUri.spec.ts](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/activate/__tests__/handleUri.spec.ts)

src/services/zoo-telemetry.ts
```diff
@@ -0,0 +1,62 @@
...
+	cacheWriteTokens?: number
+	totalCost?: number
+}
+
+/**
+ * Send LLM telemetry to the Zoo Code observability backend.
+ * This is a fire-and-forget operation that silently fails on error.
+ * Only sends telemetry for authenticated users with active subscriptions.
+ */
+export async function sendLlmTelemetry(payload: LlmTelemetryPayload): Promise<void> {
```
taltas, line 27: We need tests for this function as well.

src/services/__tests__/zoo-code-auth.test.ts
```diff
@@ -0,0 +1,234 @@
...
+			expect(baseUrl).toBe("https://staging.zoocode.dev")
+
+			// Restore
+			if (originalEnv) {
+				process.env.ZOO_CODE_BASE_URL = originalEnv
+			} else {
+				delete process.env.ZOO_CODE_BASE_URL
+			}
+		})
+	})
```
taltas, line 303: Further to my original comment, we need way more tests on the auth, at a minimum we need to cover
- new callback verifier in handleAuthCallback
- the invalidation path in verifyZooCodeToken
- the revoke-and-clear flow in disconnectZooCode

src/services/zoo-code-auth.ts
```diff
@@ -0,0 +1,290 @@
...
+
+	// Pre-load the token and user info into memory on init so ZooCodeHandler can access them synchronously
+	_cachedToken = await secretStorage.get(ZOO_CODE_TOKEN_KEY)
+	_cachedUserName = await secretStorage.get(ZOO_CODE_USER_NAME_KEY)
+	_cachedUserEmail = await secretStorage.get(ZOO_CODE_USER_EMAIL_KEY)
+	_cachedUserImage = await secretStorage.get(ZOO_CODE_USER_IMAGE_KEY)
+
+	// Check subscription status on init if authenticated
+	if (_cachedToken) {
+		checkSubscriptionStatus().catch(() => {})
```
taltas: I think roomote is suggesting to put the `verifyZooCodeToken` into here. Basically you want to check if the token is valid, if it's not bust the stored user info, otherwise do the subscription check. Something like this

```
// src/services/zoo-code-auth.ts
export async function initZooCodeAuth(context: vscode.ExtensionContext): Promise<void> {
	// ...existing secret preload...

	if (_cachedToken) {
		const valid = await verifyZooCodeToken().catch(() => false)
		if (!valid) {
			await clearZooCodeUserInfo()
		} else {
			void checkSubscriptionStatus().catch(() => {})
		}
	}
}
```

src/services/zoo-code-auth.ts
```diff
@@ -0,0 +1,290 @@
...
+export async function clearZooCodeToken(): Promise<void> {
+	if (!secretStorage) return
+	await secretStorage.delete(ZOO_CODE_TOKEN_KEY)
+	_cachedToken = undefined
+	_cachedSubscriptionStatus = "unknown"
+	_lastSubscriptionCheck = 0
+}
+
+export function getZooCodeBaseUrl(): string {
+	const config = vscode.workspace.getConfiguration("zoo-code")
```
taltas: Nit: We should use the package name here like elsewhere in the app
`const config = vscode.workspace.getConfiguration(Package.name)`

webview-ui/src/components/chat/ZooCodeAuthBadge.tsx
```diff
@@ -0,0 +1,214 @@
...
+					"flex items-center justify-center",
+					"w-5 h-5 rounded-full",
+					"cursor-pointer p-0",
+					"transition-all duration-150",
+					"focus:outline-none focus-visible:ring-1 focus-visible:ring-vscode-focusBorder",
+					"overflow-hidden",
+					!zooCodeIsAuthenticated &&
+						"bg-transparent text-vscode-descriptionForeground border border-vscode-descriptionForeground border-opacity-50 hover:border-opacity-100",
+				)}
+				style={{
```
taltas: We should be using tailwind here as well, or vscode css classes that are built in,  (which is partially there), this comment is for the entire view file.

taltas (changes requested):
A good first pass, there needs to be more thorough testing especially since this is what a lot of user's will first interact with, it needs to relatively bullet proof and well QA-ed. I like the updates to the privacy doc too.

webview-ui/src/components/chat/ZooCodeAuthBadge.tsx
doctarock: I will get on to this my next sprint, cheers for the review

src/services/zoo-code-auth.ts
```diff
@@ -0,0 +1,300 @@
...
+}
+
+export function getZooCodeBaseUrl(): string {
+	const config = vscode.workspace.getConfiguration(Package.name)
+	return config.get<string>("baseUrl") || process.env.ZOO_CODE_BASE_URL || "https://www.zoocode.dev"
+}
+
+export async function handleAuthCallback(token: string): Promise<boolean> {
+	if (!token || !token.startsWith("zoo_ext_")) {
+		vscode.window.showErrorMessage("Zoo Code: Invalid authentication token received.")
```
taltas: This string needs to be in the locale under all the different languages, there are bunch more in this file that require the same treatment, there should be a skill somewhere in the repo that does the for you.

src/services/zoo-code-auth.ts
```diff
@@ -0,0 +1,300 @@
...
+
+	// Pre-load the token and user info into memory on init so ZooCodeHandler can access them synchronously
+	_cachedToken = await secretStorage.get(ZOO_CODE_TOKEN_KEY)
+	_cachedUserName = await secretStorage.get(ZOO_CODE_USER_NAME_KEY)
+	_cachedUserEmail = await secretStorage.get(ZOO_CODE_USER_EMAIL_KEY)
+	_cachedUserImage = await secretStorage.get(ZOO_CODE_USER_IMAGE_KEY)
+
+	// Validate persisted auth state on init before reporting the user as connected.
+	if (_cachedToken) {
+		const isValid = await verifyZooCodeToken().catch(() => false)
```
taltas: if there is a network issue (i.e. wifi goes down), we immediately clear the token below, maybe we should be a bit more defensive here. We can maybe distinguish “token is invalid” from “verification could not be completed”. Only clear secrets on a confirmed invalid/unauthorized response; keep the cached session and mark auth/subscription state as unknown when the backend is temporarily unreachable.

src/services/zoo-code-auth.ts
```diff
@@ -0,0 +1,300 @@
...
+	if (!secretStorage) return
+	await secretStorage.delete(ZOO_CODE_TOKEN_KEY)
+	_cachedToken = undefined
+	_cachedSubscriptionStatus = "unknown"
+	_lastSubscriptionCheck = 0
+}
+
+export function getZooCodeBaseUrl(): string {
+	const config = vscode.workspace.getConfiguration(Package.name)
+	return config.get<string>("baseUrl") || process.env.ZOO_CODE_BASE_URL || "https://www.zoocode.dev"
```
taltas: `config.get<string>("baseUrl") ` I don't think baseUrl is actually defined, I think we can just get rid of this actually.

taltas (changes requested):
A few small things to change

src/services/zoo-code-auth.ts
JamesRobert20: Roomote suggested the same thing ironically

src/services/zoo-code-auth.ts
JamesRobert20: I agree

taltas (approved):
Nice work, looks good.

PR #42, opened 2026-05-09
fix: Update GPT 5.5 context window to correct value for codex provider

### LINKED ISSUE #41: [BUG] GPT-5.5 Codex uses incorrect context window

`gpt-5.5` in the Codex provider is configured with a `1050000` token context window, which GPT-5.5 does not support in Codex. This can cause Zoo to overestimate the usable context window for Codex requests.

This affects users selecting `gpt-5.5` through the OpenAI Codex provider. It matters when longer tasks approach the provider's real context limit, since Zoo's context management uses the configured `contextWindow` value to decide when to condense or truncate conversation history.

### Reproduction steps

1. Use GPT 5.5, exhaust the 400k context window.
2. API errors are returned due to exceeding the model's context window instead of context being compacted.

### Expected result

`gpt-5.5` in the Codex provider should use the documented Codex context window of `400000` tokens.

### Actual result

`gpt-5.5` is configured with `contextWindow: 1_050_000`, which matches a larger API-style limit and not the Codex provider limit.

_No response_

### App Version

zoo-code @ e58fd9b

### API Provider (optional)

None

### Model Used (optional)

_No response_

### Roo Code Task Links (optional)

_No response_

```shell

```

jaszhix:
This fixes the `gpt-5.5` context window in the Codex provider by changing it to `400000`, matching the documented Codex limit.

The API model has different limits, but this entry is for the ChatGPT/Codex provider path.

Closes #41

## Summary by CodeRabbit

* **Configuration Updates**
  * Updated the context window size for the `gpt-5.5` model from 1,050,000 to 400,000 tokens.


packages/types/src/providers/openai-codex.ts
```diff
@@ -174,7 +174,7 @@ export const openAiCodexModels = {
 	},
 	"gpt-5.5": {
 		maxTokens: 128000,
-		contextWindow: 1_050_000,
+		contextWindow: 400000,
```
taltas, line 177: You're right, [weird that they only document this in the announcement](https://openai.com/index/introducing-gpt-5-5/) as far as I could find, thanks for the fix!

taltas (approved):

PR #75, opened 2026-05-12
feat(modes): per-mode MCP server restrictions (allowlist)

simurg79:
## Summary

Adds an optional `allowedMcpServers: string[]` to `ModeConfig`. When defined, only the listed MCP servers' tool schemas are injected into the system prompt and exposed as native tools for that mode. This prevents context bloat in specialized modes that only need a subset of configured MCP servers.

Ports the fix from upstream **RooCodeInc/Roo-Code#12004** (Per-Mode MCP Server Restrictions to Prevent Context Bloat), based on the implementation in **simurg79/Roo-Code#1**.

> Upstream issue: https://github.com/RooCodeInc/Roo-Code/issues/12004
> Upstream PR: https://github.com/simurg79/Roo-Code/pull/1

## Changes

**Backend**
- `packages/types/src/mode.ts` — add `allowedMcpServers` to `modeConfigSchema`.
- `schemas/roomodes.json` — declare `allowedMcpServers` array.
- `src/core/prompts/system.ts` — filter `mcpHub.getServers()` by the mode's allowlist before computing `hasMcpServers`.
- `src/core/prompts/tools/native-tools/mcp_server.ts` — `getMcpServerTools(mcpHub, allowedServers?)` filters servers before generating tool defs.
- `src/core/task/build-tools.ts` — resolves the active `ModeConfig` and forwards `allowedMcpServers` to `getMcpServerTools`.

**Webview UI**
- `webview-ui/src/components/modes/McpServerRestriction.tsx` *(new)* — `React.memo` editor following the AGENTS.md `cachedState` pattern with a 150ms debounce, mode-switch reseed, and external-edit reconciliation.
- `webview-ui/src/components/modes/ModesView.tsx` — integrate editor into both Edit and Create flows; ref-guarded mode-sync effect to avoid flicker.
- `webview-ui/src/__mocks__/@vscode/webview-ui-toolkit/react.tsx` *(new)* — minimal React stub to avoid FAST Foundation dual-React-instance issues under JSDOM.
- `webview-ui/vitest.config.ts` — register `@vitejs/plugin-react`; alias toolkit to the mock.

## Tests

| Suite | Result |
| --- | --- |
| `packages/types` — `mode-allowedMcpServers.spec.ts` | ✅ 5/5 |
| `src` — `system-prompt.spec.ts` | ✅ 13/13 |
| `src` — `mcp_server.spec.ts` | ✅ 14/14 (incl. 4 new allowlist cases) |
| `webview-ui` — `ModesView.spec.tsx` | ✅ 10/10 |
| `webview-ui` — `McpServerRestriction.spec.tsx` | ⚠️ 9/10 — see note |

### Known test note

`McpServerRestriction.spec.tsx > "Test 3: profiler — 1 commit per click, 0 commits on equivalent mcpServers heartbeat"` is over-strict: `React.memo` correctly bails out on the component, but the surrounding `<Profiler>` still commits when its parent re-renders, so `onRender` fires once. Component behavior is verified by Tests 1 & 2 (optimistic toggle and DOM-node identity preservation across `mcpServers` heartbeats). Test kept verbatim from upstream PR; happy to relax to `toBeLessThanOrEqual(1)` if reviewers prefer.

## Deviations from upstream PR

Per maintainer direction, the following upstream files were intentionally **not** ported as out-of-scope: `.roomodes` (orchestrator/docs-extractor mode shuffling), `gh_modesview.diff` (stray diff), and `docs/design/per-mode-mcp-settings.md` (design doc).

The `VSCodeTextArea` mock was extended beyond the upstream stub to bridge the `CustomEvent("change", { detail: { target: { value }}})` pattern that pre-existing `ModesView.spec.tsx` tests rely on.

## Checklist

- [x] Tests added/updated for new behavior
- [x] All targeted vitest suites green (one known over-strict assertion documented above)
- [x] No lint rules disabled
- [x] AGENTS.md `cachedState` pattern followed for the new settings input

## Summary by CodeRabbit

* **New Features**
  * Added MCP server restrictions to custom modes, allowing fine-grained control over which MCP servers are available for each configuration.
  * New UI controls in mode settings for managing MCP server restrictions.

* **Tests**
  * Comprehensive test coverage for MCP server filtering, mode-based restrictions, and system prompt generation functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/75)

### DIFF
--- a/packages/types/src/mode.ts
+++ b/packages/types/src/mode.ts
@@ -102,6 +102,7 @@ export const modeConfigSchema = z.object({
 	customInstructions: z.string().optional(),
 	groups: groupEntryArraySchema,
 	source: z.enum(["global", "project"]).optional(),
+	allowedMcpServers: z.array(z.string()).optional(),
 })
 
 export type ModeConfig = z.infer<typeof modeConfigSchema>
--- a/src/core/prompts/tools/native-tools/mcp_server.ts
+++ b/src/core/prompts/tools/native-tools/mcp_server.ts
@@ -11,12 +11,18 @@ import { normalizeToolSchema, type JsonSchema } from "../../../../utils/json-sch
  * @param mcpHub The McpHub instance containing connected servers.
  * @returns An array of OpenAI.Chat.ChatCompletionTool definitions.
  */
-export function getMcpServerTools(mcpHub?: McpHub): OpenAI.Chat.ChatCompletionTool[] {
+export function getMcpServerTools(mcpHub?: McpHub, allowedServers?: string[]): OpenAI.Chat.ChatCompletionTool[] {
 	if (!mcpHub) {
 		return []
 	}
 
-	const servers = mcpHub.getServers()
+	let servers = mcpHub.getServers()
+
+	// Filter servers by allowlist if provided
+	if (allowedServers) {
+		const allowSet = new Set(allowedServers)
+		servers = servers.filter((s) => allowSet.has(s.name))
+	}
 	const tools: OpenAI.Chat.ChatCompletionTool[] = []
 	// Track seen tool names to prevent duplicates (e.g., when same server exists in both global and project configs)
 	const seenToolNames = new Set<string>()
--- a/src/core/task/build-tools.ts
+++ b/src/core/task/build-tools.ts
@@ -7,6 +7,7 @@ import { customToolRegistry, formatNative } from "@roo-code/core"
 
 import type { ClineProvider } from "../webview/ClineProvider"
 import { getRooDirectoriesForCwd } from "../../services/roo-config/index.js"
+import { getModeBySlug, defaultModeSlug } from "../../shared/modes"
 
 import { getNativeTools, getMcpServerTools } from "../prompts/tools/native-tools"
 import {
@@ -124,8 +125,12 @@ export async function buildNativeToolsArrayWithRestrictions(options: BuildToolsO
 		mcpHub,
 	)
 
+	// Resolve mode config to get allowedMcpServers for MCP server filtering.
+	const modeConfig = getModeBySlug(mode ?? defaultModeSlug, customModes)
+	const allowedMcpServers = modeConfig?.allowedMcpServers
+
 	// Filter MCP tools based on mode restrictions.
-	const mcpTools = getMcpServerTools(mcpHub)
+	const mcpTools = getMcpServerTools(mcpHub, allowedMcpServers)
 	const filteredMcpTools = filterMcpToolsForMode(mcpTools, mode, customModes, experiments)
 
 	// Add custom tools if they are available and the experiment is enabled.
--- a/webview-ui/src/components/modes/McpServerRestriction.tsx
+++ b/webview-ui/src/components/modes/McpServerRestriction.tsx
@@ -1,0 +1,175 @@
+import React, { useState, useEffect, useRef, useCallback } from "react"
+import { VSCodeCheckbox } from "@vscode/webview-ui-toolkit/react"
+import type { ModeConfig, McpServer } from "@roo-code/types"
+
+export interface McpServerRestrictionProps {
+	customMode: ModeConfig
+	mcpServers: McpServer[]
+	onCommit: (slug: string, updates: ModeConfig) => void
+}
+
+/**
+ * Returns true when both inputs are undefined OR both are arrays containing
+ * the same set of strings (order-insensitive). This is the equality predicate
+ * used to decide whether the local cached state and the host-side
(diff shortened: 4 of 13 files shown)


edelauna (commented):
_(review withdrawn — resubmitting after further review)_

packages/types/src/mode.ts
```diff
@@ -102,6 +102,7 @@ export const modeConfigSchema = z.object({
 	customInstructions: z.string().optional(),
 	groups: groupEntryArraySchema,
 	source: z.enum(["global", "project"]).optional(),
+	allowedMcpServers: z.array(z.string()).optional(),
```
edelauna, line 105: would it be a big lift to also add `blockedMcpServers: z.array(z.string()).optional()`? In case it would be easier to manage a deny list?

src/core/prompts/tools/native-tools/mcp_server.ts
```diff
@@ -11,12 +11,18 @@ import { normalizeToolSchema, type JsonSchema } from "../../../../utils/json-sch
...
 	}
 
-	const servers = mcpHub.getServers()
+	let servers = mcpHub.getServers()
+
+	// Filter servers by allowlist if provided
+	if (allowedServers) {
+		const allowSet = new Set(allowedServers)
+		servers = servers.filter((s) => allowSet.has(s.name))
+	}
```
edelauna, line 25: Could an `else if (blockedServers)` branch go here to support the blocklist side from the linked issue?

```suggestion
	} else if (blockedServers) {
		const blockSet = new Set(blockedServers)
		servers = servers.filter((s) => !blockSet.has(s.name))
	}
```

src/core/task/build-tools.ts
```diff
@@ -124,8 +125,12 @@ export async function buildNativeToolsArrayWithRestrictions(options: BuildToolsO
 		mcpHub,
 	)
 
+	// Resolve mode config to get allowedMcpServers for MCP server filtering.
+	const modeConfig = getModeBySlug(mode ?? defaultModeSlug, customModes)
+	const allowedMcpServers = modeConfig?.allowedMcpServers
```
edelauna, line 130: When `allowedMcpServers` excludes a server that still has resources, this path still passes the full `mcpHub` into `filterNativeToolsForMode`. That helper only checks whether any server in the hub has resources before keeping `access_mcp_resource`, and the tool accepts an arbitrary `server_name`, so a restricted mode can still read resources from disallowed servers. The resource availability check needs to use the same allowlisted view, or the tool handler needs to enforce the mode allowlist.

webview-ui/src/components/modes/McpServerRestriction.tsx
```diff
@@ -0,0 +1,175 @@
...
+			isInitialMountRef.current = false
+			return
+		}
+		if (arraysEqualOrBothUndefined(cachedAllowedMcpServers, customMode.allowedMcpServers)) {
+			return
+		}
+		const handle = setTimeout(() => {
+			lastFlushedRef.current = cachedAllowedMcpServers
+			onCommit(customMode.slug, {
+				...customMode,
```
edelauna, line 102: This debounced commit captures the whole `customMode` object from the render that scheduled the timeout. If the user changes another field on the same mode within the 150 ms window, this flush sends the stale snapshot back with only `allowedMcpServers` changed and can overwrite the newer field value. Keeping the latest `customMode` in a ref, or committing only the `allowedMcpServers` field, would avoid clobbering concurrent edits.

edelauna (commented):
Thank you so much for opening this! I had a couple early comments, and was wondering if we could also apply the same logic to a deny list as well.

simurg79:
Thanks for the valuable comments, I will check them out this weekend.

simurg79:
Is the ask here to swap from allowlist to denylist or support both?

From usability perspective, allowlist makes more sense to me since the UI is a checkbox allowing you to use MCP. Usage patterns here would be also allowing few mcp server for a particular mode out of several mcps. 

Again the ask is not clear to me.

---
In reply to: [4277512916](https://github.com/Zoo-Code-Org/Zoo-Code/pull/75#issuecomment-4277512916) [](http://example.com/codeflow?ancestors=4277512916)

packages/types/src/mode.ts
simurg79: I think the key question here is not maintaining lists but UI/UX. I was considering a simple UI/UX where no MCP is  allowed by default

I still don't quite understand the direction you are pointing out. Should we allow all mcps by default but maintain a list of denylist of MCPs per mode?

PR #81, opened 2026-05-13
Upstream/feature/Add Xiaomi MiMo as a first-class API provider

### LINKED ISSUE #80: Add Xiaomi MiMo models
Name of model: MiMo V2.5 Pro, MiMo V2.5, MiMo V2 Flash

Model creator: Xiaomi

Link to model: https://platform.xiaomimimo.com/

Additional context:

Previously this provider could be used via the "OpenAI Compatible" option with the base URL https://token-plan-sgp.xiaomimimo.com/v1, but it broke because MiMo's proxy rejects OpenAI-specific extensions that Zoo Code sends by default (strict: true on tool schemas, additionalProperties: false, parallel_tool_calls, stream_options). Additionally, MiMo requires reasoning_content to be echoed back in multi-turn tool calling conversations. Without it the API returns 400 "Param Incorrect".

capitanfeeder:
### Related GitHub Issue

Closes: #80

### Description

Add Xiaomi MiMo as a first-class API provider with full multi-turn tool calling support.

MiMo's API requires `reasoning_content` to be echoed back in assistant messages during tool calling conversations. Without this, the API returns 400 "Param Incorrect". The existing message conversion (`convertToOpenAiMessages`) was dropping reasoning content blocks entirely, and the base tool schema conversion was adding OpenAI-specific extensions (`strict: true`, `additionalProperties: false`) that MiMo's proxy rejects.

Key implementation details:

- **MimoHandler** (`src/api/providers/mimo.ts`): Extends `OpenAiHandler` with custom `convertMessagesForMiMo()` that extracts `type: "reasoning"` content blocks from Anthropic format and maps them to `reasoning_content` in OpenAI assistant messages. Overrides `convertToolsForOpenAI()` to strip `strict` and `additionalProperties` from tool schemas. Sends `thinking: { type: "enabled" }` via `extra_body`.
- **Model definitions** (`packages/types/src/providers/mimo.ts`): mimo-v2.5-pro, mimo-v2.5, mimo-v2-flash with official pricing, correct context windows (1M/1M/256K), and max output tokens (131K/131K/65K).
- **UI settings** (`webview-ui/src/components/settings/providers/Mimo.tsx`): Base URL dropdown with 4 options (3 Token Plan regions + Pay-as-you-go), API key input.
- **Full provider wiring**: types schema, provider name registration, model picker, i18n keys, provider config, useSelectedModel hook.
- Also fixed a pre-existing type error in `openrouter.ts` where `convertToOpenAiMessages` was called with 3 arguments instead of 2.

Previously this provider could be used via the "OpenAI Compatible" option with the base URL `https://token-plan-sgp.xiaomimimo.com/v1`, but it broke because MiMo's proxy rejects OpenAI-specific extensions that Roo Code sends by default. This PR adds MiMo as a dedicated provider that handles all these compatibility issues.

### Test Procedure

- Built the extension with `pnpm build` (web-evals failure is pre-existing and unrelated)
- Type checks pass: `tsc --noEmit` for types, src, and webview-ui packages with zero new errors
- 1228 of 1235 tests pass; 1 pre-existing timeout failure in SettingsView TTS test (unrelated)
- Manual testing: configured MiMo provider in Roo Code settings, sent multi-turn conversations with tool calls (read_file, list_files), verified reasoning_content is preserved in history and tool calling works end to end

### Pre-Submission Checklist

### Screenshots / Videos

N/A - Provider settings UI follows existing patterns (Moonshot, MiniMax). The MiMo settings tab shows a base URL dropdown with 4 regions and an API key input field.

### Documentation Updates

### Additional Notes

This provider is critical for users in regions where access to Western AI providers is limited. MiMo V2.5 Pro offers competitive performance with 1M context window at $1.00/M input pricing, making it a cost-effective alternative.

### Get in Touch

[gabrielsosa_](https://discord.com/users/gabrielsosa_)

## Summary by CodeRabbit

* **New Features**
  * Added Xiaomi MiMo as a supported AI provider with configurable API key and base URL selection for different regions and pricing plans
  * Full streaming and model integration support

* **Internationalization**
  * Added provider configuration translations across 16 languages

* **Tests**
  * Comprehensive test coverage for the new provider implementation

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/81?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


edelauna:
Thanks for the thorough implementation — the MiMo handler looks solid and the test coverage is great. Before we can merge though, this PR bundles three separate concerns and we'd like them split into individual PRs: 
  
1. Git LFS hooks (dev/null/)
    The four hook files are unrelated to MiMo. Please open a separate PR for these, or drop them if they were included accidentally.

3. convertToOpenAiMessages signature fix
The fix to openai-format.ts and the cascading changes across the existing providers (openrouter.ts, openai.ts, lm-studio.ts, lite-llm.ts, qwen-code.ts, requesty.ts, roo.ts, base-openai-compatible-provider.ts) is a self-contained bug fix that's valuable on its own. It touches existing providers and deserves its own focused review.

4. MiMo provider (everything else)

Once the bug fix lands, the MiMo feature PR will be much cleaner — just the model definitions, handler, UI, i18n, and tests.

The split keeps the git history bisectable and makes each piece easier to review and revert independently if needed.

src/api/transform/openai-format.ts
```diff
@@ -256,7 +256,8 @@ export function sanitizeGeminiMessages(
 /**
  * Options for converting Anthropic messages to OpenAI format.
  */
-export interface ConvertToOpenAiMessagesOptions {
+interface ConvertToOpenAiMessagesOptions {
```
taltas: I believe there is still some Mimo related plumbing we need to take out here so that @edelauna mentioned we should put in it's own PR.

src/api/providers/qwen-code.ts
```diff
@@ -223,7 +223,7 @@ export class QwenCodeHandler extends BaseProvider implements SingleCompletionHan
 			content: systemPrompt,
 		}
 
-		const convertedMessages = [systemMessage, ...convertToOpenAiMessages(messages)]
+		const convertedMessages = [systemMessage, ...convertToOpenAiMessages(messages, { modelId: model.id })]
```
taltas: same with all the `convertToOpenAiMessages` in the variuos providers, we need to put this in a seperate PR.

webview-ui/src/components/settings/providers/Mimo.tsx
```diff
@@ -0,0 +1,74 @@
...
+	return (
+		<>
+			<div>
+				<label className="block font-medium mb-1">{t("settings:providers.mimoBaseUrl")}</label>
+				<VSCodeDropdown
+					value={apiConfiguration.mimoBaseUrl}
+					onChange={handleInputChange("mimoBaseUrl")}
+					className={cn("w-full")}>
+					<VSCodeOption value="https://token-plan-sgp.xiaomimimo.com/v1" className="p-2">
+						Token Plan - Singapore (Default)
```
taltas: we should put these as locales for multiple langues

packages/types/src/providers/mimo.ts
```diff
@@ -0,0 +1,39 @@
...
+		contextWindow: 1_048_576,
+		supportsImages: false,
+		supportsPromptCache: true,
+		preserveReasoning: true,
+		inputPrice: 0.4, // $0.40 per million tokens (cache miss, ≤256K)
+		outputPrice: 2.0, // $2.00 per million tokens
+		cacheReadsPrice: 0.08, // $0.08 per million tokens (cache hit)
+		cacheWritesPrice: 0, // Free for limited time
+		description:
+			"MiMo V2.5 - Full modal understanding model with 1M context, deep thinking, tool calling, and structured output.",
```
taltas: It's weird that the description `full modal understand model` but there is no images supported... is this official description?

src/api/providers/mimo.ts
```diff
@@ -0,0 +1,303 @@
...
+		const convertedMessages = this.convertMessagesForMiMo(messages)
+
+		const tools = this.convertToolsForOpenAI(metadata?.tools)
+
+		// Build request per MiMo's OpenAI-compatible API
+		const params: Record<string, any> = {
+			model: modelId,
+			temperature,
+			messages: [{ role: "system", content: systemPrompt }, ...convertedMessages],
+			stream: true,
```
taltas, line 90: We should include usage here like other providers so we can calculate the costs.
```
			stream_options: {
				include_usage: true,
			},
```
Make sure you updare the test as well.

taltas (dismissed):
A few comments. @edelauna will give you the approval when its good, he can just dismiss my review when he is satisfied.

hannesrudolph:
@roomote review this

packages/types/src/providers/mimo.ts
capitanfeeder: Fixed. mimo-v2.5 is confirmed by MiMo's official docs as a full-modal model that processes text, image, audio, and video  updated supportsImages to true and corrected the description to reflect this. mimo-v2.5-pro stays text-only since it's the reasoning/agentic variant. See https://developer.puter.com/ai/xiaomi/mimo-v2.5/

packages/types/src/providers/mimo.ts
```diff
@@ -0,0 +1,46 @@
...
+		// Overseas pricing (≤256K input) — https://platform.xiaomimimo.com/static/docs/pricing.md
+		inputPrice: 0.4, // $0.40/1M tokens (cache miss)
+		outputPrice: 2.0, // $2.00/1M tokens
+		cacheReadsPrice: 0.08, // $0.08/1M tokens (cache hit)
+		cacheWritesPrice: 0, // Free for limited time
+		// "Full Modal Understanding Model" is the official category name from MiMo docs
+		description:
+			"MiMo V2.5 - Full-modal understanding model (text, image, audio, video) with 1M context, deep thinking, tool calling, and structured output.",
+	},
+} as const satisfies Record<string, ModelInfo>
```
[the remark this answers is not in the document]
capitanfeeder: Intentionally excluded. mimo-v2-flash has thinking mode disabled by default and doesn't reliably handle reasoning_content passthrough during multi-turn tool calling, which causes 400 errors from the proxy. The whole point of this provider is agentic tool use with thinking enabled  flash can't do that. Added a comment in the code documenting this rationale. If flash support is needed later it should be validated against the API first.

src/api/providers/mimo.ts
capitanfeeder: Done. Added stream_options: { include_usage: true } to the request params and updated the test suite to verify it.

src/api/providers/qwen-code.ts
capitanfeeder: Done. Reverted the { modelId: model.id } changes from all providers (openrouter, openai, lm-studio, lite-llm, qwen-code, requesty, roo, base-openai-compatible-provider). All back to original calls.

src/api/transform/openai-format.ts
capitanfeeder: Done. Removed both MiMo guards and restored the export interface. MimoHandler uses its own convertMessagesForMiMo and doesn't touch convertToOpenAiMessages at all.

webview-ui/src/components/settings/providers/Mimo.tsx
capitanfeeder: Done. Added mimoBaseUrlSingapore, mimoBaseUrlChina, mimoBaseUrlEurope, mimoBaseUrlPayg as i18n keys and translated across all 18 locales.

src/api/providers/lite-llm.ts
```diff
@@ -116,9 +116,7 @@ export class LiteLLMHandler extends RouterProvider implements SingleCompletionHa
 	): ApiStream {
 		const { id: modelId, info } = await this.fetchModel()
 
-		const openAiMessages = convertToOpenAiMessages(messages, {
-			normalizeToolCallId: sanitizeOpenAiCallId,
-		})
+		const openAiMessages = convertToOpenAiMessages(messages)
```
edelauna: This removes `sanitizeOpenAiCallId` which truncates tool call IDs to 64 chars and strips invalid characters (dots, colons from MCP tool IDs). Was this intentional? The import on line 12 is still present, suggesting it was accidental. Without it, MCP tools with long IDs like `call_mcp--posthog--dashboard_create_...` could cause 400s from LiteLLM-proxied providers.

```suggestion
		const openAiMessages = convertToOpenAiMessages(messages, {
			normalizeToolCallId: sanitizeOpenAiCallId,
		})
```

src/api/providers/base-openai-compatible-provider.ts
```diff
@@ -224,7 +224,7 @@ export abstract class BaseOpenAiCompatibleProvider<ModelName extends string>
 
 		const params: OpenAI.Chat.Completions.ChatCompletionCreateParams = {
 			model: modelId,
-			messages: [{ role: "user", content: prompt }],
+			messages: [{ role: "user", content: prompt }], // No history, no MiMo strip needed
```
edelauna: This comment is MiMo-specific context in a base class shared by all providers. Could we drop it?

```suggestion
			messages: [{ role: "user", content: prompt }],
```

src/api/providers/mimo.ts
```diff
@@ -0,0 +1,346 @@
...
+
+				// CRITICAL: Add reasoning_content if present
+				if (reasoningParts.length > 0) {
+					assistantMsg.reasoning_content = reasoningParts.join("\n")
+				}
+
+				// Add tool_calls if present
+				if (toolUseParts.length > 0) {
+					assistantMsg.tool_calls = toolUseParts.map((block) => ({
+						id: block.id,
```
edelauna: Does MiMo's API have constraints on tool call ID format or length? Other providers in this codebase use `sanitizeOpenAiCallId` to enforce OpenAI's 64-char alphanumeric limit. MCP tools can generate IDs like `call_mcp--posthog--dashboard_create_12345678-1234-...` — worth checking if those pass through cleanly or if you need sanitization here and on the `tool_call_id` at line 213.

packages/types/src/providers/mimo.ts
```diff
@@ -0,0 +1,47 @@
...
+export type MimoModelId = keyof typeof mimoModels
+
+export const mimoDefaultModelId: MimoModelId = "mimo-v2.5-pro"
+
+export const mimoModels = {
+	"mimo-v2.5-pro": {
+		maxTokens: 131_072,
+		contextWindow: 1_048_576,
+		supportsImages: false, // Pro series is text-only
+		supportsPromptCache: true,
```
edelauna: Has prompt caching been tested against the MiMo API? If the API doesn't return `cached_tokens` in usage, this flag could lead to misleading cost calculations.

src/api/providers/mimo.ts
```diff
@@ -0,0 +1,346 @@
...
+		if (tools && tools.length > 0) {
+			params.tools = tools
+		}
+
+		let stream: AsyncIterable<OpenAI.Chat.Completions.ChatCompletionChunk>
+		try {
+			stream = (await this.client.chat.completions.create(params as any)) as any
+		} catch (error) {
+			const { handleOpenAIError } = await import("./utils/openai-error-handler")
+			throw handleOpenAIError(error, "MiMo")
```
edelauna: `handleOpenAIError` is marked `@deprecated` and just delegates to `handleProviderError`. Could use that directly:

```suggestion
			const { handleProviderError } = await import("./utils/error-handler")
			throw handleProviderError(error, "MiMo")
```

webview-ui/src/components/ui/hooks/useSelectedModel.ts
```diff
@@ -250,6 +251,11 @@ function getSelectedModel({
 			const info = minimaxModels[id as keyof typeof minimaxModels]
 			return { id, info }
 		}
+		case "mimo": {
+			const id = apiConfiguration.apiModelId ?? defaultModelId
+			const info = mimoModels[id as keyof typeof mimoModels]
```
edelauna: If the user types a custom model ID, `info` will be `undefined` here (unlike `getModel()` in the handler which falls back to the default). Could this cause issues in the settings UI?

edelauna (commented):
Thanks so much for your continued contributions. I have a couple additional comments. Also review the output from CI.

src/api/providers/mimo.ts
```diff
@@ -0,0 +1,355 @@
+import { Anthropic } from "@anthropic-ai/sdk"
+import OpenAI from "openai"
+
+import { mimoModels, mimoDefaultModelId, MIMO_DEFAULT_TEMPERATURE, type ModelInfo } from "@roo-code/types"
+
+import type { ApiHandlerOptions } from "../../shared/api"
+
+import { ApiStream } from "../transform/stream"
+import { convertToOpenAiMessages } from "../transform/openai-format"
```
edelauna: Is this import used? I only see `convertMessagesForMiMo` called in this file — `convertToOpenAiMessages` appears only in the JSDoc comment on line 253.

packages/types/src/providers/mimo.ts
```diff
@@ -0,0 +1,47 @@
...
+export type MimoModelId = keyof typeof mimoModels
+
+export const mimoDefaultModelId: MimoModelId = "mimo-v2.5-pro"
+
+export const mimoModels = {
+	"mimo-v2.5-pro": {
+		maxTokens: 131_072,
+		contextWindow: 1_048_576,
+		supportsImages: false, // Pro series is text-only
+		supportsPromptCache: false,
```
edelauna, line 22: Could you add a comment explaining the relationship between `supportsPromptCache: false` and the non-zero `cacheReadsPrice` below? `supportsPromptCache` controls client-side `cache_control` injection (which MiMo doesn't support), but MiMo's server-side cache can still return `cached_tokens` in usage — so the prices are correct for cost calculation. Without a note, a future contributor might "fix" this by flipping the flag to `true` and accidentally enabling the `cache_control` injection path.

src/api/providers/mimo.ts
```diff
@@ -0,0 +1,355 @@
...
+				// When both tool_results and text exist in the same user turn (happens
+				// during resume/delegate flows where <environment_details> is appended),
+				// fold the text into the last tool message instead of creating a separate
+				// user message. A fresh role:"user" after tool results drops reasoning
+				// continuity for thinking models.
+				if (toolResults.length > 0 && textBlocks.length > 0 && mediaParts.length === 0) {
+					const lastToolIdx = converted.length - 1
+					const lastTool = converted[lastToolIdx] as any
+					if (lastTool?.role === "tool") {
+						lastTool.content += "\n" + textBlocks.join("\n")
```
edelauna: The other providers that do this (`convertToOpenAiMessages` with `mergeToolResultText`, DeepSeek's `convertToR1Format`, Z.ai's `convertToZAiFormat`) all use `\n\n` as the separator between tool content and appended text. This uses `\n`. Should it match the convention?

```suggestion
						lastTool.content += "\n\n" + textBlocks.join("\n")
```

src/api/providers/mimo.ts
```diff
@@ -0,0 +1,355 @@
...
+						type: "tool_call_partial",
+						index: toolCall.index,
+						id: toolCall.id,
+						name: toolCall.function?.name,
+						arguments: toolCall.function?.arguments,
+					}
+				}
+			}
+
+			if (chunk.usage) {
```
edelauna, line 130: Other providers that override the stream loop (see `openai.ts:489-496`, `base-openai-compatible-provider.ts:178-182`) emit `tool_call_end` events when `finish_reason === "tool_calls"`. There's a safety net in `NativeToolCallParser.finalizeRawChunks()` that catches these at stream end, so it won't break — but MCP tools benefit from explicit finalization. Worth adding the same pattern here?

src/api/providers/mimo.ts
```diff
@@ -0,0 +1,355 @@
...
+
+		if (tools && tools.length > 0) {
+			params.tools = tools
+		}
+
+		let stream: AsyncIterable<OpenAI.Chat.Completions.ChatCompletionChunk>
+		try {
+			stream = (await this.client.chat.completions.create(params as any)) as any
+		} catch (error) {
+			const { handleProviderError } = await import("./utils/error-handler")
```
edelauna: Any reason for the dynamic import? A static `import { handleProviderError } from "./utils/error-handler"` at the top of the file would avoid the deferred resolution on first error.

```suggestion
			const { handleProviderError } = await import("./utils/error-handler")
```

(Or just move it to a top-level import.)

webview-ui/src/components/settings/providers/Mimo.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { useAppTranslation } from "@src/i18n/TranslationContext"
+import { VSCodeButtonLink } from "@src/components/common/VSCodeButtonLink"
+
+import { inputEventTransform } from "../transforms"
+import { cn } from "@/lib/utils"
+
+type MimoProps = {
+	apiConfiguration: ProviderSettings
+	setApiConfigurationField: (field: keyof ProviderSettings, value: ProviderSettings[keyof ProviderSettings]) => void
+	simplifySettings?: boolean
```
edelauna: `simplifySettings` is declared here and passed from `ApiOptions.tsx`, but the component never reads it. Remove it from the type, or did you intend to use it?

edelauna (commented):
Review of latest state (66cf689). A few cleanup items before this is ready to approve — see inline comments.

src/api/providers/mimo.ts
edelauna: Going further on this — `convertMessagesForMiMo` reimplements most of what `convertToR1Format` (in `src/api/transform/r1-format.ts`) already does:

| Feature | `convertToR1Format` | `convertMessagesForMiMo` |
|---|---|---|
| `reasoning_content` passthrough | ✅ | ✅ |
| `type: "reasoning"` block extraction | ✅ (line 184) | ✅ |
| `mergeToolResultText` | ✅ | ✅ (new in 66cf689) |
| Image block conversion | ✅ | ✅ |
| Same-role message merging | ✅ | ❌ |
| `normalizeToolCallId` | ❌ | ✅ (via `sanitizeOpenAiCallId`) |

The only thing `convertToR1Format` is missing is a `normalizeToolCallId` option. If you add that (same pattern as `convertToOpenAiMessages`), MiMo could just call:

```ts
const convertedMessages = convertToR1Format(messages, {
    mergeToolResultText: true,
    normalizeToolCallId: sanitizeOpenAiCallId,
})
```

This would eliminate ~120 lines of reimplemented conversion logic, keep MiMo in sync with future bug fixes to the shared converter, and naturally resolve the dead import. The `stripOpenAiExtensions` / `convertToolsForOpenAI` logic is MiMo-specific and would stay in the handler.

src/api/providers/mimo.ts
edelauna: Yes — `OpenAiHandler` already has a `processToolCalls` helper (line 466) that handles both `tool_call_partial` and `tool_call_end` emission, including tracking active IDs. It's currently `private`, but if you change it to `protected`, MiMo's stream loop could replace the inline tool_calls block with:

```ts
const activeToolCallIds = new Set<string>()

for await (const chunk of stream) {
    const delta = chunk.choices?.[0]?.delta ?? {}
    const finishReason = chunk.choices?.[0]?.finish_reason

    // ... text and reasoning_content handling ...

    yield* this.processToolCalls(delta, finishReason, activeToolCallIds)

    // ... usage handling ...
}
```

That gets you `tool_call_end` for free and keeps the behavior in sync with the base class.

edelauna: Actually, scratch that — changing `private` to `protected` just for one subclass isn't great practice. And looking more closely, `DeepSeekHandler` (the only other `OpenAiHandler` subclass) has the same pattern: inline `tool_call_partial` without `tool_call_end`. Both rely on `NativeToolCallParser.finalizeRawChunks()` as the safety net.

So as-is, MiMo is consistent with DeepSeek. The missing `tool_call_end` is a pre-existing gap across both subclasses, not something to fix in this PR alone. If the maintainers want to address it, the right fix would be to refactor `processToolCalls` to `protected` and have both DeepSeek and MiMo use it — but that's a separate cleanup.

Withdrawing this as a request for this PR.

packages/types/src/providers/mimo.ts
```diff
@@ -0,0 +1,47 @@
...
+		cacheReadsPrice: 0.08, // $0.08/1M tokens (cache hit)
+		cacheWritesPrice: 0, // Free for limited time
+		description:
+			"MiMo V2.5 - Full-modal understanding model (text, image, audio, video) with 1M context, deep thinking, tool calling, and structured output.",
+	},
+} as const satisfies Record<string, ModelInfo>
+
+export const mimoDefaultModelInfo: ModelInfo = mimoModels[mimoDefaultModelId]
+
+export const MIMO_DEFAULT_TEMPERATURE = 1.0
```
edelauna, line 62: for my knowledge - why is the default 1 and not 0?

src/api/providers/mimo.ts
```diff
@@ -0,0 +1,212 @@
...
+		// Build request per MiMo's OpenAI-compatible API
+		// https://developer.puter.com/ai/xiaomi/mimo-v2.5-pro/
+		const params: Record<string, any> = {
+			model: modelId,
+			temperature,
+			messages: [{ role: "system", content: systemPrompt }, ...convertedMessages],
+			stream: true,
+			stream_options: { include_usage: true },
+			// MiMo requires thinking to be enabled via extra_body
+			extra_body: { thinking: { type: "enabled" } },
```
edelauna: Per the [official API docs](https://platform.xiaomimimo.com/docs/en-US/api/chat/openai-api), the `temperature` field under the `thinking` parameter:

> In thinking mode, the mimo-v2.5-pro and mimo-v2.5 models do not support customizing the temperature parameter. Even if this parameter is passed in, it will be forcibly overridden and take effect with the model's recommended default value of 1.0.

Since thinking is always enabled here, the temperature in the request is dead code — users adjusting it in settings won't see any effect. Could we either omit it from the request when thinking is on, or at minimum add a comment noting it's ignored by the API?

src/api/providers/mimo.ts
```diff
@@ -0,0 +1,212 @@
...
+		})
+		return { id, info, ...params }
+	}
+
+	/**
+	 * Strip OpenAI-specific extensions that MiMo's proxy rejects:
+	 * - strict: true on tools
+	 * - additionalProperties: false on schemas
+	 */
+	protected override convertToolsForOpenAI(tools: any[] | undefined): any[] | undefined {
```
edelauna: The [official API reference](https://platform.xiaomimimo.com/docs/en-US/api/chat/openai-api) documents `strict` as a supported boolean field (default `false`) on function tool definitions:

> Whether to enable strict schema adherence when generating the function call.

Is the stripping still needed for the Token Plan proxy endpoints? If this was a proxy bug that's since been fixed, we could remove `convertToolsForOpenAI` and `stripOpenAiExtensions` entirely (~50 lines), which would be a nice diff reduction.

packages/types/src/providers/mimo.ts
```diff
@@ -0,0 +1,47 @@
...
+	"mimo-v2.5-pro": {
+		maxTokens: 131_072,
+		contextWindow: 1_048_576,
+		supportsImages: false, // Pro series is text-only
+		supportsPromptCache: false,
+		preserveReasoning: true,
+		inputPrice: 1.0, // $1.00/1M tokens (cache miss, ≤256K)
+		outputPrice: 3.0, // $3.00/1M tokens
+		cacheReadsPrice: 0.2, // $0.20/1M tokens (cache hit)
+		cacheWritesPrice: 0, // Free for limited time
```
edelauna, line 27: MiMo uses [2-tier pricing](https://platform.xiaomimimo.com/static/docs/pricing.md) — input >256K costs 2x ($1→$2 input, $3→$6 output for Pro; $0.40→$0.80 input, $2→$4 output for Omni). The framework already supports this via `tiers` (see `packages/types/src/providers/bedrock.ts`). Could we add something like:

```suggestion
		inputPrice: 1.0, // $1.00/1M tokens (cache miss, ≤256K)
		outputPrice: 3.0, // $3.00/1M tokens (≤256K)
		cacheReadsPrice: 0.2, // $0.20/1M tokens (cache hit, ≤256K)
		cacheWritesPrice: 0, // Free for limited time
		tiers: [
			{
				contextWindow: 1_048_576,
				inputPrice: 2.0, // $2.00/1M tokens (>256K)
				outputPrice: 6.0, // $6.00/1M tokens (>256K)
				cacheReadsPrice: 0.4, // $0.40/1M tokens (>256K)
			},
		],
```

packages/types/src/providers/mimo.ts
edelauna: Looks like this is from mimo documentation - resolving.

edelauna (commented):
Couple more comments - thanks for your continued contributions. This should be the last round.

packages/types/src/providers/mimo.ts
```diff
@@ -0,0 +1,66 @@
...
+		// MiMo charges 2x for input >256K context
+		tiers: [
+			{
+				contextWindow: 256_001,
+				inputPrice: 2.0, // $2.00/1M tokens (>256K)
+				outputPrice: 6.0, // $6.00/1M tokens (>256K)
+				cacheReadsPrice: 0.4, // $0.40/1M tokens (>256K)
+			},
+		],
```
edelauna: Sorry, my earlier suggestion to use `tiers` was wrong — `calculateApiCostOpenAI` (which MiMo's handler calls) doesn't read `tiers`. It reads `longContextPricing`, which is designed exactly for this "prices double above a threshold" pattern. See `openai.ts` line 39 for an example.

```suggestion
		// MiMo charges 2x above 256K context (https://platform.xiaomimimo.com/static/docs/pricing.md)
		longContextPricing: {
			thresholdTokens: 256_000,
			inputPriceMultiplier: 2,
			outputPriceMultiplier: 2,
			cacheReadsPriceMultiplier: 2,
		},
```

packages/types/src/providers/mimo.ts
```diff
@@ -0,0 +1,66 @@
...
+		cacheWritesPrice: 0, // Free for limited time
+		// MiMo charges 2x for input >256K context
+		tiers: [
+			{
+				contextWindow: 256_001,
+				inputPrice: 0.8, // $0.80/1M tokens (>256K)
+				outputPrice: 4.0, // $4.00/1M tokens (>256K)
+				cacheReadsPrice: 0.16, // $0.16/1M tokens (>256K)
+			},
+		],
```
edelauna: Same here — `longContextPricing` instead of `tiers`.

```suggestion
		// MiMo charges 2x above 256K context (https://platform.xiaomimimo.com/static/docs/pricing.md)
		longContextPricing: {
			thresholdTokens: 256_000,
			inputPriceMultiplier: 2,
			outputPriceMultiplier: 2,
			cacheReadsPriceMultiplier: 2,
		},
```

edelauna (approved):
Thanks for adding this.

PR #155, opened 2026-05-17
Integrate Zoo portable core CLI, SDK, and VS Code bridge

mojomast:
## Summary
- Introduces the Zoo portable-core monorepo layout with `packages/zoo-cli`, `packages/zoo-sdk`, and relocated VS Code/webview packages.
- Adds the Zoo CLI, typed SDK facade, default-off VS Code portable-core bridge, local `zoo.jsonc` schema validation, and broad focused route/adapter coverage.
- Removes active bundled Kilo gateway/indexing dependencies, preserves migration fallbacks, and rebrands visible CLI/TUI surfaces to Zoo Code.

## Maintainer Handoff
- A dedicated PR guide is included in `PR_README.md` with architecture notes, review map, verification history, known gaps, and suggested next steps.
- `HANDOFF.md`, `DEVPLAN.md`, and `CHANGELOG.md` were kept current throughout the integration work.
- The portable-core VS Code path remains gated behind `zoo-code.usePortableCore` and defaults to `false`.

## Verification
- `pnpm --filter @zoo-code/cli build`
- `pnpm --filter @zoo-code/cli check-types`
- `pnpm --filter @zoo-code/sdk test`
- `pnpm --filter @zoo-code/sdk check-types`
- `pnpm --filter @zoo-code/sdk build`
- `pnpm --filter zoo-code check-types`
- `pnpm --filter zoo-code exec vitest run services/portable-core/__tests__/PortableSessionAdapter.test.ts core/webview/__tests__/ClineProvider.spec.ts`
- `pnpm build`
- Push hook `turbo check-types`

## Known Follow-Ups
- Keep portable-core VS Code mode default-off until complete extension e2e coverage is added.
- Hosted public `$schema` URL for `zoo.jsonc` is still deferred; this PR ships local VSIX schema validation.
- Some internal compatibility names still contain `kilo` because they are inherited dependency/API seams; user-facing TUI strings were rebranded, but internal cleanup should be done in dedicated follow-up slices.
- All-platform CLI release packaging still needs dedicated release-pipeline work; current-platform CLI build is functional.

PR #158, opened 2026-05-17
Restructure extension into workspace packages

mojomast:
## Summary

- Moves the extension source from root `src` into `packages/zoo-vscode/src`.
- Moves the webview from root `webview-ui` into `packages/zoo-vscode/webview-ui`.
- Wires the monorepo workspace/package layout so later Zoo CLI and SDK work can land as separate packages.

## Scope

This is intended as the mechanical workspace/package-layout foundation for the portable-core follow-up stack. It does not add the portable-core CLI implementation, SDK implementation, or VS Code portable-core bridge behavior.

Placeholder `@zoo-code/cli` and `@zoo-code/sdk` package manifests are included only so later stacked PRs can add those implementations without changing the workspace shape again.

## Verification

- `pnpm lint`
- `pnpm check-types`
- push hook `turbo check-types`

## Review Notes

Most of this PR is file movement. Reviewing with rename detection enabled is recommended.

Umbrella/reference PR: #155


mojomast:
Closing this split because the workspace move will not be pursued as a prerequisite. I’ll replace it with independent PRs based directly on upstream main, starting with the portable support packages only.

PR #197, opened 2026-05-19
test: migrate workspaces to Vitest 4

maxdewald:
## Summary

Contains the Vitest 4/test-runner migration only. The Vite 8 webview production build migration was split out into #214.

This PR includes:

- upgrade Vitest to 4.1.x across workspaces, including matching `@vitest/coverage-v8` and `@vitest/ui` packages
- update Vitest config options for Vitest 4, including the Windows CI worker limit option
- add Node ambient types alongside `vitest/globals` in package tsconfigs that compile test/config files
- update test mocks for Vitest 4 typing and constructor behavior
- keep existing path alias config where needed for test resolution
- add the Vite 6-compatible React test transform for `apps/web-roo-code` TSX tests without pulling in Vite 8

The Vite 8/webview build changes are intentionally not part of this PR and are isolated in #214.

No changeset is included because this is a tooling/test migration and does not change published package behavior.

Related PRs:

- #214 — Vite 8 webview build migration


edelauna:
Thanks for taking this on — getting the webview/test tooling onto current versions is valuable.

I’m a bit concerned about the review/merge risk from bundling several major migrations together here. This PR currently combines at least three distinct upgrade surfaces:

1. Vite 6 → Vite 8 / Rolldown webview build changes
2. Vitest 3 → Vitest 4 test-runner and mock/type updates
3. Tailwind tooling updates / dependency reshuffling

Do these need to land together? If there’s a concrete compatibility reason, it would be helpful to document it in the PR description. Otherwise, I think this would be much easier and safer to review if split roughly as:

- Vitest 4 migration + test/mock fixes
- Vite 8 / webview build output migration
- Tailwind tooling/dependency cleanup

The Vite/Rolldown changes in particular seem worth isolating because they affect the production webview bundle/chunking/output assumptions, while the Vitest changes are mostly CI/test-harness behavior. Keeping those separate would make regressions much easier to attribute.

Before merging, I’d also want CI green and a quick webview smoke test confirming the extension still produces/loads the expected stable assets, especially `assets/index.js` and `assets/index.css`, plus any lazy-loaded/Mermaid-related paths.

maxdewald:
Thanks for the review. I split the changes into two PRs:

- #214 contains the Vite 8 webview build migration.
- #197 contains the Vitest 4 test-runner migration.

The PR descriptions and changed files have been checked and updated so the split matches the intended scope.

src/vitest.config.ts
```diff
@@ -15,13 +15,7 @@ export default defineConfig({
...
 		hookTimeout: 20_000,
 		onConsoleLog,
-		poolOptions: isWindowsCI
-			? {
-					forks: {
-						singleFork: true,
-					},
-				}
-			: undefined,
+		maxWorkers: isWindowsCI ? 1 : undefined,
```
edelauna, line 18: The previous `poolOptions.forks.singleFork: true` (added in `dee7adaf7` to stabilize Windows CI coverage) ran all test files in a single shared process — no cross-process coverage merging. `maxWorkers: 1` serializes forks but still creates a separate process per file, so coverage is still merged across processes.

Was `singleFork` removed/deprecated in Vitest 4, making this the closest equivalent? Worth a short comment here so future readers know this isn't an accidental weakening of the Windows CI fix.

edelauna (changes requested):
Thanks for this much needed upgrade. I had one comment on windows ci changes, otherwise could you address the merge conflicts? I wanted to wait until we had merged the final upstream roo changes before performing a more indepth review of your PRs.

src/vitest.config.ts
maxdewald: Yes, `singleFork` was fully removed in Vitest 4 — per the [Pool Rework section of the migration guide](https://vitest.dev/guide/migration#pool-rework):

> `singleThread` and `singleFork` are now `maxWorkers: 1, isolate: false`. `poolOptions` is removed.

So `maxWorkers: 1` alone was not the complete equivalent — `isolate: false` is also needed to collapse everything into a single shared process (without it, a separate process is still spawned per file). Fixed in f1c2f4c with both options set.

edelauna:
Ended up covering this via renovate: https://github.com/Zoo-Code-Org/Zoo-Code/pull/443

PR #214, opened 2026-05-20
chore(webview): migrate build to Vite 8

maxdewald:
## Summary

Splits the Vite/webview build migration out of #197 so the Vitest 4 test-runner migration can stack separately.

This PR includes only the production webview build surface:

- upgrade webview Vite from 6.3.6 to 8.0.13 and React plugin to 5.2.0
- move the webview build config from `rollupOptions` to `rolldownOptions`
- preserve stable asset names for `assets/index.js` and `assets/index.css`
- bump the Node 20 typings used by Vite-facing workspaces to satisfy Vite 8's peer range

## Summary by CodeRabbit

* **Chores**
  * Updated TypeScript Node type definitions and Vite build toolchain to more specific, stable versions for improved compatibility.

* **Refactor**
  * Revised build configuration: simplified bundling, minification behavior, and asset output patterns; removed custom pre-bundling and special-chunk handling for certain large visualization dependencies to streamline builds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/214?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


proyectoauraorg (changes requested):
## Code Review: Migrate webview build to Vite 8

**Overall Assessment: Clean migration with good simplification. Request changes for one blocking concern.**

### Strengths

1. **Rolldown migration**: The switch from `rollupOptions` to `rolldownOptions` correctly follows Vite 8's move to Rolldown/Oxc as the default bundler. Good to stay current.

2. **Simplified chunk strategy**: Removing the complex `manualChunks` function for mermaid/dagre/cytoscape is a win — Rolldown handles this better natively, and the old logic was fragile.

3. **tsconfigPaths plugin**: Replacing the manual `resolve.alias` with `tsconfigPaths: true` is cleaner and keeps paths in sync with tsconfig automatically.

4. **minify simplification**: Using `minify: mode === "production"` instead of the conditional esbuild string is cleaner.

### Blocking Concern

1. **`@vitejs/plugin-react` major version bump (4.x → 5.x)**: This is a major version change that may have breaking changes. The diff doesn't show any changes to the plugin configuration, but v5 may require adjustments to the React compiler or babel plugin settings. Has this been tested with the `babel-plugin-react-compiler` that's also in the devDependencies? The React Compiler integration changed significantly in plugin-react v5.

### Non-blocking Suggestions

1. **`@types/node` specifier change (20.x → ^20.19.25)**: This is fine but unrelated to the Vite migration. Consider splitting into a separate commit for cleaner git history.

2. **`lightningcss` version bump in lockfile** (1.30.1 → 1.32.0): Similarly transitive — fine but worth noting in case of CSS rendering differences.

3. **`jiti` version bump** (2.4.2 → 2.7.0): Lockfile-level change, likely harmless but affects all packages that use jiti for config loading.

4. **Missing e2e smoke test**: Given the bundler change, consider running the `e2e-mock` workflow specifically on this PR to catch any runtime regressions in the webview.

### Questions

- Was the `mermaid-bundle` chunk naming strategy intentionally removed, or did Rolldown handle it differently? If mermaid is still dynamically imported, the chunk splitting behavior may differ and affect load performance.

webview-ui/vite.config.ts
```diff
@@ -188,11 +148,6 @@ export default defineConfig(({ mode }) => {
 		},
 		define,
 		optimizeDeps: {
-			include: [
-				"mermaid",
-				"dagre", // Explicitly include dagre for pre-bundling
-				// Add other known large mermaid dependencies if identified
-			],
 			exclude: ["@vscode/codicons", "vscode-oniguruma", "shiki"],
```
edelauna, line 151: The old config had `include: ["mermaid", "dagre"]` here for dev-mode pre-bundling. Mermaid is fairly large — was its removal intentional, or did it just get dropped as part of the `manualChunks` cleanup? Dev cold-start could be noticeably slower without it.

webview-ui/vite.config.ts
```diff
@@ -98,79 +98,39 @@ export default defineConfig(({ mode }) => {
...
-						const moduleInfo = getModuleInfo(id)
-						if (moduleInfo?.importers.some((importer) => importer.includes("node_modules/mermaid"))) {
-							return "mermaid-bundle"
-						}
-						if (
-							moduleInfo?.dynamicImporters.some((importer) => importer.includes("node_modules/mermaid"))
-						) {
-							return "mermaid-bundle"
-						}
+						return "assets/[name][extname]"
```
edelauna, line 133: The previous `assetFileNames` had an explicit branch for `.map` files (`name.endsWith(".map") → "assets/[name]"`). Rolldown emits source maps as sidecar files so it likely doesn't matter, but worth confirming source maps still land where the VS Code extension expects them.

edelauna (changes requested):
looks good, this is a much needed update - can you resolve the lock conflicts - I had to update it as part of the final upstream Roo changes / detaching from fork.

regarding @proyectoauraorg review: 
* I think because mermaid is statically loaded there no lazy loading benefit
* regarding plugin-react v5 + babel-plugin-react-compiler compatibility - I'm hoping ci would've caught this, but all checks were green, is there a compatibility issue here, should we update ci?

edelauna (commented):
Looks good - thank you!

edelauna (approved):
Lockfile is clean, CI is green, changes are correct for a Vite 8 migration. Filed #280 for the pre-existing sourcemapPlugin mode detection bug.

PR #222, opened 2026-05-20
test(readFileTool): add comprehensive test coverage for ReadFileTool

proyectoauraorg:
## Summary

Adds comprehensive test coverage for ReadFileTool, increasing from **33 to 77 tests** (+44 new tests).

### New Test Coverage

**handlePartial error handling:**
- Validates error when mode parameter is missing
- Validates error when fullFileRead is missing from partial data
- Validates error when lineSnippet is missing from partial data
- Validates error when path cannot be retrieved from approval item
- Returns empty result with success flag when all validations pass

**getStartLine fallback logic:**
- Returns undefined when diff parameters are null
- Returns undefined when only insertions in diff (no deletions)
- Uses start line from edit content when no diff match

**getLineSnippet edge cases:**
- Returns empty string when startLine exceeds total lines
- Handles single-line files correctly
- Correctly reads middle line ranges

**Path normalization:**
- Resolves relative paths against cwd
- Resolves tilde paths to home directory
- Trims whitespace from paths

**Legacy parameters:**
- Supports legacy fullContent parameter
- Supports legacy lineFrom parameter
- Supports legacy initLine parameter

**Direct file reading:**
- Reads files in non-legacy mode with auto-detected ranges

**Batch approval flow:**
- Processes first item in batch and queues remainder
- Sets isReadingFile on pending approval items
- Skips next_item processing when not reading a file

**params getter:**
- Returns current params object

### Test Results
- **77 total tests** (was 33, added 44)
- **0 failures**
- Full test suite: **5637 tests pass**

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for file-reading flows: legacy and new input shapes, many edge cases (legacy detection/warnings, line-range slicing, indentation/slice modes, image validation, directory/permission/ignored-file handling), approval/denial messaging, truncation and fallback behaviors, and prevention of cross-test side effects.
* **Bug Fixes**
  * Improved detection/compatibility with legacy file-parameter formats so older payloads are recognized.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/222?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/types/src/tool-params.ts
+++ b/packages/types/src/tool-params.ts
@@ -89,7 +89,14 @@ export type ReadFileToolParams = ReadFileParams | LegacyReadFileParams
  * Type guard to check if params are in legacy format.
  */
 export function isLegacyReadFileParams(params: ReadFileToolParams): params is LegacyReadFileParams {
-	return "_legacyFormat" in params && params._legacyFormat === true
+	// `NativeToolCallParser` always tags freshly parsed legacy calls with `_legacyFormat: true`.
+	// The bare-`files` fallback only matters for chat history persisted before that flag was
+	// introduced (commit cc86049f1) and re-hydrated on a later run. Note that params matched via
+	// that fallback narrow to `LegacyReadFileParams` but leave `_legacyFormat` `undefined`, so
+	// callers should branch on the presence of `files`, not on `_legacyFormat === true`.
+	const hasLegacyFlag = "_legacyFormat" in params && params._legacyFormat === true
+	const hasFilesArray = "files" in params && Array.isArray((params as unknown as Record<string, unknown>).files)
+	return hasLegacyFlag || hasFilesArray
 }
 
 export interface Coordinate {
--- a/src/core/tools/ReadFileTool.ts
+++ b/src/core/tools/ReadFileTool.ts
@@ -668,9 +668,6 @@ export class ReadFileTool extends BaseTool<"read_file"> {
 		const { pushToolResult } = callbacks
 		const modelInfo = task.api.getModel().info
 
-		// Temporary indicator for testing legacy format detection
-		console.warn("[read_file] Legacy format detected - using backward compatibility path")
-
 		if (!fileEntries || fileEntries.length === 0) {
 			task.consecutiveMistakeCount++
 			task.recordToolError("read_file")
@@ -694,6 +691,8 @@ export class ReadFileTool extends BaseTool<"read_file"> {
 				await task.say("rooignore_error", relPath)
 				const errorMsg = formatResponse.rooIgnoreError(relPath)
 				results.push(`File: ${relPath}\nError: ${errorMsg}`)
+				// Mirror the native path: a blocked file marks the tool turn as failed.
+				task.didToolFailInCurrentTurn = true
 				continue
 			}
 
@@ -731,6 +730,8 @@ export class ReadFileTool extends BaseTool<"read_file"> {
 					const errorMsg = `Cannot read '${relPath}' because it is a directory.`
 					results.push(`File: ${relPath}\nError: ${errorMsg}`)
 					await task.say("error", `Error reading file ${relPath}: ${errorMsg}`)
+					// Mirror the native path: a failed read marks the tool turn as failed.
+					task.didToolFailInCurrentTurn = true
 					continue
 				}
 
@@ -802,6 +803,8 @@ export class ReadFileTool extends BaseTool<"read_file"> {
 				const errorMsg = error instanceof Error ? error.message : String(error)
 				results.push(`File: ${relPath}\nError: ${errorMsg}`)
 				await task.say("error", `Error reading file ${relPath}: ${errorMsg}`)
+				// Mirror the native path: a failed read marks the tool turn as failed.
+				task.didToolFailInCurrentTurn = true
 			}
 		}
 
--- a/src/core/tools/__tests__/readFileTool.spec.ts
+++ b/src/core/tools/__tests__/readFileTool.spec.ts
@@ -730,4 +730,760 @@ describe("ReadFileTool", () => {
 			expect(description).toBe("[read_file with missing path]")
 		})
 	})
+
+	describe("legacy format handling", () => {
+		it("should detect legacy format and use backward compatibility path", async () => {
+			const mockTask = createMockTask()
+			const callbacks = createMockCallbacks()
+
+			mockTask.ask.mockResolvedValue({ response: "yesButtonClicked", text: undefined, images: undefined })
+			mockedFsReadFile.mockResolvedValue(Buffer.from("legacy content"))
+
+			await readFileTool.execute({ files: [{ path: "legacy.ts" }] } as any, mockTask as any, callbacks)
+
+			// The legacy path emits "File: <path>" entries — proof the backward-compat branch ran.
+			expect(callbacks.pushToolResult).toHaveBeenCalledWith(expect.stringContaining("File: legacy.ts"))
+		})
+
+		it("should return error when legacy files array is empty", async () => {
+			const mockTask = createMockTask()
+			const callbacks = createMockCallbacks()
+
+			await readFileTool.execute({ files: [] } as any, mockTask as any, callbacks)
+
+			expect(mockTask.consecutiveMistakeCount).toBe(1)
+			expect(mockTask.recordToolError).toHaveBeenCalledWith("read_file")
+			expect(callbacks.pushToolResult).toHaveBeenCalledWith(expect.stringContaining("Error:"))
+		})
+
+		it("should handle multiple files in legacy format", async () => {
+			const mockTask = createMockTask()
+			const callbacks = createMockCallbacks()
+
+			mockTask.ask.mockResolvedValue({ response: "yesButtonClicked", text: undefined, images: undefined })
+			mockedFsReadFile.mockResolvedValueOnce(Buffer.from("file1 content"))
+			mockedFsReadFile.mockResolvedValueOnce(Buffer.from("file2 content"))
+
+			// Override readWithSlice to return content that reflects the actual file data
+			mockedReadWithSlice
+				.mockReturnValueOnce({
+					content: "1 | file1 content",
+					returnedLines: 1,
+					totalLines: 1,
+					wasTruncated: false,
+					includedRanges: [[1, 1]],
+				})
+				.mockReturnValueOnce({
+					content: "1 | file2 content",
+					returnedLines: 1,
+					totalLines: 1,
+					wasTruncated: false,
+					includedRanges: [[1, 1]],
+				})
+
+			await readFileTool.execute(
+				{ files: [{ path: "file1.ts" }, { path: "file2.ts" }] } as any,
+				mockTask as any,
+				callbacks,
+			)
+
+			expect(callbacks.pushToolResult).toHaveBeenCalledWith(expect.stringContaining("file1 content"))
+			expect(callbacks.pushToolResult).toHaveBeenCalledWith(expect.stringContaining("file2 content"))
+		})
+
+		it("should block rooignore-protected files in legacy format", async () => {
+			const mockTask = createMockTask({ rooIgnoreAllowed: false })
+			const callbacks = createMockCallbacks()
+
+			await readFileTool.execute({ files: [{ path: "secret.env" }] } as any, mockTask as any, callbacks)
+
+			expect(mockTask.say).toHaveBeenCalledWith("rooignore_error", "secret.env")
+			expect(callbacks.pushToolResult).toHaveBeenCalledWith(expect.stringContaining("blocked by the .rooignore"))
+			// Consistent with the native path: a blocked file fails the tool turn.
+			expect(mockTask.didToolFailInCurrentTurn).toBe(true)
+		})
+
+		it("should deny legacy file read when user clicks no", async () => {
+			const mockTask = createMockTask()
+			const callbacks = createMockCallbacks()
+
+			mockTask.ask.mockResolvedValue({ response: "noButtonClicked", text: undefined, images: undefined })
+
+			await readFileTool.execute({ files: [{ path: "protected.ts" }] } as any, mockTask as any, callbacks)
+
+			expect(mockTask.didRejectTool).toBe(true)
+			expect(callbacks.pushToolResult).toHaveBeenCalledWith(expect.stringContaining("Denied by user"))
+		})
+
+		it("should handle directory path in legacy format", async () => {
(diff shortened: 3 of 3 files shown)


proyectoauraorg:
Addressed the spy leakage @coderabbitai — added a scoped \`afterEach\` that restores \`console.warn\` (guarded by \`vi.isMockFunction\`) so it no longer leaks across tests, while leaving the suite-level module mocks intact. 75 tests pass locally.

packages/types/src/tool-params.ts
```diff
@@ -89,7 +89,10 @@ export type ReadFileToolParams = ReadFileParams | LegacyReadFileParams
  * Type guard to check if params are in legacy format.
  */
 export function isLegacyReadFileParams(params: ReadFileToolParams): params is LegacyReadFileParams {
-	return "_legacyFormat" in params && params._legacyFormat === true
+	// Detect explicit flag (new code path) or bare `files` array (real legacy chat data)
+	const hasLegacyFlag = "_legacyFormat" in params && params._legacyFormat === true
+	const hasFilesArray = "files" in params && Array.isArray((params as unknown as Record<string, unknown>).files)
+	return hasLegacyFlag || hasFilesArray
```
edelauna, line 99: Does a bare `files` array without `_legacyFormat` actually occur in practice? `NativeToolCallParser` already sets `_legacyFormat: true` whenever it parses a legacy `files` array (lines 422 and 753 of `NativeToolCallParser.ts`). Is there a specific deserialization path (e.g. old chat history saved before `_legacyFormat` was introduced in `cc86049f1`) that produces params with `files` but no flag? If so, it'd be worth noting in the comment.

src/core/tools/__tests__/readFileTool.spec.ts
```diff
@@ -730,4 +739,847 @@ describe("ReadFileTool", () => {
...
+			const callbacks = createMockCallbacks()
+
+			mockTask.ask.mockResolvedValue({ response: "yesButtonClicked", text: undefined, images: undefined })
+			mockedFsReadFile.mockResolvedValue(Buffer.from("legacy content"))
+
+			const consoleWarnSpy = vi.spyOn(console, "warn").mockImplementation(() => {})
+
+			await readFileTool.execute({ files: [{ path: "legacy.ts" }] } as any, mockTask as any, callbacks)
+
+			expect(consoleWarnSpy).toHaveBeenCalledWith(expect.stringContaining("Legacy format detected"))
```
edelauna: This asserts on a `console.warn` that's labeled "Temporary indicator for testing legacy format detection" in `ReadFileTool.ts:671`. If the warn gets removed (as the comment implies it should), this test will fail. Worth noting — or better, removing the production `console.warn` and this assertion together.

src/core/tools/__tests__/readFileTool.spec.ts
```diff
@@ -730,4 +739,847 @@ describe("ReadFileTool", () => {
...
+		it("should block rooignore-protected files in legacy format", async () => {
+			const mockTask = createMockTask({ rooIgnoreAllowed: false })
+			const callbacks = createMockCallbacks()
+
+			vi.spyOn(console, "warn").mockImplementation(() => {})
+
+			await readFileTool.execute({ files: [{ path: "secret.env" }] } as any, mockTask as any, callbacks)
+
+			expect(mockTask.say).toHaveBeenCalledWith("rooignore_error", "secret.env")
+			expect(callbacks.pushToolResult).toHaveBeenCalledWith(expect.stringContaining("blocked by the .rooignore"))
```
edelauna, line 801: The equivalent new-format test (`should set didToolFailInCurrentTurn on rooignore block` at line 1565) asserts `mockTask.didToolFailInCurrentTurn` is `true`. Could this test also check that flag? If `executeLegacy` doesn't set it (and it currently doesn't), the assertion would fail — which would surface an inconsistency between the two code paths.

edelauna (commented):
Nice! Thanks for your contribution had some test comments.

packages/types/src/tool-params.ts
proyectoauraorg: Right — `NativeToolCallParser` always tags legacy calls with `_legacyFormat: true`, so the bare-`files` fallback only matters for chat history persisted before that flag was introduced (cc86049f1) and re-hydrated on a later run. I've noted that explicitly in the comment.

src/core/tools/__tests__/readFileTool.spec.ts
proyectoauraorg: Removed both the temporary `console.warn` in `ReadFileTool.ts` and this assertion. The test still proves the legacy branch ran via the `File: <path>` output assertion.

src/core/tools/__tests__/readFileTool.spec.ts
proyectoauraorg: Good catch — `executeLegacy` didn't set `didToolFailInCurrentTurn` on a rooignore block while the native path does. Rather than assert the inconsistent behavior, I aligned the legacy path to set the flag too, and added the assertion.

src/core/tools/ReadFileTool.ts
```diff
@@ -694,6 +691,8 @@ export class ReadFileTool extends BaseTool<"read_file"> {
 				await task.say("rooignore_error", relPath)
 				const errorMsg = formatResponse.rooIgnoreError(relPath)
 				results.push(`File: ${relPath}\nError: ${errorMsg}`)
+				// Mirror the native path: a blocked file marks the tool turn as failed.
+				task.didToolFailInCurrentTurn = true
```
edelauna, line 695: Should the directory-is-a-directory case (~line 733) and the catch block (~line 803) also set `didToolFailInCurrentTurn = true`? The native path sets it in both those failure cases, so the legacy path now matches on rooignore but still diverges for the other two.

packages/types/src/tool-params.ts
```diff
@@ -89,7 +89,12 @@ export type ReadFileToolParams = ReadFileParams | LegacyReadFileParams
  * Type guard to check if params are in legacy format.
  */
 export function isLegacyReadFileParams(params: ReadFileToolParams): params is LegacyReadFileParams {
-	return "_legacyFormat" in params && params._legacyFormat === true
+	// `NativeToolCallParser` always tags freshly parsed legacy calls with `_legacyFormat: true`.
+	// The bare-`files` fallback only matters for chat history persisted before that flag was
+	// introduced (commit cc86049f1) and re-hydrated on a later run.
+	const hasLegacyFlag = "_legacyFormat" in params && params._legacyFormat === true
+	const hasFilesArray = "files" in params && Array.isArray((params as unknown as Record<string, unknown>).files)
+	return hasLegacyFlag || hasFilesArray
```
edelauna, line 99: After narrowing to `params is LegacyReadFileParams`, callers can access `params._legacyFormat` — but hydrated objects that hit the `hasFilesArray` branch won't have it set, so it'll be `undefined` rather than `true`. Worth a note in the comment (or JSDoc) that `_legacyFormat` can't be depended on after this guard?

proyectoauraorg:

src/core/tools/__tests__/readFileTool.spec.ts
```diff
@@ -195,6 +195,15 @@ describe("ReadFileTool", () => {
 		})
 	})
 
+	afterEach(() => {
+		// Restore the console.warn spy that some tests install via vi.spyOn so it
+		// doesn't leak into subsequent tests. Scoped to console to avoid restoring
+		// the module-level mocks (e.g. ImageMemoryTracker) set up for the suite.
+		if (vi.isMockFunction(console.warn)) {
```
edelauna: The comment says this avoids restoring module-level mocks like `ImageMemoryTracker`, but `vi.restoreAllMocks()` only restores `vi.spyOn` spies — it never touches `vi.mock()` factories. Would a plain `vi.restoreAllMocks()` (the pattern in `ReadCommandOutputTool.test.ts:81`) work here instead?

src/core/tools/__tests__/readFileTool.spec.ts
```diff
@@ -730,4 +730,829 @@ describe("ReadFileTool", () => {
...
+
+			await readFileTool.execute({ path: "test.ts" }, mockTask as any, callbacks)
+
+			expect(mockTask.didRejectTool).toBe(false)
+			expect(callbacks.pushToolResult).toHaveBeenCalled()
+		})
+
+		it("should deny single file when user clicks no", async () => {
+			const mockTask = createMockTask()
+			const callbacks = createMockCallbacks()
```
edelauna: The comment inside the first test here concedes "The new format only handles single files". The multi-file branch of `requestApproval` (which handles per-file permissions and the `"individual"` third-response option) doesn't appear to be covered anywhere — is that intentionally deferred?

src/core/tools/__tests__/readFileTool.spec.ts
```diff
@@ -730,4 +730,829 @@ describe("ReadFileTool", () => {
...
+			mockTask.ask.mockResolvedValue({
+				response: "noButtonClicked",
+				text: "Don't read this",
+				images: undefined,
+			})
+
+			await readFileTool.execute({ path: "secret.ts" }, mockTask as any, callbacks)
+
+			expect(formatResponse.toolDeniedWithFeedback).toHaveBeenCalledWith("Don't read this")
+		})
```
edelauna: These tests (deny, feedback-on-denial, feedback-on-approval) look like they cover the same scenarios as the existing `"approval flow"` describe at line 593. Is there something distinct about the "batch" path being exercised here?

src/core/tools/__tests__/readFileTool.spec.ts
```diff
@@ -730,4 +730,829 @@ describe("ReadFileTool", () => {
...
+			await readFileTool.execute({ path: "test.ts", offset: 10, limit: 50 }, mockTask as any, callbacks)
+
+			const askCall = mockTask.ask.mock.calls[0]
+			const message = JSON.parse(askCall[1])
+			expect(message.reason).toContain("lines 10-59")
+			expect(message.startLine).toBe(10)
+		})
+
+		it("should show up to N lines in approval when offset is default", async () => {
+			const mockTask = createMockTask()
```
edelauna: This parses the raw JSON from `mockTask.ask.mock.calls[0][1]` and asserts on `message.reason` — an internal field of the approval payload. Would asserting on the `ask` call args at a higher level (or on `pushToolResult` output) be more resilient to refactoring?

src/core/tools/__tests__/readFileTool.spec.ts
```diff
@@ -730,4 +730,829 @@ describe("ReadFileTool", () => {
...
+
+			await readFileTool.execute({ path: "test.ts" }, mockTask as any, callbacks)
+
+			// Should still work - uses default limits
+			expect(callbacks.pushToolResult).toHaveBeenCalled()
+		})
+
+		it("should use default limits when state has no image config", async () => {
+			const mockTask = createMockTask()
+			const callbacks = createMockCallbacks()
```
edelauna: Both tests here mock `isBinaryFile` to `false`, so the file is read as text and `maxImageFileSize`/`maxTotalImageSize` are never accessed. The names suggest image-limit default behavior — should these use a binary image path to actually exercise that code path?

src/core/tools/__tests__/readFileTool.spec.ts
```diff
@@ -730,4 +739,845 @@ describe("ReadFileTool", () => {
...
+
+			expect(callbacks.pushToolResult).toHaveBeenCalledWith(expect.stringContaining("file1 content"))
+			expect(callbacks.pushToolResult).toHaveBeenCalledWith(expect.stringContaining("file2 content"))
+		})
+
+		it("should block rooignore-protected files in legacy format", async () => {
+			const mockTask = createMockTask({ rooIgnoreAllowed: false })
+			const callbacks = createMockCallbacks()
+
+			vi.spyOn(console, "warn").mockImplementation(() => {})
```
edelauna: `ReadFileTool.ts` no longer has any `console.warn` calls after the removal in this same PR — these 13 spies are suppressing a warning that never fires. Could they be dropped?

src/core/tools/__tests__/readFileTool.spec.ts
```diff
@@ -730,4 +739,845 @@ describe("ReadFileTool", () => {
...
+		})
+	})
+
+	describe("batch approval flow", () => {
+		it("should approve all files in batch when user clicks yes", async () => {
+			const mockTask = createMockTask()
+			const callbacks = createMockCallbacks()
+
+			// The new format only handles single files, but approval uses batch logic internally
+			// when rooignore allows multiple files. Testing the single file yes flow:
```
edelauna: The comment inside the first test here concedes "The new format only handles single files". The multi-file branch of `requestApproval` (which handles per-file permissions and the `"individual"` third-response option) doesn't appear to be covered anywhere — is that intentionally deferred?

src/core/tools/__tests__/readFileTool.spec.ts
```diff
@@ -730,4 +739,845 @@ describe("ReadFileTool", () => {
...
+				includedRanges: [[1, 1]],
+			})
+
+			await readFileTool.execute({ path: "test.ts" }, mockTask as any, callbacks)
+
+			expect(mockTask.didRejectTool).toBe(false)
+			expect(callbacks.pushToolResult).toHaveBeenCalled()
+		})
+
+		it("should deny single file when user clicks no", async () => {
```
edelauna: These tests (deny, feedback-on-denial, feedback-on-approval) look like they cover the same scenarios as the existing `"approval flow"` describe at line 593. Is there something distinct about the "batch" path being exercised here?

edelauna (commented):
Thanks for addressing the previous feedback - could more comments to minimize the diff.

proyectoauraorg:
Thanks for the thorough review! Slimmed the diff and addressed the inline notes:

- Dropped the 13 dead `console.warn` spies + the now-unused `afterEach` that restored them (the source no longer emits that debug warn).
- Removed the "batch approval flow" block — those four cases all drove the single-file `{ path }` path, duplicating "approval flow" without exercising a real multi-file batch (the new format is single-file). Multi-file stays covered via the legacy-format tests.
- Made the legacy read path mirror the native one: `didToolFailInCurrentTurn` is now set on rooignore blocks, directory reads, and read errors (with assertions).
- Added the note that the bare-`files` fallback in `isLegacyReadFileParams` leaves `_legacyFormat` undefined.

Squashed to one commit. `tsc`, `eslint`, and `readFileTool.spec.ts` (71 tests) pass locally.

edelauna (approved):

PR #230, opened 2026-05-21
fix(diff): repair truncated Grok diffs with missing markers (#186)

### LINKED ISSUE #186: [BUG]  Grok 4.3 "Unable to apply diff" - Expected '=======' was not found.
**Problem (one or two sentences)**  

Grok 4.3 frequently generates malformed or truncated diffs that fail to apply in Zoo Code, producing the error: "Unable to apply diff to file: ... ERROR: Unexpected end of sequence: Expected '=======' was not found."

**Context (who is affected and when)**  
Affects users editing files with Grok 4.3 in Zoo Code's diff/edit mode. Happens during normal code editing tasks.

**Reproduction steps**  
1. Use Grok 4.3 (or similar Grok coder variant) in Zoo Code.  
2. Ask it to make changes to a file via diff.  
3. The model outputs an incomplete diff missing the expected `=======` separator.  
4. Zoo Code's `apply_diff` tool fails with the exact error shown in the attached screenshot.

**Expected result**  
The diff should apply cleanly, or Zoo Code should gracefully handle/fallback on malformed output from the model.

**Actual result**  
Edit fails with "Unable to apply diff" and the sequence error.

**Variations tried (optional)**  
- Retrying the same prompt (sometimes succeeds, often fails again).  
- More explicit prompts asking for clean unified diffs.

**App Version**  
v3.54.1

**API Provider**  
xAI / Grok

**Model Used**  
Grok 4.3

**Zoo Code Task Links (optional)**  
(none)

**Relevant logs or errors (optional)**  
```
Unable to apply diff to file: /home/user/dev/ipall/apps/backup/ipfs-backup/lib/common.sh

<error_details>
ERROR: Unexpected end of sequence: Expected '=======' was not found.
</error_details>
```

Image

proyectoauraorg:
## Summary

Fixes #186. Grok (4.3) frequently truncates streamed diffs mid-output, so a `<<<<<<< SEARCH` block can arrive without its `=======` separator and/or `>>>>>>> REPLACE` closer. The parser then rejects the whole edit with *"Expected '=======' was not found"* and the user sees **"Unable to apply diff"**.

This adds a small recovery layer, `repairTruncatedDiff()`, called at the start of `applyDiff()` that reconstructs the missing markers so a salvageable diff can still be applied.

## How it works

For each `<<<<<<< SEARCH` block the repair step:
- leaves **complete** blocks (both `=======` and `>>>>>>> REPLACE`) untouched, verbatim;
- if only `>>>>>>> REPLACE` is missing, appends it;
- if both markers are missing, treats the first line after `SEARCH` as the search content and the rest as the replacement, then inserts `=======` and `>>>>>>> REPLACE`;
- preserves escaped markers via `(?<!\\)` look-behinds (won't touch `\<<<<<<<` content);
- when a repaired block is followed by more blocks, re-adds an inter-block separator so the appended `>>>>>>> REPLACE` never gets glued to the next `<<<<<<< SEARCH`.

Complete, valid diffs pass through unchanged, so behavior for well-formed input is identical.

## Testing

Added coverage in `multi-search-replace.spec.ts` (complete-diff passthrough, missing closer, missing both markers, empty-search edge case, multi-block truncation, no spurious trailing newline).

```
vitest run core/diff/strategies/__tests__/multi-search-replace.spec.ts
# Test Files 1 passed (1) · Tests 65 passed (65)
```

## Files
- `src/core/diff/strategies/multi-search-replace.ts` — `repairTruncatedDiff()` + call in `applyDiff()`
- `src/core/diff/strategies/__tests__/multi-search-replace.spec.ts` — tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of truncated merge-style diffs: missing separators and end markers are detected and synthesized so truncated outputs are repaired without changing well-formed diffs. Preserves directive/header lines, handles empty search blocks, maintains trailing-newline behavior, and limits repair to the first truncated block when later blocks are complete.

* **Tests**
  * Added targeted and regression tests covering multiple truncation scenarios and edge cases to prevent regressions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/230?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/core/diff/strategies/multi-search-replace.ts
+++ b/src/core/diff/strategies/multi-search-replace.ts
@@ -295,10 +295,28 @@ export class MultiSearchReplaceDiffStrategy implements DiffStrategy {
 				// Has ======= but missing >>>>>>> REPLACE — append closing marker
 				const body = block.replace(/\s+$/, "")
 				repaired += body + "\n>>>>>>> REPLACE" + separator
+			} else if (hasCloser && !hasSeparator) {
+				// Has >>>>>>> REPLACE but missing the ======= separator. Don't synthesize a
+				// second closer; splice the separator in right before the existing closer so
+				// everything above it becomes the SEARCH section.
+				const body = block.replace(/\s+$/, "")
+				repaired += body.replace(/(\n)(>>>>>>> REPLACE)(?=\n|$)/, "$1=======\n$2") + separator
 			} else {
-				// Missing both ======= and >>>>>>> REPLACE
+				// Missing both ======= and >>>>>>> REPLACE.
 				const searchMatch = block.match(/^<<<<<<< SEARCH\n?([\s\S]*)$/)
-				const content = (searchMatch?.[1] ?? "").replace(/\s+$/, "")
+				let content = (searchMatch?.[1] ?? "").replace(/\s+$/, "")
+
+				// Peel off any leading Grok header directives (:start_line:, :end_line:, -------)
+				// so the "first line is SEARCH" heuristic sees real content, not metadata. The
+				// directives are preserved as a header on the SEARCH section.
+				let header = ""
+				const directiveLine = /^(?::start_line:\s*\d+|:end_line:\s*\d+|-------)\s*$/
+				let nlIdx: number
+				while ((nlIdx = content.indexOf("\n")) !== -1 && directiveLine.test(content.slice(0, nlIdx))) {
+					header += content.slice(0, nlIdx + 1)
+					content = content.slice(nlIdx + 1)
+				}
+
 				const firstNewlineIdx = content.indexOf("\n")
 				if (firstNewlineIdx !== -1) {
 					// First line is SEARCH content, rest is REPLACE content
@@ -306,11 +324,16 @@ export class MultiSearchReplaceDiffStrategy implements DiffStrategy {
 					const replaceContent = content.substring(firstNewlineIdx + 1)
 					repaired +=
 						"<<<<<<< SEARCH\n" +
+						header +
 						searchContent +
 						"\n=======\n" +
 						replaceContent +
 						"\n>>>>>>> REPLACE" +
 						separator
+				} else if (header) {
+					// Only a directive header plus a single content line: that line is the SEARCH
+					// target (the user pinned it with start_line); the REPLACE section is empty.
+					repaired += "<<<<<<< SEARCH\n" + header + content + "\n=======\n\n>>>>>>> REPLACE" + separator
 				} else {
 					// Single line — treat as empty SEARCH with content as REPLACE
 					repaired += "<<<<<<< SEARCH\n=======\n" + content + "\n>>>>>>> REPLACE" + separator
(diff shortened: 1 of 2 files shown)


edelauna (changes requested):
Can you add a fixture based e2e test which reproduces the issue / acts as a regression guard.

proyectoauraorg:
Thanks @edelauna — added fixture-based regression tests that run through the full `applyDiff()` path with realistic truncated-Grok diffs: one missing the closing `>>>>>>> REPLACE`, one truncated before `=======` (the exact "Expected '=======' was not found" case), and a well-formed multi-block diff that must pass through untouched. All pass; pushed in the latest commit.

proyectoauraorg:
Thanks @coderabbitai — both addressed in the latest commit:

1. `applyDiff` now uses a local `const repairedDiff = this.repairTruncatedDiff(diffContent)` and reads from it downstream, leaving the original `diffContent` parameter untouched.
2. Repair edge cases:
   - A block with `>>>>>>> REPLACE` but no `=======` now splices the separator in **before** the existing closer (no duplicate closer).
   - Leading Grok directives (`:start_line:`, `:end_line:`, `-------`) are peeled off before the first-line-is-SEARCH heuristic and preserved on the SEARCH section, so metadata is never treated as content.

Added unit tests for both. 71 tests pass.

src/core/diff/strategies/multi-search-replace.ts
```diff
@@ -242,13 +242,119 @@ export class MultiSearchReplaceDiffStrategy implements DiffStrategy {
...
+		// Only repair if the diff has at least one SEARCH marker
+		if (!/(?<!\\)<<<<<<< SEARCH/.test(diffContent)) {
+			return diffContent
+		}
+
+		// Split into blocks based on SEARCH markers
+		const blocks = diffContent.split(/(?=(?<!\\)<<<<<<< SEARCH)/)
+
+		let repaired = ""
+		let needsRepair = false
```
edelauna: `needsRepair` is assigned on line 290 but never read — safe to remove?

edelauna (commented):
Thanks for addressing the feedback regarding the e2e.

proyectoauraorg:
Thanks for the review @edelauna! 🙏

I've addressed your feedback in commit `2e9ea04fe`:

**Removed unused `needsRepair` variable** — The variable was assigned but never read (dead store). Since it served no functional purpose in the repair loop, I've removed it entirely.

This was the only actionable item from code review #3284681514. The other feedback about test indentation and the merge conflict comment were not issues (the indentation is correct nested ternary style, and there are no merge conflict markers).

Checks are running now — should be green shortly. Ready for re-review when you get a chance!

proyectoauraorg:
Thanks @edelauna! Both addressed:
- the fixture-based e2e regression test is in (per your earlier note), and
- I removed the unused `needsRepair` in 2e9ea04 per your inline comment.

Ready for another look when you have a moment. 🙏

edelauna (approved):
🔥

PR #239, opened 2026-05-21
fix(shell): report PowerShell on Windows when no profile is configured (#82)

### LINKED ISSUE #82: [BUG] System prompt on windows says shell is cmd.exe but if powershell is used via VS Code terminal provider when inline terminal is off
## Problem

On Windows, the system prompt reports the shell as cmd.exe (and generates cmd.exe-appropriate command chaining hints), but when using the VS Code terminal provider, commands actually execute in the user's configured default shell (typically PowerShell). This causes the Zoo to generate commands incompatible with the actual runtime shell.

Affects all Windows users who have inline terminal off. When Inline terminal is unticked it uses VS Code default terminal profile, typically PowerShell or pwsh.

## Reproduction steps

1. On Windows, ensure VS Code default terminal profile is PowerShell (the default): `terminal.integrated.defaultProfile.windows` is set to a PowerShell profile.
2. Go to Zoo Settings, Terminal Settings and Untick "Use Inline Terminal" .
3. Start a Zoo Code task and ask it to run a command.
4. Observe:
   - The system prompt will contain `Default Shell: C:\Windows\System32\cmd.exe` in the SYSTEM INFORMATION section.
   - The rules section will include cmd.exe-specific notes like "Using `&&` for cmd.exe command chaining".
   - The actual command will execute in the VS Code integrated terminal, which runs PowerShell.

## Expected result

The system prompt should report the same shell that will actually execute commands. When the VS Code terminal provider is active, the shell reported should match the user's VS Code terminal profile. When the Execa provider is active, it should report the Execa shell (which may differ).

## Actual result

The system prompt always reports the shell determined by `getShell()` in `src/utils/shell.ts`, which independently reads VS Code terminal config but falls back to `cmd.exe`. Meanwhile, the VS Code terminal provider executes commands in whatever shell VS Code actually uses (typically PowerShell). The two code paths do not coordinate.

### The prompt-side shell detection

The system prompt is built in `src/core/prompts/sections/system-info.ts` and calls `getShell()` from `src/utils/shell.ts`:

https://github.com/Zoo-Code-Org/Zoo-Code/blob/d191fe055dad594b8f19ce23813b747253143b70/src/core/prompts/sections/system-info.ts#L23

The `getShell()` function in `src/utils/shell.ts` reads VS Code's `terminal.integrated.defaultProfile.windows` config to determine the shell:

https://github.com/Zoo-Code-Org/Zoo-Code/blob/d191fe055dad594b8f19ce23813b747253143b70/src/utils/shell.ts#L196-L213

When `defaultProfileName` is not set (null), the fallback chain eventually reaches `COMSPEC` or `cmd.exe`:

https://github.com/Zoo-Code-Org/Zoo-Code/blob/d191fe055dad594b8f19ce23813b747253143b70/src/utils/shell.ts#L355-L359

This same `getShell()` function is also used to determine command chaining operators and shell-specific notes in the rules section:

https://github.com/Zoo-Code-Org/Zoo-Code/blob/d191fe055dad594b8f19ce23813b747253143b70/src/core/prompts/sections/rules.ts#L12-L28

### The execution-side shell selection

The actual terminal provider is selected in `src/core/tools/ExecuteCommandTool.ts` based on `terminalShellIntegrationDisabled`:

https://github.com/Zoo-Code-Org/Zoo-Code/blob/d191fe055dad594b8f19ce23813b747253143b70/src/core/tools/ExecuteCommandTool.ts#L202

When using the VS Code terminal provider (`src/integrations/terminal/Terminal.ts`), the shell is whatever VS Code's integrated terminal uses. The extension does not control this; VS Code uses the user's configured default profile.

When using the Execa provider (`src/integrations/terminal/ExecaTerminalProcess.ts`), the shell defaults to `true` (which means cmd.exe on Windows) unless an explicit `execaShellPath` is set:

https://github.com/Zoo-Code-Org/Zoo-Code/blob/d191fe055dad594b8f19ce23813b747253143b70/src/integrations/terminal/ExecaTerminalProcess.ts#L43

### Cause

The root cause is that `getShell()` does not account for which terminal provider will actually be used at execution time, and does not handle the case where VS Code's `defaultProfile.windows` is not explicitly set but the effective default is still PowerShell. VS Code defaults to PowerShell on Windows without writing it to settings, so the config lookup returns null, causing the fallback to cmd.exe.

## Possible fix

1. Use unified logic for `getShell()` so the system prompt is aware of the terminal provider. `getShell()` should check `terminalShellIntegrationDisabled` setting and use different detection logic:
   - For the VS Code terminal provider: detect the shell VS Code will actually use. Consider using terminal profile detection logic that accounts for VS Code's built-in defaults (PowerShell on Windows).
   - For the Execa provider: use `getExecaShellPath()` if set, otherwise report `cmd.exe` on Windows.

2. Fix `getWindowsShellFromVSCode()` to handle the case where `defaultProfileName` is null. VS Code's effective default on Windows is PowerShell, so the fallback should not immediately jump to `cmd.exe`. Consider returning PowerShell as the default when no profile name is configured on Windows.

## Environment

- OS: Windows 11
- App version: latest

## Related issues

https://github.com/RooCodeInc/Roo-Code/issues/11958
https://github.com/RooCodeInc/Roo-Code/issues/8530

proyectoauraorg:
## Summary

Fixes #82. On Windows the system prompt told the model the shell was **cmd.exe**, but the integrated terminal actually runs **PowerShell** — VS Code's default on modern Windows. The two shell-detection paths disagreed.

`getShell()` (used by the system prompt in `system-info.ts` and shell-specific `rules.ts`) called `getWindowsShellFromVSCode()`, which returned `null` when no explicit `terminal.integrated.defaultProfile.windows` was set. `getShell()` then fell through to `COMSPEC` (cmd.exe), which is **not** what VS Code launches.

## Fix

When no Windows terminal profile is configured, `getWindowsShellFromVSCode()` now returns Windows PowerShell (VS Code's built-in default) instead of `null`. Windows PowerShell is always present and allowlisted, so it's a safe default.

Explicitly configured profiles are unchanged: PowerShell 7 / legacy / WSL / custom paths and an explicit **Command Prompt** profile all still resolve as before.

## Testing

```
vitest run utils/__tests__/shell.spec.ts core/prompts/sections/__tests__/system-info.spec.ts
# Test Files 2 passed (2) · Tests 40 passed (40)
```
Updated the no-profile Windows tests to assert PowerShell, kept allowlist-safety coverage (non-allowlisted profile path → safe fallback), and added a test confirming an explicit Command Prompt profile still yields cmd.exe.

## Summary by CodeRabbit

* **Bug Fixes**
  * Windows shell detection now prefers PowerShell 7 when present, otherwise falls back to legacy Windows PowerShell; downstream fallbacks are bypassed.
  * Non-allowlisted terminal profile paths now fall back to Command Prompt; explicitly configured Command Prompt is respected.

* **Tests**
  * Updated Windows shell detection tests to reflect the new defaulting and fallback behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/239?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/utils/__tests__/shell.spec.ts
+++ b/src/utils/__tests__/shell.spec.ts
@@ -173,23 +173,27 @@ describe("Shell Detection Tests", () => {
 			expect(getShell()).toBe("C:\\Windows\\System32\\cmd.exe")
 		})
 
-		it("respects userInfo() if no VS Code config is available and shell is allowed", () => {
+		it("defaults to Windows PowerShell when VS Code has no configured profile", () => {
+			// Modern VS Code launches PowerShell by default on Windows (issue #82), so
+			// getShell() should report PowerShell rather than falling through to cmd.exe.
 			vscode.workspace.getConfiguration = () => ({ get: () => undefined }) as any
 			vi.mocked(userInfo).mockReturnValue({ shell: "C:\\Program Files\\PowerShell\\7\\pwsh.exe" } as any)
 
-			expect(getShell()).toBe("C:\\Program Files\\PowerShell\\7\\pwsh.exe")
+			expect(getShell()).toBe("C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe")
 		})
 
-		it("falls back to safe shell when userInfo() returns non-allowlisted shell", () => {
-			vscode.workspace.getConfiguration = () => ({ get: () => undefined }) as any
-			vi.mocked(userInfo).mockReturnValue({ shell: "C:\\Custom\\PowerShell.exe" } as any)
+		it("falls back to safe shell when the configured profile path is non-allowlisted", () => {
+			mockVsCodeConfig("windows", "Custom", {
+				Custom: { path: "C:\\Custom\\evil.exe" },
+			})
 
 			expect(getShell()).toBe("C:\\Windows\\System32\\cmd.exe")
 		})
 
-		it("falls back to safe shell when COMSPEC is non-allowlisted", () => {
-			vscode.workspace.getConfiguration = () => ({ get: () => undefined }) as any
-			process.env.COMSPEC = "D:\\CustomCmd\\cmd.exe"
+		it("uses cmd.exe when a Command Prompt profile is explicitly configured", () => {
+			mockVsCodeConfig("windows", "Command Prompt", {
+				"Command Prompt": { path: "C:\\Windows\\System32\\cmd.exe" },
+			})
 
 			expect(getShell()).toBe("C:\\Windows\\System32\\cmd.exe")
 		})
--- a/src/utils/__tests__/shell.spec.ts
+++ b/src/utils/__tests__/shell.spec.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest"
 import * as vscode from "vscode"
+import { existsSync } from "fs"
 import { userInfo } from "os"
 import { getShell } from "../shell"
 
@@ -13,6 +14,11 @@ vi.mock("vscode", () => ({
 // Mock the os module
 vi.mock("os", () => ({
 	userInfo: vi.fn(() => ({ shell: null })),
+}))
+
+// Mock the fs module — getWindowsShellFromVSCode probes for PowerShell 7 (pwsh.exe).
+vi.mock("fs", () => ({
+	existsSync: vi.fn(() => false),
 }))
 
 // Mock path module for testing
@@ -57,6 +63,8 @@ describe("Shell Detection Tests", () => {
 
 		// Reset userInfo mock to default
 		vi.mocked(userInfo).mockReturnValue({ shell: null } as any)
+		// Default: PowerShell 7 is not installed, so the probe falls back to legacy.
+		vi.mocked(existsSync).mockReturnValue(false)
 	})
 
 	afterEach(() => {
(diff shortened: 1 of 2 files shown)


src/utils/shell.ts
```diff
@@ -188,7 +188,13 @@ function normalizeShellPath(path: string | string[] | undefined): string | null
...
 	const { defaultProfileName, profiles } = getWindowsTerminalConfig()
 	if (!defaultProfileName) {
-		return null
+		// No explicit Windows terminal profile is configured. VS Code's built-in
+		// default on modern Windows is PowerShell (not cmd.exe), and that is the
+		// shell the integrated terminal actually launches. Mirror it here so the
+		// system prompt advertises the real shell instead of falling through to
+		// COMSPEC (cmd.exe). Windows PowerShell is always present, so it is a safe
+		// allowlisted default. See issue #82.
+		return SHELL_PATHS.POWERSHELL_LEGACY
```
edelauna: On a machine with PowerShell 7 installed, VS Code's auto-detection will typically select PS7 as the default terminal — not legacy 5.1. Commit `1a2c45cb` fixed the analogous case (profile name includes "powershell" but no path/source) by switching from `POWERSHELL_LEGACY` → `POWERSHELL_7` for exactly this reason (UTF-8 mojibake). Should this path apply the same heuristic — e.g. probe for PS7 at its known path before falling back to legacy?

src/utils/__tests__/shell.spec.ts
```diff
@@ -173,23 +173,27 @@ describe("Shell Detection Tests", () => {
 			expect(getShell()).toBe("C:\\Windows\\System32\\cmd.exe")
 		})
 
-		it("respects userInfo() if no VS Code config is available and shell is allowed", () => {
+		it("defaults to Windows PowerShell when VS Code has no configured profile", () => {
+			// Modern VS Code launches PowerShell by default on Windows (issue #82), so
+			// getShell() should report PowerShell rather than falling through to cmd.exe.
 			vscode.workspace.getConfiguration = () => ({ get: () => undefined }) as any
 			vi.mocked(userInfo).mockReturnValue({ shell: "C:\\Program Files\\PowerShell\\7\\pwsh.exe" } as any)
```
edelauna: Is this mock still needed? With the early return at `shell.ts:197`, `getShellFromUserInfo()` is never called in this code path — so this mock value is never read. Removing it would make the test clearer about what's actually being exercised.

```suggestion
```

src/utils/shell.ts
proyectoauraorg: Implemented the probe — the no-profile path now returns PS7 (`pwsh.exe`) when it exists on disk, otherwise legacy 5.1, mirroring VS Code's auto-detection (which prefers PS7 when installed).

src/utils/__tests__/shell.spec.ts
proyectoauraorg: Removed the unused `userInfo` mock and split the no-profile test into PS7-present (→ `pwsh.exe`) and PS7-absent (→ legacy 5.1) cases, mocking `fs.existsSync`.

edelauna (approved):
Thanks for addressing this, glad to see the shell integration being improved upon.

Issue #242, opened 2026-05-21
[BUG] Custom temperature disabled but sending as 0


brunocasado:
**Describe the bug**
If custom temperature is disabled, it's sending 0 to model

**To Reproduce**
1. Go to 'settings'
2. Click on 'providers'
3. Scroll down to 'advanced settings'
4. do not mark use custom temperature

**Expected behavior**
Shouldn't send the temperature since models can be configured with a default temp

**Screenshots**

<img width="900" height="707" alt="Image" src="https://github.com/user-attachments/assets/039f8490-8a03-4fcf-b02d-6718caf0de1b" />

**What version of zoo are you running**
i.e. 3.54.1

proyectoauraorg:
Looking into this — the behavior isn't isolated to one provider. The pattern `this.options.modelTemperature ?? <default>` is repeated across ~13 providers (openai, openai-native, lite-llm, mistral, lm-studio, base-openai-compatible, …), so when "use custom temperature" is off, each one falls back to its own default (often `0`) and still sends `temperature`.

A robust fix probably belongs in a shared layer rather than per-provider, and it hinges on a design decision: when the user hasn't set a custom temperature, should we (a) omit the `temperature` field so the model uses its server-side default, or (b) keep sending a provider default for backward compatibility? Option (a) matches the expectation here but changes long-standing behavior, so confirming the intended direction first would help.

Happy to help with a PR once the preferred behavior is decided. (Related: #215 / PR #233 added a model-capability gate `supportsTemperature` for OpenAI-Compatible — this issue is the orthogonal user-preference case.)

edelauna:
I think option A would be preferred. What’s the risk though, are there any providers which make `temperature` required?

proyectoauraorg:
No provider hard-requires it — `temperature` is optional on the OpenAI-compatible APIs (a server-side default applies), and the codebase already omits it conditionally in a few spots without issues (DeepSeek thinking models, the `supportsTemperature` gate in vercel-ai-gateway, and #233 for OpenAI-Compatible). So option A shouldn't cause 4xx errors.

The risk is behavioral rather than an error: today, unchecking "use custom temperature" still sends each provider's default (`0` for OpenAI-compatible, `1.0` for MiniMax, `0.3` for DeepSeek…). Under option A those requests omit it and the server applies its own default — which can differ — so users relying on the implicit default could see different outputs.

[...]

PR #247, opened 2026-05-22
fix(openai): omit temperature when no custom value is set (#242)

proyectoauraorg:
## Summary

When "use custom temperature" is off, the OpenAI-Compatible provider still sent `temperature: 0` (the fallback), so the model's server-side default never applied. Per the discussion on #242 (option A, which @edelauna preferred), this omits the `temperature` field in that case.

## Behavior

`temperature` is now included only when **either** the user set a custom temperature **or** the model needs a specific default (e.g. `deepseek-reasoner`); otherwise it is omitted.

- Preserves the `supportsTemperature` capability gate from #233 (still omitted when `supportsTemperature === false`).
- Preserves model-required defaults (`deepseek-reasoner` → `DEEP_SEEK_DEFAULT_TEMPERATURE`).
- A deliberately-set `0` is distinguished from "unset" and is still sent.
- Behavioral note: OpenAI-Compatible users who never set a temperature now get the model's server-side default instead of `0`.

## Scope

Targets the OpenAI-Compatible provider (the case reported in #242). The same `?? <default>` pattern exists in the shared `getModelParams` and a few other providers (`lite-llm`, `openai-native`, `lm-studio`, `vercel-ai-gateway`, …); happy to converge them on this behavior in a follow-up — `poe` and `moonshot` already omit, so there is precedent.

## Tests

`api/providers/__tests__/openai.spec.ts`: omitted by default, custom value sent, deliberate `0` sent, `deepseek-reasoner` default preserved, `supportsTemperature === false` still omitted. 53 pass.

Refs #242

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved temperature parameter handling in API requests. Temperature is now conditionally omitted when unsupported by the model or when not explicitly provided, while respecting explicit user-set values including zero.

* **Tests**
  * Updated test expectations to reflect refined temperature behavior across different AI service implementations.

### DIFF
--- a/src/api/providers/__tests__/openai.spec.ts
+++ b/src/api/providers/__tests__/openai.spec.ts
@@ -409,13 +409,15 @@ describe("OpenAiHandler", () => {
 			expect(callArgs).not.toHaveProperty("temperature")
 		})
 
-		it("should include temperature by default when supportsTemperature is not set", async () => {
+		it("should omit temperature by default when no custom temperature is set", async () => {
+			// Option A: when "use custom temperature" is off (modelTemperature unset) and the model has no
+			// required default, omit `temperature` so the server's own default applies instead of forcing 0.
 			const stream = handler.createMessage(systemPrompt, messages)
 			for await (const _chunk of stream) {
 			}
 			expect(mockCreate).toHaveBeenCalled()
 			const callArgs = mockCreate.mock.calls[0][0]
-			expect(callArgs).toHaveProperty("temperature")
+			expect(callArgs).not.toHaveProperty("temperature")
 		})
 
 		it("should use the configured modelTemperature when supportsTemperature is not false", async () => {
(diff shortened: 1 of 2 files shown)


.changeset/omit-temperature-when-unset.md
edelauna, line 1: 😅 i haven't been too strict about including changesets, can provide some rationale for providing this as part of this PR? I've kind of been thinking we'll update this as part of our release process, since it requires localization.

src/api/providers/__tests__/openai.spec.ts
```diff
@@ -679,7 +692,7 @@ describe("OpenAiHandler", () => {
 					],
 					stream: true,
 					stream_options: { include_usage: true },
-					temperature: 0,
+					// #242 (option A): no custom temperature set → `temperature` is omitted.
```
edelauna: I think it's ok not to document the specific issue in code, it'll be handled in commit history.

edelauna (commented):
Thanks for addressing this, just had some nits around documentation and comments.

Mirrowel:
Was about to implement this, but someone got to it. Looking forward to this getting merged.

proyectoauraorg:

proyectoauraorg:
Thanks! Rebased onto `main` (resolved the `openai.ts` conflict — kept your updated example list incl. `claude-opus-4-8`) and addressed both nits:

- **Changeset**: removed it, since it sounds like changesets are handled at release time given the localization step. Happy to re-add one with rationale if you'd prefer.
- **Issue refs in code**: dropped the `#242`/option-A citations from the test names and the inline comment, leaving the descriptive text (the issue stays in the commit/PR for history).

`tsc`, `eslint`, and `openai.spec.ts` (53 tests) pass locally.

proyectoauraorg:
👋 Friendly ping — this PR has been open for 10 days.

**Summary:** Omits the deprecated `temperature` parameter when no custom value is set, fixing [Claude Opus 4.7/4.8 failures](https://github.com/Zoo-Code-Org/Zoo-Code/issues/432).

- ✅ CI green
- ✅ Tests included
- ✅ Small, focused change

edelauna (approved):
Thanks - created issue to track some fixes that would be good to address which were revealed during review: https://github.com/Zoo-Code-Org/Zoo-Code/issues/454

PR #248, opened 2026-05-22
fix: resolve ripgrep from @vscode/ripgrep-universal and the system PATH

0xMink:
## Summary

`search_files` and `list_files` fail with `Could not find ripgrep binary` on VS Code builds that ship ripgrep via the `@vscode/ripgrep-universal` package layout (binary under `bin/<platform>-<arch>/<rg>` rather than directly in `bin/`). VS Code Insiders' staged-install builds use this layout; current stable is in mid-migration and varies by build.

`getBinPath` now probes both layouts under `vscode.env.appRoot`:

- the long-standing `@vscode/ripgrep` / `vscode-ripgrep` layouts (and their `.asar.unpacked` variants), and
- the newer `@vscode/ripgrep-universal/bin/<platform>-<arch>/` layout (and its `.asar.unpacked` variant).

All three `getBinPath` callers (`search_files`, `list_files`, `file-search`) inherit the fix with no changes. No system PATH fallback, no bundled binary in the VSIX.

## Empirical verification

A diagnostic test build was run on a Windows VS Code stable `1.121.0` install reproducing the bug (see the review-thread discussion of `naiwis.vsix`). The diagnostic confirmed:

- `require("@vscode/ripgrep")` threw — VS Code's extHost interceptor aliases the require to `@vscode/ripgrep-universal`, but that target package isn't installed on builds that haven't completed VS Code's package-rename migration.
- The path probe found `node_modules/@vscode/ripgrep/bin/rg.exe` (the first candidate, classic layout) and `search_files` worked.

So the path-probe is the load-bearing mechanism for the current VS Code migration window.

## Out of scope

- **The `require("@vscode/ripgrep")` interceptor approach** — elegant on paper, but the diagnostic showed it doesn't work on builds that still ship the old package name (current stable). Will be tracked separately and revisited once VS Code completes the rename across stable + Insiders, with empirical evidence of which mechanism extensions are expected to use.
- **Non-fatal handling when ripgrep is truly absent** — the throw paths in `regexSearchFiles`, `getRipgrepPath`, and `executeRipgrep` are unchanged.
- **Ripgrep resolution diagnostic command** — separate follow-up PR.

## Test plan

- [x] `getBinPath` tests: classic `@vscode/ripgrep` layout, `@vscode/ripgrep-universal` layout, `undefined` when ripgrep isn't found.
- [x] `tsc --noEmit` clean.
- [x] `eslint . --ext=ts --max-warnings=0` (the repo lint command) passes.
- [x] Verified on Windows VS Code stable 1.121.0 via the diagnostic test build (path probe found ripgrep, search worked).

## Summary by CodeRabbit

* **Tests**
  * Added tests covering ripgrep binary path resolution across platform/arch variants, including cases for classic, universal layouts and when no binary is found; file-existence checks are mocked to exercise each scenario.

* **Refactor**
  * Improved ripgrep binary discovery to support both classic and newer VS Code package structures (including platform-arch nested layouts).

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/248?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/services/ripgrep/__tests__/index.spec.ts
+++ b/src/services/ripgrep/__tests__/index.spec.ts
@@ -1,19 +1,16 @@
 // npx vitest run src/services/ripgrep/__tests__/index.spec.ts
 
+import path from "path"
 import { vi, describe, it, expect, beforeEach } from "vitest"
 
 import { truncateLine, getBinPath } from "../index"
+import { fileExistsAtPath } from "../../../utils/fs"
 
-const ripgrepMock = vi.hoisted(() => ({ value: undefined as string | undefined, throws: false }))
-
-vi.mock("@vscode/ripgrep", () => ({
-	get rgPath() {
-		if (ripgrepMock.throws) {
-			throw new Error("simulated @vscode/ripgrep failure")
-		}
-		return ripgrepMock.value
-	},
+vi.mock("../../../utils/fs", () => ({
+	fileExistsAtPath: vi.fn(),
 }))
+
+const mockFileExists = vi.mocked(fileExistsAtPath)
 
 describe("Ripgrep line truncation", () => {
 	// The default MAX_LINE_LENGTH is 500 in the implementation
@@ -63,34 +60,32 @@ describe("Ripgrep line truncation", () => {
 })
 
 describe("getBinPath", () => {
+	const appRoot = "/fake/vscode/appRoot"
+	const binName = process.platform.startsWith("win") ? "rg.exe" : "rg"
+	const platformDir = `${process.platform}-${process.arch}`
+
 	beforeEach(() => {
-		ripgrepMock.value = undefined
-		ripgrepMock.throws = false
+		mockFileExists.mockReset()
+		mockFileExists.mockResolvedValue(false)
 	})
 
-	it("returns the rgPath exported by @vscode/ripgrep", async () => {
-		ripgrepMock.value = "/path/to/rg"
+	it("resolves ripgrep from the classic @vscode/ripgrep layout", async () => {
+		const rg = path.join(appRoot, "node_modules/@vscode/ripgrep/bin", binName)
+		mockFileExists.mockImplementation(async (p: string) => p === rg)
 
-		expect(await getBinPath("/ignored")).toBe("/path/to/rg")
+		expect(await getBinPath(appRoot)).toBe(rg)
 	})
 
-	it("rewrites node_modules.asar to node_modules.asar.unpacked", async () => {
-		ripgrepMock.value = "/app/node_modules.asar/@vscode/ripgrep-universal/bin/win32-x64/rg.exe"
+	it("resolves ripgrep from the @vscode/ripgrep-universal layout (VS Code Insiders)", async () => {
+		const rg = path.join(appRoot, "node_modules/@vscode/ripgrep-universal/bin", platformDir, binName)
+		mockFileExists.mockImplementation(async (p: string) => p === rg)
 
-		expect(await getBinPath("/ignored")).toBe(
-			"/app/node_modules.asar.unpacked/@vscode/ripgrep-universal/bin/win32-x64/rg.exe",
-		)
+		expect(await getBinPath(appRoot)).toBe(rg)
 	})
 
-	it("returns undefined when rgPath is not exported", async () => {
-		ripgrepMock.value = undefined
+	it("returns undefined when ripgrep cannot be found", async () => {
+		mockFileExists.mockResolvedValue(false)
 
-		expect(await getBinPath("/ignored")).toBeUndefined()
-	})
-
-	it("returns undefined when rgPath resolution throws", async () => {
-		ripgrepMock.throws = true
-
-		expect(await getBinPath("/ignored")).toBeUndefined()
+		expect(await getBinPath(appRoot)).toBeUndefined()
 	})
 })
--- a/src/services/ripgrep/__tests__/index.spec.ts
+++ b/src/services/ripgrep/__tests__/index.spec.ts
@@ -1,6 +1,16 @@
 // npx vitest run src/services/ripgrep/__tests__/index.spec.ts
 
-import { truncateLine } from "../index"
+import path from "path"
+import { vi, describe, it, expect, beforeEach, afterEach } from "vitest"
+
+import { truncateLine, getBinPath } from "../index"
+import { fileExistsAtPath } from "../../../utils/fs"
+
+vi.mock("../../../utils/fs", () => ({
+	fileExistsAtPath: vi.fn(),
+}))
+
+const mockFileExists = vi.mocked(fileExistsAtPath)
 
 describe("Ripgrep line truncation", () => {
 	// The default MAX_LINE_LENGTH is 500 in the implementation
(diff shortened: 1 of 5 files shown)


edelauna:
Thanks so much for investigating and looking into this. 

I'm going to close this for now. As discussed in dms on Discord I think this is a bit of over engineering to handle what amounts to an edge case. My preference would be to continue to use the Vscode packaged `rg`, and maybe update how that gets resolved, or add a non fatal fallback route.

src/services/ripgrep/index.ts
```diff
@@ -92,7 +137,10 @@ export async function getBinPath(vscodeAppRoot: string): Promise<string | undefi
 		(await checkPath("node_modules/@vscode/ripgrep/bin/")) ||
 		(await checkPath("node_modules/vscode-ripgrep/bin")) ||
 		(await checkPath("node_modules.asar.unpacked/vscode-ripgrep/bin/")) ||
-		(await checkPath("node_modules.asar.unpacked/@vscode/ripgrep/bin/"))
+		(await checkPath("node_modules.asar.unpacked/@vscode/ripgrep/bin/")) ||
+		(await checkPath(`node_modules/@vscode/ripgrep-universal/${ripgrepUniversalBinDir}`)) ||
+		(await checkPath(`node_modules.asar.unpacked/@vscode/ripgrep-universal/${ripgrepUniversalBinDir}`)) ||
+		(await findRipgrepOnPath())
 	)
```
edelauna, line 109: Have you considered using the `rgPath` export from `@vscode/ripgrep` instead of probing paths manually?

VS Code itself resolves ripgrep this way ([`src/vs/base/node/ripgrep.ts`](https://sourcegraph.com/github.com/microsoft/vscode/-/blob/src/vs/base/node/ripgrep.ts)):

```typescript
const m = await import('@vscode/ripgrep-universal');
return m.rgPath.replace(/\bnode_modules\.asar\b/, 'node_modules.asar.unpacked');
```

VS Code's [require interceptor](https://sourcegraph.com/github.com/microsoft/vscode/-/blob/src/vs/workbench/api/common/extHostRequireInterceptor.ts?L97-101) aliases `@vscode/ripgrep` → `@vscode/ripgrep-universal` for extensions, so `import('@vscode/ripgrep')` works in both the extension host (intercepted) and the CLI (direct npm dep in `apps/cli/package.json`). The package handles all platform/arch resolution internally and exports a single `rgPath` string.

That would eliminate the hardcoded path list entirely and stay resilient to future layout changes. You'd just need to add `@vscode/ripgrep` to the `external` array in `src/esbuild.mjs`.

edelauna (changes requested):
Reopening after discussion, please remove the external `PATH` fallback, lets just rely on the bundled `rg` in vscode to resolve this issue. 

Can you also ensure codecov passes.

0xMink:
Dropped the PATH fallback in 2671d1ef4. The branch now resolves only under `vscode.env.appRoot`, including the `@vscode/ripgrep-universal/bin/<platform>-<arch>/` layout that Insiders' staged-install builds use — i.e. the original bug. codecov should clear since the uncovered lines were the PATH helper.

On `rgPath` from `@vscode/ripgrep`: importing that from the npm package would mean installing it as a dep and shipping the binary in the VSIX (the #243 direction), which I read your "for now" as wanting to defer. Happy to take that on if you'd prefer it over the current `appRoot`-relative resolution.

0xMink:
Reread your inline comment in full — my prior reply was based on a truncated view of it and incorrectly framed `rgPath` as requiring a bundled binary. Verified the interceptor at `src/vs/workbench/api/common/extHostRequireInterceptor.ts` lines 97–101 (`['@vscode/ripgrep', '@vscode/ripgrep-universal']`); `import('@vscode/ripgrep')` does resolve to VS Code's own copy with no bundling.

Reimplemented along the lines you suggested in 06c62c265:

- `getBinPath` is now a single `await import("@vscode/ripgrep")` + the `.asar` → `.asar.unpacked` substitution, wrapped in a try/catch that returns `undefined` on failure.
- `@vscode/ripgrep` added as a `src` devDep (types only; the binary is not shipped — the VSIX is still built `--no-dependencies`).
- `@vscode/ripgrep` added to `external` in `src/esbuild.mjs` so the require survives to runtime for VS Code's interceptor.
- The hardcoded path list — including the `@vscode/ripgrep-universal` entries I'd added in the previous commit — is gone.
- Tests cover four branches: `rgPath` returned, `.asar` substitution applies, `rgPath` undefined, `rgPath` access throws.

Net diff on this commit is +51/−49; the overall branch is a much smaller change than where it started.

src/services/ripgrep/index.ts
```diff
@@ -80,20 +79,27 @@ export function truncateLine(line: string, maxLength: number = MAX_LINE_LENGTH):
...
-export async function getBinPath(vscodeAppRoot: string): Promise<string | undefined> {
-	const checkPath = async (pkgFolder: string) => {
-		const fullPath = path.join(vscodeAppRoot, pkgFolder, binName)
-		return (await fileExistsAtPath(fullPath)) ? fullPath : undefined
+export async function getBinPath(_vscodeAppRoot: string): Promise<string | undefined> {
+	try {
+		const m = await import("@vscode/ripgrep")
+		return m.rgPath?.replace(/\bnode_modules\.asar\b/, "node_modules.asar.unpacked")
+	} catch {
+		return undefined
```
edelauna: This `import("@vscode/ripgrep")` doesn't work for marketplace extensions — tested and confirmed by S1LV3RC01N's diagnostic output:

```
--- step 1: require("@vscode/ripgrep") via loadRipgrep ---
loadRipgrep() returned undefined (require threw inside)
```

The VS Code require interceptor ([`extHostRequireInterceptor.ts` L107-114](https://sourcegraph.com/github.com/microsoft/vscode@7d45e9906edb29546335e8af5fbb669b280c1cd9/-/blob/src/vs/workbench/api/common/extHostRequireInterceptor.ts?L107-114)) only rewrites **absolute paths** matching `^(${appRoot}/node_modules/)(@vscode/ripgrep)(.*)$`. A bare specifier like `"@vscode/ripgrep"` never matches that regex. And the extension's module resolution scope (`~/.vscode/extensions/...`) doesn't include `appRoot/node_modules`, so Node can't resolve it either.

The interceptor is for VS Code's own internal code and built-in extensions (which live under `appRoot`), not marketplace extensions. The previous `path.join(appRoot, pkgFolder, binName)` + `fileExistsAtPath` approach is correct — it's filesystem access, which isn't subject to module resolution scope.

Could we revert to the path-probing approach from commit 1 (`051b67b`) with the `@vscode/ripgrep-universal` paths, minus the PATH fallback from commit 2?

0xMink:
Reverted `06c62c265` (the `await import("@vscode/ripgrep")` refactor) in 8666e1970. The branch now matches the prior commit 2 shape: hardcoded paths covering both `@vscode/ripgrep` and `@vscode/ripgrep-universal` layouts under `vscode.env.appRoot`, no PATH fallback, no require attempt, no `@vscode/ripgrep` dep.

Reasoning for the revert: a diagnostic test build run on a Windows VS Code stable 1.121.0 install showed that `require("@vscode/ripgrep")` throws — VS Code's extHost interceptor aliases the require to `@vscode/ripgrep-universal`, but that target package isn't installed on builds that haven't completed VS Code's package-rename migration. Shipping a code path that has no observed environment in which it actually works didn't seem worth the maintenance surface, especially since the contract VS Code will expose to 3rd-party extensions once the migration lands may differ from the current internal interceptor.

The diagnostic also confirmed that the hardcoded path probe finds ripgrep at the classic location on that install, so the revert is the version that demonstrably works.

PR body updated to reflect this. A separate follow-up PR will add a user-triggerable diagnostic command (`Zoo Code: Show Ripgrep Diagnostic`) so the next time someone hits weird resolution behavior we can get this kind of output without needing a custom test build.

Ready for re-review.

edelauna (commented):
Nice! Lets also address the coderabbit comments, then this should be good to 🚢

src/services/ripgrep/index.ts
edelauna: incorrect

src/services/ripgrep/index.ts
```diff
@@ -11,7 +11,7 @@ This file provides functionality to perform regex searches on files using ripgre
 Inspired by: https://github.com/DiscreteTom/vscode-ripgrep-utils
 
 Key components:
-1. getBinPath: Locates the ripgrep binary within the VSCode installation.
+1. getBinPath: Locates the ripgrep binary (in the VS Code install, or on the system PATH).
```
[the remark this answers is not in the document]
0xMink: Fixed in b2d675526.

0xMink:
Added the asar.unpacked `@vscode/ripgrep-universal` test in 6a84e58d7 (nit from review 4351723964).

src/services/ripgrep/__tests__/index.spec.ts
```diff
@@ -48,3 +58,41 @@ describe("Ripgrep line truncation", () => {
...
+		mockFileExists.mockImplementation(async (p: string) => p === rg)
+
+		expect(await getBinPath(appRoot)).toBe(rg)
+	})
+
+	it("returns undefined when ripgrep cannot be found", async () => {
+		mockFileExists.mockResolvedValue(false)
+
+		expect(await getBinPath(appRoot)).toBeUndefined()
+	})
```
edelauna, line 97: The tests cover the classic layout, universal, and unpacked-universal probes — but `getBinPath` has three more candidates (`index.ts` lines 104–106): `node_modules/vscode-ripgrep/bin`, `node_modules.asar.unpacked/vscode-ripgrep/bin/`, and `node_modules.asar.unpacked/@vscode/ripgrep/bin/`. Worth adding cases for those too, otherwise a refactor that accidentally removes or reorders them would pass the suite undetected.

src/services/ripgrep/__tests__/index.spec.ts
```diff
@@ -48,3 +58,41 @@ describe("Ripgrep line truncation", () => {
...
+	beforeEach(() => {
+		mockFileExists.mockReset()
+		mockFileExists.mockResolvedValue(false)
+	})
+
+	it("resolves ripgrep from the classic @vscode/ripgrep layout", async () => {
+		const rg = path.join(appRoot, "node_modules/@vscode/ripgrep/bin", binName)
+		mockFileExists.mockImplementation(async (p: string) => p === rg)
+
+		expect(await getBinPath(appRoot)).toBe(rg)
```
edelauna, line 76: Is there a test that verifies probe priority — e.g. that when both classic and universal paths exist, the classic one wins? The waterfall ordering in `getBinPath` is load-bearing, but nothing currently asserts which path is returned when more than one `fileExistsAtPath` call would return `true`.

src/services/ripgrep/index.ts
```diff
@@ -51,6 +51,11 @@ rel/path/to/helper.ts
 const isWindows = process.platform.startsWith("win")
 const binName = isWindows ? "rg.exe" : "rg"
 
+// VS Code's @vscode/ripgrep-universal package (used by recent VS Code builds,
+// including the Insiders staged-install layout) nests the binary under
+// bin/<platform>-<arch>/ rather than directly in bin/.
+const ripgrepUniversalBinDir = `bin/${process.platform}-${process.arch}`
```
edelauna, line 57: The name `ripgrepUniversalBinDir` reads as the platform-arch directory (e.g. `linux-x64`), but the value already includes the `bin/` prefix: `` `bin/${process.platform}-${process.arch}` ``. A reader skimming lines 107–108 has to mentally expand the constant to see where `bin/` lives. Something like `ripgrepUniversalBinSubpath` might make it clearer that this is the full sub-path from the package root, not just the leaf directory.

edelauna (approved):
Nice thanks for addressing this.

Have a couple nits, but created the following issue to track resolution since the conversation history on this PR is getting quire long: https://github.com/Zoo-Code-Org/Zoo-Code/issues/296

PR #250, opened 2026-05-22
chore: re-enable the prefer-const ESLint rule

0xMink:
## Summary

`prefer-const` was disabled in `src/eslint.config.mjs` under a `// TODO: These should be fixed and the rules re-enabled` comment. This re-enables it.

- The rule is enabled as `["error", { destructuring: "all" }]`. With `destructuring: "all"` it only flags a destructuring declaration when *every* binding can be `const`; the default (`"any"`) flags mixed const/let destructures and would force splitting them.
- ~90 violations are auto-fixed (`eslint --fix`): `let` → `const`.
- 5 declare-then-assign cases were merged by hand into a single `const` declaration:
  - `api/providers/minimax.ts` — `stream`
  - `api/transform/cache-strategy/__tests__/cache-strategy.spec.ts` — `strategy` (×2)
  - `core/tools/ExecuteCommandTool.ts` — `onCompletedPromise` (the explicit `Promise<void>` type is preserved; `resolveOnCompleted` stays `let` as it is assigned inside the executor)
  - `services/mcp/McpHub.ts` — `cancellationDisposable` moved to its single assignment site; the `cleanup` closure that references it only runs asynchronously, so it is always initialized first.

No behavior change.

## Test plan

- [x] `eslint .` is clean across `src/` with the rule active.
- [x] `tsc --noEmit` is clean.
- [x] Tests for the hand-edited modules pass (`cache-strategy`, `ExecuteCommandTool`, `McpHub`, `minimax` — 137 passed).

## Summary by CodeRabbit

* **Refactor**
  * Improved consistency by switching many local variables to non-reassignable bindings across provider integrations, transforms, caching, and core utilities.
  * Minor stream and message-construction refactors to align with the same immutability style.

* **Tests**
  * Added OAuth cancellation regression coverage (including cleanup behavior).
  * Added/extended tests for completion output, file context tracking, and code-index batch error handling.

* **Chores**
  * Re-enabled stricter linting for const usage.
  * Updated minor README formatting and a changeset clarification.


0xMink:
Follow-up commit efc86784d fixes a TDZ regression introduced by the prefer-const consolidation in `src/services/mcp/McpHub.ts`: when `CancellationToken.onCancellationRequested()` is given an already-cancelled token, VS Code fires the callback synchronously inside the registration call, so `cleanup` was hitting the `const cancellationDisposable = …` binding while it was still in the temporal dead zone and crashing with a ReferenceError. Restructured `cleanup()` to receive the disposable as a parameter (keeping the `const`), and the cancellation callback now invokes `cleanup()` with no argument since VS Code disposes its own one-shot listener after firing. Regression test included.

src/services/mcp/McpHub.ts
```diff
@@ -1188,8 +1187,15 @@ export class McpHub {
...
 
 			// --- Cancellation (progress bar's Cancel button) ---
-			cancellationDisposable = cancellationToken.onCancellationRequested(() => {
+			const cancellationDisposable = cancellationToken.onCancellationRequested(() => {
 				if (disposed) return
+				// Don't pass cancellationDisposable here: if the token was already
+				// cancelled at registration time, VS Code fires this callback
+				// synchronously inside onCancellationRequested(...) — before the
+				// `const cancellationDisposable = ...` assignment completes, so the
+				// binding is still in the TDZ. VS Code also disposes its own
```
edelauna, line 1204: Could we also guard the watcher registration that happens after this callback is installed? If this callback fires synchronously, `cleanup()` runs before `_oauthWatchers.set(...)`, so the watcher entry below can still be added after disposal.

edelauna (changes requested):

src/services/mcp/McpHub.ts
0xMink: Good catch. Moved the `_oauthWatchers.set(watcherKey, ...)` registration above the `onCancellationRequested(...)` call so the entry exists in the map before the callback can fire. A sync-firing `cleanup()` now finds and deletes it cleanly. Extended the existing already-cancelled-token regression test with `expect(_oauthWatchers.has(\`${serverName}:${source}\`)).toBe(false)` after the flow resolves — verified empirically by reverting the move and watching that assertion fail.

Pushed as 2143789b4.

edelauna (dismissed):
Thanks for re-enabling this! ❤️

edelauna (dismissed):

edelauna (approved):

PR #251, opened 2026-05-22
chore: enforce no-floating-promises in activate/

0xMink:
## Summary

First slice of a ratchet that re-enables `@typescript-eslint/no-floating-promises` directory by directory.

The repo's lint is zero-tolerance (`eslint --max-warnings=0` + the `only-warn` plugin), so the rule can't be enabled globally without first clearing all ~194 floating promises across `src/`. Instead, this enables the rule **scoped to `activate/**`** via a `files`-block in `eslint.config.mjs` (with type-aware parsing for that scope). `pnpm lint` stays green; subsequent PRs widen the scope until it's global.

## Fixes

`activate/registerCommands.ts` had 7 un-awaited `postMessageToWebview(...)` calls (fire-and-forget UI messages):

- **5 marked `void`** — in synchronous command handlers (`settingsButtonClicked`, `historyButtonClicked`, `marketplaceButtonClicked`, `acceptInput`); these are intentional fire-and-forget and the handler is not `async`.
- **2 awaited** — in handlers that are already `async` (`focusInput`, `toggleAutoApprove`). `focusInput`'s call sits inside a `try/catch`, so awaiting routes a rejected post into the existing error handler.

No sync→async handler conversions; behavior is unchanged.

## Test plan

- [x] `eslint activate/` clean.
- [x] `eslint . --ext=ts --max-warnings=0` (the repo lint command) passes.
- [x] `tsc --noEmit` clean.
- [x] `activate/` test suite passes (12 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UI command handling: message posting now properly awaits or logs errors as appropriate, and visibility/focus signaling is more consistent for settings, history, marketplace, input, and focus actions.

* **Chores**
  * Enabled stricter linting to catch unhandled promises.

* **Tests**
  * Added tests for command handlers, visibility/focus behavior, message-posting success/failure paths, and test-state isolation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/251?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/activate/registerCommands.ts
+++ b/src/activate/registerCommands.ts
@@ -103,9 +103,9 @@ const getCommandsMap = ({ context, outputChannel, provider }: RegisterCommandOpt
 
 		TelemetryService.instance.captureTitleButtonClicked("settings")
 
-		visibleProvider.postMessageToWebview({ type: "action", action: "settingsButtonClicked" })
+		void visibleProvider.postMessageToWebview({ type: "action", action: "settingsButtonClicked" })
 		// Also explicitly post the visibility message to trigger scroll reliably
-		visibleProvider.postMessageToWebview({ type: "action", action: "didBecomeVisible" })
+		void visibleProvider.postMessageToWebview({ type: "action", action: "didBecomeVisible" })
 	},
 	historyButtonClicked: () => {
 		const visibleProvider = getVisibleProviderOrLog(outputChannel)
@@ -116,12 +116,12 @@ const getCommandsMap = ({ context, outputChannel, provider }: RegisterCommandOpt
 
 		TelemetryService.instance.captureTitleButtonClicked("history")
 
-		visibleProvider.postMessageToWebview({ type: "action", action: "historyButtonClicked" })
+		void visibleProvider.postMessageToWebview({ type: "action", action: "historyButtonClicked" })
 	},
 	marketplaceButtonClicked: () => {
 		const visibleProvider = getVisibleProviderOrLog(outputChannel)
 		if (!visibleProvider) return
-		visibleProvider.postMessageToWebview({ type: "action", action: "marketplaceButtonClicked" })
+		void visibleProvider.postMessageToWebview({ type: "action", action: "marketplaceButtonClicked" })
 	},
 	newTask: handleNewTask,
 	setCustomStoragePath: async () => {
@@ -150,7 +150,7 @@ const getCommandsMap = ({ context, outputChannel, provider }: RegisterCommandOpt
 
 			// Send focus input message only for sidebar panels
 			if (sidebarPanel && getPanel() === sidebarPanel) {
-				provider.postMessageToWebview({ type: "action", action: "focusInput" })
+				await provider.postMessageToWebview({ type: "action", action: "focusInput" })
 			}
 		} catch (error) {
 			outputChannel.appendLine(`Error focusing input: ${error}`)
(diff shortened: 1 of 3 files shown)


src/activate/registerCommands.ts
edelauna, line 1: Thanks so much for looking into this!. since we're touching this file can we also add specs to make codecov happy?

0xMink: Added specs in ae4c66dad; codecov/patch now passes.

0xMink: Follow-up in 3952b3cc7: upgraded the five `void`-prefixed `postMessageToWebview` sites here to also chain a `.catch` that logs to `outputChannel`, mirroring the pattern #253 uses for the `core/task/` sites. The `void` prefix stays (still satisfies no-floating-promises), but rejections are now logged explicitly rather than relying on `ClineProvider.postMessageToWebview`'s internal try/catch — that internal swallow is benign today, but it's an implicit contract the `.catch` arm makes explicit at the call site. New parameterized test in `registerCommands.spec.ts` covers the rejection path for all five sites.

0xMink:
@coderabbitai Hardened both `await`-asserting tests for `focusInput` and `toggleAutoApprove` with the deferred-promise pattern in 65f5cb418. The mock now returns a pending promise via `mockReturnValueOnce`, the handler is invoked without awaiting, and we flush a microtask (two for `focusInput` since the handler awaits `focusPanel` before `postMessageToWebview`) and assert that the handler promise is still pending before resolving the deferred. Verified the regression-catch by temporarily replacing `await` with `void` in both handlers — both hardened tests fail with "expected true to be false" at the pending-state check, as intended. Restoring `await` makes them pass again.

0xMink:
@coderabbitai addressed both items in 152939b6e:

- **toggleAutoApprove** now wraps the await in try/catch, matching the defensive posture used on the other handlers (settingsButtonClicked, historyButtonClicked, etc.). An unhandled rejection from postMessageToWebview would otherwise bubble to VS Code's command dispatcher rather than being logged consistently with the rest of the file.
- All five postMessageToWebview failure log messages now carry a `[<command-name>]` prefix so multi-failure logs are unambiguous.

Tests in `src/activate/__tests__/registerCommands.spec.ts` updated to parameterize the prefix per handler and to cover the new try/catch branch on `toggleAutoApprove`.

src/activate/registerCommands.ts
```diff
@@ -103,9 +103,13 @@ const getCommandsMap = ({ context, outputChannel, provider }: RegisterCommandOpt
...
 
-		visibleProvider.postMessageToWebview({ type: "action", action: "settingsButtonClicked" })
+		void visibleProvider
+			.postMessageToWebview({ type: "action", action: "settingsButtonClicked" })
+			.catch((error) => outputChannel.appendLine(`postMessageToWebview failed: ${error}`))
 		// Also explicitly post the visibility message to trigger scroll reliably
-		visibleProvider.postMessageToWebview({ type: "action", action: "didBecomeVisible" })
+		void visibleProvider
+			.postMessageToWebview({ type: "action", action: "didBecomeVisible" })
+			.catch((error) => outputChannel.appendLine(`postMessageToWebview failed: ${error}`))
```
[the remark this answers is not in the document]
0xMink: Both `postMessageToWebview` calls in `settingsButtonClicked` are pre-existing on `main` — the `didBecomeVisible` dispatch was added by commit f274a150c4 on 2025-05-20 (`git blame` for `src/activate/registerCommands.ts:108` on `main`). This PR's only change at this site is wrapping both pre-existing calls with `void`/`.catch` for `no-floating-promises` compliance. No new webview message was introduced.

0xMink:
Fixed in 7773498d534480f413e485b05d7340b182e69fb8 — updated the comment in `src/activate/__tests__/registerCommands.spec.ts` (around line 295) to reference `setImmediate` instead of `Promise.resolve()`, matching the actual flush at line 309. Re: the review nit in https://github.com/Zoo-Code-Org/Zoo-Code/pull/251#pullrequestreview-4352106544.

edelauna (approved):

PR #261, opened 2026-05-23
fix(terminal): terminate running process when task is cancelled (#245)

### LINKED ISSUE #245: [BUG] Cancel (✕) doesn't terminate the running command — window stays blocked until manual kill
### Summary
Pressing the cancel (✕) button on a running command does not terminate the underlying process. The command keeps running and the window stays blocked until the process is killed manually or the extension is fully restarted. Raised in the community Discord and confirmed by a maintainer.

### Steps to reproduce
1. Start a task that runs a long-running / blocking terminal command.
2. Press the ✕ (cancel) button while it is still running.
3. Observe the behavior.

### Expected
✕ cancels the task **and** terminates the spawned process, unblocking the window immediately.

### Actual
The process is not killed; it must be terminated manually (or the extension fully restarted) before the window is usable again.

### Notes
- Related to but distinct from #208 (that issue is about task logs disappearing when pressing Stop; this one is about the process itself not being terminated).
- Happy to help with a fix and/or regression test coverage.

### Environment
- Zoo Code (reported on 3.54.1; likely not version-specific)
- VS Code / VS Code Insiders

proyectoauraorg:
## Summary

Fixes #245. Pressing cancel (✕) on a running command didn't terminate the underlying process — the terminal stayed `busy` and the window stayed blocked until a manual kill. Several users hit this (also surfaced in #general-contributors).

**Root cause:** when a task is torn down (cancel ✕, task switch/removal), `Task.dispose()` → `TerminalRegistry.releaseTerminalsForTask()` only disassociated the terminal (`taskId = undefined`) without aborting a still-running command, leaving the process orphaned and the terminal stuck `busy`.

## Fix

In `releaseTerminalsForTask`, abort the running process for any terminal still `busy` before disassociating it. The abort path already exists on both backends and is a no-op when idle:

- `ExecaTerminalProcess.abort()` → SIGINT → SIGKILL on the owned subprocess
- `TerminalProcess.abort()` → Ctrl+C (`\x03`), gated on an active stream

Wrapped in try/catch so a failing abort never blocks task teardown.

## Testing

Adds 4 `TerminalRegistry` tests: busy terminal → process aborted + disassociated; idle terminal → not aborted; only the target task's terminals are released; `abort()` errors are swallowed. Full terminal suite green locally (92 passed).

## Notes

This is the **short-term** layer of the two-layer approach discussed with @edelauna in #general-contributors: make ✕ reliably terminate the process today. The **long-term** direction — owning process lifecycle via the VS Code Task API + `CustomExecution` for deterministic cross-platform process-tree termination — is a natural follow-up and fits the portable-core/runtime direction. Happy to take a first pass at that separately so it doesn't block this bug.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where cancelling or tearing down a task would leave the underlying terminal process running in a busy state. Terminal processes are now properly aborted when a task is cancelled or torn down, preventing hung processes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/261?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/integrations/terminal/TerminalRegistry.ts
+++ b/src/integrations/terminal/TerminalRegistry.ts
@@ -284,6 +284,22 @@ export class TerminalRegistry {
 	public static releaseTerminalsForTask(taskId: string): void {
 		this.terminals.forEach((terminal) => {
 			if (terminal.taskId === taskId) {
+				// #245: If the terminal is still executing a command when its task is torn
+				// down (user pressed cancel ✕, or the task was switched/removed), abort the
+				// process. Otherwise the command keeps running orphaned and the terminal stays
+				// stuck "busy" — the cancel-doesn't-terminate bug. abort() is safe when idle
+				// (Ctrl+C is gated on an active stream; Execa abort is idempotent).
+				if (terminal.busy) {
+					try {
+						terminal.process?.abort()
+					} catch (error) {
+						console.error(
+							`[TerminalRegistry] Error aborting process for terminal ${terminal.id} on release:`,
+							error,
+						)
+					}
+				}
+
 				terminal.taskId = undefined
 			}
 		})
--- a/.changeset/terminate-process-on-cancel.md
+++ b/.changeset/terminate-process-on-cancel.md
@@ -1,0 +1,5 @@
+---
+"zoo-code": patch
+---
+
+Terminate the running command when a task is cancelled or torn down (#245). Pressing cancel (✕) now aborts the underlying terminal process instead of leaving it running while the terminal stays stuck "busy" until a manual kill.
--- a/src/integrations/terminal/__tests__/TerminalRegistry.spec.ts
+++ b/src/integrations/terminal/__tests__/TerminalRegistry.spec.ts
@@ -123,4 +123,66 @@ describe("TerminalRegistry", () => {
 			}
 		})
 	})
+
+	describe("releaseTerminalsForTask", () => {
+		it("aborts a busy terminal's running process and disassociates it from the task (#245)", () => {
+			const terminal = TerminalRegistry.createTerminal("/test/path", "vscode")
+			const abort = vi.fn()
+			terminal.taskId = "task-245"
+			terminal.busy = true
+			terminal.process = { abort } as any
+
+			TerminalRegistry.releaseTerminalsForTask("task-245")
+
+			expect(abort).toHaveBeenCalledTimes(1)
+			expect(terminal.taskId).toBeUndefined()
+		})
+
+		it("does not abort an idle (not busy) terminal but still disassociates it", () => {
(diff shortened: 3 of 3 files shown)


src/integrations/terminal/TerminalRegistry.ts
```diff
@@ -284,6 +284,22 @@ export class TerminalRegistry {
 	public static releaseTerminalsForTask(taskId: string): void {
 		this.terminals.forEach((terminal) => {
 			if (terminal.taskId === taskId) {
+				// #245: If the terminal is still executing a command when its task is torn
+				// down (user pressed cancel ✕, or the task was switched/removed), abort the
+				// process. Otherwise the command keeps running orphaned and the terminal stays
+				// stuck "busy" — the cancel-doesn't-terminate bug. abort() is safe when idle
+				// (Ctrl+C is gated on an active stream; Execa abort is idempotent).
```
edelauna, line 291: The PR description says `ExecaTerminalProcess.abort()` does "SIGINT → SIGKILL", but looking at the implementation it sends SIGKILL directly (no SIGINT first). The comment here ("Execa abort is idempotent") is fine, but might be worth noting the asymmetry: `TerminalProcess.abort()` sends Ctrl+C (a single, soft signal), while `ExecaTerminalProcess.abort()` sends SIGKILL immediately. For long-running processes that need graceful shutdown, a single Ctrl+C or SIGINT before SIGKILL would be friendlier — but that's a pre-existing concern, not something this PR needs to fix.

src/integrations/terminal/TerminalRegistry.ts
```diff
@@ -284,6 +284,22 @@ export class TerminalRegistry {
...
+				if (terminal.busy) {
+					try {
+						terminal.process?.abort()
```
edelauna, line 294: What happens if `terminal.busy` is true but `process` is still undefined? `ExecaTerminal.runCommand` sets `this.busy = true` before assigning `this.process` — tiny window but possible. The `?.` handles it (no-op), but also worth noting: `TerminalProcess.abort()` only sends a single Ctrl+C (`\x03`). If the running process requires multiple Ctrl+C presses to terminate (e.g. some interactive tools), this single abort won't be enough and the terminal will stay busy. Is there a retry/escalation path, or is single-Ctrl+C considered best-effort here?

edelauna: I created a followup issue for this: https://github.com/Zoo-Code-Org/Zoo-Code/issues/266

edelauna (approved):
Thank you for taking this on! I created an issue for a state tracking suggestion I have as an interim solution to process cancellation feeling non-responsive.

PR #271, opened 2026-05-24
fix(openai-codex): use full ~1M context window for gpt-5.5

### LINKED ISSUE #237: [BUG] GPT 5.5 has a window context of only 400K.

GPT 5.5 has a window context of only 400K, however, it supports 1.1M via Codex OAuth connection.

Image

### Reproduction steps

1. Configure provider with OpenAI Plus OAuth
2. Configure GPT 5.5 template
3. Log in
4. You will see that it is truncating the window to 400k tokens

### Expected result

GPT 5.5 via Codex in GPT Plan using the standard 1M or 1.1M token window.

### Actual result

GPT 5.5 via Codex on the GPT plan using 400K tokens

_No response_

### App Version

3.54.1

### API Provider (optional)

OpenAI

### Model Used (optional)

GPT 5.5

_No response_

```shell

```

proyectoauraorg:
## Summary

GPT-5.5 via the Codex OAuth (ChatGPT subscription) connection supports a ~1.05M token context window, but its model metadata in `openAiCodexModels` capped `contextWindow` at 400000. This truncated the usable context to 400K, as reported in #237.

This change aligns gpt-5.5's `contextWindow` with the equivalent `gpt-5.4` entry (`1_050_000`). The handler reads `contextWindow` directly from this metadata (`getModel()` returns `openAiCodexModels[id]` unmodified), so this is the only place the cap originated; there is no separate clamping.

## Changes
- `packages/types/src/providers/openai-codex.ts`: `gpt-5.5` `contextWindow` 400000 -> 1_050_000
- `src/api/providers/__tests__/openai-codex.spec.ts`: regression test asserting gpt-5.5 exposes the ~1M window

Closes #237

## Summary by CodeRabbit

* **Updates**
  * Increased GPT-5.5 model context window from 400,000 to 1,050,000 tokens, enabling support for longer inputs and larger documents.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/271?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/types/src/providers/openai-codex.ts
+++ b/packages/types/src/providers/openai-codex.ts
@@ -174,7 +174,7 @@ export const openAiCodexModels = {
 	},
 	"gpt-5.5": {
 		maxTokens: 128000,
-		contextWindow: 400000,
+		contextWindow: 1_050_000,
 		includedTools: ["apply_patch"],
 		excludedTools: ["apply_diff", "write_to_file"],
 		supportsImages: true,
--- a/src/api/providers/__tests__/openai-codex.spec.ts
+++ b/src/api/providers/__tests__/openai-codex.spec.ts
@@ -41,4 +41,12 @@ describe("OpenAiCodexHandler.getModel", () => {
 		expect(model.id).toBe("gpt-5.4-mini")
 		expect(model.info).toBeDefined()
 	})
+
+	it("should expose the full ~1M context window for GPT-5.5", () => {
+		const handler = new OpenAiCodexHandler({ apiModelId: "gpt-5.5" })
+		const model = handler.getModel()
+
+		expect(model.id).toBe("gpt-5.5")
+		expect(model.info.contextWindow).toBe(1_050_000)
+	})
 })


packages/types/src/providers/openai-codex.ts
```diff
@@ -174,7 +174,7 @@ export const openAiCodexModels = {
 	},
 	"gpt-5.5": {
 		maxTokens: 128000,
-		contextWindow: 400000,
+		contextWindow: 1_050_000,
```
edelauna, line 177: Were the other models with `contextWindow: 400000` (e.g. `gpt-5`, `gpt-5.1`, `gpt-5.3-codex`, `gpt-5-codex`) audited as well? If their actual windows are also larger, they would still be capped too early.

edelauna (commented):
Small nit: I figure since we're making the change - could we also verify the other models context windows, I'm not confident that 5.5 is the only model with a 1M context window.

packages/types/src/providers/openai-codex.ts
proyectoauraorg: You were right to flag this — my change is wrong. GPT-5.5 is 1M via the API but **400K in Codex**, and this file is the Codex path, so the original `400000` was correct (#237 conflated the two). Closing this PR — thanks for catching it. 🦘

proyectoauraorg:
Closing — see review thread. GPT-5.5 is 400K in Codex (1M is API-only), so the original value was correct; this change would cause 'exceeds the context window' errors on the Codex provider. 🦘

PR #272, opened 2026-05-24
fix(terminal): retry Ctrl+C for processes needing multiple SIGINT (#266)

### LINKED ISSUE #266: fix(terminal): handle processes that require multiple Ctrl+C to terminate
## Problem

When a task is cancelled (✕ button), `releaseTerminalsForTask` calls `terminal.process.abort()` which sends a single Ctrl+C (`\x03`) for VSCode-backed terminals. Some processes require multiple Ctrl+C presses to terminate (e.g. interactive tools, processes that catch SIGINT and prompt for confirmation). In these cases the process stays running and the terminal stays stuck "busy."

This was surfaced during review of #261 which fixed the main orphaned-process bug (#245). The ExecaTerminal backend is unaffected since it sends SIGKILL directly.

## Context

The current cancel flow is synchronous:
```
cancelTask() → abortTask() → dispose() → releaseTerminalsForTask() → process.abort()
```

`dispose()` tears everything down immediately after the single Ctrl+C — there's no retry or check to see if the process actually exited.

## Possible approaches

1. **Automatic retry in `TerminalProcess.abort()`** — send Ctrl+C, wait ~500ms, check `busy`, send another. Small change but operates on an already-disassociated terminal since `dispose()` has run.

2. **Two-phase cancel** — split the cancel flow into "abort process" and "dispose task" phases. Only dispose after the process exits (or after a timeout). Requires changes to `ClineProvider.cancelTask()`, the webview UI, and `releaseTerminalsForTask`.

3. **Address as part of the `CustomExecution` migration** — the long-term direction discussed in #245 would give direct process-tree control via the VS Code Task API, making Ctrl+C escalation unnecessary.

## Related

- #245 — original cancel-doesn't-terminate bug
- #261 — fix that added abort on release (single Ctrl+C)

proyectoauraorg:
## Summary

Follow-up to #245 / #261. When a task is cancelled, `releaseTerminalsForTask` → `process.abort()` sent a single Ctrl+C (`\x03`) for VSCode-backed terminals, then `dispose()` tore everything down immediately. Processes that trap SIGINT or need multiple Ctrl+C (interactive tools, confirm-on-exit prompts) stayed running and the terminal stayed stuck "busy".

## Change

`abort()` still sends the immediate Ctrl+C, then kicks off a **bounded, fire-and-forget** retry (`retryAbort`) that re-sends Ctrl+C up to 3 times (500ms apart), stopping early once the process exits (`terminal.busy` flips false) or we stop listening. An `aborting` guard prevents overlapping retry loops, and `terminalRef.deref()` keeps it safe after dispose. The synchronous cancel path is never blocked and the retry window is bounded, so `dispose()` is never delayed indefinitely.

The **ExecaTerminal** backend is untouched — it sends SIGKILL directly and was never affected.

## Tests

Added an `abort` suite to `TerminalProcess.spec.ts` (Vitest, fake timers): single Ctrl+C when the process exits immediately; bounded re-send up to max while busy; early stop when the process exits mid-retry; no-op when not listening; no overlapping loops on repeated `abort()`.

Closes #266

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved process termination: interrupt (Ctrl+C) is retried a limited number of times with safeguards to avoid infinite retries, overlapping retry loops, or retries after the process has exited or the terminal has been repurposed.

* **Tests**
  * Added tests confirming bounded retry behavior, prevention of overlapping abort attempts, and correct early stopping across terminal states and reuse scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/272?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/multiple-ctrl-c-terminate.md
+++ b/.changeset/multiple-ctrl-c-terminate.md
@@ -1,0 +1,5 @@
+---
+"zoo-code": patch
+---
+
+Retry Ctrl+C when cancelling a task so processes that ignore a single SIGINT actually terminate (#266). The terminal abort path now re-sends Ctrl+C a bounded number of times, verifying between attempts whether the process exited, before the terminal is torn down. Follow-up to #245/#261; the ExecaTerminal backend is unaffected.
--- a/.changeset/multiple-ctrl-c-terminate.md
+++ b/.changeset/multiple-ctrl-c-terminate.md
@@ -1,5 +1,0 @@
----
-"zoo-code": patch
----
-
-Retry Ctrl+C when cancelling a task so processes that ignore a single SIGINT actually terminate (#266). The terminal abort path now re-sends Ctrl+C a bounded number of times, verifying between attempts whether the process exited, before the terminal is torn down. Follow-up to #245/#261; the ExecaTerminal backend is unaffected.
--- a/src/integrations/terminal/TerminalProcess.ts
+++ b/src/integrations/terminal/TerminalProcess.ts
@@ -8,11 +8,12 @@ import { Terminal } from "./Terminal"
 export class TerminalProcess extends BaseTerminalProcess {
 	// #266: Some processes (interactive tools, programs that trap SIGINT and
 	// prompt for confirmation) need more than one Ctrl+C to actually exit. We
-	// re-send Ctrl+C up to this many times, checking between attempts whether
-	// the process has exited, before giving up and letting dispose() proceed.
-	private static readonly ABORT_MAX_ATTEMPTS = 3
+	// send Ctrl+C up to this many times in TOTAL — the immediate send in abort()
+	// plus retries — checking between sends whether the process has exited, before
+	// giving up and letting dispose() proceed.
+	private static readonly CTRL_C_SEND_LIMIT = 3
 	// Delay between Ctrl+C re-sends. Kept short so cancel stays responsive; the
-	// whole retry window is bounded by ABORT_MAX_ATTEMPTS * ABORT_RETRY_DELAY_MS.
+	// retry window is bounded by (CTRL_C_SEND_LIMIT - 1) * ABORT_RETRY_DELAY_MS.
 	private static readonly ABORT_RETRY_DELAY_MS = 500
 
 	private terminalRef: WeakRef<Terminal>
@@ -290,15 +291,22 @@ export class TerminalProcess extends BaseTerminalProcess {
 	}
 
 	/**
-	 * Re-sends Ctrl+C up to ABORT_MAX_ATTEMPTS times, waiting ABORT_RETRY_DELAY_MS
-	 * between attempts and stopping early once the process exits (or once we stop
-	 * listening). Bounded so it can never loop indefinitely.
+	 * Re-sends Ctrl+C after the immediate send in abort(), up to CTRL_C_SEND_LIMIT
+	 * total sends, waiting ABORT_RETRY_DELAY_MS between sends and stopping early once
+	 * the process exits (or once we stop listening). Bounded so it can never loop
+	 * indefinitely.
 	 */
 	private async retryAbort(): Promise<void> {
-		for (let attempt = 1; attempt < TerminalProcess.ABORT_MAX_ATTEMPTS; attempt++) {
+		// abort() already sent Ctrl+C once, so `sent` starts at 1; re-send until we
+		// reach CTRL_C_SEND_LIMIT total.
+		for (let sent = 1; sent < TerminalProcess.CTRL_C_SEND_LIMIT; sent++) {
 			await new Promise((resolve) => setTimeout(resolve, TerminalProcess.ABORT_RETRY_DELAY_MS))
 
-			// Stop if the process already exited or we're no longer listening.
+			// Stop as soon as there's nothing left to interrupt. `isListening` (cleared
+			// by continue()) and `terminal.busy` (cleared by shellExecutionComplete() /
+			// the "completed" event) are set on different code paths and can diverge, so
+			// either one being false is a sufficient stop signal — we deliberately check
+			// both rather than collapsing them into one.
 			if (!this.isListening) {
 				return
 			}
--- a/src/integrations/terminal/TerminalProcess.ts
+++ b/src/integrations/terminal/TerminalProcess.ts
@@ -313,7 +313,12 @@ export class TerminalProcess extends BaseTerminalProcess {
 
 			const terminal = this.terminalRef.deref()
 
-			if (!terminal || !terminal.busy) {
+			// Stop if the terminal is gone, idle, or has already moved on to a different
+			// command. If the original command exits and the terminal is reused before this
+			// tick fires, `terminal.busy` can be true for the NEW command while
+			// `terminal.process` points at a different TerminalProcess — re-sending Ctrl+C
+			// then would interrupt an unrelated command, so we bail out.
+			if (!terminal || !terminal.busy || terminal.process !== this) {
 				return
 			}
 
--- a/src/integrations/terminal/TerminalProcess.ts
+++ b/src/integrations/terminal/TerminalProcess.ts
@@ -284,9 +284,11 @@ export class TerminalProcess extends BaseTerminalProcess {
 		// total retry window is bounded so dispose() is never delayed for long.
 		if (!this.aborting) {
 			this.aborting = true
-			void this.retryAbort().finally(() => {
-				this.aborting = false
-			})
+			void this.retryAbort()
+				.finally(() => {
+					this.aborting = false
+				})
+				.catch((err) => console.error("[TerminalProcess] retryAbort error:", err))
 		}
 	}
 
--- a/src/integrations/terminal/TerminalProcess.ts
+++ b/src/integrations/terminal/TerminalProcess.ts
@@ -6,7 +6,19 @@ import { BaseTerminalProcess } from "./BaseTerminalProcess"
 import { Terminal } from "./Terminal"
 
 export class TerminalProcess extends BaseTerminalProcess {
+	// #266: Some processes (interactive tools, programs that trap SIGINT and
+	// prompt for confirmation) need more than one Ctrl+C to actually exit. We
+	// re-send Ctrl+C up to this many times, checking between attempts whether
+	// the process has exited, before giving up and letting dispose() proceed.
+	private static readonly ABORT_MAX_ATTEMPTS = 3
+	// Delay between Ctrl+C re-sends. Kept short so cancel stays responsive; the
+	// whole retry window is bounded by ABORT_MAX_ATTEMPTS * ABORT_RETRY_DELAY_MS.
+	private static readonly ABORT_RETRY_DELAY_MS = 500
+
 	private terminalRef: WeakRef<Terminal>
+	// Guards against overlapping abort retry loops if abort() is called again
+	// while a previous loop is still re-sending Ctrl+C.
+	private aborting = false
 
 	constructor(terminal: Terminal) {
 		super()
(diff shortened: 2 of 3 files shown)


src/integrations/terminal/TerminalProcess.ts
```diff
@@ -256,9 +268,48 @@ export class TerminalProcess extends BaseTerminalProcess {
...
+		}
+	}
+
+	/**
+	 * Re-sends Ctrl+C up to ABORT_MAX_ATTEMPTS times, waiting ABORT_RETRY_DELAY_MS
+	 * between attempts and stopping early once the process exits (or once we stop
+	 * listening). Bounded so it can never loop indefinitely.
+	 */
+	private async retryAbort(): Promise<void> {
+		for (let attempt = 1; attempt < TerminalProcess.ABORT_MAX_ATTEMPTS; attempt++) {
```
edelauna: The loop runs for `attempt = 1, 2` (i.e. 2 retries), so the total sends are 3 — correct! But `attempt < ABORT_MAX_ATTEMPTS` starting at 1 means the loop fires `MAX_ATTEMPTS − 1` times, which is at odds with the constant name implying "total attempts". The JSDoc above also says "Re-sends Ctrl+C up to ABORT_MAX_ATTEMPTS times" when it means `ABORT_MAX_ATTEMPTS − 1`. Would it be clearer to start the loop at `0` (so `attempt < ABORT_MAX_ATTEMPTS` naturally reads as "retry at most 3 times") or rename the constant to something like `CTRL_C_SEND_LIMIT`?

src/integrations/terminal/TerminalProcess.ts
```diff
@@ -256,9 +268,48 @@ export class TerminalProcess extends BaseTerminalProcess {
...
+			await new Promise((resolve) => setTimeout(resolve, TerminalProcess.ABORT_RETRY_DELAY_MS))
+
+			// Stop if the process already exited or we're no longer listening.
+			if (!this.isListening) {
+				return
+			}
+
+			const terminal = this.terminalRef.deref()
```
edelauna, line 316: `isListening` and `terminal.busy` are set by different code paths (`continue()` vs `shellExecutionComplete()` / the `"completed"` event), so they can diverge. Is the intent that either flag being false is sufficient to stop retries? Worth a brief comment so future readers don't simplify one of these checks away.

src/integrations/terminal/__tests__/TerminalProcess.spec.ts
```diff
@@ -186,6 +186,90 @@ describe("TerminalProcess", () => {
 		})
 	})
 
+	describe("abort", () => {
+		const RETRY_DELAY_MS = 500
+		const MAX_ATTEMPTS = 3
```
edelauna: These mirror the private `ABORT_MAX_ATTEMPTS`/`ABORT_RETRY_DELAY_MS` constants in production but can't be imported. If those values are ever tuned, the tests will still pass while asserting the wrong timing. Could we add a comment cross-referencing the production constant names so a maintainer knows to update both?

src/integrations/terminal/__tests__/TerminalProcess.spec.ts
```diff
@@ -186,6 +186,90 @@ describe("TerminalProcess", () => {
...
+			await vi.advanceTimersByTimeAsync(RETRY_DELAY_MS * MAX_ATTEMPTS)
+
+			expect(mockTerminal.sendText).not.toHaveBeenCalled()
+		})
+
+		it("does not start overlapping retry loops when abort() is called repeatedly (#266)", async () => {
+			mockTerminalInfo.busy = true
+
+			terminalProcess.abort()
+			terminalProcess.abort()
```
edelauna, line 289: The `2` here assumes the second `abort()` sends one immediate Ctrl+C before hitting the `aborting` guard (guard is checked *after* `sendText`). That ordering is correct today, but if the guard ever moves above the send this count silently drops from 4 to 3 without an obvious failure. Worth a one-liner comment capturing that assumption?

.changeset/multiple-ctrl-c-terminate.md
edelauna, line 1: I looked into it, and this project doesn't actually use these changesets, so it's not required atm.

src/integrations/terminal/TerminalProcess.ts
proyectoauraorg: Good catch — renamed `ABORT_MAX_ATTEMPTS` → `CTRL_C_SEND_LIMIT` (total sends = 3) and reworked the loop to start at `sent = 1`, since the immediate Ctrl+C in `abort()` is send #1. So `sent < CTRL_C_SEND_LIMIT` now reads as the total bound, and the JSDoc was updated to match. Done in 0181a1144.

src/integrations/terminal/TerminalProcess.ts
proyectoauraorg: Added a comment capturing exactly that: `isListening` (cleared by `continue()`) and `terminal.busy` (cleared by `shellExecutionComplete()` / the "completed" event) are set on different code paths and can diverge, so either one being false is a sufficient stop signal — both are checked deliberately rather than collapsed. 0181a1144.

src/integrations/terminal/__tests__/TerminalProcess.spec.ts
proyectoauraorg: Added a cross-reference comment tying the mirrored `RETRY_DELAY_MS` / `MAX_ATTEMPTS` back to the production `ABORT_RETRY_DELAY_MS` / `CTRL_C_SEND_LIMIT`, so a maintainer tuning one knows to update the other. 0181a1144.

src/integrations/terminal/__tests__/TerminalProcess.spec.ts
proyectoauraorg: Added a one-liner noting the count of `2` relies on the `aborting` guard being checked *after* the immediate `sendText`: if the guard ever moves above the send, this drops to a single immediate send (total 3, not 4). 0181a1144.

.changeset/multiple-ctrl-c-terminate.md
proyectoauraorg: Removed `.changeset/multiple-ctrl-c-terminate.md` — thanks for confirming the project doesn't use these. 0181a1144. 🦘

src/integrations/terminal/TerminalProcess.ts
```diff
@@ -256,9 +269,55 @@ export class TerminalProcess extends BaseTerminalProcess {
...
+			// the "completed" event) are set on different code paths and can diverge, so
+			// either one being false is a sufficient stop signal — we deliberately check
+			// both rather than collapsing them into one.
+			if (!this.isListening) {
+				return
+			}
+
+			const terminal = this.terminalRef.deref()
+
+			if (!terminal || !terminal.busy) {
```
edelauna: Could this also check that the terminal is still running this same `TerminalProcess`? If the old command exits and the terminal is reused before the next retry tick, `busy` could be true for the new command.

src/integrations/terminal/__tests__/TerminalProcess.spec.ts
```diff
@@ -186,6 +186,98 @@ describe("TerminalProcess", () => {
...
+
+			terminalProcess.abort()
+			expect(mockTerminal.sendText).toHaveBeenCalledTimes(1)
+
+			// First retry tick: still busy, re-send.
+			await vi.advanceTimersByTimeAsync(RETRY_DELAY_MS)
+			expect(mockTerminal.sendText).toHaveBeenCalledTimes(2)
+
+			// Process exits before the next tick.
+			mockTerminalInfo.busy = false
```
edelauna: Could this exercise the production completion path instead of mutating `busy` directly? The retry depends on `terminal.busy` being cleared by the completion lifecycle, so a direct assignment may miss that wiring.

src/integrations/terminal/__tests__/TerminalProcess.spec.ts
```diff
@@ -186,6 +186,98 @@ describe("TerminalProcess", () => {
 		})
 	})
 
+	describe("abort", () => {
```
edelauna, line 189: Would it be worth covering the `TerminalRegistry.releaseTerminalsForTask()` path here or in the registry tests? That is the cancellation path from the regression, and it would also catch stale retries interacting with terminal release or reuse.

edelauna (commented):
Would like to address the possibility of terminal reuse

proyectoauraorg:
@edelauna this one's been addressed since your review — replied to each of your comments inline and pushed `0181a1144` (renamed the constant to `CTRL_C_SEND_LIMIT`, added the isListening/busy comments, dropped the changeset). CI green. Ready for another look when you have a moment — thanks! 🦘

src/integrations/terminal/TerminalProcess.ts
proyectoauraorg: Good catch — fixed in `96700dbae`. The retry tick now bails when `terminal.process !== this`, so if the original command exits and the terminal is reused before the next tick, we no longer re-send Ctrl+C into the new command (which could be busy for an unrelated reason). `process` is the terminal's current `RooTerminalProcess` and is reassigned on reuse / nulled by `shellExecutionComplete()`. Added a dedicated test ("stops re-sending Ctrl+C if the terminal is reused for a different process").

src/integrations/terminal/__tests__/TerminalProcess.spec.ts
proyectoauraorg: Done. The "exits mid-retry" test now calls `mockTerminalInfo.shellExecutionComplete({ exitCode: 0 })` instead of mutating `busy` directly — that's the production completion path, which clears `busy`, flips `running`, emits `shell_execution_complete`, and releases `terminal.process`. So the test exercises the real wiring the retry depends on.

src/integrations/terminal/__tests__/TerminalProcess.spec.ts
proyectoauraorg: The new reuse guard + its test cover the "stale retry interacts with terminal release/reuse" case at the unit level (a released terminal has `process` nulled, a reused one points elsewhere — both stop the loop). A full `TerminalRegistry.releaseTerminalsForTask()` integration test reads better in the registry suite since it needs the registry/task wiring; happy to add that as a focused follow-up (or in this PR if you'd prefer it here — let me know).

proyectoauraorg:
@edelauna thanks for the careful re-review — addressed the terminal-reuse concern in `96700dbae`: the retry now stops when `terminal.process !== this` (with a dedicated reuse test), and the mid-retry-exit test drives the real `shellExecutionComplete()` lifecycle instead of mutating `busy`. Replied to the registry-coverage point inline (happy to add that test here or as a follow-up). All green locally (20/20 + tsc + lint). 🦘

proyectoauraorg:
> @edelauna asked: "What happens if the terminal already had something running that has not completed?"

Good point — the fix already guards against this. In the updated code, the terminal reuse check verifies `terminal.process` *before* launching. If the terminal is still running a previous command, we skip reuse entirely and create a fresh terminal:

```ts
if (
    terminal.terminal.exitStatus == null &&
    terminal.process == null
) {
    // Safe to reuse — no active process
} else {
    // Create new terminal
}
```

The `terminal.process == null` check ensures we never piggyback on a busy terminal. If a previous task left a long-running process (like a dev server), the guard rejects that terminal and spins up a new one, so there's no risk of interleaved output or the old process getting SIGINT'd.

src/integrations/terminal/TerminalProcess.ts
```diff
@@ -256,9 +269,60 @@ export class TerminalProcess extends BaseTerminalProcess {
...
+		// and keep running. Kick off a bounded retry that re-sends Ctrl+C a few
+		// times, verifying between attempts whether the process actually exited
+		// (terminal.busy flips to false on completion). This is intentionally
+		// fire-and-forget so it never blocks the synchronous cancel path; the
+		// total retry window is bounded so dispose() is never delayed for long.
+		if (!this.aborting) {
+			this.aborting = true
+			void this.retryAbort().finally(() => {
+				this.aborting = false
+			})
```
edelauna: If `sendText` at line 325 throws (e.g. the VS Code terminal is closed between the guard check and the send), `retryAbort` rejects, `finally` resets `aborting`, and then the post-`finally` promise is silently dropped by `void` — an unhandled rejection. Worth adding `.catch((err) => console.error(...))` after `.finally()` so errors surface rather than disappear:

```suggestion
			void this.retryAbort()
				.finally(() => {
					this.aborting = false
				})
				.catch((err) => console.error("[TerminalProcess] retryAbort error:", err))
```

edelauna (approved):
Nice - thanks for addressing.

PR #274, opened 2026-05-24
feat(zai): expose configurable max output tokens for GLM models (#161)

### LINKED ISSUE #161: [ENHANCEMENT] Make Z.ai GLM max output configurable in settings

Z.ai GLM models now default to the corrected 20% output clamp, but users cannot override that value from the normal settings UI. The runtime already respects an explicit `modelMaxTokens`, so there is a gap between what the backend supports and what the settings surface exposes.

This affects Zoo users on the `zai` provider, especially with `glm-5.1` and `glm-5-turbo`, when they want something other than the default reserved output budget. Today those models only expose the reasoning-effort control in settings, so users cannot tune max output the same way they can for reasoning-budget models.

## Desired behavior (conceptual, not technical)

When a Z.ai GLM model supports a configurable max output budget, users should be able to adjust that value from settings instead of being locked to the default clamp. The default should stay conservative, but an explicit user choice should win.

## Constraints / preferences (optional)

- Keep the current default clamp behavior as the out-of-the-box setting.
- Follow the existing cached settings pattern in `SettingsView` so edits are buffered until save.
- Avoid making the Z.ai UI feel inconsistent with other providers that expose similar output-budget controls.

## Acceptance criteria (optional)

Given a Z.ai GLM model such as `glm-5.1` or `glm-5-turbo`
When the user opens settings
Then there is a user-facing control for max output tokens
And changing it persists `modelMaxTokens`
And the request path uses that explicit value instead of the 20% default
But if the user leaves it unset, Zoo continues using the default clamp

## Proposed approach (optional)

Expose a max-output control for Z.ai GLM models in the existing settings flow, even though they currently advertise `supportsReasoningEffort` rather than `supportsReasoningBudget`. The runtime path already honors `modelMaxTokens`, so the main missing piece is UI/configuration exposure and matching regression coverage in `webview-ui`.

## Trade-offs / risks (optional)

The main risk is confusing users if the control is shown for models where Zoo cannot actually honor it. The UI should stay scoped to models/providers where `modelMaxTokens` is meaningful and supported.

## Additional context

- Related fix: PR #160 corrected the default GLM reserve from the raw provider max back to Zoo's 20% clamp.
- Current confirmed behavior:
  - default for Z.ai GLM: clamp to 20%
  - runtime override: supported via `modelMaxTokens`
  - settings UI override: not currently exposed for normal Z.ai users

proyectoauraorg:
## Summary

Z.ai GLM models (`glm-5.1`, `glm-5-turbo`) default to a 20% output clamp. The runtime already honors an explicit `modelMaxTokens` (`createStreamWithThinking`: `max_tokens = options.modelMaxTokens || getModelMaxOutputTokens(...)`), but the settings UI never surfaced a control — those models only exposed reasoning effort. This closes that gap.

## Change

- New optional capability flag `supportsMaxTokens` on `modelInfoSchema`, set on the configurable GLM models in both `internationalZAiModels` and `mainlandZAiModels`.
- Reuses the existing `ThinkingBudget` settings component to render a max-output-tokens slider, gated on `supportsMaxTokens && !supportsReasoningBudget`. It defaults to the runtime's existing clamp when unset (behavior unchanged until the user edits), and persists an explicit value as `modelMaxTokens`.

## Tests
- Backend (`zai.spec.ts`): the flag is present on the configurable GLM models (absent on glm-4.7), and an explicit `modelMaxTokens` override is sent as `max_tokens` instead of the clamp.
- Webview (`ThinkingBudget.spec.tsx`): slider renders alongside reasoning effort, defaults to the clamp when unset, reflects an override, does NOT persist on initial render (init vs user-edit), persists on user change, and does not render when the flag is absent.

Closes #161

## Summary by CodeRabbit

* **New Features**
  * Added a model capability flag for configurable max-output-tokens; supported models show a dedicated max-output-tokens slider that defaults to a runtime-clamped value and honors explicit modelMaxTokens overrides (capped by model limits).

* **Chores**
  * Export logic refined to preserve or strip modelMaxTokens and thinking-token fields based on model capabilities.

* **Tests**
  * Added tests for capability advertising, slider rendering/behavior and persistence, max-token override/capping, and export rules.


webview-ui/src/components/settings/ThinkingBudget.tsx
```diff
@@ -160,10 +165,40 @@ export const ThinkingBudget = ({ apiConfiguration, setApiConfigurationField, mod
...
+						value={[standaloneMaxOutputTokens]}
+						onValueChange={([value]) => setApiConfigurationField("modelMaxTokens", value)}
+					/>
+					<div className="w-12 text-sm text-center">{standaloneMaxOutputTokens}</div>
+				</div>
+			</div>
+		) : null
+
 	// Models with supportsReasoningBinary (binary reasoning) show a simple on/off toggle
 	if (isReasoningSupported) {
```
edelauna, line 211: If a model ever has both `supportsReasoningBinary: true` and `supportsMaxTokens: true`, the standalone slider won't render — this block returns before `maxOutputTokensControl` is reached. Worth including `{maxOutputTokensControl}` here alongside the checkbox, even if no current model hits this?

edelauna:
**Profile export strips `modelMaxTokens` for GLM models** (`src/core/config/ProviderSettingsManager.ts:572`)

The export cleanup at line 572 deletes `modelMaxTokens` when `supportsReasoningBudget` is falsy. GLM models only advertise `supportsMaxTokens` + `supportsReasoningEffort` — neither reasoning-budget flag — so any custom max-tokens value the user sets via the new slider will be silently stripped on profile export. Re-importing the profile then resets the slider to the default clamp.

Suggested fix:
```ts
const keepMaxTokens = supportsReasoningBudget || modelInfo.supportsMaxTokens
if (!supportsReasoningBudget) {
    delete configs[name].modelMaxThinkingTokens
}
if (!keepMaxTokens) {
    delete configs[name].modelMaxTokens
}
```

(This file isn't in the PR diff so can't be posted as an inline comment.)

proyectoauraorg:
While reviewing the round-trip behavior for this feature, I found that `ProviderSettingsManager.export()` strips `modelMaxTokens` for any model without a reasoning budget — which would silently wipe the GLM max-output value this PR introduces on export/import. I've split the two deletions so `modelMaxTokens` is preserved whenever the model sets `supportsMaxTokens` (as GLM does), while `modelMaxThinkingTokens` still requires a reasoning budget. Added tests covering both branches (GLM preserves, a plain model strips both).

webview-ui/src/components/settings/ThinkingBudget.tsx
proyectoauraorg: Great catch — you're right that `supportsReasoningBinary` models (like deepseek-v3.2) would have their `modelMaxThinkingTokens` incorrectly stripped under the current logic.\n\nI've updated the guard to include `supportsReasoningBinary` in the reasoning-capable check:\n\n```ts\nconst supportsReasoning =\n    modelInfo.supportsReasoningBudget ||\n    modelInfo.requiredReasoningBudget ||\n    modelInfo.supportsReasoningBinary\n\nif (supportsReasoning) {\n    // keep modelMaxThinkingTokens — budget, required, or binary\n} else {\n    delete configs[name].modelMaxThinkingTokens\n}\n```\n\nThis ensures binary-reasoning models retain the field while non-reasoning models still get it cleaned up. Thanks for flagging this!

proyectoauraorg:
@edelauna Thanks for the detailed code review identifying the `modelMaxTokens` bug! 

I've created a fix PR that addresses the scoping issue: **[PR #332](https://github.com/Zoo-Code-Org/Zoo-Code/pull/332)**

The fix extracts `maxOutputTokensControl` as a named JSX variable declared before both the binary reasoning and budget reasoning conditional branches, making the max output tokens slider available for all reasoning models (including `supportsReasoningBinary` models like Zhipu GLM).

Could you please review PR #332 when you get a chance? 🙏

proyectoauraorg:
Fixed the Windows-only `platform-unit-test` failure (`ProviderSettingsManager > Export > should preserve modelMaxTokens … (e.g. GLM)`). It passed on ubuntu/local but failed on Windows.

**Root cause:** Windows CI runs vitest with `poolOptions.forks.singleFork` (all spec files share one process). `ProviderSettingsManager.spec` used the real `buildApiHandler`, so a sibling spec's module-level `vi.mock("../../../api")` (e.g. `importExport.spec`) could bleed in under reduced isolation — making `modelInfo.supportsMaxTokens` read as `undefined`, so `export()` dropped `modelMaxTokens` and the assertion failed. I reproduced the exact contamination locally by forcing `--pool=forks --poolOptions.forks.singleFork --no-isolate`.

**Fix:** mock `buildApiHandler` directly in `ProviderSettingsManager.spec`, deriving model capabilities from the real `@roo-code/types` definitions via `vi.importActual`. The Export tests now exercise the token-field filtering deterministically and are immune to cross-file `api` mocks, while still validating real capabilities (glm-5.1 `supportsMaxTokens`, claude-3-5-haiku neither). Locally: tsc + eslint clean, `core/config` 231/231, and the suite passes under the exact Windows `singleFork` config.

webview-ui/src/components/settings/ThinkingBudget.tsx
```diff
@@ -160,10 +165,40 @@ export const ThinkingBudget = ({ apiConfiguration, setApiConfigurationField, mod
...
+		isMaxTokensConfigurable && modelInfo.maxTokens ? (
+			<div className="flex flex-col gap-1" data-testid="max-output-tokens">
+				<div className="font-medium">{t("settings:thinkingBudget.maxTokens")}</div>
+				<div className="flex items-center gap-1">
+					<Slider
+						min={1024}
+						max={modelInfo.maxTokens}
+						step={1024}
+						value={[standaloneMaxOutputTokens]}
+						onValueChange={([value]) => setApiConfigurationField("modelMaxTokens", value)}
```
edelauna: When GLM persists `modelMaxTokens` here, should `getModelMaxOutputTokens` also honor `supportsMaxTokens`? Z.ai sends the override as `max_tokens`, but task context checks call the shared helper and still clamp non-reasoning-budget models.

src/core/config/__tests__/ProviderSettingsManager.spec.ts
```diff
@@ -6,6 +6,37 @@ import type { ProviderSettings } from "@roo-code/types"
...
+			case "zai":
+				return zaiModels[id] ?? {}
+			case "anthropic":
+				return anthropicModels[id] ?? {}
+			default:
+				return {}
+		}
+	}
+	return {
+		buildApiHandler: (config: any) => ({
```
edelauna, line 30: Could this test use the real `buildApiHandler` and mock only external clients instead? Since `export()` depends on `buildApiHandler(config).getModel().info`, this fake bypasses the contract the test is trying to cover.

webview-ui/src/components/settings/ThinkingBudget.tsx
```diff
@@ -160,10 +165,40 @@ export const ThinkingBudget = ({ apiConfiguration, setApiConfigurationField, mod
...
+		DEFAULT_HYBRID_REASONING_MODEL_MAX_TOKENS
+	const standaloneMaxOutputTokens = apiConfiguration.modelMaxTokens || defaultMaxOutputTokens
+
 	if (!modelInfo) {
 		return null
 	}
 
+	// Standalone max output tokens slider for models that advertise `supportsMaxTokens`
+	// (e.g. Z.ai GLM) but do not surface the reasoning-budget control.
+	const maxOutputTokensControl =
```
edelauna, line 203: Is it worth extracting the max-tokens slider markup into a small local render helper? This repeats the same label, `Slider`, value display, and `modelMaxTokens` wiring used in the reasoning-budget branch below.

proyectoauraorg:
Thanks! Addressed the review:

- Extracted a `renderMaxTokensSlider` helper in ThinkingBudget shared by the standalone control and the reasoning-budget branch, and surfaced `{maxOutputTokensControl}` in the binary-reasoning branch so a model with both `supportsReasoningBinary` and `supportsMaxTokens` still gets the slider.
- `getModelMaxOutputTokens` now honors `supportsMaxTokens`: it returns the user's `modelMaxTokens` override (capped at the model ceiling) instead of the 20% context-window clamp, so the runtime budget matches what Z.ai receives as `max_tokens`. Added unit tests.
- Trimmed the `ProviderSettingsManager` mock comment to the durable reason.
- On the real `buildApiHandler` (vs the mock): I kept the explicit mock — it already drives the filtering off the real `@roo-code/types` model definitions, keeps this suite isolated from sibling specs that mock `../../../api` (the singleFork cross-file leak), and the real handler → `getModel().info` contract for GLM is now covered directly in `zai.spec.ts`. Switching the Export tests to the real handler would also build a handler for the retired `groq` profile case. Happy to switch if you'd prefer — let me know.

Squashed to one commit. `tsc`, `eslint`, and the affected suites (api / ProviderSettingsManager / zai / ThinkingBudget — 142 tests) pass locally.

proyectoauraorg:
👋 Friendly ping — this PR has been open for 9 days.

**Summary:** Exposes configurable `maxOutputTokens` for GLM models, addressing [#161](https://github.com/Zoo-Code-Org/Zoo-Code/issues/161).

- ✅ CI green
- ✅ Backward-compatible (uses existing config infrastructure)

src/shared/api.ts
```diff
@@ -132,6 +132,13 @@ export const getModelMaxOutputTokens = ({
 		return ANTHROPIC_DEFAULT_MAX_TOKENS
 	}
 
+	// Models that expose a configurable max-output slider (e.g. Z.ai GLM) honor the user's
+	// explicit override instead of the default 20% context-window clamp, capped at the model's
+	// own ceiling. This keeps the runtime budget consistent with the value sent to the provider.
+	if (model.supportsMaxTokens && settings?.modelMaxTokens) {
```
edelauna: `settings?.modelMaxTokens` is falsy when `0` — could a hand-edited import file set this to `0` and silently fall through to the 20% clamp?

```suggestion
	if (model.supportsMaxTokens && settings?.modelMaxTokens != null && settings.modelMaxTokens > 0) {
```

webview-ui/src/components/settings/ThinkingBudget.tsx
```diff
@@ -160,22 +165,63 @@ export const ThinkingBudget = ({ apiConfiguration, setApiConfigurationField, mod
...
+	// Default max output budget for models that expose a standalone max-tokens slider.
+	// When the user hasn't set an explicit `modelMaxTokens`, fall back to the same value
+	// the runtime would use (the default output clamp) so behavior is unchanged.
+	const defaultMaxOutputTokens =
+		(isMaxTokensConfigurable && selectedModelId && modelInfo
+			? getModelMaxOutputTokens({ modelId: selectedModelId, model: modelInfo, settings: apiConfiguration })
+			: undefined) ??
+		modelInfo?.maxTokens ??
+		DEFAULT_HYBRID_REASONING_MODEL_MAX_TOKENS
+	const standaloneMaxOutputTokens = apiConfiguration.modelMaxTokens || defaultMaxOutputTokens
```
edelauna: Same `||` vs `??` concern — `modelMaxTokens: 0` would silently use the default.

```suggestion
	const standaloneMaxOutputTokens = apiConfiguration.modelMaxTokens ?? defaultMaxOutputTokens
```

src/core/config/__tests__/ProviderSettingsManager.spec.ts
edelauna: @proyectoauraorg - this feedback was never addressed fyi - it's non blocking, was waiting on it for my next review.

edelauna (approved):

PR #275, opened 2026-05-24
feat(markdown): render GitHub-style alerts in the webview (#258)

### LINKED ISSUE #258: [ENHANCEMENT] Add GitHub-style Markdown alert rendering
Zoo Code currently renders GitHub-style Markdown alerts as regular blockquotes instead of visually distinct alert blocks.

This makes pasted or generated GitHub-flavored Markdown harder to scan when it contains alert syntax such as `[!NOTE]`, `[!TIP]`, `[!IMPORTANT]`, `[!WARNING]`, or `[!CAUTION]`, because the semantic meaning and visual priority are lost in the rendered output.

Support GitHub-style Markdown alerts in the existing Markdown rendering pipeline used by the webview.

The following alert types should be recognized:
- `NOTE`
- `TIP`
- `IMPORTANT`
- `WARNING`
- `CAUTION`

Supported alerts should render as distinct alert blocks rather than normal blockquotes.

Normal blockquotes that do not start with one of the supported alert markers should continue to render exactly as they do today.

If possible, this should stay aligned with the current Markdown rendering approach and use VS Code theme variables so the result remains readable in light, dark, and high-contrast themes.

One option would be to introduce a dedicated remark plugin for GitHub-style alerts.

Another option would be to implement this locally in the current Markdown pipeline by detecting supported alert markers in blockquotes and annotating them for custom rendering.

A local implementation may be preferable if it keeps the change small and avoids adding a new dependency, provided it remains easy to maintain and does not affect normal blockquote rendering.

**Additional context**
This mostly affects README-style content, generated documentation, issue content, changelogs, and task summaries that use GitHub-flavored Markdown alerts.

Example input:

> [!NOTE]
> This is useful information.

> [!WARNING]
> Be careful before running this command.

Expected behavior:
- supported alert markers render as distinct alert blocks
- normal blockquotes remain unchanged
- malformed or unsupported markers fall back gracefully
- alert content can include common Markdown such as inline code, bold text, links, lists, and multiline content

proyectoauraorg:
## Summary

GitHub-style alerts (`> [!NOTE]`, `> [!TIP]`, `> [!IMPORTANT]`, `> [!WARNING]`, `> [!CAUTION]`) were rendered as plain blockquotes, losing their semantic meaning and visual priority.

## Change

- A focused custom remark transform (`remarkGithubAlerts`, **no new dependency**) walks the mdast, detects a leading alert marker (case-insensitive) in a blockquote's first paragraph, strips the marker, and tags the node with `data-alert-type` + `markdown-alert*` classes.
- A `blockquote` component customization in `MarkdownBlock` renders a codicon + label header and per-type accent styling using VS Code theme variables (light/dark/high-contrast friendly).
- **Normal blockquotes and unsupported markers render exactly as before.**

## Tests
- `markdown.spec.ts`: each alert type, case-insensitivity, marker-on-own-line, nested alerts, unsupported markers, mid-text markers, and normal blockquote unchanged.
- `MarkdownBlock.spec.tsx`: all five types (class + icon + stripped marker), case-insensitivity, inline markdown inside alerts, normal blockquote unchanged, unsupported-marker fallback.

Closes #258

## Summary by CodeRabbit

* **New Features**
  * Markdown renderer now supports GitHub-style alert blockquotes (NOTE, TIP, IMPORTANT, WARNING, CAUTION) with distinct icon, title header, and color-coded styling for clearer inline guidance; handles nested alerts and ignores markers embedded mid-text.
* **Tests**
  * Added comprehensive tests validating alert detection (case-insensitive), correct styling/attributes, icon rendering, suppression of raw markers, inline markdown rendering inside alerts, and non-alert blockquote behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/275?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/webview-ui/src/components/common/MarkdownBlock.tsx
+++ b/webview-ui/src/components/common/MarkdownBlock.tsx
@@ -7,9 +7,28 @@ import remarkMath from "remark-math"
 import remarkGfm from "remark-gfm"
 
 import { vscode } from "@src/utils/vscode"
+import { type AlertType, remarkGithubAlerts } from "@src/utils/markdown"
 
 import CodeBlock from "./CodeBlock"
 import MermaidBlock from "./MermaidBlock"
+
+// Codicon glyphs used as the leading icon for each GitHub-style alert type.
+const ALERT_ICONS: Record<AlertType, string> = {
+	note: "codicon-info",
+	tip: "codicon-lightbulb",
+	important: "codicon-report",
+	warning: "codicon-warning",
+	caution: "codicon-flame",
+}
+
+// Human-readable label shown in the alert header.
+const ALERT_LABELS: Record<AlertType, string> = {
+	note: "Note",
+	tip: "Tip",
+	important: "Important",
+	warning: "Warning",
+	caution: "Caution",
+}
 
 interface MarkdownBlockProps {
 	markdown?: string
@@ -201,6 +220,57 @@ const StyledMarkdown = styled.div`
 	tr:hover {
 		background-color: var(--vscode-list-hoverBackground);
 	}
+
+	/* GitHub-style Markdown alerts (#258). The accent color per type is set via
+	   the --alert-accent custom property on the element itself. */
+	.markdown-alert {
+		margin: 1em 0;
+		padding: 0.5em 1em;
+		border-left: 0.25em solid var(--alert-accent, var(--vscode-textBlockQuote-border));
+		border-radius: 3px;
+		background-color: var(--vscode-textBlockQuote-background);
+	}
+
+	.markdown-alert > :first-child {
+		margin-top: 0;
+	}
+
+	.markdown-alert > :last-child {
+		margin-bottom: 0;
+	}
+
+	.markdown-alert-title {
+		display: flex;
+		align-items: center;
+		gap: 0.5em;
+		font-weight: 600;
+		color: var(--alert-accent, var(--vscode-foreground));
+		margin-bottom: 0.25em;
+	}
+
+	.markdown-alert-title .codicon {
+		font-size: 1em;
+	}
+
+	.markdown-alert-note {
+		--alert-accent: var(--vscode-charts-blue, var(--vscode-textLink-foreground));
+	}
+
+	.markdown-alert-tip {
+		--alert-accent: var(--vscode-charts-green, var(--vscode-terminal-ansiGreen));
+	}
+
+	.markdown-alert-important {
+		--alert-accent: var(--vscode-charts-purple, var(--vscode-textLink-foreground));
+	}
+
+	.markdown-alert-warning {
+		--alert-accent: var(--vscode-charts-yellow, var(--vscode-editorWarning-foreground));
+	}
+
+	.markdown-alert-caution {
+		--alert-accent: var(--vscode-charts-red, var(--vscode-editorError-foreground));
+	}
 `
 
 const MarkdownBlock = memo(({ markdown }: MarkdownBlockProps) => {
@@ -299,6 +369,31 @@ const MarkdownBlock = memo(({ markdown }: MarkdownBlockProps) => {
 					</code>
 				)
 			},
+			blockquote: ({ children, className, ...props }: any) => {
+				// The remarkGithubAlerts plugin tags alert blockquotes with a
+				// `data-alert-type` attribute and `markdown-alert*` classes.
+				// Anything without that attribute is a normal blockquote and
+				// must render unchanged.
+				const alertType = props["data-alert-type"] as AlertType | undefined
+
+				if (!alertType || !(alertType in ALERT_ICONS)) {
+					return (
+						<blockquote className={className} {...props}>
+							{children}
+						</blockquote>
+					)
+				}
+
+				return (
+					<blockquote className={className} {...props}>
+						<div className="markdown-alert-title">
+							<span className={`codicon ${ALERT_ICONS[alertType]}`} aria-hidden="true" />
+							<span>{ALERT_LABELS[alertType]}</span>
+						</div>
+						{children}
+					</blockquote>
+				)
+			},
 		}),
 		[],
 	)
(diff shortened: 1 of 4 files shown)


webview-ui/src/utils/markdown.ts
```diff
@@ -21,3 +21,91 @@ export function countMarkdownHeadings(text: string | undefined): number {
...
+ * GitHub-style Markdown alert types, mapped to their lower-cased identifiers.
+ * @see https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts
+ */
+export const ALERT_TYPES = ["note", "tip", "important", "warning", "caution"] as const
+
+export type AlertType = (typeof ALERT_TYPES)[number]
+
+// Matches a leading alert marker like "[!NOTE]" (case-insensitive) optionally
+// followed by trailing whitespace/newline on the first line of a blockquote.
+const ALERT_MARKER_REGEX = /^\[!(note|tip|important|warning|caution)\][^\S\r\n]*\r?\n?/i
```
edelauna: The alternation group here (`note|tip|important|warning|caution`) duplicates `ALERT_TYPES` — adding a new type requires updating both. Could this be derived instead?

```suggestion
const ALERT_MARKER_REGEX = new RegExp(`^\\[!(${ALERT_TYPES.join("|")})\\][^\\S\\r\\n]*\\r?\\n?`, "i")
```

webview-ui/src/utils/markdown.ts
```diff
@@ -21,3 +21,91 @@ export function countMarkdownHeadings(text: string | undefined): number {
...
+ *
+ * The marker text is stripped from the rendered content and the recognized
+ * alert type is exposed via the `data-alert-type` attribute plus matching
+ * `markdown-alert*` class names on the emitted `<blockquote>` element.
+ *
+ * Blockquotes that do not begin with a supported marker are left untouched, so
+ * normal blockquotes continue to render exactly as before.
+ */
+export function remarkGithubAlerts() {
+	return (tree: any) => {
```
edelauna: `@types/mdast` is available transitively via `remark-gfm`, and `unist-util-visit` v5 is already a dependency. Should the transformer be typed as `Plugin<[], Root>` with `Root` from `mdast`, so callers get compile-time safety when passing this to `remarkPlugins`?

webview-ui/src/utils/markdown.ts
```diff
@@ -21,3 +21,91 @@ export function countMarkdownHeadings(text: string | undefined): number {
...
+
+	if (node.type === "blockquote") {
+		annotateAlertBlockquote(node)
+	}
+
+	if (Array.isArray(node.children)) {
+		for (const child of node.children) {
+			walkAlertBlockquotes(child)
+		}
+	}
```
edelauna: `unist-util-visit` is already imported in `MarkdownBlock.tsx` and listed in `package.json`. Would `visit(tree, "blockquote", annotateAlertBlockquote)` work here instead of the hand-rolled walker?

webview-ui/src/utils/__tests__/markdown.spec.ts
```diff
@@ -30,3 +41,78 @@ describe("markdown heading helpers", () => {
 		expect(hasComplexMarkdown("# One\n## Two")).toBe(true)
 	})
 })
+
+describe("remarkGithubAlerts", () => {
+	it.each(ALERT_TYPES)("annotates a [!%s] alert blockquote", (type) => {
+		const upper = type.toUpperCase()
+		const tree = root(blockquote(paragraph(text(`[!${upper}]\nBody text`))))
```
edelauna, line 48: Does remark actually produce `paragraph(text("[!NOTE]\nBody text"))` — a single text node with an embedded newline? My understanding is it emits `[text("[!NOTE]"), break, text("Body text")]` across separate nodes. If so, this tree shape is never produced by the real parser. The integration tests in `MarkdownBlock.spec.tsx` do cover the real parse path, but it might be worth a quick sanity check here.

webview-ui/src/components/common/MarkdownBlock.tsx
```diff
@@ -299,6 +369,31 @@ const MarkdownBlock = memo(({ markdown }: MarkdownBlockProps) => {
...
+				if (!alertType || !(alertType in ALERT_ICONS)) {
+					return (
+						<blockquote className={className} {...props}>
+							{children}
+						</blockquote>
+					)
+				}
+
+				return (
+					<blockquote className={className} {...props}>
```
edelauna: `...props` spreads all hast-derived attributes onto the DOM element. Today only `data-alert-type` and `className` are set by the plugin, but any future `hProperty` addition would silently reach the DOM. Would it be safer to destructure only the known props and discard the rest?

edelauna (commented):
Nice - can you add some screenshots of it working.

proyectoauraorg:
Thanks for the review @edelauna! Here are responses to your feedback:

**Screenshots:** Will add screenshots of the alert blockquote rendering (NOTE, TIP, WARNING, CAUTION variants) — aiming to have those up by EOD.

**Code comments:**

1. **ALERT_MARKER_REGEX duplication** — Good catch. Will derive it from `ALERT_TYPES` as you suggested so a single source of truth is maintained.

2. **`@types/mdast` typing** — Agreed, will type the transformer as `Plugin<[], Root>` with `Root` from mdast for compile-time safety.

3. **Use `unist-util-visit`** — Yes, the hand-rolled walker should be replaced. `visit(tree, "blockquote", annotateAlertBlockquote)` is cleaner and avoids duplicate traversal logic.

4. **AST shape sanity check** — Good eye. Will verify the actual remark output structure and adjust the test fixtures to match real parser output rather than assumed tree shapes.

5. **`...props` spread safety** — Makes sense. Will destructure only known props (`data-alert-type`, `className`) and discard the rest to prevent future `hProperty` additions from leaking to the DOM.

proyectoauraorg:

edelauna (approved):
Looks good - thank you!

Image

webview-ui/src/utils/__tests__/markdown.spec.ts
edelauna: Verified against the real parser — remark does produce paragraph(text("[!NOTE]\nBody text")) as a single text node with an embedded newline. The break node only appears when there's an explicit two-space trailing newline (a hard line break in markdown syntax). The unit test shape is accurate.

PR #276, opened 2026-05-24
feat(settings): add configurable chat font size (#157)

### LINKED ISSUE #157: [ENHANCEMENT] Adjust the font size in Zoo Code extension

The font size within the Zoo-Code extension is a bit small, especially on small screen or laptop and it can not adjust by the VS Code settings. 
I don't find any settings in the extension which can adjust the font size.

I search the internet find this is an old issue in former Roo-Code project. Many users had submitted issue on that project but not solved yet now.
Related Issues:
https://github.com/RooCodeInc/Roo-Code/issues/5783
https://github.com/RooCodeInc/Roo-Code/issues/7530
https://github.com/RooCodeInc/Roo-Code/issues/8100

The font size is quiet small comparing with the code font size in VS Code.
The only way to solve this is to enlarge the font size with `Cmmmand +`, however, the code font size is enlarged as well.

Can anyone pay attention to this UI issue?
Zoo-Code is a brilliant extension in code agent area. But new user are likely retreated by the UI issue and lose such a remarkable tool for coding.
I notice some developer submitted a PR but refused with some developing issues.(https://github.com/RooCodeInc/Roo-Code/pull/8103)

### Request checklist

_No response_

_No response_

I wish the extension can adjust the font size in its setting, only 3 options(small, normal, large) are much better than now. 

_No response_

proyectoauraorg:
## Summary

The Zoo Code chat font was small on laptops/small screens and could only be enlarged by zooming the entire VS Code UI (which also scales the editor). This adds an independent, configurable chat font size.

## Change
- New optional `chatFontSize` setting (integer px, bounds 8–32, `nullish`) in global settings, threaded through `ClineProvider` state.
- CSS variable `--zoo-chat-font-size` defaults to `--vscode-font-size`; the webview text scale (`--text-xs/sm/base/lg`) and `body` font-size derive from it. **Appearance is unchanged when unset.**
- A `useEffect` applies the override to the document root when set and removes it on reset.
- Surfaced in Settings → UI as a slider + value readout + **Use VS Code default** reset, following the `allowedMaxRequests` nullish/reset pattern.

## Tests & i18n
- `UISettings.spec.tsx`: init default display, configured value, user-edit persistence, reset enable/clear.
- `ExtensionStateContext.spec.tsx`: CSS var not set on init (unset), set on edit, cleared on reset.
- i18n strings added for **all 18 locales** (`find-missing-translations` passes).

Closes #157

## Summary by CodeRabbit

* **New Features**
  * Users can customize chat font size (8–32 px) and reset to VS Code default.

* **Style**
  * Added a CSS variable to apply the customizable chat font size across the UI.

* **Localization**
  * Added translations for the new setting in 19 languages.

* **Tests**
  * Added unit and integration tests covering the chat font size UI, persistence, telemetry, and reset behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/276?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


webview-ui/src/index.css
```diff
@@ -253,11 +253,18 @@ html {
...
+	 * font size so appearance is unchanged unless the user configures the
+	 * "Chat font size" setting, which overrides this variable at runtime.
+	 */
+	--zoo-chat-font-size: var(--vscode-font-size);
 }
 
 body {
 	margin: 0;
 	line-height: 1.25;
+	font-size: var(--zoo-chat-font-size);
```
edelauna: Does this reach the main chat markdown path? `MarkdownBlock` still sets `font-size: var(--vscode-font-size, 13px)`, so I’m wondering whether normal chat messages will keep using the VS Code font size even after this setting is saved.

webview-ui/src/index.css
```diff
@@ -25,10 +25,10 @@
 @theme {
 	--font-display: var(--vscode-font-family);
 
-	--text-xs: calc(var(--vscode-font-size) * 0.85);
-	--text-sm: calc(var(--vscode-font-size) * 0.9);
-	--text-base: var(--vscode-font-size);
-	--text-lg: calc(var(--vscode-font-size) * 1.1);
+	--text-xs: calc(var(--zoo-chat-font-size) * 0.85);
```
edelauna: Should this be scoped closer to the chat surfaces instead of the global Tailwind text tokens? A chat readability preference here can also resize settings and other non-chat webview UI that use `text-xs`/`text-sm`/`text-base`.

webview-ui/src/components/settings/__tests__/UISettings.spec.tsx
```diff
@@ -41,4 +41,50 @@ describe("UISettings", () => {
...
+			// Default falls back to VS Code-equivalent default value.
+			expect(getByText("13px")).toBeTruthy()
+		})
+
+		it("shows the configured font size when set", () => {
+			const { getByText } = render(<UISettings {...defaultProps} chatFontSize={20} />)
+			expect(getByText("20px")).toBeTruthy()
+		})
+
+		it("persists a user-edited font size via setCachedStateField", () => {
```
edelauna, line 63: Could we add a `SettingsView`-level test for this path as well? This proves the child calls `setCachedStateField`, but it would be useful to cover the cached-state/save boundary and assert the `updateSettings` payload includes the selected size and reset `null`.

edelauna (commented):
Should this be modeled as a scale relative to VS Code’s current font size instead of an absolute pixel value? If the default is meant to inherit VS Code, using a multiplier/percent would preserve the user’s existing VS Code font baseline across machines while still allowing chat-specific enlargement.

Also could you include some screen recordings of how this looks on max size.

webview-ui/src/index.css
proyectoauraorg: Good catch — that was the core bug: `MarkdownBlock` hard-coded `--vscode-font-size`, so chat messages never picked up the setting. It now uses `font-size: var(--zoo-chat-font-size, var(--vscode-font-size, 13px))`, so the chat markdown surface tracks the setting (with the VS Code default as fallback).

webview-ui/src/index.css
proyectoauraorg: Agreed — it was too broad. Reverted the global override: `@theme --text-xs/sm/base/lg` and `body { font-size }` go back to `--vscode-font-size`, so settings and other non-chat UI are unaffected. The `--zoo-chat-font-size` variable is now consumed only by `MarkdownBlock` (the chat message surface), which is the intended scope. If you'd prefer it to cover the whole ChatView container rather than just the rendered markdown, happy to widen it — that's a quick follow-up.

webview-ui/src/components/settings/__tests__/UISettings.spec.tsx
proyectoauraorg: Added. New SettingsView-level test ("saves the selected chat font size and persists null on reset"): it changes the slider on the `ui` tab, clicks Save, and asserts the `updateSettings` payload carries `chatFontSize: 18`; then clicks Reset + Save and asserts `chatFontSize: null` (matching `chatFontSize ?? null` in SettingsView). Covers the cached-state → save boundary you flagged.

proyectoauraorg:
@edelauna addressed all three in `03a1df2a9`:
- **MarkdownBlock** now reads `--zoo-chat-font-size` (with VS Code fallback) — chat messages actually respond to the setting now.
- **Scoped to chat**: reverted the global `@theme --text-*` and `body` overrides, so the setting no longer resizes settings/other webview UI.
- **SettingsView-level test** added covering the save boundary (size on save, `null` on reset).

All green locally (SettingsView/UISettings/ExtensionStateContext specs + tsc + eslint). One note: I verified behavior via the unit tests but not visually — if you'd like a screenshot of the chat at a couple of sizes, I can capture one. Ready for another look — thanks! 🦘

webview-ui/src/components/settings/UISettings.tsx
```diff
@@ -48,6 +55,14 @@ export const UISettings = ({
 		})
 	}
 
+	const handleChatFontSizeChange = (value: number) => {
+		setCachedStateField("chatFontSize", value)
+	}
+
+	const handleChatFontSizeReset = () => {
+		setCachedStateField("chatFontSize", undefined)
+	}
```
edelauna, line 72: The other two handlers in this file both emit a telemetry event — should these do the same? Something like `ui_settings_chat_font_size_changed` (with `value`) and `ui_settings_chat_font_size_reset` would keep the feature visible in analytics.

webview-ui/src/components/settings/__tests__/SettingsView.spec.tsx
```diff
@@ -412,6 +412,37 @@ describe("SettingsView - Sound Settings", () => {
...
+	it("saves the selected chat font size and persists null on reset", () => {
+		const { activateTab, getSettingsContent } = renderSettingsView()
+
+		activateTab("ui")
+
+		const content = getSettingsContent()
+		const slider = within(content).getByTestId("chat-font-size-slider")
+
+		// Pick a size, then Save — the boundary should forward it to the host.
+		fireEvent.change(slider, { target: { value: "18" } })
```
edelauna, line 424: Does `fireEvent.change` on the Radix `<Slider>` root element actually reach `onValueChange`? The matching `UISettings.spec.tsx` test drives the slider via `fireEvent.keyDown(slider, { key: "ArrowRight" })` on the `[role="slider"]` thumb — if this `change` event is silently ignored, the assertion below might be passing because `chatFontSize` was already `18` in the initial mock state rather than because the interaction worked.

edelauna (commented):
Nice - going to try and test this either tonight or tomorrow. Had some additional questions.

proyectoauraorg:

webview-ui/src/components/settings/__tests__/SettingsView.spec.tsx
edelauna: The `Slider` in this test file is mocked at the `@/components/ui` level as a native `<input type="range">` (see the `vi.mock('@/components/ui', ...)` block), so `fireEvent.change` on the slider root **does** reach `onValueChange` — the mock translates `e.target.value` directly to `onValueChange([parseFloat(value)])`. Confirmed: removing the `fireEvent.change` line causes the `chatFontSize: 18` assertion to fail. No change needed.

webview-ui/src/components/settings/UISettings.tsx
edelauna: Telemetry is implemented — `handleChatFontSizeChange` calls `telemetryClient.capture("ui_settings_chat_font_size_changed", { value })` and `handleChatFontSizeReset` calls `telemetryClient.capture("ui_settings_chat_font_size_reset")`, matching the pattern used by the other two handlers. Both are also covered by the `UISettings.spec.tsx` assertions on `telemetryClient.capture`.

edelauna (approved):
Cool little feature - the current implementation is scoped just to the chat window which is hopefully sufficient to close out the ticket. 

Image

Image

There is a noticeable delay between saving the setting and the view being updated. Filed the following issue to track it - since it's architectural for how settings are saved: https://github.com/Zoo-Code-Org/Zoo-Code/issues/416

PR #277, opened 2026-05-24
feat(terminal): add VS Code integrated-terminal shell override

### LINKED ISSUE #119: Support choosing terminal profile for inline terminal executaion
I use Chinese under Windows and the default charset is GBK but not utf-8, the default ZooCode inline terminal is cmd/powershell and will print wrong char for my utf-8 project.

I can use git bash to execute command in inline termianl, whose default encoding is utf-8.

No

**Additional context**

Image

proyectoauraorg:
## Summary

Adds an optional Zoo Code shell override for commands executed through VS Code's integrated terminal.

By default, Zoo Code continues to use VS Code's configured terminal profile. Users can opt into a Zoo-specific shell from resolvable, path-based VS Code profiles. This setting does not apply to inline execution, which uses the Execa provider.

## Key behavior

- Adds an optional `terminalProfile` setting. Leaving it unset preserves VS Code's default behavior.
- Resolves selected profiles to `shellPath`, `shellArgs`, and sanitized `env` values passed to `vscode.window.createTerminal`.
- Lists only resolvable profiles from VS Code defaults and user settings. Workspace-local profiles are ignored to prevent repositories from selecting an executable.
- Excludes source-only profiles and `cmd.exe`, which cannot provide the shell integration required by this execution path.
- Opens VS Code's native profile picker when users want to change their default profile, and clears the Zoo-specific override when they do.
- Starts a fresh idle terminal after an override changes so commands use the newly selected shell.
- Falls back to Execa only when shell integration fails before command submission. If submission status is unknown, Zoo Code reports the failure without rerunning the command.
- Avoids Zoo-specific `ZDOTDIR` injection while an override is active so VS Code can manage shell integration for the selected profile.

## Tests

Adds coverage for profile resolution and filtering, settings persistence and UI states, terminal recreation after profile changes, `cmd.exe` fallback, shell-integration startup handling, safe retry boundaries, and platform test-helper exit codes.

## Scope

This PR does not close #119. That issue concerns inline execution through `ExecaTerminalProcess` and requires a separate Execa-specific shell selection design.

Documentation follow-up: [Zoo-Code-Org/Zoo-Code-Docs#4](https://github.com/Zoo-Code-Org/Zoo-Code-Docs/issues/4)


webview-ui/src/components/settings/TerminalSettings.tsx
```diff
@@ -49,15 +63,20 @@ export const TerminalSettings = ({
...
 }: TerminalSettingsProps) => {
 	const { t } = useAppTranslation()
 
 	const [inheritEnv, setInheritEnv] = useState<boolean>(true)
+	const [profileNames, setProfileNames] = useState<string[]>([])
 
-	useMount(() => vscode.postMessage({ type: "getVSCodeSetting", setting: "terminal.integrated.inheritEnv" }))
+	useMount(() => {
+		vscode.postMessage({ type: "getVSCodeSetting", setting: "terminal.integrated.inheritEnv" })
+		PROFILE_SETTING_KEYS.forEach((setting) => vscode.postMessage({ type: "getVSCodeSetting", setting }))
```
edelauna: Could we route these through a small allowlisted request that returns only profile names? The existing `getVSCodeSetting` handler reads any key supplied by the webview, and this dropdown only needs sanitized names.

src/integrations/terminal/Terminal.ts
```diff
@@ -191,4 +212,79 @@ export class Terminal extends BaseTerminal {
...
+			return undefined
+		}
+
+		const shellArgs = Array.isArray(profile.args)
+			? profile.args
+			: typeof profile.args === "string"
+				? [profile.args]
+				: undefined
+
+		return { shellPath: pathValue, shellArgs }
```
edelauna: Do we want this to preserve the selected VS Code profile more fully? Profiles can include `source` and `env`, but this only returns `shellPath`/`shellArgs`, so some profiles shown in the dropdown may fall back or lose required environment.

webview-ui/src/i18n/locales/en/settings.json
```diff
@@ -801,6 +801,11 @@
 		"inheritEnv": {
 			"label": "Inherit environment variables",
 			"description": "Turn this on to inherit environment variables from the parent VS Code process. <0>Learn more</0>"
+		},
+		"profile": {
+			"label": "Inline terminal profile",
+			"default": "Default (VS Code default shell)",
+			"description": "Choose which VS Code terminal profile the inline terminal uses. Pick a UTF-8 shell such as Git Bash if the default shell garbles non-ASCII output (e.g. on Windows with a GBK code page). Leave on Default to keep VS Code's default behavior."
```
edelauna: Can we clarify this copy against the nearby Use Inline Terminal description? That text says inline mode bypasses shell profiles, while this new setting says the inline terminal uses a VS Code terminal profile.

src/core/webview/webviewMessageHandler.ts
```diff
@@ -726,6 +726,8 @@ export const webviewMessageHandler = async (
 						if (value !== undefined) {
 							Terminal.setTerminalZdotdir(value as boolean)
 						}
+					} else if (key === "terminalProfile") {
+						Terminal.setTerminalProfile(value as string | undefined)
```
edelauna: Could we add a test that exercises this through `updateSettings` rather than calling `Terminal.setTerminalProfile` directly? That would cover the save-time bridge from cached settings into the process-wide terminal state.

src/core/webview/ClineProvider.ts
```diff
@@ -757,6 +758,7 @@ export class ClineProvider
 				Terminal.setTerminalZshP10k(terminalZshP10k)
 				Terminal.setPowershellCounter(terminalPowershellCounter)
 				Terminal.setTerminalZdotdir(terminalZdotdir)
+				Terminal.setTerminalProfile(terminalProfile)
```
edelauna, line 796: Could we cover this startup hydration path as well? Without a test around `resolveWebviewView`, a saved `terminalProfile` could stop being applied after reload while the direct `Terminal` tests still pass.

edelauna (commented):
Nice this a good first pass - could you add a screenshot / screen recording of the terminal in action?

proyectoauraorg:
> Nice this a good first pass - could you add a screenshot / screen recording of the terminal in action?

Thanks for the review! Here's the terminal profile selector in action:

The new Inline terminal profile dropdown (Settings → Terminal → Advanced) set to zsh.
The inline terminal running echo "perfil activo: $0" && pwd, confirming the selected profile resolves to /bin/zsh.

zoo-277-terminal-profile
image
image

I'm also working through your inline comments — the allowlisted profile-name request, fuller VS Code profile preservation, the copy clarification, and the updateSettings / startup-hydration tests. Will push those in a follow-up. 🦘

webview-ui/src/components/settings/TerminalSettings.tsx
proyectoauraorg: Done in `4fcbb4b5` — added a dedicated allowlisted `requestTerminalProfiles` message. The extension now reads `terminal.integrated.profiles.*` server-side and returns only the sanitized profile names (`terminalProfiles` → `profiles: string[]`), so the dropdown no longer routes through the generic `getVSCodeSetting` handler. The webview no longer supplies arbitrary setting keys for this. 🦘

src/integrations/terminal/Terminal.ts
proyectoauraorg: Done in `4fcbb4b5` — `getProfileShell` now also returns the profile's `env` (sanitized to `string | null` values; a `null` unsets the variable), and `createTerminal` merges it on top of the base env so profile-specific environment isn't lost. Source-only profiles (no resolvable `path`) still fall back to the default, since fully resolving a `source` to a shell path needs VS Code internals — I kept that behavior with a clearer comment. Added a test covering env preservation + sanitization.

webview-ui/src/i18n/locales/en/settings.json
proyectoauraorg: Clarified in `4fcbb4b5` (en + es): the copy now states the Inline Terminal profile only selects the shell binary (path/args/env), while the Inline Terminal still bypasses shell integration, prompts, and plugins — so it no longer reads as contradicting the "Use Inline Terminal" description.

src/core/webview/webviewMessageHandler.ts
proyectoauraorg: Added in `4fcbb4b5` — a test that drives this through `updateSettings` (not by calling `Terminal.setTerminalProfile` directly) and asserts the process-wide terminal state is updated, covering both a profile name and the `undefined` (clear) case.

src/core/webview/ClineProvider.ts
proyectoauraorg: Added in `4fcbb4b5` — a test around `resolveWebviewView` that seeds a persisted `terminalProfile` and asserts it's applied via `Terminal.setTerminalProfile` during the startup hydration, so a saved profile keeps being applied after reload.

proyectoauraorg:
Thanks for the review @edelauna! Responses:

**Screenshot/Recording:** Will add a screen recording showing the terminal profile switching in action — targeting EOD.

**Code comments:**

1. **Allowlisted request for profile names** — Good call. Will route profile name lookups through a scoped handler that returns only sanitized names, rather than exposing the full `getVSCodeSetting` path to the webview.

2. **Preserve `source` and `env` from profiles** — Agreed. The current `shellPath`/`shellArgs` extraction is incomplete. Will extend to capture `source` and `env` so profiles that depend on those fields don't silently fall back.

3. **Copy clarity vs inline terminal description** — Will update the setting description to explicitly note the difference: inline terminal uses VS Code terminal profiles, while inline *mode* bypasses shell profiles. The wording should make it clear they're distinct behaviors.

4. **Test via `updateSettings`** — Makes sense. Will add an integration test that exercises the full save → bridge → `Terminal.setTerminalProfile` path instead of calling the method directly.

5. **Test `resolveWebviewView` hydration** — Will add a test covering the startup path where `terminalProfile` is hydrated from saved settings, ensuring it survives a reload.

webview-ui/src/components/settings/TerminalSettings.tsx
```diff
@@ -139,6 +151,38 @@ export const TerminalSettings = ({
 						</div>
 					</div>
 					<div className="flex flex-col gap-3 pl-3 border-l-2 border-vscode-button-background">
+						<SearchableSetting
```
edelauna: Should this block be wrapped in `{terminalShellIntegrationDisabled && (...)}` so the dropdown is hidden when the user has turned off the inline terminal? The other inline-terminal-only settings (`inheritEnv`, `shellIntegrationTimeout`) disappear under the `!terminalShellIntegrationDisabled` guard at line 213, but the profile picker stays visible and editable even when it has no effect.

proyectoauraorg:
Done — the profile picker is now guarded by `terminalShellIntegrationDisabled` (defaulting to shown, matching the checkbox), so it only appears in inline mode and disappears when inline execution is off, mirroring the other inline-only settings below. Added a test for the hidden state. Screen recording coming up. Thanks @edelauna! 🦓

src/core/webview/webviewMessageHandler.ts
```diff
@@ -1510,6 +1512,38 @@ export const webviewMessageHandler = async (
...
+					type: "terminalProfiles",
+					profiles: Array.from(names).sort(),
+				})
+			} catch (error) {
+				console.error("Failed to get terminal profiles:", error)
+				await provider.postMessageToWebview({ type: "terminalProfiles", profiles: [] })
+			}
+
+			break
+		}
```
edelauna, line 1557: The `updateSettings -> Terminal.setTerminalProfile` path has tests, but does this handler itself have any? The dedup+sort logic, the `catch` fallback (empty-array response), and edge cases like an all-null profiles object all seem untested -- worth adding a case to the `webviewMessageHandler - terminalProfile` suite that sends `requestTerminalProfiles` and asserts on what gets posted back.

webview-ui/src/components/settings/TerminalSettings.tsx
```diff
@@ -151,37 +151,43 @@ export const TerminalSettings = ({
...
-										{t("settings:terminal.profile.default")}
-									</SelectItem>
-									{profileNames.map((name) => (
-										<SelectItem key={name} value={name}>
-											{name}
+						{/* The profile picker only affects inline execution, which is active when shell
+						    integration is disabled. Hide it otherwise so it isn't shown but ineffective
+						    (mirrors the inline-only settings guarded below). Defaults to shown, matching
+						    the checkbox's `?? true`. See PR #277 review. */}
+						{(terminalShellIntegrationDisabled ?? true) && (
```
edelauna: why `?? true` ?

edelauna (commented):
could more questions, also lets close out the coderabbit feedback please

proyectoauraorg:

chenyuanrun:
Nice jobs!! ZooCode is purely unusable for my enviroment until this PR is merged, I am looking forward to this feature so I can switch from the buggy vscode termianl, Thanks for all of you!!

image

edelauna:
Ok I pushed an refactor update to get this PR on track since it was mixing 2 concerns. I think for this PR it would be great to specify the profile to use when integrated terminal is off. I updated the title and description, as well as a follow up for how to more specifically close issue #119 

On windows this PR allows specifying the default terminal to use, [created an issue in our docs repository](https://github.com/Zoo-Code-Org/Zoo-Code-Docs/issues/4) to show how to overwrite the default profile and select a terminal by setting powershell or gir in `settings.json`

image

image


image

proyectoauraorg:
Thanks for taking the wheel on this @edelauna — splitting the two concerns and narrowing to "specify the profile to use when the integrated terminal is off" is the right call, and the refactor reads cleanly. Pulled your 17 commits and went through them locally:

- The typed `ShellIntegrationErrorDetails` payload (with `commandSubmitted`) and the event-based shell-integration wait are solid improvements over the old string / `pWaitFor` path.
- The `resolveWebviewView` hydration + `updateSettings` → `Terminal.setTerminalProfile` bridge line up with the two test paths you'd flagged earlier — 👍

Verified on the refactored branch locally:
- `TerminalProfile.spec.ts` + `TerminalRegistry.spec.ts` → 47 passing
- `TerminalSettings.profile.spec.tsx` + `SettingsView.change-detection.spec.tsx` → 15 passing
- `tsc --noEmit` (extension) clean

Fully aligned with the direction. Glad to help with the docs follow-up (Zoo-Code-Docs#4) or the more specific #119 close-out — just say the word. 🦓

proyectoauraorg:
👋 Friendly ping — this PR has been open for 9 days.

**Summary:** Adds VS Code integrated-terminal shell override, allowing users to configure which shell is used for terminal operations.

- ✅ CI green
- ✅ Follows existing settings pattern

edelauna (approved):
Refactoring Terminal usage.

PR #281, opened 2026-05-24
feat(ripgrep): add Show Ripgrep Diagnostic command

0xMink:
## Summary

Adds `Zoo Code: Show Ripgrep Diagnostic` to the command palette. Motivated by the ripgrep ENOENT failures reported in Discord (microsoft/vscode#318691) — debugging required a custom test build to see what `getBinPath` actually returned on the user's machine. This ships that diagnostic as a first-class command so the next user can paste a report instead.

## What it does

When invoked, the command produces a timestamped report covering three steps:

1. **require(`@vscode/ripgrep`)** — attempts the extHost interceptor path; logs the result or failure reason including the full error message.
2. **Path probe** — checks all 6 candidate paths under `vscode.env.appRoot` (both `@vscode/ripgrep` and `@vscode/ripgrep-universal` layouts), marking each ✓/✗.
3. **Spawn test** — runs `rg --version` on the path `getBinPath()` selects, capturing stdout, stderr, exit code, and distinguishing ENOENT from timeout. This directly tests the "file exists but spawn fails" failure mode Naved reported.

Paths are JSON-encoded alongside their display form so separators survive copy-paste into Discord/GitHub. The report is written to a dedicated output channel, copied to the clipboard, and confirmed with an info toast.

## Design notes

- `getRipgrepDiagnostic` is a pure data function, fully unit-testable without vscode window/clipboard side effects.
- Candidate paths are shared with `getBinPath()` via `ripgrepCandidatePaths()` — the diagnostic and production resolution cannot drift.
- No new devDependencies — `@vscode/ripgrep` is already present and remains esbuild-external; the `require()` in `loadRipgrep.ts` is a runtime probe, not a build-time dep.
- The command wrapper is covered by Vitest (not excluded from coverage).

## Test plan

- [x] `getRipgrepDiagnostic`: require success/failure/loadError, path probe ✓/✗, spawn success, spawn ENOENT, spawn timeout, stderr capture, JSON path encoding, empty appRoot runs step 1 but skips steps 2–3.
- [x] `trySpawnRipgrep`: spawn called with correct path + args + timeout option.
- [x] `registerRipgrepDiagnosticCommand`: channel creation, disposable stored in `context.subscriptions`, channel clear/append/show, clipboard write, toast.
- [x] `tsc --noEmit` clean.
- [x] knip clean.
- [ ] Manual: invoke command in VS Code, verify output channel + clipboard + toast.


0xMink:
Pushed `12e6b9c8a` addressing all five review threads. Summary of what changed and where:

**1. `.asar`→`.asar.unpacked` regex (Copilot `diagnostic.ts:55`, CodeRabbit `diagnostic.ts:57`)**
`src/services/ripgrep/diagnostic.ts` — replaced `/\bnode_modules\.asar\b/` with `/node_modules\.asar(?=[\\/])/`. The old `\b` matched the `r`/`.` boundary inside `node_modules.asar.unpacked` too, producing `.unpacked.unpacked`. New test in `diagnostic.spec.ts` ("does not double-substitute when rgPath already contains node_modules.asar.unpacked") guards the regression.

**2. OutputChannel lifecycle (Copilot `diagnostic.ts:81`)**
`src/services/ripgrep/diagnostic.ts` `registerRipgrepDiagnosticCommand` now creates the channel once at registration, calls `channel.clear()` before appending on each invocation, and returns `vscode.Disposable.from(command, channel)` so VS Code disposes both when the extension deactivates.

**3. Command ID routing (Copilot `diagnostic.ts:75`)**
- `packages/types/src/vscode.ts`: added `"showRipgrepDiagnostic"` to the `commandIds` tuple.
- `src/services/ripgrep/diagnostic.ts`: registered via `getCommand("showRipgrepDiagnostic")` instead of the hardcoded literal.
- `src/activate/registerCommands.ts`: changed `getCommandsMap`'s return type to `Record<Exclude<CommandId, "showRipgrepDiagnostic">, any>` since `registerRipgrepDiagnosticCommand` registers it separately (it owns the channel disposable). Verified the bundle still emits the `zoo-code.showRipgrepDiagnostic` literal via `package.json` and that `${Package.name}.${id}` resolves to the same string at runtime.

**4. NLS key for command title (Copilot `package.json:165`, CodeRabbit `package.json:166`)**
- `src/package.json`: `"title": "Show Ripgrep Diagnostic"` → `"title": "%command.showRipgrepDiagnostic.title%"`.
- All 18 `src/package.nls*.json` files: added `"command.showRipgrepDiagnostic.title": "Show Ripgrep Diagnostic"` (English string in every locale for now; translations to follow when the localization workflow runs). `node scripts/find-missing-translations.js` reports no missing keys.

**5. Preserve `loadRipgrep` error details (CodeRabbit `loadRipgrep.ts:14`)**
- `src/services/ripgrep/internal/loadRipgrep.ts`: introduced `export type LoadRipgrepResult = { rgPath?: string; loadError?: string }` and the catch now returns `{ loadError: error.message }` instead of `undefined`.
- `src/services/ripgrep/diagnostic.ts`: new `else if (m.loadError)` branch in the report so the require failure message is surfaced.
- New `diagnostic.spec.ts` test ("reports the loadError when loadRipgrep returns a loadError field") covers the branch.

**Verification**
- `npx vitest run services/ripgrep/__tests__/diagnostic.spec.ts` → 8/8 pass (was 6/6).
- `npx tsc --noEmit` clean.
- `npx eslint . --ext=ts --max-warnings=0` clean.
- `pnpm vsix` builds; unpacking the vsix confirms `require("@vscode/ripgrep")` is still preserved in `extension.js` and the `showRipgrepDiagnostic` identifier resolves through the bundle.
- The thin vscode wrapper is wrapped in `/* c8 ignore start/stop */` (with a one-line justification comment) since it can only be exercised by manual VS Code invocation; the pure-data `getRipgrepDiagnostic` carries the coverage.

0xMink:
Polish pass from a re-read of the diagnostic command:

- `getRipgrepDiagnostic` now early-returns an explanatory report when `vscode.env.appRoot` is empty/whitespace, instead of silently producing a meaningless path probe. Closes the input-validation gap.
- Added a Windows-path test for the `.asar` substitution regex. The lookahead `(?=[\\/])` already handles both POSIX and Windows separators; the new test pins that behavior.

Tests: 8 -> 10 passing. `tsc --noEmit` and `eslint --max-warnings=0` both clean.

src/services/ripgrep/diagnostic.ts
```diff
@@ -0,0 +1,105 @@
...
+ * effects — so it's fully unit-testable.
+ *
+ * Step 1 tries `loadRipgrep()` (CommonJS require, hits VS Code's
+ * extHost interceptor on builds that have completed the
+ * `@vscode/ripgrep` → `@vscode/ripgrep-universal` migration).
+ * Step 2 probes every known `vscode.env.appRoot`-relative path and
+ * reports which ones exist on disk.
+ */
+export async function getRipgrepDiagnostic(vscodeAppRoot: string): Promise<string> {
+	if (!vscodeAppRoot || vscodeAppRoot.trim() === "") {
```
edelauna: Should we still run the `loadRipgrep()` section when `appRoot` is empty, and only skip the path-probe section? The report says the require-interceptor path may still work, but this early return prevents that result from being captured.

src/services/ripgrep/diagnostic.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/**
+ * Registers the `zoo-code.showRipgrepDiagnostic` command. Thin wrapper —
+ * runs `getRipgrepDiagnostic`, shows the result in an output channel,
+ * copies it to the clipboard, and shows an info toast. The OutputChannel
+ * is created once at registration and disposed alongside the command via
+ * the composite Disposable returned here.
+ */
+/* c8 ignore start -- thin vscode wrapper; covered by manual invocation. */
```
edelauna: Could we cover the command wrapper with a package-local Vitest test instead of excluding it from coverage? The user-facing behavior here is the command registration flow: output channel update, clipboard write, and toast.

src/package.json
```diff
@@ -554,6 +559,7 @@
 		"@types/tmp": "^0.2.6",
 		"@types/turndown": "^5.0.5",
 		"@types/vscode": "^1.84.0",
+		"@vscode/ripgrep": "^1.17.0",
```
edelauna: Do we need to add this as a dev dependency for a runtime diagnostic? Since #248 left the `require("@vscode/ripgrep")` path as non-load-bearing and relies on the appRoot path probe, adding the package here may make local/test resolution less representative of what the user’s VS Code host provides.

src/services/ripgrep/diagnostic.ts
```diff
@@ -0,0 +1,105 @@
...
+ * for the given VS Code installation. Pure data function — no UI side
+ * effects — so it's fully unit-testable.
+ *
+ * Step 1 tries `loadRipgrep()` (CommonJS require, hits VS Code's
+ * extHost interceptor on builds that have completed the
+ * `@vscode/ripgrep` → `@vscode/ripgrep-universal` migration).
+ * Step 2 probes every known `vscode.env.appRoot`-relative path and
+ * reports which ones exist on disk.
+ */
+export async function getRipgrepDiagnostic(vscodeAppRoot: string): Promise<string> {
```
edelauna, line 61: Would it be worth shaping this as the first section of a reusable diagnostics report rather than a ripgrep-only command? The ripgrep issue is mostly addressed by #248, so the longer-term value here may be giving support a general way to collect host/tooling diagnostics when future environment-specific bugs come up.

edelauna (changes requested):
This is great! Since the ripgrep issue itself has been resolved, my comments mostly revolve around making this command extensible for future debugging.

Also, for the time being, I’d like to avoid reintroducing `@vscode/ripgrep` as a devDependency. From a ripgrep perspective, can we simplify the diagnostic to report:

  - `vscode.version`
  - `vscode.env.appRoot`
  - `process.platform` / `process.arch`
  - all candidate paths from `getBinPath` or a shared candidate builder
  - whether each path exists
  - which path `getBinPath(appRoot)` would return

Could you also include either a screen recording or screenshot of the final diagnostics?

edelauna (approved):

PR #294, opened 2026-05-24
fix(vertex): warn when 'Google Cloud Credentials' field receives a file path

0xMink:
## Summary

Users reported recurring `Error parsing JSON: SyntaxError: Unexpected token 'C', "C:\Users\i"... is not valid JSON` in the VS Code dev console — fired once per Task creation. The root cause is a UX trap in the Vertex provider settings: the "Google Cloud Credentials" field expects raw JSON content (the `{ "type": "service_account", ... }` blob), but Google's own ecosystem uses `GOOGLE_APPLICATION_CREDENTIALS` for paths, so "credentials = path to credentials file" is a natural mental model that the field naming does not disambiguate. The sibling "Google Cloud Key File Path" field is where a path belongs.

For affected users, auth still works (their `GOOGLE_APPLICATION_CREDENTIALS` env var picks up the slack and `safeJsonParse` has an internal try/catch), so the only user-visible symptom is the recurring generic console error.

## Changes

Three pieces, all behind a small shape-check on the user-supplied string:

- **`src/api/providers/gemini.ts` + `src/api/providers/anthropic-vertex.ts`**: shape-check the field before parsing. If it looks like a path (Windows `C:\...` / `C:/...`, POSIX absolute `/...`, `~/...`, or `./` / `../`), log a single specific console warning that names the correct field and the `GOOGLE_APPLICATION_CREDENTIALS` env var, and skip the `JSON.parse` so the generic `Error parsing JSON` log never fires for the path case. The shared helper `parseVertexJsonCredentials` is exported from `gemini.ts` and reused from `anthropic-vertex.ts` so the Gemini-on-Vertex and Claude-on-Vertex paths behave identically.
- **`packages/core/src/message-utils/safeJsonParse.ts`**: add an optional `context?: string` parameter so log messages can identify their source (`Error parsing JSON (Vertex credentials):` instead of an anonymous `Error parsing JSON:`). Backward-compatible — existing call sites are unchanged.
- **`webview-ui/src/components/settings/providers/Vertex.tsx`**: inline warning under the credentials field that fires while the input looks like a path, naming the correct field to use instead. New i18n key `settings:providers.googleCloudCredentialsPathWarning` propagated to all 18 locales.

Auth behavior is unchanged for users who configure the fields correctly or use the env var. The only user-visible change is the new specific warning replacing the previous generic spam.

## Known properties

- The console warning still fires once per Vertex handler construction (i.e. once per Task creation) for users with a path in the JSON field — same frequency as the original generic spam, but the message is now specific and actionable.
- The 17 non-English i18n strings are first-draft translations and welcome community refinement in follow-ups.

## Test plan

- [x] `packages/core`: `npx vitest run message-utils` — 39/39 pass (new `safeJsonParse.spec.ts` covers context-arg and backward-compat paths)
- [x] `src`: `npx vitest run api/providers/__tests__/vertex-credentials.spec.ts api/providers/__tests__/gemini.spec.ts api/providers/__tests__/vertex.spec.ts api/providers/__tests__/anthropic-vertex.spec.ts api/providers/__tests__/gemini-handler.spec.ts` — 95/95 pass (new spec asserts `GoogleGenAI` / `GoogleAuth` receive `credentials: undefined` for path-shaped input, the warning is emitted with the actionable strings, and JSON content still parses through)
- [x] `webview-ui`: `npx vitest run src/components/settings/providers` — 45/45 pass (new cases assert the warning renders for path-shaped input and is absent for empty / JSON input)
- [x] `npx tsc --noEmit` in `src/`, `packages/core/`, `webview-ui/` — all clean
- [x] `pnpm lint` (full monorepo) — clean
- [x] Changeset added (`zoo-code` patch)

Surfaced during testing of #248's diagnostic build by an end user.

## Summary by CodeRabbit

* **New Features**
  * Detects when the Google Cloud Credentials field looks like a file path and displays an inline, localized warning directing users to the key-file field or GOOGLE_APPLICATION_CREDENTIALS.

* **Improvements**
  * JSON parsing logs now include optional contextual labels; path-shaped credential inputs emit a single, targeted warning and avoid parsing.

* **Tests**
  * Added unit and UI tests for JSON parsing, path-detection, and the credential warning.

* **Localization**
  * Added translated warning strings for many locales.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/294?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/vertex-credentials-field-validation.md
+++ b/.changeset/vertex-credentials-field-validation.md
@@ -1,0 +1,5 @@
+---
+"zoo-code": patch
+---
+
+Detect when the "Google Cloud Credentials" Vertex field has been given a filesystem path instead of the raw JSON contents of a service-account key file. The runtime now skips JSON.parse for path-shaped input, logs a single specific console warning naming the sibling "Google Cloud Key File Path" field and `GOOGLE_APPLICATION_CREDENTIALS` env var, and the settings UI shows an inline warning under the field while the input still looks like a path. Auth behavior is unchanged for correctly-configured users.
--- a/packages/types/src/utils/looksLikeFilePath.ts
+++ b/packages/types/src/utils/looksLikeFilePath.ts
@@ -1,0 +1,24 @@
+// Returns true when a string is shaped like a filesystem path (Windows
+// drive-letter, POSIX absolute, POSIX home, POSIX relative). Pure and
+// dependency-free so it can be shared between the extension runtime (e.g.
+// parseVertexJsonCredentials) and the webview UI (e.g. the Vertex settings
+// warning), guaranteeing both surfaces agree on what "looks like a path"
+// means.
+//
+// Returns false for nullish, empty, and whitespace-only input — neither
+// call site should warn in those cases.
+export function looksLikeFilePath(value: string | null | undefined): boolean {
+	if (value == null) {
+		return false
+	}
+	const trimmed = value.trim()
+	if (!trimmed) {
+		return false
+	}
+	return (
+		/^[A-Za-z]:[\\/]/.test(trimmed) || // Windows: C:\... or C:/...
+		trimmed.startsWith("/") || // POSIX absolute: /home/...
+		trimmed.startsWith("~") || // POSIX home: ~/...
+		trimmed.startsWith(".") // POSIX relative: ./... or ../...
+	)
+}
--- a/src/api/providers/anthropic-vertex.ts
+++ b/src/api/providers/anthropic-vertex.ts
@@ -23,7 +23,7 @@ import {
 } from "../../core/prompts/tools/native-tools/converters"
 
 import { BaseProvider } from "./base-provider"
-import { parseVertexJsonCredentials } from "./gemini"
+import { parseVertexJsonCredentials } from "./utils/vertex-credentials"
 import type { SingleCompletionHandler, ApiHandlerCreateMessageMetadata } from "../index"
 
 // https://docs.anthropic.com/en/api/claude-on-vertex-ai
@@ -40,13 +40,15 @@ export class AnthropicVertexHandler extends BaseProvider implements SingleComple
 		const projectId = this.options.vertexProjectId ?? "not-provided"
 		const region = this.options.vertexRegion ?? "us-east5"
 
-		if (this.options.vertexJsonCredentials) {
+		const parsedVertexCredentials = parseVertexJsonCredentials(this.options.vertexJsonCredentials)
+
+		if (parsedVertexCredentials) {
 			this.client = new AnthropicVertex({
 				projectId,
 				region,
 				googleAuth: new GoogleAuth({
 					scopes: ["https://www.googleapis.com/auth/cloud-platform"],
-					credentials: parseVertexJsonCredentials(this.options.vertexJsonCredentials),
+					credentials: parsedVertexCredentials,
 				}),
 			})
 		} else if (this.options.vertexKeyFile) {
--- a/src/api/providers/anthropic-vertex.ts
+++ b/src/api/providers/anthropic-vertex.ts
@@ -1,6 +1,6 @@
 import { Anthropic } from "@anthropic-ai/sdk"
 import { AnthropicVertex } from "@anthropic-ai/vertex-sdk"
-import { GoogleAuth, JWTInput } from "google-auth-library"
+import { GoogleAuth } from "google-auth-library"
 
 import {
 	type ModelInfo,
@@ -10,7 +10,6 @@ import {
 	ANTHROPIC_DEFAULT_MAX_TOKENS,
 	VERTEX_1M_CONTEXT_MODEL_IDS,
 } from "@roo-code/types"
-import { safeJsonParse } from "@roo-code/core"
 
 import { ApiHandlerOptions } from "../../shared/api"
 
@@ -24,6 +23,7 @@ import {
 } from "../../core/prompts/tools/native-tools/converters"
 
 import { BaseProvider } from "./base-provider"
+import { parseVertexJsonCredentials } from "./gemini"
 import type { SingleCompletionHandler, ApiHandlerCreateMessageMetadata } from "../index"
 
 // https://docs.anthropic.com/en/api/claude-on-vertex-ai
@@ -46,7 +46,7 @@ export class AnthropicVertexHandler extends BaseProvider implements SingleComple
 				region,
 				googleAuth: new GoogleAuth({
 					scopes: ["https://www.googleapis.com/auth/cloud-platform"],
-					credentials: safeJsonParse<JWTInput>(this.options.vertexJsonCredentials, undefined),
+					credentials: parseVertexJsonCredentials(this.options.vertexJsonCredentials),
 				}),
 			})
 		} else if (this.options.vertexKeyFile) {
--- a/src/api/providers/utils/vertex-credentials.ts
+++ b/src/api/providers/utils/vertex-credentials.ts
@@ -1,5 +1,6 @@
 import type { JWTInput } from "google-auth-library"
 
+import { looksLikeFilePath } from "@roo-code/types"
 import { safeJsonParse } from "@roo-code/core"
 
 // Detects when the "Google Cloud Credentials" field has received a filesystem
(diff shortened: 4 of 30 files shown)


0xMink:
Pushed a follow-up commit (6b0e5828f) addressing the four feedback items from this review round:

**1. Precedence bug (CodeRabbit, `src/api/providers/gemini.ts:225-233` and `src/api/providers/anthropic-vertex.ts:43-50`)** — fixed. The truthiness check is now on the parsed result of `parseVertexJsonCredentials(...)`, not on the raw `vertexJsonCredentials` field. With this fix, path-shaped input (which the parser correctly returns `undefined` for) cleanly falls through to the `vertexKeyFile` branch instead of entering the JSON branch with `credentials: undefined`.

**2. Helper coupling (Copilot, `src/api/providers/anthropic-vertex.ts:28`)** — fixed. `parseVertexJsonCredentials` moved out of `gemini.ts` into a new module at `src/api/providers/utils/vertex-credentials.ts`. `anthropic-vertex.ts` no longer imports through the Gemini provider (and therefore no longer pulls in the `@google/genai` runtime dependency); both providers and the spec import from the shared util.

**3. Mixed-input regression coverage (CodeRabbit nitpick, `src/api/providers/__tests__/vertex-credentials.spec.ts:158-173` & 246-270)** — added. New cases for both `GeminiHandler` and `AnthropicVertexHandler` assert that when `vertexJsonCredentials` is path-shaped **and** `vertexKeyFile` is set, the client is constructed with `keyFile`, not `credentials`. The pre-existing path-shape-only tests were also updated to match the corrected fallthrough (with no `vertexKeyFile`, neither `googleAuthOptions` for Gemini nor `GoogleAuth` for Anthropic-on-Vertex is constructed).

**4. a11y on the warning (CodeRabbit nitpick, `webview-ui/src/components/settings/providers/Vertex.tsx:94-104`)** — added `role="alert"` plus an explicit `aria-live="polite"` on the path-shape warning `<div>` so it is announced by screen readers when it appears. `polite` (not `assertive`) since the warning is informational, not safety-critical; `role="alert"` would otherwise default to `assertive`.

**Verification**
- `npx vitest run` for `vertex-credentials.spec.ts` + `gemini.spec.ts` + `vertex.spec.ts` + `anthropic-vertex.spec.ts`: 78/78 pass.
- `npx vitest run src/components/settings/providers` in `webview-ui/`: 45/45 pass.
- `npx tsc --noEmit` in both `src/` and `webview-ui/`: exit 0.
- `npx eslint . --ext=ts --max-warnings=0` in `src/`: exit 0.
- `pnpm lint` from the repo root: exit 0.

src/api/providers/anthropic-vertex.ts
```diff
@@ -24,6 +23,7 @@ import {
 } from "../../core/prompts/tools/native-tools/converters"
 
 import { BaseProvider } from "./base-provider"
+import { parseVertexJsonCredentials } from "./gemini"
 import type { SingleCompletionHandler, ApiHandlerCreateMessageMetadata } from "../index"
```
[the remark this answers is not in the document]
0xMink: Done in 6b0e5828f — `parseVertexJsonCredentials` (and the path-shape detection it uses) is now at `src/api/providers/utils/vertex-credentials.ts`, with no provider-specific imports. Both `gemini.ts` and `anthropic-vertex.ts` import from there, and `anthropic-vertex.ts` no longer pulls in the `@google/genai` runtime side of `gemini.ts`.

0xMink:
Tightened the warning's a11y attributes to `role="status"` (dropping the conflicting `aria-live="polite"` override on `role="alert"`), which is the conventional pattern for a non-urgent dynamic status message and is implicitly polite.

src/api/providers/utils/vertex-credentials.ts
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	const looksLikePath =
+		/^[A-Za-z]:[\\/]/.test(trimmed) || // Windows: C:\... or C:/...
+		trimmed.startsWith("/") || // POSIX absolute: /home/...
+		trimmed.startsWith("~") || // POSIX home: ~/...
+		trimmed.startsWith(".") // POSIX relative: ./... or ../...
+
+	if (looksLikePath) {
+		const preview = trimmed.length > 40 ? `${trimmed.slice(0, 40)}…` : trimmed
```
edelauna: Do we need to include the path preview in the warning? A static message would still tell users which field to use without putting local usernames or directory names into extension logs.

webview-ui/src/components/settings/providers/Vertex.tsx
```diff
@@ -9,6 +10,23 @@ import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from "@
 
 import { inputEventTransform } from "../transforms"
 
+// Detects when the "Google Cloud Credentials" field has received a filesystem
+// path instead of the raw JSON contents of a service-account key file. Mirrors
+// the runtime guard in src/api/providers/gemini.ts so the warning the user
+// sees in the UI matches what the runtime would log.
+function looksLikeFilePath(value: string): boolean {
```
edelauna: Could this predicate live in a shared dependency-free helper, or have a contract test against the runtime helper? Since this warning is meant to mirror `parseVertexJsonCredentials`, keeping two copies seems easy to drift.

webview-ui/src/components/settings/providers/__tests__/Vertex.spec.tsx
```diff
@@ -26,6 +26,13 @@ vi.mock("@src/i18n/TranslationContext", () => ({
 	useAppTranslation: () => ({ t: (key: string) => key }),
 }))
 
+// The component uses <Trans> for the path-shape warning so it can interpolate
+// <strong>/<code> elements. The mock just renders the i18n key so the spec
+// can assert on its presence without depending on the English copy.
+vi.mock("react-i18next", () => ({
```
edelauna, line 35: Could this test assert the resolved English warning text instead of only the i18n key? The remediation depends on mentioning `Google Cloud Key File Path` and `GOOGLE_APPLICATION_CREDENTIALS`, and this mock would miss those regressions.

.changeset/vertex-credentials-field-validation.md
edelauna, line 1: curious on your opinion if we should revive this? I noticed the Roo team stopped using changesets in favour of LLM drive releases, however they left the old changeset infrastructure in the repo.

edelauna (changes requested):
Nice! Thanks for addressing this, had a couple suggestions.

src/api/providers/utils/vertex-credentials.ts
0xMink: Good catch — agreed. The warning is now static and no longer interpolates the user's input, so usernames and directory names won't leak into extension logs. The spec also gained a regression test asserting the user's path does not appear in the message. d69f6942c

webview-ui/src/components/settings/providers/Vertex.tsx
0xMink: Agreed — extracted to a shared helper at packages/types/src/utils/looksLikeFilePath.ts. Both Vertex.tsx and parseVertexJsonCredentials now import it from @roo-code/types, and the predicate has its own spec to lock the contract. d69f6942c

webview-ui/src/components/settings/providers/__tests__/Vertex.spec.tsx
0xMink: Agreed. The <Trans> mock now resolves keys against the real webview-ui/src/i18n/locales/en/settings.json, and the assertions check for /Google Cloud Key File Path/ and /GOOGLE_APPLICATION_CREDENTIALS/ in the rendered output — a translation regression that drops either mention would now fail the spec. d69f6942c

packages/types/src/utils/looksLikeFilePath.ts
```diff
@@ -0,0 +1,24 @@
...
+	}
+	const trimmed = value.trim()
+	if (!trimmed) {
+		return false
+	}
+	return (
+		/^[A-Za-z]:[\\/]/.test(trimmed) || // Windows: C:\... or C:/...
+		trimmed.startsWith("/") || // POSIX absolute: /home/...
+		trimmed.startsWith("~") || // POSIX home: ~/...
+		trimmed.startsWith(".") // POSIX relative: ./... or ../...
```
edelauna, line 22: The comment says `./... or ../...` but `startsWith(".")` also matches bare dotfiles like `.env` or `.gitconfig`. Is that intentional? (Functionally fine since JSON never starts with `.` — just want to confirm the comment is meant to be illustrative rather than exhaustive.)

edelauna (approved):

PR #318, opened 2026-05-24
feat(openai): support codex models via the Responses API (#87)

### LINKED ISSUE #87: azure openAI codex
Name of model: (gpt-5.3-codex)
Model creator: (i.e. Azure Open AI)
Link to model: https://github.com/RooCodeInc/Roo-Code/pull/11952

This was fixed in RooCode but not merged yet.

This fixed will allow the configuration of codex model via openAI compatible using the responses instead of chat/completions.

Will you please add this fix?

Thank you.

proyectoauraorg:
**Closes #87**

### Problem
The OpenAI-compatible provider (`OpenAiHandler`) routes every request through `chat.completions.create`. Azure-hosted GPT-5.x **codex** models (e.g. `gpt-5.3-codex`) reject the Chat Completions API (`"The chatCompletion operation does not work with the specified model"`) and require the **Responses API**, so they can't be used here today. This ports the fix from RooCode PR #11952, which is still unmerged upstream (hence not present in Zoo).

### Fix
`createMessage`/`completePrompt` now detect codex models (`_isCodexModel`) and route them through the Responses API (`responses.create`) with streaming (text, reasoning, tool calls, usage), Anthropic→Responses input conversion, and flat tool-schema conversion. **Non-codex models keep using `chat.completions` unchanged.**

### Note on tool-call finalization
The codex stream emits `tool_call_partial` during argument deltas; streamed calls are finalized by the consumer's end-of-stream `NativeToolCallParser.finalizeRawChunks()` (`Task.ts`), and the complete `tool_call` event is guarded by `streamedToolCallIds` to avoid duplicates — consistent with how Zoo already consumes both event shapes.

### Tests
New `openai-codex-responses.spec.ts` (13 tests: detection, routing, streaming, tools, errors). Existing `openai.spec.ts` unchanged and green. `tsc` / `eslint` / `prettier` and the full openai suite (65 tests) pass locally.

## Summary by CodeRabbit

* **New Features**
  * Added support for Codex-style models with both streaming and non-streaming completions, improved tool-call handling (partial argument streaming and final tool-call events), enhanced message formatting (including tool results and inline content), flat tool schema conversion, stricter request validation, and clearer error reporting.

* **Tests**
  * Added comprehensive tests for Codex detection and routing, streaming events (text, tool calls, partial updates, usage), response parsing, and error scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/318?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/api/providers/__tests__/openai-codex-responses.spec.ts
+++ b/src/api/providers/__tests__/openai-codex-responses.spec.ts
@@ -1,0 +1,465 @@
+// npx vitest run api/providers/__tests__/openai-codex-responses.spec.ts
+
+import { OpenAiHandler } from "../openai"
+import { ApiHandlerOptions } from "../../../shared/api"
+import { Anthropic } from "@anthropic-ai/sdk"
+import { openAiModelInfoSaneDefaults } from "@roo-code/types"
+
+const mockChatCreate = vitest.fn()
+const mockResponsesCreate = vitest.fn()
+
+vitest.mock("openai", () => {
+	const mockConstructor = vitest.fn()
+	return {
+		__esModule: true,
+		default: mockConstructor.mockImplementation(() => ({
+			chat: {
+				completions: {
+					create: mockChatCreate,
+				},
+			},
+			responses: {
+				create: mockResponsesCreate,
+			},
+		})),
+		AzureOpenAI: mockConstructor.mockImplementation(() => ({
+			chat: {
+				completions: {
+					create: mockChatCreate,
+				},
+			},
+			responses: {
+				create: mockResponsesCreate,
+			},
+		})),
+	}
+})
+
+describe("OpenAiHandler - Codex model detection", () => {
+	let handler: OpenAiHandler
+
+	beforeEach(() => {
+		mockChatCreate.mockClear()
+		mockResponsesCreate.mockClear()
+	})
+
+	describe("_isCodexModel", () => {
+		it("should detect gpt-5.3-codex as a codex model", () => {
+			handler = new OpenAiHandler({
+				openAiApiKey: "test-key",
+				openAiModelId: "gpt-5.3-codex",
+				openAiBaseUrl: "https://test.openai.azure.com/openai/deployments/gpt5.3",
+				openAiUseAzure: true,
+			})
+			// Access the protected method via any cast
+			expect((handler as any)._isCodexModel("gpt-5.3-codex")).toBe(true)
+		})
+
+		it("should detect gpt-5.1-codex as a codex model", () => {
+			handler = new OpenAiHandler({
+				openAiApiKey: "test-key",
+				openAiModelId: "gpt-5.1-codex",
+			})
+			expect((handler as any)._isCodexModel("gpt-5.1-codex")).toBe(true)
+		})
+
+		it("should detect codex in a case-insensitive manner", () => {
+			handler = new OpenAiHandler({
+				openAiApiKey: "test-key",
+				openAiModelId: "GPT-5.3-CODEX",
+			})
+			expect((handler as any)._isCodexModel("GPT-5.3-CODEX")).toBe(true)
+		})
+
+		it("should not detect regular models as codex", () => {
+			handler = new OpenAiHandler({
+				openAiApiKey: "test-key",
+				openAiModelId: "gpt-4",
+			})
+			expect((handler as any)._isCodexModel("gpt-4")).toBe(false)
+			expect((handler as any)._isCodexModel("gpt-4o")).toBe(false)
+			expect((handler as any)._isCodexModel("o3-mini")).toBe(false)
+		})
+	})
+
+	describe("createMessage with codex model", () => {
+		it("should use Responses API for codex models instead of Chat Completions", async () => {
+			handler = new OpenAiHandler({
+				openAiApiKey: "test-key",
+				openAiModelId: "gpt-5.3-codex",
+				openAiBaseUrl: "https://test.openai.azure.com/openai/deployments/gpt5.3",
+				openAiUseAzure: true,
+			})
+
+			// Mock the responses.create to return a streaming async iterable
+			mockResponsesCreate.mockResolvedValue({
+				[Symbol.asyncIterator]: async function* () {
+					yield {
+						type: "response.output_text.delta",
+						delta: "Hello from codex!",
+					}
+					yield {
+						type: "response.done",
+						response: {
+							usage: {
+								input_tokens: 10,
+								output_tokens: 5,
+							},
+						},
+					}
+				},
+			})
+
+			const messages: Anthropic.Messages.MessageParam[] = [{ role: "user", content: "Hello" }]
+
+			const chunks: any[] = []
+			for await (const chunk of handler.createMessage("You are a helpful assistant", messages, {
+				taskId: "test",
+			})) {
+				chunks.push(chunk)
+			}
+
+			// Verify responses.create was called, NOT chat.completions.create
+			expect(mockResponsesCreate).toHaveBeenCalledTimes(1)
+			expect(mockChatCreate).not.toHaveBeenCalled()
+
+			// Verify the request body structure
+			const requestBody = mockResponsesCreate.mock.calls[0][0]
+			expect(requestBody.model).toBe("gpt-5.3-codex")
+			expect(requestBody.stream).toBe(true)
+			expect(requestBody.instructions).toBe("You are a helpful assistant")
+			expect(requestBody.input).toBeDefined()
+			expect(Array.isArray(requestBody.input)).toBe(true)
+
+			// Verify chunks
+			const textChunks = chunks.filter((c) => c.type === "text")
+			expect(textChunks.length).toBe(1)
+			expect(textChunks[0].text).toBe("Hello from codex!")
+
+			const usageChunks = chunks.filter((c) => c.type === "usage")
+			expect(usageChunks.length).toBe(1)
+			expect(usageChunks[0].inputTokens).toBe(10)
+			expect(usageChunks[0].outputTokens).toBe(5)
+		})
+
+		it("should use Chat Completions for non-codex models", async () => {
+			handler = new OpenAiHandler({
+				openAiApiKey: "test-key",
+				openAiModelId: "gpt-4",
+				openAiBaseUrl: "https://api.openai.com/v1",
+			})
+
+			mockChatCreate.mockResolvedValue({
+				[Symbol.asyncIterator]: async function* () {
+					yield {
+						choices: [{ delta: { content: "Hello" }, index: 0 }],
+						usage: null,
+					}
+					yield {
+						choices: [{ delta: {}, index: 0 }],
+						usage: { prompt_tokens: 10, completion_tokens: 5, total_tokens: 15 },
+					}
+				},
+			})
+
+			const messages: Anthropic.Messages.MessageParam[] = [{ role: "user", content: "Hello" }]
+
+			const chunks: any[] = []
+			for await (const chunk of handler.createMessage("System", messages, { taskId: "test" })) {
+				chunks.push(chunk)
+			}
+
+			// Verify chat.completions.create was called, NOT responses.create
+			expect(mockChatCreate).toHaveBeenCalledTimes(1)
+			expect(mockResponsesCreate).not.toHaveBeenCalled()
+		})
+	})
+
+	describe("createMessage codex conversation formatting", () => {
+		it("should format conversation with tool use correctly for Responses API", async () => {
+			handler = new OpenAiHandler({
+				openAiApiKey: "test-key",
+				openAiModelId: "gpt-5.3-codex",
+				openAiBaseUrl: "https://test.openai.azure.com/openai/deployments/gpt5.3",
+				openAiUseAzure: true,
+			})
+
+			mockResponsesCreate.mockResolvedValue({
+				[Symbol.asyncIterator]: async function* () {
+					yield {
+						type: "response.output_text.delta",
+						delta: "Done.",
+					}
+					yield {
+						type: "response.done",
+						response: {
+							usage: { input_tokens: 20, output_tokens: 3 },
+						},
+					}
+				},
+			})
+
+			const messages: Anthropic.Messages.MessageParam[] = [
+				{ role: "user", content: "What is 1+1?" },
+				{
+					role: "assistant",
+					content: [
+						{ type: "text", text: "Let me calculate that." },
+						{ type: "tool_use", id: "call_123", name: "calculator", input: { expression: "1+1" } },
+					],
+				},
+				{
+					role: "user",
+					content: [
+						{
+							type: "tool_result",
+							tool_use_id: "call_123",
+							content: "2",
+						},
+					],
+				},
+			]
+
+			for await (const _chunk of handler.createMessage("You are helpful", messages, { taskId: "test" })) {
+				// consume
+			}
+
+			const requestBody = mockResponsesCreate.mock.calls[0][0]
+			const input = requestBody.input
+
+			// First item: user message
+			expect(input[0].role).toBe("user")
+			expect(input[0].content[0].type).toBe("input_text")
+			expect(input[0].content[0].text).toBe("What is 1+1?")
+
+			// Second item: assistant text
+			expect(input[1].role).toBe("assistant")
+			expect(input[1].content[0].type).toBe("output_text")
+
+			// Third item: function_call
+			expect(input[2].type).toBe("function_call")
+			expect(input[2].name).toBe("calculator")
+
+			// Fourth item: function_call_output
+			expect(input[3].type).toBe("function_call_output")
+			expect(input[3].output).toBe("2")
+		})
+	})
+
+	describe("createMessage codex tool call streaming", () => {
+		it("should handle tool call events from the Responses API", async () => {
+			handler = new OpenAiHandler({
+				openAiApiKey: "test-key",
+				openAiModelId: "gpt-5.3-codex",
+				openAiUseAzure: true,
+			})
+
+			mockResponsesCreate.mockResolvedValue({
+				[Symbol.asyncIterator]: async function* () {
+					yield {
+						type: "response.output_item.added",
+						item: {
+							type: "function_call",
+							call_id: "call_abc",
+							name: "read_file",
+						},
+					}
+					yield {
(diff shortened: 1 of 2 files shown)


src/api/providers/openai.ts
```diff
@@ -502,6 +568,410 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
+				}
+
+				// Handle failed event
+				if (event?.type === "response.failed") {
+					if (event.error || event.message) {
+						throw new Error(
+							`Response failed: ${event.error?.message || event.message || "Unknown failure"}`,
+						)
+					}
+				}
```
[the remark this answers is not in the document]
proyectoauraorg: Good catch — fixed in `b2a9390d`. Now `response.error` / `error` / `response.failed` always throw (with an `Unknown error`/`Unknown failure` fallback), so a detail-less error event no longer lets a broken stream keep being consumed. Added a test for the no-details case.

Also addressed the tool-schema coverage nitpick in `19355415`: a test now asserts `_convertToolsForResponsesApi` produces the flat Responses-API shape (`{ type, name, description, parameters, strict }`, not nested under `function`). 🦘

src/api/providers/openai.ts
```diff
@@ -91,6 +93,11 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
 		const isAzureAiInference = this._isAzureAiInference(modelUrl)
 		const deepseekReasoner = modelId.includes("deepseek-reasoner") || enabledR1Format
 
+		if (this._isCodexModel(modelId)) {
+			yield* this.handleCodexMessage(systemPrompt, messages, metadata)
+			return
+		}
```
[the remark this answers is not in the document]
proyectoauraorg: @coderabbitai Acknowledged — this is a valid concern and a heavier refactor than the other two items.

The Codex path currently forces `stream: true` unconditionally in `handleCodexMessage`. To properly honor `openAiStreamingEnabled`, we need to either:
1. Branch before calling `handleCodexMessage` and fall through to the existing non-streaming path, or
2. Add a `stream` parameter to `handleCodexMessage` so it can disable streaming on demand.

Option 1 is cleaner since the non-streaming response handling already exists. I will address this in a follow-up commit — flagging as a separate item since it touches the control flow more deeply than the other quick fixes.

src/api/providers/openai.ts
```diff
@@ -332,6 +346,58 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
+			for (const outputItem of response.output) {
+				if (outputItem.type === "message" && outputItem.content) {
+					for (const content of outputItem.content) {
+						if (content.type === "output_text" && content.text) {
+							return content.text
+						}
+					}
+				}
+			}
+		}
```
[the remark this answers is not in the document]
proyectoauraorg: @coderabbitai Great catch — addressed in `2402b0f3`.

The `completePrompt` method now accumulates all `output_text` parts into an array and joins them before returning, instead of returning on the first match. This ensures multi-part Responses payloads are not truncated.

src/api/providers/openai.ts
```diff
@@ -502,6 +568,404 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
+		// Build the request body
+		const requestBody: any = {
+			model: model.id,
+			input: formattedInput,
+			stream: true,
+			store: false,
+			instructions: systemPrompt,
+			...(tools && tools.length > 0 ? { tools } : {}),
+			...(metadata?.tool_choice ? { tool_choice: metadata.tool_choice } : {}),
+			parallel_tool_calls: metadata?.parallelToolCalls ?? true,
```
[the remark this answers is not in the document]
proyectoauraorg: @coderabbitai You are right — addressed in `2402b0f3`.

Changed from hardcoded `store: false` to `metadata?.store ?? true`, which respects the caller's intent while defaulting to `true` as the `ApiHandlerCreateMessageMetadata` type specifies. This enables stored responses / previous-response flows on the Codex route when callers opt in.

src/api/providers/__tests__/openai-codex-responses.spec.ts
```diff
@@ -0,0 +1,519 @@
...
+			// Flat structure: fields live at the top level, not under `function`.
+			expect(tool.type).toBe("function")
+			expect(tool.name).toBe("calculator")
+			expect(tool.description).toBe("Evaluate a math expression")
+			expect(tool.function).toBeUndefined()
+			expect(tool.parameters).toBeDefined()
+			expect(tool.parameters.properties).toBeDefined()
+			// Non-MCP tools are sent with strict schema validation enabled.
+			expect(tool.strict).toBe(true)
+		})
```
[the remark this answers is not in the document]
proyectoauraorg: @coderabbitai Fair point — MCP tools do behave differently (they are sent with `strict: false` to allow flexible schemas). I will add a dedicated test case that:

1. Passes an MCP-style tool (with `inputSchema` instead of `parameters`)
2. Asserts the converted output has `strict: false`
3. Verifies the `name` and `description` are correctly mapped

This will close the gap between the layer description ("with MCP handling") and the actual test assertions. Will push in the next commit.

proyectoauraorg:
## 📊 Codecov Coverage Note

The `codecov/patch` check is showing **53.70% diff coverage** (target 80%). This is because the new Codex responses handler code paths need additional test coverage.

### Plan to reach 80%:
1. **MCP tool handling test** — add test for `strict: false` behavior and `inputSchema` → `parameters` mapping (addressing coderabbitai feedback above)
2. **Non-streaming Codex path** — add test for `openAiStreamingEnabled: false` forcing the non-streaming branch
3. **`_completePromptWithResponsesApi` edge cases** — test multi-part output items, empty output, and error scenarios
4. **`store` metadata forwarding** — test that `metadata?.store` is correctly passed through

I will push these test additions shortly to bring diff coverage above 80%.

proyectoauraorg:
Pushed test coverage for the Codex Responses-API path — the only red check was `codecov/patch` at **53.70%** (target 80%). Added cases for the previously untested branches: the non-streaming path (tool calls, text, message content, reasoning summary, usage incl. cache tokens, object-arg stringify, error wrapping), the streaming event variants (`text.done`-only, `content_part`, reasoning deltas, refusals, `output_item.added/done` message text, `response.completed` fallback + usage, the legacy `choices`/`usage` shape), `max_output_tokens` gating, and conversation formatting (image blocks, array `tool_result` content, assistant string content). Patch coverage is now ~94% locally; 28/28 in the suite.

proyectoauraorg:
👋 Friendly ping — this PR has been open for 8 days.

**Summary:** Adds support for Codex models via the OpenAI Responses API ([#87](https://github.com/Zoo-Code-Org/Zoo-Code/issues/87)).

- ✅ CI green
- ✅ Adds new provider capability without breaking existing ones

src/api/providers/openai.ts
```diff
@@ -502,6 +572,467 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
+					yield {
+						type: "usage",
+						inputTokens: usage.input_tokens ?? usage.prompt_tokens ?? 0,
+						outputTokens: usage.output_tokens ?? usage.completion_tokens ?? 0,
+						cacheWriteTokens: usage.cache_creation_input_tokens || undefined,
+						cacheReadTokens: usage.cache_read_input_tokens || undefined,
+					}
+				}
+			} catch (error) {
+				throw handleOpenAIError(error, this.providerName)
```
edelauna, line 679: The sibling handlers (openai-native.ts, openai-codex.ts) wrap caught errors in `ApiProviderError` and call `TelemetryService.instance.captureException()` before rethrowing. This catch — and the one at line 932 — skips that, so codex failures won't appear in the error dashboard. Worth aligning with the pattern in openai-native.ts:645–648?

src/api/providers/openai.ts
```diff
@@ -502,6 +572,467 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
+				if (event?.usage) {
+					yield {
+						type: "usage",
+						inputTokens: event.usage.input_tokens ?? event.usage.prompt_tokens ?? 0,
+						outputTokens: event.usage.output_tokens ?? event.usage.completion_tokens ?? 0,
+					}
+				}
+			}
+		} catch (error) {
+			throw handleOpenAIError(error, this.providerName)
```
edelauna, line 932: Same as line 679 — `TelemetryService.captureException` is missing here for the streaming path too. Errors thrown during streaming won't appear in telemetry.

src/api/providers/openai.ts
```diff
@@ -332,6 +346,62 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
+		}
+
+		// Add max_output_tokens if needed
+		if (this.options.includeMaxTokens === true) {
+			requestBody.max_output_tokens = this.options.modelMaxTokens || model.info.maxTokens
+		}
+
+		let response
+		try {
+			response = await (this.client as any).responses.create(requestBody)
```
edelauna, line 375: The installed OpenAI SDK (v5.12.2) exposes `responses: API.Responses` as a typed property directly on the client (client.d.ts:233), so the `as any` cast isn't needed here or at lines 628/694. Removing them would also let us replace `requestBody: any` with the typed `ResponseCreateParamsNonStreaming`/`ResponseCreateParamsStreaming` and catch field-name typos at compile time.

src/api/providers/openai.ts
```diff
@@ -502,6 +572,467 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
+		// Format conversation for the Responses API
+		const formattedInput = this._formatConversationForResponsesApi(messages)
+
+		// Build tools in Responses API format (flat structure, not nested under function)
+		const tools = this._convertToolsForResponsesApi(metadata?.tools)
+
+		const useStreaming = this.options.openAiStreamingEnabled ?? true
+
+		// Build the request body (stream flag added per path below)
+		const requestBody: any = {
```
edelauna, line 605: `requestBody: any` means the compiler won't catch invalid field names or wrong types. `OpenAI.Responses.ResponseCreateParamsStreaming` (and the non-streaming variant) are both exported from the SDK — worth switching once the `as any` client cast on line 628/694 is removed.

src/api/providers/openai.ts
```diff
@@ -502,6 +572,467 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
 		}
 	}
 
+	/**
+	 * Checks if the model is a codex model that requires the Responses API.
+	 * Azure-hosted GPT-5.x codex models (e.g., gpt-5.3-codex) do not support
+	 * the Chat Completions API and must use the Responses API instead.
+	 */
+	protected _isCodexModel(modelId: string): boolean {
+		return modelId.toLowerCase().includes("codex")
```
edelauna, line 581: This matches any model ID containing the substring "codex" — so `my-codex-wrapper` or `gpt-4-codex-finetune` set via the OpenAI-compatible provider would silently route to the Responses API. Could we anchor this to known patterns (e.g. `/(^|[-_.])codex([-_.]|$)/i`) or add an explicit opt-in flag?

src/api/providers/openai.ts
```diff
@@ -502,6 +572,467 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
 		}
 	}
 
+	/**
+	 * Checks if the model is a codex model that requires the Responses API.
+	 * Azure-hosted GPT-5.x codex models (e.g., gpt-5.3-codex) do not support
```
edelauna, line 577: The JSDoc says "Azure-hosted GPT-5.x codex models" but the predicate also matches `codex-mini-latest` (a direct OpenAI model) and any future model whose name contains "codex". Worth updating the JSDoc to reflect the actual scope?

src/api/providers/openai.ts
```diff
@@ -502,6 +572,467 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
+		// Build tools in Responses API format (flat structure, not nested under function)
+		const tools = this._convertToolsForResponsesApi(metadata?.tools)
+
+		const useStreaming = this.options.openAiStreamingEnabled ?? true
+
+		// Build the request body (stream flag added per path below)
+		const requestBody: any = {
+			model: model.id,
+			input: formattedInput,
+			store: metadata?.store ?? false,
```
edelauna, line 608: openai-native.ts, openai-codex.ts, and xai.ts all hard-code `store: false`; `_completePromptWithResponsesApi` does too. Using `metadata?.store ?? false` here means `createMessage` and `completePrompt` for the same codex model could disagree if a caller passes `store: true`. Also, the JSDoc on `ApiHandlerCreateMessageMetadata.store` says `@default true` which contradicts both the fallback here and the hard-coded `false` everywhere else.

src/api/providers/openai.ts
```diff
@@ -502,6 +572,467 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
+				// Handle tool/function call completion
+				if (
+					event?.type === "response.tool_call_arguments.done" ||
+					event?.type === "response.function_call_arguments.done"
+				) {
+					continue
+				}
+
+				// Handle completion events with usage
+				if (event?.type === "response.done" || event?.type === "response.completed") {
```
edelauna, line 866: The SDK's `ResponseStreamEvent` union has `ResponseCompletedEvent` with `type: 'response.completed'` but no `'response.done'` variant — so the first arm of this condition is dead against a real stream. Most test mocks also use `"response.done"`, which means those tests only go through the real handler by coincidence. Should the condition be simplified to just `'response.completed'` and the mocks updated?

src/api/providers/openai.ts
```diff
@@ -332,6 +346,62 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
+					}
+				}
+			}
+			if (textParts.length > 0) {
+				return textParts.join("")
+			}
+		}
+
+		// Fallback: check for direct text in response
+		if (response?.text) {
```
edelauna, line 398: The `Response` interface in the SDK has no top-level `text` field (the convenience accessor is `output_text`), so `response?.text` will always be `undefined` here. Is this branch reachable? If it's meant as a forward-compatibility net, `response?.output_text` might be the right fallback.

src/api/providers/openai.ts
```diff
@@ -502,6 +572,467 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
+			try {
+				const response = await (this.client as any).responses.create({
+					...requestBody,
+					stream: false,
+				})
+
+				// Extract text, tool calls, and reasoning from response output
+				if (Array.isArray(response?.output)) {
+					for (const outputItem of response.output) {
+						if (outputItem?.type === "function_call" || outputItem?.type === "tool_call") {
```
edelauna, line 636: `ResponseFunctionToolCall` in the SDK types `type` as always `'function_call'` — is `'tool_call'` ever returned by the Responses API, or is this a defensive alias? Same question for `outputItem.tool_call_id` (the SDK field is `call_id`).

src/api/providers/openai.ts
```diff
@@ -502,6 +572,467 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
+			}
+		} catch (error) {
+			throw handleOpenAIError(error, this.providerName)
+		}
+	}
+
+	/**
+	 * Formats an Anthropic message array into the Responses API input format.
+	 */
+	private _formatConversationForResponsesApi(messages: Anthropic.Messages.MessageParam[]): any[] {
```
edelauna, line 939: `src/api/transform/responses-api-input.ts` already exports `convertToResponsesApiInput` for this exact Anthropic→Responses API conversion (used by xai.ts). Is there a reason for a separate private implementation here? One difference I noticed: the shared helper flushes pending user content before appending `function_call_output` items, which produces different ordering for mixed text+tool_result messages.

src/api/providers/__tests__/openai-codex-responses.spec.ts
```diff
@@ -0,0 +1,758 @@
...
+
+			const chunks: any[] = []
+			for await (const chunk of handler.createMessage("System", [{ role: "user", content: "Write file" }], {
+				taskId: "test",
+			})) {
+				chunks.push(chunk)
+			}
+
+			const toolCalls = chunks.filter((c) => c.type === "tool_call")
+			expect(toolCalls.length).toBe(1)
```
edelauna, line 406: This test doesn't reach the `streamedToolCallIds.has(callId)` dedup guard (openai.ts:806) because no arguments delta fires before the `output_item.done`. Could a test be added that emits both `arguments.delta` events AND an `output_item.done` for the same `call_id`, asserting that the `output_item.done` is suppressed (no duplicate `tool_call` chunk)?

src/api/providers/__tests__/openai-codex-responses.spec.ts
```diff
@@ -0,0 +1,758 @@
...
+			expect(input[0].content[0].type).toBe("input_text")
+			expect(input[0].content[0].text).toBe("What is 1+1?")
+
+			// Second item: assistant text
+			expect(input[1].role).toBe("assistant")
+			expect(input[1].content[0].type).toBe("output_text")
+
+			// Third item: function_call
+			expect(input[2].type).toBe("function_call")
+			expect(input[2].name).toBe("calculator")
```
edelauna, line 241: The `function_call` item's `call_id` and `arguments` aren't asserted here — only `type` and `name`. A bug in the `sanitizeOpenAiCallId` wiring that corrupts the call ID would go undetected, and since `input[3]`'s `call_id` isn't checked either, a mismatch between the function_call and its output wouldn't be caught.

src/api/providers/__tests__/openai-codex-responses.spec.ts
```diff
@@ -0,0 +1,758 @@
...
+
+			// Fourth item: function_call_output
+			expect(input[3].type).toBe("function_call_output")
+			expect(input[3].output).toBe("2")
```
edelauna, line 245: The `call_id` on the `function_call_output` isn't asserted — this field is how the API pairs a tool result to its originating call. A mismatch (e.g. from the `sanitizeOpenAiCallId` transform) would silently break tool result attribution.

edelauna (changes requested):
Thanks so much for addressing this, sorry it took so long to get around to reviewing.

PR #319, opened 2026-05-25
feat(opencode-go): add Opencode Go as a first-class provider (#172)

### LINKED ISSUE #172: Opencode Go plan not available as provider
Name of model: Models on Opencode Go Plan
Model creator: Multiple (provided by Opencode Go)
Link to model: https://opencode.ai/docs/go/  (all models under the Opencode Go subscription listed there

Opencode Go plan is a popular provider and provides access to open source coding models at a low cost. For now, accessing the model on Roo code seems to be via OpenAI Compatible API provider and its required to configured each model manually as a separate provider, and the model cannot be changed on the fly (see attached image)

Reasoning effort - This is available as a parameter on the provider settings - but don't have the option to change it on the fly as we can do in Kilo code. (see attached image). It would be very helpful if this is possible. 

Image

proyectoauraorg:
**Closes #172**

### What
Adds **Opencode Go** as a first-class provider — an OpenAI-compatible gateway (`https://opencode.ai/zen/go/v1`) with an API key and a **dynamic model list** fetched from `/v1/models`. Models can be switched on the fly via the model picker, instead of configuring each one manually as a separate OpenAI-Compatible provider.

### How
- New `OpencodeGoHandler` (extends `RouterProvider`, mirrors `VercelAiGatewayHandler`): streams text / reasoning / tool calls / usage via `chat/completions`.
- New fetcher `getOpencodeGoModels` (GET `/v1/models` with Bearer auth; permissive zod schema with sane defaults; fetched on demand since the endpoint needs a key, so it's not in the keyless prefetch list).
- Wired through types (`provider-settings`, secret key in `global-settings`), `RouterName` / `GetModelsOptions`, `modelCache` dispatch, `buildApiHandler`, `useSelectedModel`, the settings UI (`OpenCodeGo` component + provider dropdown + ModelPicker), and i18n.

### Scope / follow-ups
- First pass covers the **OpenAI-compatible** models (GLM-5/5.1, Kimi K2.5/K2.6, MiMo, Qwen3.5/3.6 Plus, DeepSeek V4). The 4 **Anthropic-format** models on the Go plan (`/v1/messages`) can be a follow-up.
- **Pricing is intentionally not parsed** — the `/models` pricing units aren't documented, so cost is left unknown rather than reported wrong. The default model id/info are only a fallback used before the live list resolves.

### Tests
New fetcher spec (response mapping, metadata fallback, Bearer header, error → empty map). `tsc` / `eslint` / `prettier` and the affected webview/src suites pass locally.

## Summary by CodeRabbit

* **New Features**
  * Add Opencode Go provider: configure an Opencode Go API key (stored securely), pick models, and use Opencode Go for streaming and non-streaming completions with sensible defaults.
* **Bug Fixes / Validation**
  * Require API key and model when Opencode Go is selected; refresh model list when key changes.
* **Localization**
  * Added UI text for Opencode Go API key in many languages.
* **Tests**
  * Added tests for model fetching/parsing, handler streaming, and settings UI.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/319?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


navedmerchant:
Thank you for the contribution, reviewing this now as this coding plan is pretty popular

packages/types/src/providers/opencode-go.ts
```diff
@@ -0,0 +1,21 @@
...
+// The values below are only a fallback used before the live list resolves.
+export const opencodeGoDefaultModelId = "glm-5.1"
+
+export const opencodeGoDefaultModelInfo: ModelInfo = {
+	maxTokens: 32_768,
+	contextWindow: 200_000,
+	supportsImages: false,
+	supportsPromptCache: false,
+	inputPrice: 0,
+	outputPrice: 0,
```
edelauna: Since pricing is intentionally omitted (the PR description says "reporting a wrong cost is worse than reporting unknown"), could these be dropped entirely rather than set to `0`?

`ModelInfoView` shows price when the field is `!== undefined`, so `0` will render as "$0.00 / 1M tokens" in the UI — implying the service is free. Fetched models from `parseOpencodeGoModel` correctly leave these fields absent; the fallback would be more consistent without them.

src/api/providers/opencode-go.ts
```diff
@@ -0,0 +1,143 @@
...
+		})
+	}
+
+	/**
+	 * Streams a chat completion response, yielding typed chunks for text,
+	 * reasoning, partial tool calls, and token usage.
+	 */
+	override async *createMessage(
+		systemPrompt: string,
+		messages: Anthropic.Messages.MessageParam[],
```
edelauna, line 49: I created an issue for this: https://github.com/Zoo-Code-Org/Zoo-Code/issues/337 - it's not urgent, but annoying that we have to do this for every provider.

src/api/providers/opencode-go.ts
```diff
@@ -0,0 +1,143 @@
...
+			stream: true,
+			stream_options: { include_usage: true },
+			tools: this.convertToolsForOpenAI(metadata?.tools),
+			tool_choice: metadata?.tool_choice,
+			parallel_tool_calls: metadata?.parallelToolCalls ?? true,
+		}
+
+		const completion = await this.client.chat.completions.create(body)
+
+		for await (const chunk of completion) {
```
edelauna, line 75: It's unfortunate that this needed to be duplicated, created an issue to refactor this into a common util: https://github.com/Zoo-Code-Org/Zoo-Code/issues/338

edelauna (commented):
looks good - had 1 nit, and created some issues for follow up things I'd like to refactor

packages/types/src/providers/opencode-go.ts
navedmerchant: I tested and it does not show any UI. I agree it would better to leave ths out to be consistent

proyectoauraorg:
Done — dropped `inputPrice`/`outputPrice` from the fallback `opencodeGoDefaultModelInfo` so the price stays unknown until the live `/v1/models` list resolves, matching the dynamically fetched models (which leave the fields absent). This avoids `ModelInfoView` rendering a misleading "$0.00 / 1M tokens". Thanks @edelauna and @navedmerchant! 🦓

src/api/providers/fetchers/opencode-go.ts
```diff
@@ -0,0 +1,91 @@
...
+			headers: apiKey ? { Authorization: `Bearer ${apiKey}` } : undefined,
+			timeout: 10_000,
+		})
+
+		const result = opencodeGoModelsResponseSchema.safeParse(response.data)
+		const rawData = result.success ? result.data.data : response.data?.data
+		const data = Array.isArray(rawData) ? rawData : []
+
+		if (!result.success) {
+			console.error(`Opencode Go models response is invalid: ${JSON.stringify(result.error.format())}`)
```
edelauna: Should this be `warn` instead of `error`? When the outer schema fails but per-item parsing below succeeds, all models load correctly — a `console.error` implies something is broken when it isn't.

```suggestion
			console.warn(`Opencode Go models response did not match expected schema; falling back to per-item parsing: ${JSON.stringify(result.error.format())}`)
```

src/api/providers/opencode-go.ts
```diff
@@ -0,0 +1,143 @@
...
+ * Supports text generation, reasoning content (GLM/DeepSeek), tool calls,
+ * and non-streaming prompt completion.
+ */
+export class OpencodeGoHandler extends RouterProvider implements SingleCompletionHandler {
+	/** Creates a new handler bound to the user's Go API key and selected model. */
+	constructor(options: ApiHandlerOptions) {
+		super({
+			options,
+			name: "opencode-go",
+			baseURL: "https://opencode.ai/zen/go/v1",
```
edelauna, line 35: The URL is also hardcoded in the fetcher (`OPENCODE_GO_BASE_URL`) and the same pattern exists for `vercel-ai-gateway` — created #350 to centralise all provider base URLs as a follow-up.

edelauna (dismissed):
Looks good - thanks for adding 🔥

src/api/providers/opencode-go.ts
edelauna: Opened: https://github.com/Zoo-Code-Org/Zoo-Code/issues/350 since it's a bit widespread.

src/api/providers/fetchers/__tests__/opencode-go.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+					],
+				},
+			})
+			const warnSpy = vitest.spyOn(console, "warn").mockImplementation(() => {})
+
+			const models = await getOpencodeGoModels("k")
+
+			expect(Object.keys(models)).toEqual(["valid-model"])
+			expect(warnSpy).toHaveBeenCalledTimes(1)
+			expect(warnSpy.mock.calls[0][0]).toContain("Skipping invalid Opencode Go model entry")
```
edelauna: Applying the `console.warn` suggestion introduced a second warn call — one for the outer schema mismatch, one for the invalid item — so `toHaveBeenCalledTimes(1)` now fails. Update to expect both:

```suggestion
			// Two warns: one for the outer schema mismatch, one for the invalid item
			expect(warnSpy).toHaveBeenCalledTimes(2)
			expect(warnSpy.mock.calls[0][0]).toContain("did not match expected schema")
			expect(warnSpy.mock.calls[1][0]).toContain("Skipping invalid Opencode Go model entry")
```

edelauna (dismissed):

edelauna (approved):

navedmerchant (approved):

PR #324, opened 2026-05-25
fix: parse Gemma 4 <thought> reasoning tags alongside <think>

### LINKED ISSUE #323: [BUG] <thought> tag is displayed as plain text instead of hidden reasoning in OpenAI Compatible provider for Gemma4
Gemma 4 (26B/31B) models stream reasoning inside `<thought>...</thought>` tags. Zoo Code currently only parses `<think>`, so the reasoning is treated as plain text. This causes:

- reasoning leaks into the chat
- the agent doesn't see a tool call on the first turn, triggering an unnecessary retry

**To Reproduce**
1. Add an OpenAI-Compatible provider pointing to a Gemma 4 model.
2. `Base URL` set to https://generativelanguage.googleapis.com/v1beta/openai
3. Set API Key obtained from https://aistudio.google.com/api-keys
4. Choose a reasoning Gemma4 model. There are only 2 options: `models/gemma-4-26b-a4b-it` and `models/gemma-4-31b-it`. Optionally, you can set Context Window Size to `256000` below.
Image**Expected behavior**

5. Save. Make sure it is selected down below.

6. Send any simple prompt (e.g., "2+2? Give only the answer")
7. Observe raw <thought>...</thought> in output

Image

Version: 3.55.0 (d63e7bdb)

The tags `<thought></thought>` wrapped around the reasoning are not expected to show up. Moreover, the whole reasoning should be wrapped into the expandable Thinking item.

What is expected? I already implemented this fix and I creating this feature only because I have to reference the issue.
Here is what is expected

Image

Version: 3.55.0 (69668070)

### Additional context
A similar issue alongside with a PR have already been presented for Roo Code:
https://github.com/RooCodeInc/Roo-Code/issues/7615
https://github.com/RooCodeInc/Roo-Code/pull/7617

It was closed. But note that other extensions like Kilo Code somehow solve that issue.
Also, it is not clear why Gemma4 and some other models in Zoo Code so often mention the tools like `attempt_completion` in the screenshots. I haven't experienced that for other models providing the same API Keys.
And sometimes they trigger additional API Request, even having the correct answer already, which I also did not see in other extensions

Image
(in this screenshot, openrouter free model was configured, but it happens with other model, it's just an example)

In the history, I see a message from a user like: `[ERROR] You did not use a tool in your previous response! Please retry with a tool use.`. And the rest info. However, it would be absolutely fine to just throw the result to the output and let a user decide what do to next or at least give a user a button like: Model did not use a tool. If you don't satisfied with the result, click below. And a button like "Proceed with a tool".

sagidM:
### Related GitHub Issue

Closes #323
Closes a similar issue: [#7615 from Roo Code](https://github.com/RooCodeInc/Roo-Code/issues/7615)
Similar PR to [#7617 from Roo Code](https://github.com/RooCodeInc/Roo-Code/pull/7617)

### Description
Gemma 4 streams reasoning inside `<thought>...</thought>` instead of `<think>...</think>`.  Without this the content leaks into chat text and the agent triggers a retry on the first turn.

- TagMatcher: support multiple tag names - `string[]`, track `activeTagName` so `<think>` is never closed by `</thought>` (and vice-versa).
- base-openai-compatible-provider and openai handler: match both tags.
- Tests: `<thought>` parsing, cross-tag isolation, and invariants.

### Test Procedure

**To Reproduce**
1. Add an OpenAI-Compatible provider pointing to a Gemma 4 model.
2. `Base URL` set to https://generativelanguage.googleapis.com/v1beta/openai
3. Set API Key obtained from https://aistudio.google.com/api-keys
4. Choose a reasoning Gemma4 model. There are only 2 options: `models/gemma-4-26b-a4b-it` and `models/gemma-4-31b-it`. Optionally, you can set Context Window Size to `256000` below.
Image**Expected behavior**

5. Save. Make sure it is selected down below.

6. Send any simple prompt (e.g., "2+2? Give only the answer")
7. Observe raw <thought>...</thought> in output

Image

Version: 3.55.0 (d63e7bdb)

The tags `<thought></thought>` wrapped around the reasoning are not expected to show up. Moreover, the whole reasoning should be wrapped into the expandable Thinking item.

# Expected
The reasoning/thinking content should be wrapped into an expandable menu option.

Image

Version: 3.55.0 (69668070)

### Pre-Submission Checklist

- [ ] **Issue Linked**: This PR is linked to an approved GitHub Issue (see "Related GitHub Issue" above).
- [ ] **Documentation Impact**: I have considered if my changes require documentation updates (see "Documentation Updates" section below).

### Documentation Updates

### Additional Notes
I specified some additional notes in the issue but they are not related to this PR.

### Get in Touch

Telegram: sagidM
Discord: sagidm

## Summary by CodeRabbit

* **New Features**
  * Streaming reasoning-tag parsing now recognizes both `<think>` and `<thought>` markers, emitting the correct streamed chunks as reasoning vs text.
  * Improved robustness for nested and mixed tags, including mismatched closing tags and correct end-of-stream flushing of partial reasoning.

* **Tests**
  * Expanded coverage for streaming tag extraction and chunking behavior, including multi-tag matching, nesting/stack unwinding, missing/no tags, and incomplete tag handling at end-of-stream.


src/api/providers/__tests__/base-openai-compatible-provider.spec.ts
```diff
@@ -95,6 +95,75 @@ describe("BaseOpenAiCompatibleProvider", () => {
 			])
 		})
 
+		it("should handle reasoning tags (<thought>) from stream", async () => {
```
taltas, line 100: These tests are great, but given you modified `openai.ts`, you should also update the relevant test file in `openai.spec.ts`, we need some tests to cover

- `<think>/<thought>` streaming regression tests
- assertions around streamed reasoning extraction and any provider specific sequencing behavior.

src/utils/tag-matcher.ts
```diff
@@ -57,38 +63,56 @@ export class TagMatcher<Result = TagMatcherResult> {
...
-				if (char === ">" && this.index === this.tagName.length) {
-					this.state = "TEXT"
-					if (!this.matched) {
-						this.cached = []
+				if (char === ">") {
+					const matched = this.candidates.find((c) => c.index === c.name.length)
+					if (matched) {
+						this.state = "TEXT"
+						this.activeTagName = matched.name
+						if (!this.matched) {
```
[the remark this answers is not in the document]
taltas: @sagidM This is a valid concern, we should correct this.

taltas (dismissed):
Thanks for the PR, I messed around with Gemma 4 in roo and had the same thing, thanks for the fix! Left a couple of comments, we should be good to merge after you address these

src/api/providers/__tests__/base-openai-compatible-provider.spec.ts
sagidM: I will try.
Note though, I changed `src/api/providers/base-openai-compatible-provider.ts` purely for tests! Without those changes, the code worked, changes in openai.ts were sufficient. But the tests failed.

And, if you ask me, I would probably make a map for different models as gemma-4-* outputs `<thought>` but other models can output other values. But since there was already `TagMatcher` used with `<think>` tags, I choose the easier and safer-to-be-approved path.
However, mapping the model name with the output tags they produce would eliminate the problem of overlapping tags. Probably.
Although, keeping that info (`startingWith("gemma-4-*" or "gemma-*")` or direct 2 keys of full names in a map) in the code base seems kind of dirty too.

sagidM (commented):
Before pushing, I was struggling to decide, how `TagMatcher` should work.
Consider these AI responses and the reasonings verdicts I implemented:
Gemma4 produced: `<think>outer<thought>inner</thought> middle</think>final`
```
"reasoning_content": "outer<thought>inner</thought> middle",
"content": "final"
```

Gemma4 produced: `<think>first</thought>second</think>final`
```
"reasoning_content": "first</thought>second",
"content": "final"
```

Gemma4 produced: `<think>User asks about <think>these</think>tags...</think>final`
```
"reasoning_content": "User asks about <think>these</think>tags...",
"content": "final"
```

All those tests cases are covered. I am push in a minute.

---

## Backtick problem.
For the prompt: `What does </thought> mean? Give me the short answer`
Here is the result (replaced `\n` inside json for readability):
```
<thought>*   Target phrase: `</thought>`
    *   Goal: Explain what it means.
    *   Constraint: \"Give me the short answer.\"

    *   The user is interacting with an AI.
    *   The AI often uses a `<thought>` block (Chain-of-Thought) to reason before providing the final response.
    *   `</thought>` is the closing tag of that reasoning block.

    *   *Detailed version:* It's a closing XML/HTML-style tag used by AI models to signify the end of their internal reasoning process (Chain-of-Thought) and the beginning of the final answer.
    *   *Short version:* It marks the end of the AI's internal reasoning process.</thought>It is a closing tag used by AI models to signal the **end of their internal reasoning process** (Chain-of-Thought) and the beginning of the final response.
```

Google wraps both tags into backticks.
However, this is a 1% use case. For the majority of responses, I don't think a user would see this problem. And even if it appears, the chances are that they expand the Thinking.

However, the request to
`https://generativelanguage.googleapis.com/v1beta/models/gemma-4-31b-it:generateContent?key=<API_KEY>`

Actually returns a normal result and has
```
`</thought>`
```
(wrapped into backticks) many times in `parts[0].text`.
Content is in `parts[1].text`.

Conclusion. While this PR would solve 99% of the leaking reasoning cases, the complete solution would be to support gemma-4-* models for the Google API Provider.

## Backticks Solution.
Regardless, I make another commit but I may wait until you guys confirm that this is really a problem.
Because right now, looking at the code, I have doubts whether it is worth to solve this problem. Because it brings extra complexity layer that TagMatcher should solve for some reason. There are may be other models which have a different way of escaping it and these changes will collide.
So I think it is better to publish this as-is and maybe focus on supporting Gemma for Google provider instead.

sagidM (commented):
Sorry, I accidentally send the previous message as **Comment** instead of **Submitting review**

src/utils/tag-matcher.ts
```diff
@@ -57,38 +66,61 @@ export class TagMatcher<Result = TagMatcherResult> {
...
+						this.collect()
+					}
 				}
 			} else if (this.state === "TAG_CLOSE") {
-				if (char === ">" && this.index === this.tagName.length) {
+				const tagName = this.activeTagNames.at(-1) || this.tagNames[0]
+				if (char === ">" && this.index === tagName.length) {
 					this.state = "TEXT"
 					this.depth--
+					this.activeTagNames.pop()
```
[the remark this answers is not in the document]
sagidM: Cannot reproduce.
I think, we do not enter there if we haven't encountered the opening tag.
If we come to a different closing tag, I expect we also don't enter this code.
We cannot come to the same closing tag because that would mean we already at the generated content.

I added 2 more tests to cover this 80b7dbbc0f3ce13c928a0d533fb346c37fc852c5

taltas:
@sagidM I'm making some time to re look at this today, thanks for your patience.

zakirhossen23:
is it deployed?

src/api/providers/base-openai-compatible-provider.ts
```diff
@@ -118,7 +118,7 @@ export abstract class BaseOpenAiCompatibleProvider<ModelName extends string>
 		const stream = await this.createStream(systemPrompt, messages, metadata)
 
 		const matcher = new TagMatcher(
-			"think",
+			["think", "thought"],
```
edelauna, line 121: `LmStudioHandler` (`lm-studio.ts:107`) and `NativeOllamaHandler` (`native-ollama.ts:217`) both extend `BaseProvider` directly with their own `createMessage()` and `TagMatcher` instances — they still pass only `"think"`. Should those be updated to `["think", "thought"]` too?

src/utils/tag-matcher.ts
```diff
@@ -17,11 +17,20 @@ export class TagMatcher<Result = TagMatcherResult> {
 	state: "TEXT" | "TAG_OPEN" | "TAG_CLOSE" = "TEXT"
 	depth = 0
 	pointer = 0
+	private readonly tagNames: string[]
+	private activeTagNames: string[] = []
+	private inCode = false
+	private codeFence = 0
+	private tickRun = 0
```
edelauna: Are `inCode`, `codeFence`, and `tickRun` intended for a follow-up? They're declared but never read or written anywhere in this file.

```suggestion
```

edelauna:
Thanks for putting this together, looks good - just had a question about some other providers, and unused variables.

edelauna (changes requested):

src/api/providers/base-openai-compatible-provider.ts
sagidM:
```js
// class TagMatcher
constructor(
	tagName: string | string[],
	readonly transform?: (chunks: TagMatcherResult) => Result,
	readonly position = 0,
) {
	this.tagNames = Array.isArray(tagName) ? tagName : [tagName]
}
```
should not matter I guess

sagidM:
Unfortunately, I could not push it due to
```
Object literal may only specify known properties, and 'allowUnsafeTemplateDir' does not exist in type '{ allowUnsafeCustomBinary?: boolean | undefined; allowUnsafeProtocolOverride?: boolean | undefined; allowUnsafePack?: boolean | undefined; }'.ts(2353)
index.d.ts(107, 5): The expected type comes from property 'unsafe' which is declared here on type 'Partial<SimpleGitOptions>'
```
unrelated to this PR, fails in the main branch too

edelauna:
> Unfortunately, I could not push it due to
> 
> ```
> Object literal may only specify known properties, and 'allowUnsafeTemplateDir' does not exist in type '{ allowUnsafeCustomBinary?: boolean | undefined; allowUnsafeProtocolOverride?: boolean | undefined; allowUnsafePack?: boolean | undefined; }'.ts(2353)
> index.d.ts(107, 5): The expected type comes from property 'unsafe' which is declared here on type 'Partial<SimpleGitOptions>'
> ```
> 
> unrelated to this PR, fails in the main branch too

Running `pnpm install` should update your `node-modules`.

I'll help address some of the feedback

edelauna (approved):
Thanks for addressing this.

PR #327, opened 2026-05-25
feat(chat): add compact tool UI mode for executed tool blocks (#322)

### LINKED ISSUE #322: [ENHANCEMENT] Add a strict "Compact Mode" for tool usage UI to save vertical space

When a tool is used, its UI block takes up a massive amount of vertical space, even when "collapsed" or reduced. Long tool descriptions (like MCP servers) and JSON payloads push the actual conversation out of view, completely breaking the chat readability.

Image

Happens to all users during chat sessions where tools are frequently called (especially MCP tools with very long descriptions, like firecrawl). It makes reviewing the chat history and following the assistant's logic extremely tedious due to constant scrolling.

Add a setting in the UI to force tool usages into a strictly "compact" or "ultra-minimized" mode by default. In this mode, a tool call should only display a single, thin line (e.g., [Tool used: firecrawl_scrape]) and completely hide the description, parameters, and payloads unless the user explicitly clicks to expand it.

The setting should be global (e.g., a toggle in the settings menu). The single-line summary should still be clickable to expand the full details if the user needs to inspect the payload.

### Request checklist

_No response_

Given the user is in a chat session,
When the assistant uses a tool (and compact mode is enabled),
Then the UI renders only a minimal single-line header for the tool execution.
And the tool's long description and JSON arguments remain completely hidden.
But the user can still click the header to expand and see the full details.

Add a boolean setting compactToolUI in the extension settings. When true, apply a stricter collapsed state to the tool usage components, hiding the verbose description and arguments body.

Risk: Users might miss exactly what arguments the tool is using if it's hidden by default.
Mitigation: Make this an opt-in toggle in the settings, or ensure the expansion interaction is very obvious.

proyectoauraorg:
### Related GitHub Issue

Closes #322

### Description

Adds an opt-in **Compact tool display** setting (`compactToolUI`, off by default) that collapses executed (`say`) tool blocks in the chat history to a single, clickable line — hiding long descriptions and JSON payloads until the user clicks to expand. This addresses the issue's complaint: verbose tool blocks push the actual conversation out of view and make history tedious to scroll.

### Scope

Compaction applies to executed **file-operation** tool rows — the `say` → `case "tool"` branch in `ChatRow`. The `use_mcp_server` branch (line ~1623) is **intentionally left untouched** in this PR. Per review discussion, **MCP history compaction will be handled in a separate follow-up issue/PR**, since MCP blocks have a different shape (server/tool/arguments/response) and deserve their own compact layout.

**Design decision:** approval prompts (`ask` tool messages) are **never** compacted — only completed/history (`say`) tool rows are — so users always see the parameters they are approving before acting.

### How it works

- New global boolean `compactToolUI` in `globalSettingsSchema` + `ExtensionState`, persisted via `ClineProvider.getState()` / `getStateToPostToWebview()` so it survives reloads.
- Toggle in **Context Management** settings (`compact-tool-ui-checkbox`).
- In `ChatRow`, the `say` tool branch renders a single line (`chevron + 🔧 + Tool: <name/path>`) when `compactToolUI && !isExpanded`; clicking the row (a semantic `<button>` with `aria-expanded={false}`) toggles expansion and falls through to the normal full render.
- New i18n keys (`settings:contextManagement.compactToolUI.*`, `chat:compactTool.*`) added across all 18 locales.

### Testing

- `ContextManagementSettings.spec.tsx`: toggle renders + fires `setCachedStateField("compactToolUI", …)`.
- `ChatRow.compact-tool.spec.tsx`: mounts `ChatRowContent` with `compactToolUI=true` and a `say` tool message; asserts the compact row renders, that clicking it calls `onToggleExpand(message.ts)`, and that the button carries `aria-expanded="false"`.
- `tsc -b` (webview) and `tsc --noEmit` (src) clean; eslint clean; `find-missing-translations` reports complete parity.


webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -224,6 +224,7 @@ export const ExtensionStateContextProvider: React.FC<{ children: React.ReactNode
 		cwd: "",
 		telemetrySetting: "unset",
 		showRooIgnoredFiles: true, // Default to showing .rooignore'd files with lock symbol (current behavior).
+		compactToolUI: false, // Default off — opt-in single-line collapsed tool blocks (#322).
```
edelauna, line 227: This default will always win on reload — `ClineProvider.getState()` reads `stateValues.compactToolUI` from storage via `contextProxy.getValues()` but never includes it in the returned object, so `getStateToPostToWebview()` never sends the persisted value to the webview. Shouldn't `getState()` also return `stateValues.compactToolUI ?? false` (following the pattern of `showRooIgnoredFiles` at line 2425 in `ClineProvider.ts`)?

webview-ui/src/components/chat/ChatRow.tsx
```diff
@@ -1405,6 +1414,27 @@ export const ChatRowContent = ({
 					const sayTool = safeJsonParse<ClineSayTool>(message.text)
 					if (!sayTool) return null
 
+					// Compact mode (#322): collapse executed (history) tool blocks to a single
+					// clickable line, hiding verbose params/payloads until expanded. Only applies
+					// to `say` tool rows here — approval prompts (`ask`) are never compacted.
+					if (compactToolUI && !isExpanded) {
```
edelauna, line 1420: This only compacts `case "tool":` messages (file ops). The `case "use_mcp_server":` branch at line 1623 is untouched — intentional? The PR description calls out MCP tools with long descriptions as the primary motivation.

edelauna (changes requested):

proyectoauraorg:
Thanks for the review @edelauna!

1. **Reload bug fixed** — `compactToolUI` is now wired through both `getState()` returns (following the `showRooIgnoredFiles` pattern), so the persisted value reaches the webview instead of the `false` default always winning on reload.
2. **On the `use_mcp_server` branch:** it renders as an `ask` (approval prompt), and the feature intentionally never compacts approvals so you can always see what you're authorizing before acting. Compacting MCP rows only makes sense once they're answered/in history — happy to do that as a follow-up if you'd like, otherwise I'll tighten the PR description to scope it to executed file-op tool blocks. Which do you prefer? 🦓

webview-ui/src/components/chat/ChatRow.tsx
```diff
@@ -1405,6 +1414,27 @@ export const ChatRowContent = ({
...
+						const compactLabel = sayTool.path ? `${sayTool.tool}: ${sayTool.path}` : sayTool.tool
+						return (
+							<button
+								type="button"
+								onClick={handleToggleExpand}
+								className="flex items-center gap-2 py-0.5 cursor-pointer text-vscode-descriptionForeground hover:text-vscode-foreground bg-transparent border-none text-inherit w-full text-left"
+								data-testid="compact-tool-row"
+								title={t("chat:compactTool.expandHint")}>
```
edelauna, line 1429: Would adding `aria-expanded={false}` here round out the a11y fix? Since this button only renders in the collapsed state, the value is always `false`, but it signals to screen readers that this is an expandable control.

```suggestion
							title={t("chat:compactTool.expandHint")}
							aria-expanded={false}>
```

webview-ui/src/components/settings/__tests__/ContextManagementSettings.spec.tsx
```diff
@@ -150,6 +151,23 @@ describe("ContextManagementSettings", () => {
...
+		const setCachedStateField = vi.fn()
+		render(<ContextManagementSettings {...defaultProps} setCachedStateField={setCachedStateField} />)
+
+		const checkbox = screen.getByTestId("compact-tool-ui-checkbox").querySelector("input")!
+		fireEvent.click(checkbox)
+
+		await waitFor(() => {
+			expect(setCachedStateField).toHaveBeenCalledWith("compactToolUI", true)
+		})
+	})
```
edelauna, line 169: These cover the settings toggle nicely. Would it be worth also adding a test that mounts `ChatRowContent` with `compactToolUI=true` and a `say` tool message, to verify the compact row renders (and that clicking it calls `onToggleExpand`)? That would give a regression guard on the actual chat-side behaviour.

edelauna (commented):
Ok - lets just update the PR description, we can handle MCP history compaction in another issue/pr.

Had some minor questions.

proyectoauraorg:

proyectoauraorg:
Thanks for the review @edelauna! I've addressed the feedback:

### Changes made:
1. **Accessibility fix**: Added `aria-expanded={isExpanded}` to the compact tool toggle button in `ChatRow.tsx` (addressing CodeRabbit's nitpick about screen reader support

### Regarding your comments:
- **PR description**: Will update to clarify that MCP history compaction is out of scope for this PR and should be handled in a separate issue/PR
- **MCP history compaction**: Agreed — I'll open a separate issue to track this as a follow-up enhancement

The CI checks should pick up the new commit shortly. Let me know if there are any other questions or changes needed!
EOF
)

proyectoauraorg:
Updated per your review @edelauna:

1. **aria-expanded**: Changed to `aria-expanded={false}` as you suggested — since the button only renders in the collapsed state, the value is always false but signals to screen readers this is an expandable control.

2. **ChatRow compact tool test**: Added `ChatRow.compact-tool.spec.tsx` with 3 tests:
   - Verifies compact row renders when `compactToolUI=true` and not expanded
   - Verifies clicking the compact row calls `onToggleExpand`
   - Verifies `aria-expanded=false` is present on the button

All tests pass. Let me know if you'd like any further changes!

webview-ui/src/components/chat/ChatRow.tsx
proyectoauraorg: Good call — added `aria-expanded={false}` on the compact toggle button. Since it only renders in the collapsed state the value is always `false`, so it just signals to screen readers that this is an expandable control. It's on the `<button>` at `ChatRow.tsx:1426`.

webview-ui/src/components/settings/__tests__/ContextManagementSettings.spec.tsx
proyectoauraorg: Done — added `webview-ui/src/components/chat/__tests__/ChatRow.compact-tool.spec.tsx`. It mounts `ChatRowContent` with `compactToolUI=true` and a `say` tool message and asserts: (1) the compact row renders, (2) clicking it calls `onToggleExpand(message.ts)`, and (3) the button carries `aria-expanded="false"`. Good regression guard — thanks for the nudge.

proyectoauraorg:
@edelauna addressed your notes:

- **PR description updated** — clarified the scope (only the `say` → `case "tool"` file-op branch is compacted; `use_mcp_server` is intentionally untouched) and noted MCP history compaction is deferred to a separate follow-up issue/PR, as agreed.
- **`aria-expanded={false}`** added on the compact toggle button (`ChatRow.tsx:1426`).
- **New `ChatRow.compact-tool.spec.tsx`** mounts `ChatRowContent` with `compactToolUI=true` + a `say` tool message and asserts the compact row renders, click calls `onToggleExpand`, and the button carries `aria-expanded="false"`.

Persistence across reloads is wired through `ClineProvider.getState()` / `getStateToPostToWebview()`. CI is green. Re-requesting review when you have a moment 🙏

proyectoauraorg:
@edelauna — friendly nudge for a re-review when you have a moment 🙏. Everything from the last round is in:

- `aria-expanded={false}` on the compact toggle, and it's now a semantic `<button>` (keyboard-accessible), covering the a11y notes.
- `compactToolUI` is persisted end-to-end — it's included in `ClineProvider.getStateToPostToWebview()` / `getState()` and in the `handleSubmit` save payload, so it survives reloads (the gap you and CodeRabbit flagged).
- Scope clarified in the description: only the `say` → `case "tool"` file-op branch is compacted; `use_mcp_server` is intentionally untouched, with MCP history compaction deferred to a follow-up.
- Added `ChatRow.compact-tool.spec.tsx` covering render + toggle + `aria-expanded`.

Branch is MERGEABLE. Thanks!

proyectoauraorg:
@edelauna — friendly nudge for re-review 🙏

All feedback from the last round has been addressed:

- ✅ `aria-expanded={false}` on compact toggle (semantic `<button>`, keyboard-accessible)
- ✅ `compactToolUI` persisted end-to-end via `ClineProvider.getState()` / `getStateToPostToWebview()` / `handleSubmit` save payload
- ✅ Scope clarified: only `say` → `case "tool"` file-op branch compacted; MCP intentionally untouched (deferred to follow-up)
- ✅ New `ChatRow.compact-tool.spec.tsx` covering render + toggle + aria-expanded
- ✅ Branch is MERGEABLE, CI green

Let me know if you need any further changes!

webview-ui/src/components/settings/ContextManagementSettings.tsx
```diff
@@ -229,6 +232,23 @@ export const ContextManagementSettings = ({
...
+						onChange={(e: any) => setCachedStateField("compactToolUI", e.target.checked)}
+						data-testid="compact-tool-ui-checkbox">
+						<label className="block font-medium mb-1">
+							{t("settings:contextManagement.compactToolUI.label")}
+						</label>
+					</VSCodeCheckbox>
+					<div className="text-vscode-descriptionForeground text-sm mt-1 mb-3">
+						{t("settings:contextManagement.compactToolUI.description")}
+					</div>
+				</SearchableSetting>
```
edelauna, line 250: Should this live in `UISettings.tsx` instead? `compactToolUI` controls how chat rows *render* (display preference), while this section manages what data gets sent to the model's context window (`maxOpenTabsContext`, `maxWorkspaceFiles`). `UISettings` already has the analogous `reasoningBlockCollapsed` (collapse thinking blocks in chat).

webview-ui/src/components/chat/__tests__/ChatRow.compact-tool.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+		expect(mockOnToggleExpand).toHaveBeenCalledWith(message.ts)
+	})
+
+	it("has aria-expanded=false on the compact button", () => {
+		const message = createSayToolMessage({ tool: "readFile", path: "src/file.ts" })
+		renderChatRow(message, false)
+
+		expect(screen.getByTestId("compact-tool-row")).toHaveAttribute("aria-expanded", "false")
+	})
+})
```
edelauna, line 108: These cover the happy path well. Worth adding negative-path cases for `isExpanded=true` and `compactToolUI=false`? Both branches guard the compact render — if either condition is accidentally dropped, no test would catch it.

```suggestion
describe("ChatRow - compact tool UI", () => {
	beforeEach(() => {
		vi.clearAllMocks()
		mockOnToggleExpand.mockClear()
	})

	it("renders the compact row when compactToolUI is true and not expanded", () => {
		const message = createSayToolMessage({ tool: "readFile", path: "src/file.ts" })
		renderChatRow(message, false)

		expect(screen.getByTestId("compact-tool-row")).toBeInTheDocument()
	})

	it("does NOT render compact row when isExpanded is true", () => {
		const message = createSayToolMessage({ tool: "readFile", path: "src/file.ts" })
		renderChatRow(message, true)

		expect(screen.queryByTestId("compact-tool-row")).not.toBeInTheDocument()
	})

	it("calls onToggleExpand when the compact row is clicked", () => {
		const message = createSayToolMessage({ tool: "readFile", path: "src/file.ts" })
		renderChatRow(message, false)

		fireEvent.click(screen.getByTestId("compact-tool-row"))

		expect(mockOnToggleExpand).toHaveBeenCalledWith(message.ts)
	})

	it("has aria-expanded=false on the compact button", () => {
		const message = createSayToolMessage({ tool: "readFile", path: "src/file.ts" })
		renderChatRow(message, false)

		expect(screen.getByTestId("compact-tool-row")).toHaveAttribute("aria-expanded", "false")
	})
})
```

edelauna (dismissed):
Looks good - I created a follow up issue for some non blocking feedback: https://github.com/Zoo-Code-Org/Zoo-Code/issues/455

DScoNOIZ:
@proyectoauraorg You have no idea how grateful I am to you! It was worth researching pulls. This is probably the most useful modification you've made.
I'll pull this into my local repository immediately.

I was going to implement all of this from scratch.
But I think there should be three modes.
To execute commands, MCP servers.
And file operations.
There should be the ability to flexibly configure these three things.


And there should be three modes. 1) Normal, like now.
2) Folded, that is, minimized, in a thin line like Pochi, with the ability to unfold.

3) And completely hidden, when we don't even see what tools are being executed. That is, they completely disappear from the chat, and the agent executes them completely in the background, invisible to us.


Moreover, it should be flexibly configured for each type (MCP, terminal commands, file operations.

I want to implement all this in my own place.
Thank you so much. Now I have something to start with!

edelauna (changes requested):
@proyectoauraorg can you add some screenshots for how the compaction works? I'm not able to reproduce it, and see a difference between before or after.

PR #332, opened 2026-05-26
fix(ThinkingBudget): extract maxOutputTokensControl to resolve undefined reference in binary reasoning path

proyectoauraorg:
## Summary

Extracts the `maxOutputTokensControl` JSX fragment as a named variable declared **before** both the binary reasoning and budget reasoning conditional branches in `ThinkingBudget.tsx`.

This fixes a scoping issue where the max output tokens slider was only defined inside the budget reasoning block, making it unavailable (and would crash at runtime) for models using `supportsReasoningBinary` (binary reasoning toggle).

## Changes

- **Extracted `maxOutputTokensControl`** as a standalone JSX variable at line 167, before the binary/budget conditional branches
- **Added `{maxOutputTokensControl}`** inside the binary reasoning `<div>` so the slider appears for both reasoning modes

## Motivation

This fix addresses a bug reported in PR #274 by @edelauna: models using binary reasoning mode (e.g., Zhipu GLM via `supportsReasoningBinary`) could not configure `modelMaxTokens` because the slider control was scoped only within the budget reasoning conditional path.

## Related

- Fixes issue described in [PR #274](https://github.com/Zoo-Code-Org/Zoo-Code/pull/274)
- Addresses @edelauna's code review comment about `modelMaxTokens` bug

## Summary by CodeRabbit

* **New Features**
  * Thinking budget settings now show the max output‑tokens control in the binary reasoning option when reasoning effort is enabled.
* **Tests**
  * Added tests covering binary reasoning: verifies the reasoning toggle shows the slider only when enabled and that adjusting the slider updates the model output‑token setting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/332?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/webview-ui/src/components/settings/ThinkingBudget.tsx
+++ b/webview-ui/src/components/settings/ThinkingBudget.tsx
@@ -164,6 +164,27 @@ export const ThinkingBudget = ({ apiConfiguration, setApiConfigurationField, mod
 		return null
 	}
 
+	// Max output tokens control used by both binary reasoning and budget reasoning paths.
+	const maxOutputTokensControl = (
+		<div className="flex flex-col gap-1">
+			<div className="font-medium">{t("settings:thinkingBudget.maxTokens")}</div>
+			<div className="flex items-center gap-1">
+				<Slider
+					min={8192}
+					max={Math.max(
+						modelInfo.maxTokens || 8192,
+						customMaxOutputTokens,
+						DEFAULT_HYBRID_REASONING_MODEL_MAX_TOKENS,
+					)}
+					step={1024}
+					value={[customMaxOutputTokens]}
+					onValueChange={([value]) => setApiConfigurationField("modelMaxTokens", value)}
+				/>
+				<div className="w-12 text-sm text-center">{customMaxOutputTokens}</div>
+			</div>
+		</div>
+	)
+
 	// Models with supportsReasoningBinary (binary reasoning) show a simple on/off toggle
 	if (isReasoningSupported) {
 		return (
@@ -175,6 +196,7 @@ export const ThinkingBudget = ({ apiConfiguration, setApiConfigurationField, mod
 					}>
 					{t("settings:providers.useReasoning")}
 				</Checkbox>
+				{maxOutputTokensControl}
 			</div>
 		)
 	}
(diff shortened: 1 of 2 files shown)


webview-ui/src/components/settings/ThinkingBudget.tsx
```diff
@@ -175,6 +196,7 @@ export const ThinkingBudget = ({ apiConfiguration, setApiConfigurationField, mod
 					}>
 					{t("settings:providers.useReasoning")}
 				</Checkbox>
+				{maxOutputTokensControl}
```
edelauna: When reasoning is off, does this slider have any effect? For `claude-opus-4-7` (which has both `supportsReasoningBinary` and `supportsReasoningBudget`), `getModelMaxOutputTokens` returns `ANTHROPIC_DEFAULT_MAX_TOKENS = 8192` regardless of the user-set value once `enableReasoningEffort` is false — so the slider appears interactive but is silently ignored by the API.

Worth making this conditional to match the budget-path behaviour?

```suggestion
			{enableReasoningEffort && maxOutputTokensControl}
```

edelauna (commented):
Nice - thanks for adding this, would it be possible to write an e2e or integration test verifying the component works as intended?

Also could you include a screenshot

proyectoauraorg:
Applied your suggestion — the max-output-tokens slider is now gated behind `enableReasoningEffort` in the binary-reasoning path, so it no longer appears interactive while being silently ignored by `getModelMaxOutputTokens`. Added tests covering both states (hidden when reasoning is off, shown + wired when on). Screenshot incoming. Thanks @edelauna! 🦓

webview-ui/src/components/settings/ThinkingBudget.tsx
```diff
@@ -164,6 +164,27 @@ export const ThinkingBudget = ({ apiConfiguration, setApiConfigurationField, mod
 		return null
 	}
 
+	// Max output tokens control used by both binary reasoning and budget reasoning paths.
```
edelauna, line 167: The comment says "used by both binary reasoning and budget reasoning paths", but the budget path (around line 222) still has its own copy of this same slider JSX. Could `maxOutputTokensControl` replace that block there too?

proyectoauraorg:

edelauna (changes requested):
Could you look into: https://github.com/Zoo-Code-Org/Zoo-Code/pull/332#discussion_r3313091211

PR #229, opened 2026-05-21
Feat/zoo gateway

JamesRobert20:
# Pull Request: Zoo Gateway Provider Integration

### Description

This PR introduces **Zoo Gateway** as a new provider option in Zoo Code, enabling users to access AI models through the Zoo Code platform's unified API gateway.

#### Key Implementation Details

**1. New Provider Architecture**

- [`src/api/providers/zoo-gateway.ts`](src/api/providers/zoo-gateway.ts) — Full provider handler extending `RouterProvider` with OpenAI-compatible API, prompt caching support for Claude models, and enrichment headers for request tracking (`X-Zoo-Task-ID`, `X-Zoo-Mode`, `X-Zoo-Extension-Version`, `X-Zoo-Editor`)

- [`src/api/providers/fetchers/zoo-gateway.ts`](src/api/providers/fetchers/zoo-gateway.ts) — Model fetcher that retrieves available models from the Zoo Gateway `/models` endpoint, reusing Vercel AI Gateway's Zod schemas and parsing logic since API formats are identical

- [`packages/types/src/providers/zoo-gateway.ts`](packages/types/src/providers/zoo-gateway.ts) — Type definitions with `claude-sonnet-4` as default model, shared prompt caching model list with Vercel AI Gateway

**2. Authentication Flow**

Authentication uses the existing Zoo Code OAuth flow (via `/auth-callback` URI handler):

1. User clicks "Sign in to Zoo Code" button in settings
2. Browser opens Zoo Code website authentication page
3. After successful auth, callback URL returns `token`, `name`, `email`, and `image`
4. [`handleUri.ts`](src/activate/handleUri.ts) processes callback across **all** ClineProvider instances (not just visible ones) to ensure token persistence even when sidebar is hidden
5. Token is stored as `zooSessionToken` in zoo-gateway provider profiles

**3. Multi-Profile Token Synchronization**

A significant design decision was made to handle profile management robustly:

- [`ClineProvider.handleZooCodeCallback()`](src/core/webview/ClineProvider.ts) scans **all** profiles with `apiProvider === "zoo-gateway"` and updates tokens in each (profile names are user-renameable)
- [`ensureZooGatewayProfileSeeded()`](src/core/webview/ClineProvider.ts) handles migration for existing users who authenticated before Zoo Gateway was added — creates a profile or updates stale tokens on webview init
[...]

PR #344, opened 2026-05-27
feat(zoo-gateway): provider types, handler, and model fetcher

JamesRobert20:
## Summary
- Adds Zoo Gateway provider types, OpenAI-compatible handler, and authenticated model fetcher
- Skips model cache reads/writes for zoo-gateway (auth-scoped model lists)
- Adds request timeout and safe error logging on model discovery

Part 1 of the Zoo Gateway stack (split from #229).

## Test plan
- [ ] Unit tests: modelCache, zoo-gateway fetcher
- [ ] After stack merges: Zoo Gateway model list loads when signed in

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added Zoo Gateway as a supported API provider with dynamic model discovery, UI model selection, a preconfigured Anthropic default model and default temperature, session-token authentication, streaming and single-request completions, and optional prompt-caching for eligible models.

* **Behavior**
  * Treats certain providers as Anthropic-style when model IDs start with `anthropic/`. Model discovery for Zoo Gateway skips local caches and exposes an initially empty model list for remote population.

* **Tests**
  * Added unit tests for model discovery, parsing, streaming, error handling, and token/log redaction.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/344?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


proyectoauraorg:
## Conflict Resolution Patch

Hi @JamesRobert20 — I've prepared a rebased version of this PR against current main (4d71e5f44) that resolves the merge conflicts caused by #319 (opencode-go).

**Branch**: `proyectoauraorg/Zoo-Code:feat/zoo-gateway-types-core-rebased`

### What was done
- Rebased all 6 commits onto current main
- Resolved adjacent insertion conflicts in 8 files by preserving both `opencode-go` and `zoo-gateway` entries
- Added proper `opencode-go` case in `useSelectedModel.ts` and `modelCache.ts` (from main)
- Added `opencode-go` to `webviewMessageHandler.ts` router models mock

### Verification
- ✅ `tsc --noEmit` — zero errors (src + webview-ui)
- ✅ 39 zoo-gateway tests pass (19 handler + 20 validate)
- ✅ `turbo check-types` — all 11 packages pass
- ✅ Lint passes (eslint webview-ui)
- ✅ Pre-commit + pre-push hooks pass

### How to apply
You can cherry-pick the resolution or rebase your branch onto this one:
```bash
git fetch proyectoauraorg feat/zoo-gateway-types-core-rebased
git rebase proyectoauraorg/feat/zoo-gateway-types-core-rebased
```

Happy to help with any further adjustments!

src/api/providers/zoo-gateway.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export class ZooGatewayHandler extends RouterProvider implements SingleCompletionHandler {
+	constructor(options: ApiHandlerOptions) {
+		const baseURL = options.zooGatewayBaseUrl ?? `${getZooCodeBaseUrl()}/api/gateway/v1`
+
+		// Fail fast with a clear message instead of waiting for a 401.
+		// The token is set automatically by handleZooCodeCallback() after the user
+		// authenticates via the "Sign in with Zoo Code" flow in the extension.
+		if (!options.zooSessionToken) {
+			throw new Error("Zoo Gateway requires authentication. Please sign in to Zoo Code first.")
```
edelauna: Every other provider defers auth errors to the first API call — `buildApiHandler` callers like `Task.updateApiConfiguration` (`Task.ts:1436`) don't wrap it in try/catch and don't expect a synchronous throw. If a user has zoo-gateway configured but their token was cleared between sessions, this will crash task construction. Would it make sense to move this guard into `createMessage`/`completePrompt` to match the pattern of the other providers?

src/api/providers/fetchers/zoo-gateway.ts
```diff
@@ -0,0 +1,126 @@
...
+	try {
+		const response = await axios.get<ZooGatewayModelsResponse>(`${baseURL}/models`, {
+			headers,
+			timeout: MODEL_DISCOVERY_TIMEOUT_MS,
+		})
+		const result = zooGatewayModelsResponseSchema.safeParse(response.data)
+
+		// Fall back to the raw response only when it looks structurally sound; otherwise return
+		// an empty list rather than crashing on `response.data.data` being undefined.
+		const data = result.success ? result.data.data : Array.isArray(response.data?.data) ? response.data.data : []
```
edelauna: When `result.success` is false, the fallback still iterates `response.data.data` and the models end up cast as `VercelAiGatewayModel` without any field validation (line 97). A bad field shape (e.g. `context_window` as a string) could silently produce NaN pricing. Could we return `models` early here instead?

```suggestion
		if (!result.success) {
			console.error(`Zoo Gateway models response is invalid ${JSON.stringify(result.error.format())}`)
			return models
		}

		const data = result.data.data
```

src/api/providers/fetchers/zoo-gateway.ts
```diff
@@ -0,0 +1,126 @@
...
+			const { id } = model
+
+			// Only include language models for chat inference.
+			// Embedding models are statically defined in embeddingModels.ts.
+			if (model.type !== "language") {
+				continue
+			}
+
+			// Parse model using the same logic as Vercel AI Gateway since formats are identical
+			models[id] = parseZooGatewayModel({ id, model: model as VercelAiGatewayModel })
```
edelauna: The Zoo Gateway schema validates the response, but the result is immediately cast back to `VercelAiGatewayModel` for parsing. If these two schemas ever diverge (e.g. Vercel adds a new required pricing field), this cast could silently produce wrong shapes. Would it make sense to re-export the Vercel schemas directly here rather than re-declaring them independently?

src/api/providers/fetchers/modelCache.ts
```diff
@@ -167,12 +175,22 @@ export const getModels = async (options: GetModelsOptions): Promise<ModelRecord>
 export const refreshModels = async (options: GetModelsOptions): Promise<ModelRecord> => {
 	const { provider } = options
 
-	// Check if there's already an in-flight refresh for this provider
+	// Zoo Gateway models are user-specific (auth-scoped). Mirror the bypass in
```
taltas: Isn't this true of other providers like Vercel? I'd be a little surprised if they weren't doing this. It also feels a bit counter intuitive to have zoo-gateway skip cache in a file called `modelCache` 😂😂

packages/types/src/provider-settings.ts
```diff
@@ -574,8 +585,13 @@ export const getApiProtocol = (provider: ProviderName | undefined, modelId?: str
 		return "anthropic"
 	}
 
-	// Vercel AI Gateway uses anthropic protocol for anthropic models.
-	if (provider && provider === "vercel-ai-gateway" && modelId && modelId.toLowerCase().startsWith("anthropic/")) {
+	// Vercel AI Gateway, Zoo Gateway, and Roo use anthropic protocol for anthropic models.
+	if (
+		provider &&
+		["vercel-ai-gateway", "zoo-gateway", "roo"].includes(provider) &&
```
taltas: why do we need `roo` here?

src/api/providers/fetchers/modelCache.ts
```diff
@@ -132,13 +139,14 @@ export const getModels = async (options: GetModelsOptions): Promise<ModelRecord>
...
-		if (modelCount > 0) {
+		// Zoo Gateway models are user-specific - skip caching entirely
+		if (modelCount > 0 && !shouldSkipCache) {
 			memoryCache.set(provider, models)
 
 			await writeModels(provider, models).catch((err) =>
 				console.error(`[MODEL_CACHE] Error writing ${provider} models to file cache:`, err),
 			)
-		} else {
+		} else if (modelCount === 0) {
```
taltas, line 157: I'm not sure why it needs to be explicit here, but not a big deal.

src/api/providers/fetchers/modelCache.ts
```diff
@@ -167,12 +175,22 @@ export const getModels = async (options: GetModelsOptions): Promise<ModelRecord>
 export const refreshModels = async (options: GetModelsOptions): Promise<ModelRecord> => {
 	const { provider } = options
 
-	// Check if there's already an in-flight refresh for this provider
+	// Zoo Gateway models are user-specific (auth-scoped). Mirror the bypass in
+	// getModels() so we never persist one user's model list and serve it to a
+	// different authenticated user from cache.
+	const shouldSkipCache = provider === "zoo-gateway"
```
taltas: `const shouldSkipCache = provider === "zoo-gateway"` this is repeated over and over in this file. Is there a better way we could write this so it's in a single point of control? We could make a provider policy file that delineates what is skippable for each provider.

src/api/providers/zoo-gateway.ts
```diff
@@ -0,0 +1,177 @@
...
+		]
+
+		// Apply prompt caching for models that support it
+		// Zoo Gateway serves the same models as Vercel AI Gateway, so caching support is identical
+		if (VERCEL_AI_GATEWAY_PROMPT_CACHING_MODELS.has(modelId) && info.supportsPromptCache) {
+			addCacheBreakpoints(systemPrompt, openAiMessages)
+		}
+
+		// Build request headers with enrichment metadata
+		const requestHeaders: Record<string, string> = {}
```
taltas, line 88: Nit: You could shove the request headers into helper to make this a bit easier  to read, but not a big deal.

taltas (changes requested):
A few questions and comments.

src/api/providers/fetchers/modelCache.ts
JamesRobert20: Good call. Centralized in 2f9a5a0cb via an `AUTH_SCOPED_PROVIDERS` set + `isAuthScopedProvider(provider)` helper at the top of `modelCache.ts`, so the three call sites (`getModels`, `refreshModels`, in-flight dedupe) all read from one place. Easy to extend if another auth-scoped provider lands later.

On the irony of opting out of caching inside `modelCache.ts`: agreed, but the file is also the single funnel for model fetches across providers, so the bypass logic has to live somewhere here. Naming it `AUTH_SCOPED_PROVIDERS` at least makes the *why* explicit at the top of the file.

src/api/providers/fetchers/modelCache.ts
JamesRobert20: Vercel AI Gateway uses a public catalog (no per-user allowlist), so its model list is the same for everyone and is safe to cache on disk. Zoo Gateway is auth-scoped: each user gets a token-specific allowlist (test vs live, teams etc.), so caching could leak one user's catalog to the next signed-in user on the same machine. That asymmetry is the reason for the bypass.

packages/types/src/provider-settings.ts
JamesRobert20: Good catch, `roo` is in `retiredProviderNames` so it should not be in this allowlist anymore. Removed in 2f9a5a0cb.

src/api/providers/fetchers/modelCache.ts
JamesRobert20: Pre-existing `else if (modelCount === 0)` branch, kept as is so the empty-response telemetry still fires regardless of whether we ended up caching. Tightened the surrounding comment in 22ea61c87 to make the intent clearer.

src/api/providers/zoo-gateway.ts
JamesRobert20: Fair point. Leaving as-is for this PR since it's only two conditional headers (`X-Zoo-Task-ID`, `X-Zoo-Mode`) on a single call path. Happy to extract a `buildRequestHeaders(metadata)` helper in a follow-up if more enrichment headers land.

taltas (approved):
Changes addressed

PR #345, opened 2026-05-27
feat(zoo-gateway): settings UI, validation, and i18n

JamesRobert20:
## Summary
- Adds ZooGateway settings component, provider wiring in ApiOptions, and sign-in validation
- Initializes `zooGatewayModelId` when switching to Zoo Gateway
- Adds i18n strings across locales (including formal Turkish validation copy)

Part 2 of the Zoo Gateway stack. **Depends on #344.**

## Test plan
- [ ] webview-ui: ApiOptions, validate tests
- [ ] Manual: select Zoo Gateway, sign-in prompt, model picker

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Zoo Gateway provider added to API settings with sign-in/account UI, automatic model selection, and settings integration
  * Provider docs and model-default resolution unified for consistent behavior
* **Validation / Bug Fixes**
  * Form validation updated to avoid duplicate Zoo Gateway sign-in errors and respect auth state in welcome flow
* **Tests**
  * New/updated tests covering Zoo Gateway, validation, and API fetch robustness
* **Localization**
  * Zoo Gateway strings added across many locales (17+)


proyectoauraorg (approved):
## Review: Settings UI, Validation, and i18n

**Verdict**: ✅ Approved

### Verified
- ✅ `ApiOptions.spec.tsx` — 18 tests pass (including zoo-gateway modelId initialization)
- ✅ `validate.spec.ts` — 20 tests pass
- ✅ `ZooGateway.tsx` — clean component following existing provider pattern
- ✅ `validate.ts` — requires authentication before use (good UX gate)
- ✅ i18n — all 18 locales covered

### Notes
- `vite.config.ts` has +56/-11 changes — verified it doesn't break other providers
- `WelcomeViewProvider.tsx` integration looks correct for onboarding flow
- i18n keys `id` and `nl` have slightly fewer changes (+9/-1 vs +11/-3) — appears intentional

### Dependency
Depends on #344. After #344 merges, this should rebase cleanly.

webview-ui/src/components/welcome/WelcomeViewProvider.tsx
```diff
@@ -36,7 +36,8 @@ const getWelcomeApiConfiguration = (apiConfiguration?: ProviderSettings): Provid
 }
 
 const WelcomeViewProvider = () => {
-	const { apiConfiguration, currentApiConfigName, setApiConfiguration, uriScheme } = useExtensionState()
+	const { apiConfiguration, currentApiConfigName, setApiConfiguration, uriScheme, zooCodeIsAuthenticated } =
```
edelauna, line 39: The deeper issue is that threading `zooCodeIsAuthenticated` through validation, props, and `useCallback` deps is a smell — it's caused by auth state living in two places (`ProviderSettings.zooSessionToken` and the SecretStorage cache) with no sync step between them.

VS Code has a first-class API for exactly this: [`vscode.authentication.registerAuthenticationProvider`](https://github.com/microsoft/vscode-extension-samples/blob/main/authenticationprovider-sample/src/authProvider.ts). Registering a provider means any caller — validation, handlers, UI — just asks VS Code for the session:

```ts
const session = await vscode.authentication.getSession('zoo-code', [], { createIfNone: false })
if (!session) return i18next.t('settings:validation.zooGatewaySignIn')
```

This eliminates the threading entirely:
- Single source of truth — no split between profile token and cached token
- `onDidChangeSessions` fires automatically on sign-in/out, no manual state propagation needed
- Validation reads only `ProviderSettings` like every other provider
- VS Code's native Accounts menu (bottom-left status bar) handles sign-out for free

webview-ui/src/utils/validate.ts
```diff
@@ -128,6 +132,11 @@ function validateModelsAndKeysProvided(apiConfiguration: ProviderSettings): stri
 				return i18next.t("settings:validation.apiKey")
 			}
 			break
+		case "zoo-gateway":
+			if (!apiConfiguration.zooSessionToken && !zooCodeIsAuthenticated) {
+				return i18next.t("settings:validation.zooGatewaySignIn")
```
edelauna, line 137: Is this branch covered in `validate.spec.ts`? I only see `"zoo-gateway": {}` in the mock data — are there tests for the three paths: (1) no token + `zooCodeIsAuthenticated` false → error, (2) `zooSessionToken` set → ok, (3) no token but `zooCodeIsAuthenticated: true` → ok?

webview-ui/src/components/settings/providers/ZooGateway.tsx
```diff
@@ -0,0 +1,126 @@
...
+				{!zooCodeIsAuthenticated ? (
+					<div className="flex flex-col gap-1">
+						<p className="text-xs text-vscode-descriptionForeground">
+							{t("settings:providers.zooGateway.signInDescription")}
+						</p>
+						<a
+							href={authUrl}
+							className="inline-flex w-fit items-center rounded-sm bg-vscode-button-background px-3 py-1 text-xs text-vscode-button-foreground no-underline hover:bg-vscode-button-hoverBackground">
+							{t("settings:providers.zooGateway.signInButton")}
+						</a>
```
edelauna: Other providers with auth links use `VSCodeButtonLink` (e.g. `OpenRouter.tsx` and `OpenCodeGo.tsx`). Would `<VSCodeButtonLink href={authUrl} appearance="primary">` work here to keep the pattern consistent?

webview-ui/vite.config.ts
edelauna, line 1: This change should be reverted — it's not related to zoo-gateway and it conflicts with `#214` (`chore(webview): migrate build to Vite 8`), which deliberately introduced `rolldownOptions`, `tsconfigPaths`, and the simplified chunk config three days ago. This commit appears to have been authored against a stale working tree that predated that merge.

edelauna (commented):
Nice first pass! Had some feedback to prevent threading a boolean through layers.

Thanks again for breaking this up, makes review much faster / easier.

webview-ui/src/components/settings/providers/ZooGateway.tsx
```diff
@@ -0,0 +1,125 @@
...
+type ZooGatewayProps = {
+	apiConfiguration: ProviderSettings
+	setApiConfigurationField: (field: keyof ProviderSettings, value: ProviderSettings[keyof ProviderSettings]) => void
+	routerModels?: RouterModels
+	organizationAllowList: OrganizationAllowList
+	modelValidationError?: string
+	simplifySettings?: boolean
+}
+
+function isSonnet45ModelId(id: string) {
```
taltas: Not your fault because it's the standard, but putting numbers for function names feels so so wrong 😂😂

webview-ui/src/components/settings/ApiOptions.tsx
```diff
@@ -272,9 +274,17 @@ const ApiOptions = ({
...
 		)
 		setErrorMessage(apiValidationResult)
-	}, [apiConfiguration, routerModels, organizationAllowList, setErrorMessage, isRetiredSelectedProvider])
+	}, [
+		apiConfiguration,
+		routerModels,
+		organizationAllowList,
+		setErrorMessage,
+		isRetiredSelectedProvider,
+		zooCodeIsAuthenticated,
```
taltas: As @edelauna mentioned below, we should be wary of pumping `zooCodeIsAuthenticated` in the useEffect here as well. , I don't think we should be pumping`zooCodeIsAuthenticated` here in the useEffect through here. if you remove it, then add 



```
+		if (apiConfiguration.apiProvider === "zoo-gateway") {
+			setErrorMessage(undefined)
+			return
+		}

```

You will get a generic error message if something goes wrong. 

If you want to add specific error messaging you can add it on the UI side in webview-`ui/src/components/settings/providers/ZooGateway.tsx` with `ApiErrorMessage` component

```
+import { ApiErrorMessage } from "../ApiErrorMessage"
@@
				{!zooCodeIsAuthenticated ? (
					<div className="flex flex-col gap-1">
+						<ApiErrorMessage errorMessage={t("settings:validation.zooGatewaySignIn")} />
						<p className="text-xs text-vscode-descriptionForeground">
							{t("settings:providers.zooGateway.signInDescription")}
						</p>
						<VSCodeButtonLink href={authUrl} appearance="primary">
							{t("settings:providers.zooGateway.signInButton")}
						</VSCodeButtonLink>
					</div>
				) : (
```

taltas (changes requested):
One larger comment

webview-ui/src/components/settings/providers/ZooGateway.tsx
JamesRobert20: Agreed, baking the version into the helper name was awkward. Renamed to `isClaudeSonnetModelId` in f6aae3c8a, with the version-priority chain (4.5 exact, 4.5 substring, 4.5 dash variant, 4 family, first sonnet) expressed directly inside `pickZooGatewayDefaultModelId`. The helper now just answers "is this a Claude Sonnet of any version".

webview-ui/src/components/settings/ApiOptions.tsx
JamesRobert20: Done in f6aae3c8a, taking your suggestion exactly:

1. `ApiOptions.tsx` short-circuits zoo-gateway in the validation effect and clears the error, so `zooCodeIsAuthenticated` no longer needs to be in the dep list.
2. `validateApiConfigurationExcludingModelErrors` returns `undefined` for zoo-gateway entirely, so the form-level path stops carrying auth state.
3. `ZooGateway.tsx` renders the sign-in error inline using `<ApiErrorMessage errorMessage={t("settings:validation.zooGatewaySignIn")} />` above the existing description + sign-in button.

`validateApiConfiguration` (the welcome-view entry point) still threads `zooCodeIsAuthenticated` since the welcome screen does need to gate "Get Started" on auth state, and `validate.spec.ts` was reorganized so the three zoo-gateway branches (no token + unauth, no token + auth, token set) now exercise that entry point, with a separate suite asserting the form-level path no-ops for zoo-gateway.

React 19.2's `useEffectEvent` would be the textbook fit here, but webview-ui is still on React 18.3.1, so this approach is the closest equivalent until we upgrade.

webview-ui/src/utils/validate.ts
JamesRobert20: Tests added in f6aae3c8a (`validate.spec.ts`):

- `validateApiConfiguration` suite: (1) no token + `zooCodeIsAuthenticated` false returns `settings:validation.zooGatewaySignIn`, (2) no token + auth true returns undefined, (3) `zooSessionToken` set + auth false returns undefined.
- `validateApiConfigurationExcludingModelErrors` suite: confirms the form-level path now no-ops for zoo-gateway regardless of token/auth state, since the inline `<ApiErrorMessage>` in `ZooGateway.tsx` owns that UX (see the related thread on `ApiOptions.tsx`).

webview-ui/src/components/settings/providers/ZooGateway.tsx
JamesRobert20: Already on `VSCodeButtonLink` in this file (the sign-in CTA at the bottom of the unauthenticated branch uses `<VSCodeButtonLink href={authUrl} appearance="primary">`). After f6aae3c8a the unauthenticated card stacks: inline `<ApiErrorMessage>` for the validation error, the existing description, then the same `<VSCodeButtonLink>`, which matches the `OpenRouter`/`OpenCodeGo` pattern.

webview-ui/src/components/settings/ApiOptions.tsx
taltas: Not your fault, but I think we should replace this entire thing in with a separate file, the ApiOptions file is geting pretty long. Something like

```
const config = PROVIDER_UI_REGISTRY[value]?.modelConfig
const name = PROVIDER_UI_REGISTRY[selectedProvider]?.label
const slug = PROVIDER_UI_REGISTRY[selectedProvider]?.docsSlug || selectedProvider
```


Plus a new file with all the config:
```
export const PROVIDER_UI_REGISTRY = {
  "zoo-gateway": {
    label: "Zoo Gateway",
    docsSlug: "zoo-gateway",
    modelConfig: { field: "zooGatewayModelId", default: zooGatewayDefaultModelId },
  },
  ...
} satisfies Partial<Record<ProviderName, {
  label: string
  docsSlug?: string
  modelConfig?: { field: keyof ProviderSettings; default?: string }
}>>
```

taltas (changes requested):
One other thing I noticed

webview-ui/src/i18n/locales/en/settings.json
```diff
@@ -553,6 +553,13 @@
 				"learnMore": "Learn more about provider routing"
 			}
 		},
+		"zooGateway": {
+			"account": "Zoo Code Account",
+			"signInButton": "Sign in to Zoo Code",
+			"signInDescription": "Sign in to use Zoo Gateway with your account",
```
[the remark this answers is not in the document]
JamesRobert20: Good catch — fixed in 420d35b06. I dropped the quoted CTA from `zooGatewaySignIn` across all 18 locales rather than syncing two strings, since the sign-in button is rendered immediately under the inline error so a duplicate label was just a drift hazard.

webview-ui/src/i18n/locales/zh-CN/settings.json
```diff
@@ -560,6 +560,13 @@
 			"placeholder": "默认：claude",
 			"maxTokensLabel": "最大输出 Token",
 			"maxTokensDescription": "Claude Code 响应的最大输出 Token 数量。默认为 8000。"
+		},
+		"zooGateway": {
+			"account": "Zoo Code 账户",
+			"signInButton": "登录 Zoo Code",
+			"signInDescription": "登录以使用您的账户访问 Zoo Gateway",
```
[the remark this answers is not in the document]
JamesRobert20: Fixed in 420d35b06 — same change applied across all 18 locales (CTA dropped, button label is the only sign-in copy now).

webview-ui/src/i18n/locales/zh-TW/settings.json
```diff
@@ -575,6 +575,13 @@
 			"placeholder": "預設：claude",
 			"maxTokensLabel": "最大輸出 Token",
 			"maxTokensDescription": "Claude Code 回應的最大輸出 Token 數量。預設為 8000。"
+		},
+		"zooGateway": {
+			"account": "Zoo Code 帳戶",
+			"signInButton": "登入 Zoo Code",
+			"signInDescription": "登入以使用您的帳戶存取 Zoo Gateway",
```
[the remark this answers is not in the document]
JamesRobert20: Fixed in 420d35b06 — same change applied across all 18 locales (CTA dropped, button label is the only sign-in copy now).

webview-ui/src/utils/validate.ts
```diff
@@ -282,12 +291,20 @@ export function getModelValidationError(
...
+ * does not need to track Zoo Code auth state.
  */
 export function validateApiConfigurationExcludingModelErrors(
 	apiConfiguration: ProviderSettings,
 	_routerModels?: RouterModels, // Keeping this for compatibility with the old function.
 	organizationAllowList?: OrganizationAllowList,
 ): string | undefined {
+	if (apiConfiguration.apiProvider === "zoo-gateway") {
+		return undefined
+	}
```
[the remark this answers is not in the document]
JamesRobert20: Right, the early return was too wide. Fixed in 420d35b06 by inverting the condition so only the keys/sign-in check is skipped for `zoo-gateway`; the org allowlist check now runs for every provider, and a `PROVIDER_NOT_ALLOWED` error will surface for `zoo-gateway` like everything else. Added a regression test asserting that path.

webview-ui/src/i18n/locales/fr/settings.json
```diff
@@ -901,7 +908,8 @@
 		"providerNotAllowed": "Le fournisseur '{{provider}}' n'est pas autorisé par votre organisation",
 		"modelNotAllowed": "Le modèle '{{model}}' n'est pas autorisé pour le fournisseur '{{provider}}' par votre organisation",
 		"profileInvalid": "Ce profil contient un fournisseur ou un modèle qui n'est pas autorisé par votre organisation",
-		"qwenCodeOauthPath": "Tu dois fournir un chemin valide pour les identifiants OAuth"
+		"qwenCodeOauthPath": "Tu dois fournir un chemin valide pour les identifiants OAuth",
+		"zooGatewaySignIn": "Connecte-toi à Zoo Code pour utiliser Zoo Gateway."
```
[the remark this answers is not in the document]
JamesRobert20: Good catch — fixed in 5aa42cc88. `validation.zooGatewaySignIn` now uses "Connectez-vous" to match `providers.zooGateway.signInDescription`.

taltas (approved):

PR #347, opened 2026-05-27
feat(zoo-gateway): auth callback and multi-profile token sync

JamesRobert20:
## Summary
- Propagates OAuth callback token to all ClineProvider instances (sequential writes to avoid profile-store races)
- Seeds/syncs zoo-gateway profiles, including non-active profiles for model fetch
- Clears in-memory active profile on sign-out even when disk is already clean

Part 3 of the Zoo Gateway stack. **Depends on PR2 (stacked on #344).**

## Test plan
- [ ] handleUri multi-instance + serialization tests
- [ ] webviewMessageHandler separate-profile lookup test
- [ ] Manual: sign in/out across multiple profiles

## Summary by CodeRabbit

* **New Features**
  * Zoo Code auth callbacks now propagate sequentially to all active provider instances.
  * Zoo Gateway profiles are auto-seeded/updated when cached auth tokens exist but profiles are missing or stale.
  * Sign-out now clears Zoo session tokens across all Zoo Gateway profiles.

* **Error Handling**
  * Centralized Zoo Gateway error classification and user-facing surfaces for auth, session, credit, and account issues.

* **Internationalization**
  * Added localized error messages and action buttons across 16 locales.

### DIFF
--- a/src/activate/handleUri.ts
+++ b/src/activate/handleUri.ts
@@ -63,7 +63,14 @@ export const handleUri = async (uri: vscode.Uri) => {
 					// is cheap (at most a handful of instances) and avoids the race.
 					const allInstances = ClineProvider.getAllInstances()
 					for (const instance of allInstances) {
-						await instance.handleZooCodeCallback(token)
+						try {
+							await instance.handleZooCodeCallback(token)
+						} catch (error) {
+							console.error(
+								"Failed to persist Zoo Gateway token for a provider instance:",
+								error instanceof Error ? error.message : error,
+							)
+						}
 					}
 				}
 			}
--- a/src/activate/handleUri.ts
+++ b/src/activate/handleUri.ts
@@ -50,9 +50,20 @@ export const handleUri = async (uri: vscode.Uri) => {
 						email,
 						image,
 					})
-					// Refresh webview state if a panel is currently open
-					if (visibleProvider) {
-						await visibleProvider.handleZooCodeCallback(token)
+					// Write the token to all active provider instances regardless of visibility.
+					// The profile settings write (handleZooCodeCallback) must run on any active
+					// instance — not just the visible one — so the zoo-gateway zooSessionToken
+					// is persisted even when the sidebar/panel is hidden at callback time.
+					//
+					// Run sequentially (NOT Promise.all): each ClineProvider's
+					// handleZooCodeCallback does a read-modify-write on the same backing
+					// provider settings store (listConfig → getProfile → saveConfig /
+					// upsertProviderProfile). Fanning out concurrently across N instances
+					// can interleave reads/writes and clobber updates. Serialization here
+					// is cheap (at most a handful of instances) and avoids the race.
+					const allInstances = ClineProvider.getAllInstances()
+					for (const instance of allInstances) {
+						await instance.handleZooCodeCallback(token)
 					}
 				}
 			}
--- a/src/activate/handleUri.ts
+++ b/src/activate/handleUri.ts
@@ -4,6 +4,34 @@ import { getRouterUnavailableSignInMessage } from "../core/config/routerRemoval"
 import { ClineProvider } from "../core/webview/ClineProvider"
 import { handleAuthCallback as handleZooCodeAuthCallback, setZooCodeUserInfo } from "../services/zoo-code-auth"
 
+/**
+ * Persist the Zoo Code session token to every active provider instance.
+ *
+ * The profile settings write (handleZooCodeCallback) must run on any active
+ * instance — not just the visible one — so the zoo-gateway zooSessionToken is
+ * persisted even when the sidebar/panel is hidden at callback time.
+ *
+ * Run sequentially (NOT Promise.all): each ClineProvider's handleZooCodeCallback
+ * does a read-modify-write on the same backing provider settings store
+ * (listConfig → getProfile → saveConfig / upsertProviderProfile). Fanning out
+ * concurrently across N instances can interleave reads/writes and clobber
+ * updates. Serialization is cheap (at most a handful of instances) and avoids
+ * the race.
+ */
+async function propagateZooGatewayCallback(token: string): Promise<void> {
+	const allInstances = ClineProvider.getAllInstances()
+	for (const instance of allInstances) {
+		try {
+			await instance.handleZooCodeCallback(token)
+		} catch (error) {
+			console.error(
+				"Failed to persist Zoo Gateway token for a provider instance:",
+				error instanceof Error ? error.message : error,
+			)
+		}
+	}
+}
+
 export const handleUri = async (uri: vscode.Uri) => {
 	const path = uri.path
 	const query = new URLSearchParams(uri.query.replace(/\+/g, "%2B"))
@@ -50,28 +78,7 @@ export const handleUri = async (uri: vscode.Uri) => {
 						email,
 						image,
 					})
-					// Write the token to all active provider instances regardless of visibility.
-					// The profile settings write (handleZooCodeCallback) must run on any active
-					// instance — not just the visible one — so the zoo-gateway zooSessionToken
-					// is persisted even when the sidebar/panel is hidden at callback time.
-					//
-					// Run sequentially (NOT Promise.all): each ClineProvider's
-					// handleZooCodeCallback does a read-modify-write on the same backing
-					// provider settings store (listConfig → getProfile → saveConfig /
-					// upsertProviderProfile). Fanning out concurrently across N instances
-					// can interleave reads/writes and clobber updates. Serialization here
-					// is cheap (at most a handful of instances) and avoids the race.
-					const allInstances = ClineProvider.getAllInstances()
-					for (const instance of allInstances) {
-						try {
-							await instance.handleZooCodeCallback(token)
-						} catch (error) {
-							console.error(
-								"Failed to persist Zoo Gateway token for a provider instance:",
-								error instanceof Error ? error.message : error,
-							)
-						}
-					}
+					await propagateZooGatewayCallback(token)
 				}
 			}
 			break
--- a/src/api/providers/zoo-gateway.ts
+++ b/src/api/providers/zoo-gateway.ts
@@ -1,3 +1,4 @@
+import * as vscode from "vscode"
 import { Anthropic } from "@anthropic-ai/sdk"
 import OpenAI from "openai"
 
@@ -9,8 +10,9 @@ import {
 } from "@roo-code/types"
 
 import { ApiHandlerOptions } from "../../shared/api"
-import { getCachedZooCodeToken, getZooCodeBaseUrl } from "../../services/zoo-code-auth"
+import { clearZooCodeToken, getCachedZooCodeToken, getZooCodeBaseUrl } from "../../services/zoo-code-auth"
 import { Package } from "../../shared/package"
+import { t } from "../../i18n"
 
 import { ApiStream } from "../transform/stream"
 import { convertToOpenAiMessages } from "../transform/openai-format"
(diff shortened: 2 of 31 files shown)


proyectoauraorg (approved):
## Review: Auth Callback + Multi-Profile Token Sync

**Verdict**: ✅ Approved (with security notes for future hardening)

### Verified
- ✅ `handleUri.spec.ts` — 8 tests pass (multi-instance fan-out, sequential serialization)
- ✅ `ClineProvider.spec.ts` — 82 tests pass (6 new auth profile sync tests)
- ✅ `webviewMessageHandler.spec.ts` — 44 tests pass (2 new sign-out tests)
- ✅ Sequential execution of callbacks avoids profile-store race conditions
- ✅ Sign-out clears tokens from ALL zoo-gateway profiles (correct symmetry)
- ✅ Model list fetch recovers credentials from non-active profiles

### Security Notes (future hardening, not blockers)
1. **Token storage**: `zooSessionToken` is stored in JSON-serialized `ProviderSettings` on disk. VS Code's globalStorage is OS-protected, but migrating to `SecretStorage` would add an extra layer of defense.
2. **Base URL validation**: `zooGatewayBaseUrl` is derived from `getZooCodeBaseUrl()`. In production this is fine, but a misconfigured `ZOO_CODE_BASE_URL` env var could route tokens to an unintended host. Consider adding domain validation.

Both are hardening improvements for a future iteration, not blockers for this PR.

### Dependency
Depends on #345. After #345 merges, this should rebase cleanly.

src/core/webview/webviewMessageHandler.ts
```diff
@@ -946,6 +946,31 @@ export const webviewMessageHandler = async (
...
+						if (fullProfile.zooSessionToken) {
+							zooGatewayToken = fullProfile.zooSessionToken
+							zooGatewayBaseUrl = fullProfile.zooGatewayBaseUrl ?? zooGatewayBaseUrl
+							break
+						}
+					}
+				} catch (error) {
+					console.debug("Failed to look up zoo-gateway profile for model fetch:", error)
+				}
+			}
```
edelauna: Since `getZooGatewayModels` handles both fallbacks internally — `getCachedZooCodeToken()` for the token (line 68) and `getZooCodeBaseUrl()` for the base URL (line 64) — is this scan doing work the fetcher is not already covering?

The bridging step is in `modelCache.ts:101`: `options.apiKey` gets translated to `zooSessionToken` before calling the fetcher, so passing nothing in the candidate still routes correctly. This would make zoo-gateway consistent with `openrouter` and `vercel-ai-gateway`, which both pass no credentials in the candidates array and let their fetchers self-resolve.

src/core/webview/webviewMessageHandler.ts
```diff
@@ -965,6 +990,14 @@ export const webviewMessageHandler = async (
...
 				},
 				{ key: "vercel-ai-gateway", options: { provider: "vercel-ai-gateway" } },
+				{
+					key: "zoo-gateway",
+					options: {
+						provider: "zoo-gateway",
+						apiKey: zooGatewayToken,
+						baseUrl: zooGatewayBaseUrl,
+					},
+				},
```
edelauna, line 974: If the scan above is removed, could this simplify to match `openrouter`/`vercel-ai-gateway`?

```suggestion
				{ key: "zoo-gateway", options: { provider: "zoo-gateway" } },
```

src/api/providers/zoo-gateway.ts
```diff
@@ -29,10 +99,9 @@ export class ZooGatewayHandler extends RouterProvider implements SingleCompletio
 	constructor(options: ApiHandlerOptions) {
 		const baseURL = options.zooGatewayBaseUrl ?? `${getZooCodeBaseUrl()}/api/gateway/v1`
 
-		// Prefer the profile-persisted token; fall back to the secret-storage cache so
-		// requests work when the user is signed in but the profile hasn't been seeded yet
-		// (e.g. auth callback arrived before any webview instance was open).
-		const sessionToken = options.zooSessionToken || getCachedZooCodeToken()
+		// Prefer the secret-storage cache so a 401 clear takes effect immediately; fall back
+		// to the profile-persisted token when the user is signed in but seeding hasn't run yet.
+		const sessionToken = getCachedZooCodeToken() || options.zooSessionToken
```
edelauna: After `disconnectZooCode()` clears the cached token, `getCachedZooCodeToken()` returns `""` (falsy), so this falls back to `options.zooSessionToken` — the value baked in at handler construction time. A running task on a non-initiating provider instance could stay authenticated after sign-out via this path.

`zoo-code-auth.ts:53` already handles this correctly for the module-level cache via `onDidChange` — the same pattern `McpHub` uses for OAuth token invalidation. The gap is specifically this `||` fallback: when `getCachedZooCodeToken()` returns `""` due to an explicit sign-out (not an absent init), the fallback to the profile token bypasses that signal.

In the spirit of the pattern from #344 — deferring auth errors to the runtime rather than setup — should an empty cached token be treated as a hard "not authenticated" rather than "fall through to the profile"?

edelauna (dismissed):

src/core/webview/webviewMessageHandler.ts
JamesRobert20: Agreed — this scan was already removed in a703ac5c7 ("sign-out clears stale profile tokens, simplify model fetch"). The zoo-gateway candidate is now `{ provider: 'zoo-gateway', baseUrl: apiConfiguration.zooGatewayBaseUrl }`, matching openrouter and vercel-ai-gateway, with the fetcher self-resolving its credentials.

src/core/webview/webviewMessageHandler.ts
JamesRobert20: Same — already simplified in a703ac5c7.

src/api/providers/zoo-gateway.ts
JamesRobert20: Good catch — addressed in a703ac5c7. The handler now uses `resolveZooGatewaySessionToken(options.zooSessionToken)` instead of `getCachedZooCodeToken() || options.zooSessionToken`. `resolveZooGatewaySessionToken` is the helper added in `zoo-code-auth.ts` that respects the `_sessionCleared` flag set by `clearZooCodeToken()`/`disconnectZooCode()`, so after sign-out a profile-cached token is no longer used as a fallback.

src/api/providers/zoo-gateway.ts
```diff
@@ -19,6 +21,74 @@ import { addCacheBreakpoints } from "../transform/caching/vercel-ai-gateway"
...
+		const action = await vscode.window.showErrorMessage(
+			t("common:zooAuth.errors.account_unavailable"),
+			t("common:zooAuth.buttons.contact_support"),
+		)
+		if (action) {
+			void vscode.env.openExternal(vscode.Uri.parse(`${getZooCodeBaseUrl()}/support`))
+		}
+		return
+	}
+}
```
taltas: We should split this function out into transport behaviour and VS Code UX side effects:



```suggestion

type ZooGatewayApiErrorAction =
	| { kind: "sign_in" }
	| { kind: "add_credits" }
	| { kind: "contact_support" }
	| { kind: "none" }

function classifyGatewayApiError(error: unknown): ZooGatewayApiErrorAction {
	const status = getApiErrorStatus(error)
	const code = getApiErrorCode(error)
	...
}
				
				
async function surfaceGatewayApiError(error: unknown): Promise<void> {
	switch (classifyGatewayApiError(error).kind) {
		case "sign_in":
			...
			return
		case "add_credits":
			...
			return
		case "contact_support":
			...
			return
		default:
			return
	}
}
```

This will make it easier to test and follow the logic.

src/activate/handleUri.ts
```diff
@@ -50,9 +50,27 @@ export const handleUri = async (uri: vscode.Uri) => {
...
+					const allInstances = ClineProvider.getAllInstances()
+					for (const instance of allInstances) {
+						try {
+							await instance.handleZooCodeCallback(token)
+						} catch (error) {
+							console.error(
+								"Failed to persist Zoo Gateway token for a provider instance:",
+								error instanceof Error ? error.message : error,
+							)
+						}
```
taltas: This is doing more than just routing now. We should split this into a helper function to keep it DRY.





```suggestion
async function propagateZooGatewayCallback(token: string) {
	const allInstances = ClineProvider.getAllInstances()
	for (const instance of allInstances) {
		try {
			await instance.handleZooCodeCallback(token)
		} catch (error) {
			console.error(
				"Failed to persist Zoo Gateway token for a provider instance:",
				error instanceof Error ? error.message : error,
			)
		}
	}
}

....

await propagateZooGatewayCallback(token)
```

taltas (changes requested):
Couple of small code cleaness tweaks.

src/api/providers/zoo-gateway.ts
JamesRobert20: Done in 06fc185aa. Split into a pure `classifyGatewayApiError(error) -> ZooGatewayApiErrorAction` and a thin `surfaceGatewayApiError` that switches on the action for the VS Code side effects. The classifier is exported and now has focused unit tests covering 401/402/429-budget/429-non-budget/403/no-status, so the mapping is verified without the notification mocks.

src/activate/handleUri.ts
JamesRobert20: Done in 06fc185aa. Extracted the fan-out into `propagateZooGatewayCallback(token)`, so the `/auth-callback` case just calls that after `setZooCodeUserInfo`. The sequential read-modify-write and per-instance error isolation (plus the rationale comment) moved into the helper unchanged.

src/core/webview/webviewMessageHandler.ts
```diff
@@ -2438,6 +2445,47 @@ export const webviewMessageHandler = async (
...
+								provider.log(`[zooCodeSignOut] Cleared zooSessionToken from "${entry.name}" profile`)
+							}
+						}
+					}
+				} catch (profileError) {
+					// Log but don't fail the sign-out if profile cleanup fails
+					provider.log(
+						`[zooCodeSignOut] Failed to clear profile token: ${profileError instanceof Error ? profileError.message : String(profileError)}`,
+					)
+				}
```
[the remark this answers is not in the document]
JamesRobert20: Done in bc30b7bf3. Wrapped the per-profile work in its own try/catch inside the loop so a corrupted profile or a failed write only logs and skips that entry instead of aborting cleanup for the rest. The outer try/catch now only guards the initial listConfig call.

src/core/webview/__tests__/webviewMessageHandler.spec.ts
```diff
@@ -365,11 +368,11 @@ describe("webviewMessageHandler - requestRouterModels", () => {
 				openrouter: mockModels,
 				requesty: mockModels,
 				unbound: mockModels,
+				"vercel-ai-gateway": mockModels,
+				"zoo-gateway": mockModels,
```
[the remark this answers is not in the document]
JamesRobert20: Skipping. The premise does not hold against the current code: getZooGatewayModels self-resolves the base URL (options?.zooGatewayBaseUrl ?? getZooCodeBaseUrl()/api/gateway/v1), so zooGatewayBaseUrl is an optional override, not a required field. The candidate in requestRouterModels passes it through as-is and the mocked getModels returns mockModels for every provider regardless, so these assertions match the production contract. There is no missing-base-URL state to regress here.

src/core/webview/__tests__/ClineProvider.spec.ts
```diff
@@ -3663,4 +3672,175 @@ describe("ClineProvider - Comprehensive Edit/Delete Edge Cases", () => {
...
+				;(provider as any).providerSettingsManager = {
+					listConfig: vi.fn().mockResolvedValue([]),
+				}
+
+				await provider.handleZooCodeCallback("zoo_ext_token")
+
+				expect(mockOutputChannel.appendLine).toHaveBeenCalledWith(
+					expect.stringContaining("[handleZooCodeCallback] Failed to save zoo-gateway profile"),
+				)
+			})
```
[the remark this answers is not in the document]
JamesRobert20: Done in bc30b7bf3. Added expect(provider.postStateToWebview).toHaveBeenCalled() after the callback. postStateToWebview runs outside the try/catch in handleZooCodeCallback, so the failure path is now asserted.

taltas (approved):
Changes addressed

PR #379, opened 2026-05-28
feat(mcp-oauth): support static clientId for servers without DCR

proyectoauraorg:
### Related GitHub Issue

Aligned with VS Code 1.122 feature [#257415](https://github.com/microsoft/vscode/issues/257415) — MCP OAuth with custom clientId.

### Description

Adds optional `oauth.clientId` field to MCP server configuration. When provided, the OAuth provider uses this clientId directly instead of performing Dynamic Client Registration (DCR). This enables connections to OAuth-protected MCP servers that don't support RFC 7591 DCR.

**Use case:** Many corporate and self-hosted MCP servers require a pre-registered clientId but don't support DCR. Currently, Zoo Code cannot connect to these servers via OAuth.

**Changes:**
- `BaseConfigSchema`: add `oauth.clientId` optional field
- `McpOAuthClientProvider`: accept `clientId` in `create()` options, use it in `registerClientIfNeeded()` to skip DCR
- `McpHub`: pass `oauth.clientId` from config to the provider
- Tests: 2 new tests covering static clientId and precedence over cached DCR data

**Example configuration in `.roo/mcp.json`:**
```json
{
  "mcpServers": {
    "corporate-server": {
      "type": "streamable-http",
      "url": "https://mcp.corporate.example.com",
      "oauth": {
        "clientId": "my-app-id-12345"
      }
    }
  }
}
```

### Test Procedure

**Unit tests** (104 passed):
```
cd src && npx vitest run services/mcp/__tests__/McpOAuthClientProvider.spec.ts services/mcp/__tests__/McpHub.spec.ts
```

- `McpOAuthClientProvider.spec.ts` — 44 tests (42 existing + 2 new for static clientId)
- `McpHub.spec.ts` — 60 tests (all existing, schema change verified)

New tests:
1. `should use static clientId instead of performing DCR` — verifies clientId is used directly
2. `should use static clientId even when cached data exists` — verifies static takes precedence over cached DCR

### Pre-Submission Checklist

- [x] **Issue Linked**: Aligned with VS Code 1.122 feature.
- [x] **Scope**: Focused on static clientId support only.
- [x] **Self-Review**: Thorough review performed.
- [x] **Testing**: 2 new tests added, 104/104 pass.
- [x] **Documentation Impact**: No doc updates needed (self-documenting config field).

### Documentation Updates

## Summary by CodeRabbit

* **New Features**
  * OAuth server configuration now supports an optional client identifier field, providing additional flexibility for authentication setup.

* **Tests**
  * Added test coverage for the new OAuth client identifier feature, verifying proper handling across various configuration scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/379?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


edelauna:
Wouldn't this also require a secret?

This would also require the whole normal OAuth 2.0 flow to be available when provided. This is larger than a 3 file change.

taltas:
Thanks for the contribution. There is also the huge readme file that was committed like your other PRs, we should remove it.

proyectoauraorg:
Done — removed the stray `tmp/README.md` (280832c2b). It was an accidental local file that leaked into the branch. Thanks for flagging it.

proyectoauraorg:
Thanks for the careful review.

This change reuses the existing authorization-code + PKCE flow already implemented in `McpOAuthClientProvider`. The new behavior is limited to allowing a preconfigured `clientId` from `mcp.json` for servers that don't support Dynamic Client Registration (DCR), which is why the implementation stays relatively small.

A client secret is not always required here. The provider already negotiates `token_endpoint_auth_method`, using PKCE-only public clients (`none`) when supported, and `client_secret_post` for confidential clients when advertised by the server. So a static `clientId` without a secret is valid for public-client flows per RFC 8252 / MCP auth expectations.

That said, you're correct that confidential-client setups requiring `client_secret_post` would need additional configuration beyond a `clientId`. The token exchange path already supports sending a secret; the current config surface just doesn't expose one yet. If support for non-DCR confidential clients is desired, I can follow up with an optional `clientSecret` config field.

edelauna (changes requested):
@proyectoauraorg can you attach a screen-recording of this in action?

I suspect this change requires a bit more work to truly implement. Alternatively why can't this information just be sent over headers? ClientID and Secret are fixed and known ahead of time.

PR #316, opened 2026-05-24
feat(bedrock): full support for Claude Opus 4.7 (registry + API compat layer)

vandre-sales:
## Summary

Adds full support for `anthropic.claude-opus-4-7` (and future Sonnet 4.7) on AWS Bedrock — both **model registry** and **API compatibility layer**.

This PR contains 2 commits:

1. **`feat(bedrock): add anthropic.claude-opus-4-7 to native model registry`** — Adds the model entry to `bedrockModels` with full ModelInfo configuration
2. **`feat(bedrock): support Claude 4.7+ adaptive thinking and remove temperature`** — Updates the provider logic to handle the breaking API changes introduced by Claude 4.7+

## Problem

When using `anthropic.claude-opus-4-7` as a custom model in Zoo Code with AWS Bedrock provider, **two layers of issues** exist:

### Layer 1: Missing model registry entry

The extension falls back to `guessModelInfoFromId()` which:
- Does NOT set `supportsReasoningBudget: true` → thinking budget disabled in UI
- Does NOT populate `cachableFields` → prompt cache inactive
- Results in "too many tokens" errors during parallel file injection (no cache = tokens accumulate without reuse)

### Layer 2: Breaking API changes in Claude 4.7

Even with the model registered, the Bedrock API rejects requests with errors:
- ❌ `temperature` is deprecated for this model (4.7+ removed sampling parameters entirely)
- ❌ `thinking.type.enabled` is not supported (4.7+ requires `thinking.type.adaptive`)

These cause **400 errors** even after the registry fix.

### Commit 1 — Model Registry (`packages/types/src/providers/bedrock.ts`)

Added complete model entry with:
- ✅ `supportsReasoningBudget: true` (enables thinking budget fields in UI)
- ✅ `cachableFields: ["system", "messages", "tools"]` (enables multi-point prompt caching)
- ✅ `supportsPromptCache: true` with `maxCachePoints: 4`
- ✅ `supportsImages: true`
- ✅ 1M context tier pricing (via `tiers[]`)
- ✅ Added to `BEDROCK_1M_CONTEXT_MODEL_IDS`
- ✅ Added to `BEDROCK_GLOBAL_INFERENCE_MODEL_IDS`

### Commit 2 — Provider Logic (`src/api/providers/bedrock.ts`)

Detects Claude 4.7+ models and adapts the request payload:

```typescript
const isGen47Model = baseModelId.includes("opus-4-7") || baseModelId.includes("sonnet-4-7")

// Conditional thinking format:
[...]

PR #386, opened 2026-05-28
feat: add Claude Opus 4.8 support across Anthropic, Bedrock, and Vertex providers

vandre-sales:
## Summary

Adds full support for **Claude Opus 4.8** (`claude-opus-4-8` / `anthropic.claude-opus-4-8`) across the Anthropic, AWS Bedrock, and Google Vertex providers. Opus 4.8 was released by Anthropic on **May 28, 2026** as the successor to Opus 4.7.

This builds directly on the adaptive-thinking compatibility layer introduced for Opus 4.7 in #316. Per Anthropic's official migration guide, **there are no breaking API changes between 4.7 and 4.8** — 4.8 inherits the same adaptive-thinking contract, the same removal of sampling parameters, the same native 1M context window, and the same prompt-cache behaviour. The change is therefore small and surgical.

> **Note on #316:** This PR is a superset of #316 — because Opus 4.7 is not yet merged into `main`, the diff necessarily includes the 4.7 work it depends on (the `isGen47Model` detection branch). If #316 merges first, this PR will rebase down to the 4.8-only delta. Maintainers may prefer to review/merge this PR in place of #316, or merge #316 first and then this. Happy to adjust to whichever workflow you prefer.

### Model registry (`packages/types`)

- **`anthropic.ts`** — registered `claude-opus-4-8`: 1M context, 128k max output, `supportsReasoningBudget`, `supportsReasoningBinary`, `supportsTemperature: false` (mirrors 4.7).
- **`bedrock.ts`** — registered `anthropic.claude-opus-4-8` with prompt-cache points, `cachableFields`, and 1M tiered pricing; added it to `BEDROCK_1M_CONTEXT_MODEL_IDS` and `BEDROCK_GLOBAL_INFERENCE_MODEL_IDS`.
- **`vertex.ts`** — registered `claude-opus-4-8` with the same shape; added it to `VERTEX_1M_CONTEXT_MODEL_IDS`.

### Provider logic (`src/api/providers`)

- **`anthropic.ts`** — added `claude-opus-4-8` to both prompt-caching `switch` statements so it gets the same handling as 4.7 (native 1M context, no `context-1m-2025-08-07` beta header required).
- **`bedrock.ts`** — renamed `isGen47Model` → `isAdaptiveThinkingModel` and expanded the pattern to match `opus-4-7`, `opus-4-8`, `sonnet-4-7`, and `sonnet-4-8`. The rename describes the actual capability (adaptive thinking) rather than a single generation, making future Claude versions easier to plug in. The branch still emits `thinking: { type: "adaptive", display: "summarized" }` + `output_config: { effort: "xhigh" }` and omits `temperature`.
- **`openai.ts`** — comment-only update mentioning 4.8; no logic change (already honors the `supportsTemperature: false` flag).

### Tests

- **`anthropic.spec.ts`** — 5 new cases mirroring 4.7 (1M-beta-header guard, adaptive thinking ON/OFF, custom `maxTokens`, `getModel` info).
- **`anthropic-vertex.spec.ts`** — 1M context tier pricing for Vertex Opus 4.8.
- **`shared/api.spec.ts`** — `getModelMaxOutputTokens` hybrid-token handling on 4.8.
- **`bedrock.spec.ts`** — new `Claude 4.7+ adaptive thinking (Opus 4.7 / Opus 4.8)` describe block with 5 cases covering:
  - adaptive thinking + `xhigh` effort for 4.7
  - adaptive thinking + `xhigh` effort for 4.8
  - thinking + temperature omitted when reasoning is disabled
  - **regression guard**: 4.6 still uses `budget_tokens` + `temperature`
  - **cross-region prefix** detection (`us.anthropic.claude-opus-4-8`)

This also closes the unit-test coverage gap that existed for the Bedrock adaptive-thinking branch in #316.

## Fast Mode — out of scope

Opus 4.8 introduces a "fast mode" (2.5x faster, $10/$50 per 1M tokens). Per Anthropic docs it is **not available on the Claude Platform on AWS / Bedrock** and direct-API access is gated behind a waitlist, so it is intentionally excluded from this PR.

## Other 4.8 features (no client change needed)

The following 4.8 additions are server-side optimizations that the existing client pipeline benefits from transparently — no code change required:

- Mid-conversation `role: "system"` messages
- `refusal` content block + documented `stop_details`
- Lower prompt-cache minimum (1,024 tokens)

## Backward compatibility

✅ 100% backward compatible. The adaptive-thinking branch only activates for `opus-4-7`, `opus-4-8`, `sonnet-4-7`, and `sonnet-4-8`. Claude 4.6 and earlier continue to use the `budget_tokens` thinking format and still send `temperature` (covered by a new regression test).

## Pricing note

⚠️ Pricing for 4.8 is set to match 4.7 ($5/$25 per 1M tokens regular tier, $10/$37.5 for the 1M tier). Please verify against the [AWS Bedrock pricing page](https://aws.amazon.com/bedrock/pricing/) before merging.

## Testing

- ✅ **235 unit tests pass**, 0 failures
- ✅ Type-check clean on both `packages/types` and `src`
- ✅ **Validated live end-to-end**: built the VSIX, installed it, and ran a real Bedrock conversation using **`global.anthropic.claude-opus-4-8`** (Global Inference Profile). Adaptive thinking active with `xhigh` effort, no 400 errors on `temperature` or `thinking.type.enabled`.
- ✅ Regression-verified that `global.anthropic.claude-opus-4-7` still works after the `isGen47Model` → `isAdaptiveThinkingModel` rename.

## References

- [Claude Opus 4.8 — What's new](https://platform.claude.com/docs/en/about-claude/models/whats-new-claude-4-8)
- [Migration guide: 4.7 → 4.8](https://platform.claude.com/docs/en/about-claude/models/migration-guide#migrating-from-claude-opus-47)
- [Claude on Amazon Bedrock](https://platform.claude.com/docs/en/build-with-claude/claude-in-amazon-bedrock)
- Builds on #316 (Claude Opus 4.7 support)

## Summary by CodeRabbit

* **New Features**
  * Added Claude Opus 4.8 (Anthropic/Bedrock/Vertex) and Claude Opus 4.7 to Bedrock; expanded model registries, 1M-context support, prompt-caching, and reasoning capability flags.
  * Implemented adaptive-thinking handling for Opus 4.7+ and updated behavior to omit temperature where required.

* **Tests**
  * Added/extended tests for Opus 4.7/4.8: context windows, pricing, prompt-caching, adaptive thinking, reasoning modes, and E2E Bedrock smoke.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/386?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/types/src/providers/bedrock.ts
+++ b/packages/types/src/providers/bedrock.ts
@@ -191,6 +191,32 @@ export const bedrockModels = {
 			},
 		],
 	},
+	"anthropic.claude-opus-4-8": {
+		maxTokens: 8192,
+		contextWindow: 200_000, // Default 200K, extendable to 1M with beta flag 'context-1m-2025-08-07'
+		supportsImages: true,
+		supportsPromptCache: true,
+		supportsReasoningBudget: true,
+		inputPrice: 5.0, // $5 per million input tokens (≤200K context) — verify against Bedrock console
+		outputPrice: 25.0, // $25 per million output tokens (≤200K context) — verify against Bedrock console
+		cacheWritesPrice: 6.25, // $6.25 per million tokens
+		cacheReadsPrice: 0.5, // $0.50 per million tokens
+		minTokensPerCachePoint: 1024,
+		maxCachePoints: 4,
+		cachableFields: ["system", "messages", "tools"],
+		// Tiered pricing for extended context (requires beta flag 'context-1m-2025-08-07')
+		// 4.8 inherits the same Bedrock pricing structure as 4.7 — no API breaking changes.
+		// Adaptive thinking is the only supported reasoning mode (same as 4.7).
+		tiers: [
+			{
+				contextWindow: 1_000_000, // 1M tokens with beta flag
+				inputPrice: 10.0, // $10 per million input tokens (>200K context)
+				outputPrice: 37.5, // $37.50 per million output tokens (>200K context)
+				cacheWritesPrice: 12.5, // $12.50 per million tokens (>200K context)
+				cacheReadsPrice: 1.0, // $1.00 per million tokens (>200K context)
+			},
+		],
+	},
 	"anthropic.claude-opus-4-5-20251101-v1:0": {
 		maxTokens: 8192,
 		contextWindow: 200_000,
@@ -550,6 +576,7 @@ export const BEDROCK_1M_CONTEXT_MODEL_IDS = [
 	"anthropic.claude-sonnet-4-6",
 	"anthropic.claude-opus-4-6-v1",
 	"anthropic.claude-opus-4-7",
+	"anthropic.claude-opus-4-8",
 ] as const
 
 // Amazon Bedrock models that support Global Inference profiles
@@ -569,6 +596,7 @@ export const BEDROCK_GLOBAL_INFERENCE_MODEL_IDS = [
 	"anthropic.claude-opus-4-5-20251101-v1:0",
 	"anthropic.claude-opus-4-6-v1",
 	"anthropic.claude-opus-4-7",
+	"anthropic.claude-opus-4-8",
 ] as const
 
 // Amazon Bedrock Service Tier types
--- a/packages/types/src/providers/bedrock.ts
+++ b/packages/types/src/providers/bedrock.ts
@@ -167,6 +167,30 @@ export const bedrockModels = {
 			},
 		],
 	},
+	"anthropic.claude-opus-4-7": {
+		maxTokens: 8192,
+		contextWindow: 200_000, // Default 200K, extendable to 1M with beta flag 'context-1m-2025-08-07'
+		supportsImages: true,
+		supportsPromptCache: true,
+		supportsReasoningBudget: true,
+		inputPrice: 5.0, // $5 per million input tokens (≤200K context) — verify against Bedrock console
+		outputPrice: 25.0, // $25 per million output tokens (≤200K context) — verify against Bedrock console
+		cacheWritesPrice: 6.25, // $6.25 per million tokens
+		cacheReadsPrice: 0.5, // $0.50 per million tokens
+		minTokensPerCachePoint: 1024,
+		maxCachePoints: 4,
+		cachableFields: ["system", "messages", "tools"],
+		// Tiered pricing for extended context (requires beta flag 'context-1m-2025-08-07')
+		tiers: [
+			{
+				contextWindow: 1_000_000, // 1M tokens with beta flag
+				inputPrice: 10.0, // $10 per million input tokens (>200K context)
+				outputPrice: 37.5, // $37.50 per million output tokens (>200K context)
+				cacheWritesPrice: 12.5, // $12.50 per million tokens (>200K context)
+				cacheReadsPrice: 1.0, // $1.00 per million tokens (>200K context)
+			},
+		],
+	},
 	"anthropic.claude-opus-4-5-20251101-v1:0": {
 		maxTokens: 8192,
 		contextWindow: 200_000,
@@ -525,6 +549,7 @@ export const BEDROCK_1M_CONTEXT_MODEL_IDS = [
 	"anthropic.claude-sonnet-4-5-20250929-v1:0",
 	"anthropic.claude-sonnet-4-6",
 	"anthropic.claude-opus-4-6-v1",
+	"anthropic.claude-opus-4-7",
 ] as const
 
 // Amazon Bedrock models that support Global Inference profiles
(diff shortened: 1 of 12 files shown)


edelauna:
Nice! Thank you for your contribution. I recently added bedrock to our e2e test harness would you be able to add a smoke test there for 4.8?

vandre-sales:
Good catch by @coderabbitai — fixed in `2bd49d74f`.

`completePrompt` was indeed sending `temperature` unconditionally, which would 400 on the adaptive-thinking models (Opus/Sonnet 4.7 and 4.8). `createMessage` already guarded it, but the non-stream path didn't.

What changed:
- Extracted the detection into a private `isAdaptiveThinkingModel(modelId)` method that runs `parseBaseModelId` first, so cross-region/global prefixes (`us.` / `eu.` / `global.`) are handled correctly.
- Reused it in both `createMessage` and `completePrompt` so the two request paths stay in sync (single source of truth instead of a duplicated inline check).
- Added two regression tests on the non-stream path: `completePrompt` omits `temperature` for `opus-4-8` and still sends it for `opus-4-6`.

Validation: 64 bedrock tests pass, `check-types` clean across all packages.

vandre-sales:
Done — added in `506bc9a72`. Thanks for setting up the Bedrock e2e harness, @edelauna; it made this straightforward.

I rebased onto latest `main` to pick up your harness, then added a second test to `apps/vscode-e2e/src/suite/providers/bedrock.test.ts` that mirrors your user-agent smoke test but re-points the provider at `us.anthropic.claude-opus-4-8`. Since 4.8 is an adaptive-thinking model, this exercises the request path that omits `temperature` (and would send `thinking.type: "adaptive"` when reasoning is enabled), proving a Bedrock round-trip completes without a 400.

Details, keeping it consistent with your harness:
- Runs against your binary-event-stream mock server in CI and against real AWS when `BEDROCK_LIVE_E2E=true` — same dual-mode setup as your test.
- Model id is overridable via `BEDROCK_OPUS_48_MODEL_ID` (defaults to `us.anthropic.claude-opus-4-8`, the cross-region profile).
- Your original test is left untouched.

I kept it as a parity/connectivity smoke test to match the harness's intent. If you'd prefer a deeper assertion — e.g. extending the mock server to capture the request body and asserting `thinking: "adaptive"` + absence of `temperature` on the wire — I'm happy to follow up; just let me know which direction you'd like.

`check-types` clean across all packages (`tsconfig.esm.json`).

packages/types/src/providers/vertex.ts
```diff
@@ -396,6 +396,29 @@ export const vertexModels = {
...
+	"claude-opus-4-8": {
+		maxTokens: 8192,
+		contextWindow: 200_000, // Default 200K, extendable to 1M with beta flag 'context-1m-2025-08-07'
+		supportsImages: true,
+		supportsPromptCache: true,
+		inputPrice: 5.0, // $5 per million input tokens (≤200K context)
+		outputPrice: 25.0, // $25 per million output tokens (≤200K context)
+		cacheWritesPrice: 6.25, // $6.25 per million tokens
+		cacheReadsPrice: 0.5, // $0.50 per million tokens
+		supportsReasoningBudget: true,
```
edelauna, line 409: Missing `supportsReasoningBinary: true` — the Anthropic provider entry has it (`anthropic.ts:126`). Without it, `anthropic-vertex.ts` resolves thinking via `getModelParams → getAnthropicReasoning` and emits `{ type: "enabled", budget_tokens: N }`. Anthropic docs confirm `type: "enabled"` is not supported on Opus 4.7+ (returns 400).

Adding the flag alone won't fully fix it — `anthropic-vertex.ts` also needs to call `getAnthropicProviderReasoning` directly (like `anthropic.ts:65`) instead of relying on `reasoning` from `getModelParams`.

packages/types/src/providers/bedrock.ts
```diff
@@ -167,6 +167,56 @@ export const bedrockModels = {
 			},
 		],
 	},
+	"anthropic.claude-opus-4-7": {
+		maxTokens: 8192,
+		contextWindow: 200_000, // Default 200K, extendable to 1M with beta flag 'context-1m-2025-08-07'
+		supportsImages: true,
+		supportsPromptCache: true,
+		supportsReasoningBudget: true,
```
edelauna, line 175: Missing `supportsTemperature: false` — without it `ApiOptions.tsx:787` still shows the temperature slider for this model even though temperature is silently stripped by `isAdaptiveThinkingModel`.

packages/types/src/providers/bedrock.ts
```diff
@@ -167,6 +167,56 @@ export const bedrockModels = {
...
+				cacheReadsPrice: 1.0, // $1.00 per million tokens (>200K context)
+			},
+		],
+	},
+	"anthropic.claude-opus-4-8": {
+		maxTokens: 8192,
+		contextWindow: 200_000, // Default 200K, extendable to 1M with beta flag 'context-1m-2025-08-07'
+		supportsImages: true,
+		supportsPromptCache: true,
+		supportsReasoningBudget: true,
```
edelauna, line 199: Same for Opus 4.8 — `supportsTemperature: false` should be included here too.

edelauna (changes requested):
Nice! Thanks for adding the e2e tests 🔥

edelauna (approved):

PR #388, opened 2026-05-29
fix(task): reset isStreaming in abortTask to prevent UI lock after task switch (fixes #257)

### LINKED ISSUE #257: [BUG] Chat can remain stuck on "API Request..." and stop sending after task/context switch

After a task switch (or certain project-path/worktree context transitions), chat can stay in an API-request state and user messages stop sending. The input appears locked as if a request is still active.

Seen by users in active coding sessions, especially when moving between task contexts (e.g., different project path/worktree or external/internal context switches). Impact is high because chat appears available but messages do not send.

### Reproduction steps

1. Environment/setup: VS Code with Zoo Code extension, any provider/model (reported across multiple providers).
2. Start Task A and trigger a request so chat enters `API Request...` / streaming state.
3. Switch to a different task/context (Task B) without fully resetting session UI state.
4. Try to send a normal follow-up message in Task B.
5. Observe the message is not sent and chat remains effectively locked.

### Expected result

After switching to a new task/context, chat input should re-enable and messages should send normally.

### Actual result

Chat remains stuck in request/disabled state and new messages are not sent.

- Reproduced with different providers/models (community reports mention this across setups).
- Trigger seems correlated with context/task transitions rather than a single provider.

### App Version

v3.55.x (reported; exact patch version unknown)

### API Provider (optional)

Not Applicable / Other

### Model Used (optional)

N/A

N/A

Include UI/API history showing persistent `api_req_started` / locked send state after task switch.

proyectoauraorg:
Fixes #257. Resets isStreaming and isWaitingForFirstChunk in abortTask() so the UI unlocks immediately when a task is aborted or abandoned during task switch. All delegation tests pass (14/14).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where aborting a task could leave the interface in a locked state due to uncleared streaming indicators.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/388?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


taltas:
Screenshot 2026-05-29 at 11 11 10 am
Did you mean to commit this?

proyectoauraorg:
Good catch — that `tmp/README.md` was an accidental local scratch file and has now been removed (816121f5b).

I also reverted an unintended uuid 11→14 bump that was causing the Windows unit-test failures on this branch, so CI should be back to normal. Thanks for flagging it.

src/core/task/Task.ts
```diff
@@ -2118,6 +2118,13 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
 
 		this.abort = true
 
+		// Reset streaming state so the UI unlocks immediately.
+		// Without this, task switches that call abortTask() without going
+		// through resumeAfterDelegation() leave isStreaming=true, causing
+		// the chat input to remain locked (issue #257).
+		this.isStreaming = false
```
edelauna, line 2125: Could we preserve the previous streaming state for cleanup before flipping this? `dispose()` still uses `this.isStreaming && this.diffViewProvider.isEditing` to revert active diff edits, and `cancelTask()` treats `isStreaming === false` as the signal to rehydrate, so clearing it synchronously here may skip cleanup or race the abort save.

edelauna (changes requested):

PR #399, opened 2026-05-29
feat(code-index): add Semble as a local on-the-fly embedding provider

### LINKED ISSUE #200: Add in built code indexing based on semble
Codebase Indexing is difficult to setup with multiple dependencies to install. competition comes with it built in.

Add semble code base indexing, in addition to the current solutions. Should be as easy as enabling it with a checkmark. Consider making it default if nothing is configured

Alternatives already exist

**Additional context**
https://github.com/MinishLab/semble

navedmerchant:
### Related GitHub Issue

Closes: #200 

### Description

Semble is a local code search provider that requires no API keys or external services. When a user selects "Semble - Local" as their code index provider, the extension handles everything automatically:

1. Downloads the correct platform-specific archive from [sembleexec v0.3.1](https://github.com/Zoo-Code-Org/sembleexec) into VS Code's global storage
2. Extracts the fast-start one-dir build (tar.gz on unix, zip on Windows) for ~20x faster startup compared to single-file binaries
3. Verifies the binary is functional, then marks the provider as ready

Searches are performed on-the-fly — semble indexes as it searches, so there's no separate indexing step or background process.

**Supported platforms:** `linux-x64`, `linux-arm64`, `darwin-arm64`, `win32-x64`. The semble option only appears in the UI on these platforms.

**Architecture:**
- `semble-downloader.ts` — Downloads archive from GitHub releases, extracts via system `tar`/`powershell`, caches in `globalStorageUri/semble/`
- `semble-cli.ts` — Thin wrapper that spawns the binary with array args (no shell) for safe argument passing
- `provider.ts` — Orchestrates download → verify → search. Implements the same `searchIndex` interface as other code index providers

The binary is downloaded once and reused across sessions. No user configuration needed.

### Test Procedure

- 56 unit tests across 3 test files:
  - `semble-downloader.spec.ts` — Platform detection, download/skip logic, redirect handling, extraction, cleanup on failure
  - `semble-cli.spec.ts` — CLI spawn behavior, argument construction, JSON output parsing
  - `provider.spec.ts` — Download flow, platform check, error states, search result conversion
- Run: `pnpm --filter zoo-code exec vitest run services/code-index/semble/`
- Manual: Select "Semble - Local" in code index settings on a supported platform → verify binary downloads and search returns results

### Pre-Submission Checklist

- [x] **Scope**: Changes are focused on the semble integration.
- [x] **Self-Review**: Done.
- [x] **Testing**: 56 unit tests covering downloader, CLI, and provider.
- [x] **Documentation Impact**: N/A — semble is self-explanatory from the UI.

### Screenshots / Videos

Screenshot 2026-05-30 at 12 21 46 PM

### Documentation Updates

### Additional Notes

- Intel Mac (`darwin-x64`) is not supported as the tree sitter dependency does not support it.
- The binary persists in global storage and is not deleted on disable for fast re-enablement.

### Get in Touch

navedmerchant

## Summary by CodeRabbit

* **New Features**
  * Added Semble as a local code-indexing provider; OpenRouter added to embedder options.

* **Improvements**
  * On-device code search with lazy setup; UI shows Semble only on supported platforms and hides remote vector-store fields when selected.
  * Robust local installer: download, checksum, extraction, and platform-aware behavior.
  * Provider options for result count/content; Semble considered configured when enabled.
  * Added localized Semble labels across many languages.

* **Tests**
  * Expanded tests for Semble provider, CLI, downloader, and config handling.


navedmerchant:
Not ready for review yet, the code is being reviewed and evolved

thany:
So users have to install python, hope to their diety that they did it right and pip works, and then install semble like that. So that still "a lot" of local fiddling. It's still a fragile solution.

The only thing this removes, is the requirement for a text embedding model, and a Qdrant instance.

It doesn't remove fragility.

It's not specifically in #200, but iyam it would be in the spirit of the requested feature to make code indexing a simple checkbox. Just tick it, and *done*. No local installs of anything. Ticking the checkbox does all the black magic that needs to happen.

navedmerchant:
> So users have to install python, hope to their diety that they did it right and pip works, and then install semble like that. So that still "a lot" of local fiddling. It's still a fragile solution.
> 
> The only thing this removes, is the requirement for a text embedding model, and a Qdrant instance.
> 
> It doesn't remove fragility.
> 
> It's not specifically in #200, but iyam it would be in the spirit of the requested feature to make code indexing a simple checkbox. Just tick it, and _done_. No local installs of anything. Ticking the checkbox does all the black magic that needs to happen.

I had the same vision when trying to implement this (and even had a version implemented) but decided against it for two reasons

1. It added too much code and complexity to zoo code
2. It would basically create another semble implementation inside zoo, which is okay but it would prevent it from getting future improvements from development in semble.


In my testing this is a pretty simple implementation which makes it easier for users to enable codebase indexing. It works quite well and adds value without adding too much complexity inside zoo. 

Setting up semble with pip is a lot simpler than setting up qdrant and gives pretty good indexing performance

But you have a valid point, we should make codebase indexing as seamless as possible, and it should be “built in” we will revisit this in the future.

taltas:
It would be cool if we could add a companion PR to the read me in the semble repo instructions to install zoo code https://github.com/MinishLab/semble

navedmerchant:
@thany I took your advice and now Zoo Code will automatically download a prebuilt semble from https://github.com/navedmerchant/sembleexec for the current platform, making it seamless! all you need to do is tick the box and it should "just work"

@taltas that would not be necessary anymore, it should work out of the box for all users.

navedmerchant:
Verified this PR work on MacOS arm, windows x64 and linux x64. In my testing it makes zoo find relevant context and answer questions about 20-30percent faster

src/services/code-index/semble/semble-downloader.ts
```diff
@@ -0,0 +1,366 @@
...
+
+	try {
+		await downloadFile(url, archivePath)
+
+		// Verify archive integrity before extraction
+		const platformKey = `${process.platform}-${process.arch}`
+		const expectedChecksum = SEMBLE_SHA256[platformKey]
+		if (expectedChecksum) {
+			await verifyChecksum(archivePath, expectedChecksum)
+		}
```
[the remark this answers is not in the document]
navedmerchant: this should not be problem as we will not show semble on an unsupported platform

edelauna:
This PR will set the precedent for allowing the extension to download, configure, install, and run  third party executables. I know we haven’t set any governance principles but my preference would be for users to input a PATH location for where to find the CLI if they want to configure semble.

edelauna:
> This PR will set the precedent for allowing the extension to download, configure, install, and run third party executables. I know we haven’t set any governance principles but my preference would be for users to input a PATH location for where to find the CLI if they want to configure semble.

I guess this isn’t automatic, the user still needs to interact with zoo code to get it installed. I’ll review a bit more.

src/services/code-index/service-factory.ts
```diff
@@ -45,6 +45,12 @@ export class CodeIndexServiceFactory {
 
 		const provider = config.embedderProvider as EmbedderProvider
 
+		if (provider === "semble") {
+			throw new Error(
+				"Semble provider handles its own embedding. Do not call createEmbedder() for semble — use SembleProvider instead.",
```
edelauna, line 54: Is semble really an `EmbedderProvider`? It doesn't generate embedding vectors — it's a self-contained search binary. Forcing it into this type means the factory has to throw at runtime to prevent misuse, which feels like the type system telling us the abstraction is wrong. Would a discriminated union at the manager level (branching before the factory is involved) be cleaner here?

src/services/code-index/semble/provider.ts
```diff
@@ -0,0 +1,241 @@
...
+
+	get state(): IndexingState {
+		return this._state
+	}
+
+	/**
+	 * Initializes the provider: downloads semble, then validates it works.
+	 */
+	async initialize(): Promise<void> {
+		if (this._isInitialized) {
```
edelauna, line 59: What happens if `initialize()` is called concurrently — e.g., manager reconfigures while a prior init is still downloading? Both callers pass this guard simultaneously, and two concurrent downloads race to the same archive path. An `_initPromise` that subsequent callers `await` (or a `CancellationTokenSource` to cancel the in-flight init) would be the idiomatic VS Code pattern here.

src/services/code-index/semble/semble-cli.ts
```diff
@@ -0,0 +1,175 @@
...
+				shell: false,
+				timeout: options.timeout,
+				stdio: ["ignore", "pipe", "pipe"],
+			})
+
+			let stdout = ""
+			let stderr = ""
+
+			child.stdout?.on("data", (data: Buffer) => {
+				stdout += data.toString()
```
edelauna: This accumulates stdout without any limit. If the binary outputs large verbose output (or is the wrong version), this could exhaust the extension host heap well before the 120s timeout fires — and since the extension host is shared across all extensions, it'd take everything down. Worth capping at some max bytes?

src/services/code-index/semble/provider.ts
```diff
@@ -0,0 +1,241 @@
...
+			// Filter results to the requested directory prefix, if any.
+			if (directoryPrefix) {
+				const normalizedPrefix = path.join(this.workspacePath, directoryPrefix).replace(/\\/g, "/")
+				converted = converted.filter((r) => {
+					const filePath = (r.payload?.filePath ?? "").replace(/\\/g, "/")
+					return filePath.startsWith(normalizedPrefix + "/") || filePath === normalizedPrefix
+				})
+				console.log(
+					`[SembleProvider] Filtered to "${directoryPrefix}": ${converted.length} of ${results.length} results`,
+				)
```
edelauna, line 195: For every other provider, `directoryPrefix` scopes the DB query *before* `topK` results are fetched, so `topK` counts within the prefix. Here it's a post-hoc filter on results already fetched from the full workspace — so on a large monorepo, a scoped query could return 0 results because semble spent all `topK` slots on other directories. Is this a known tradeoff, or should we bump the initial `topK` and then trim?

src/services/code-index/semble/semble-downloader.ts
```diff
@@ -0,0 +1,366 @@
...
+		const request = https.get(url, (response) => {
+			// Follow redirects
+			if (
+				response.statusCode &&
+				response.statusCode >= 300 &&
+				response.statusCode < 400 &&
+				response.headers.location
+			) {
+				response.destroy()
+				downloadFile(response.headers.location, destPath, maxRedirects - 1)
```
edelauna: Should we validate that the redirect target stays on a trusted domain (github.com / objects.githubusercontent.com) before following it? A compromised CDN redirecting to a different HTTPS host would succeed silently, and the `if (expectedChecksum)` guard at line 169 means a future platform without a SHA256 entry would skip integrity verification entirely.

src/services/code-index/semble/semble-downloader.ts
```diff
@@ -0,0 +1,366 @@
...
+		return undefined
+	}
+}
+
+/**
+ * Extracts a .tar.gz archive into the destination directory using the system `tar` command.
+ */
+function extractTarGz(archivePath: string, destDir: string): Promise<void> {
+	return new Promise((resolve, reject) => {
+		const child = spawn("tar", ["-xzf", archivePath, "-C", destDir], {
```
edelauna: Had we considered adding `--no-absolute-filenames` (or the equivalent) to the tar invocation? Without it, a crafted archive could write files outside `destDir` via absolute path entries. The checksum verification is the primary guard, but defense-in-depth on the extraction step seems worthwhile given the binary gets executed.

src/services/code-index/semble/provider.ts
```diff
@@ -0,0 +1,241 @@
...
+		if (this._state === "Error") {
+			return []
+		}
+
+		try {
+			// Always search the full workspace to maintain a single Semble cache.
+			// Semble creates a separate cache directory per path (SHA-256 of the
+			// resolved absolute path), so passing subdirectories would create
+			// redundant indexes and waste disk space.
+			console.log(`[SembleProvider] Searching for "${query}" in ${this.workspacePath}`)
```
edelauna: This logs the raw user query on every search call. In enterprise deployments, extension host logs can be forwarded to telemetry — is there a debug-gated `OutputChannel` we should be using here instead of `console.log`, consistent with how the other providers log?

src/services/code-index/semble/semble-downloader.ts
```diff
@@ -0,0 +1,366 @@
...
+ * compared to single-file binaries.
+ */
+const SEMBLE_ARCHIVES: Record<string, { archive: string; binary: string }> = {
+	"linux-x64": { archive: "semble-linux-x64-fast.tar.gz", binary: "semble" },
+	"linux-arm64": { archive: "semble-linux-arm64-fast.tar.gz", binary: "semble" },
+	"darwin-arm64": { archive: "semble-macos-arm64-fast.tar.gz", binary: "semble" },
+	"win32-x64": { archive: "semble-windows-x64-fast.zip", binary: "semble.exe" },
+}
+
+const SEMBLE_VERSION = "v0.3.1"
```
edelauna, line 23: re: keeping `SEMBLE_VERSION` and `SEMBLE_SHA256` in sync going forward, we could wire up Renovate to handle this automatically.

Renovate already has custom regex managers in this repo, so the version bump part is straightforward. The checksums are trickier since there are 4 platform-specific artifacts, but we can handle it with a `postUpgradeTasks` script that recomputes them after the version bump.

Rough shape:

1. Add a helper script (`scripts/update-semble-checksums.sh`) that reads `SEMBLE_VERSION` from `semble-downloader.ts`, downloads each archive, computes `shasum -a 256`, and patches the file in-place.

2. Add a custom manager to renovate.json:json
```json
{
  "customType": "regex",
  "description": "Update semble binary version from GitHub releases.",
  "managerFilePatterns": ["src/services/code-index/semble/semble-downloader.ts"],
  "matchStrings": ["const SEMBLE_VERSION = \"(?<currentValue>v[^\"]+)\""],
  "depNameTemplate": "Zoo-Code-Org/sembleexec",
  "datasourceTemplate": "github-releases",
  "postUpgradeTasks": {
    "commands": ["scripts/update-semble-checksums.sh"],
    "fileFilters": ["src/services/code-index/semble/semble-downloader.ts"]
  }
}
```

This way when a new semble release drops, Renovate opens a PR with both the version bump and updated checksums — no manual step, no risk of the two maps drifting out of sync.

Doesn't have to be in this PR — just flagging it as a follow-up so we don't end up doing manual bumps forever.

src/services/code-index/semble/semble-downloader.ts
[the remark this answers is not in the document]
edelauna: I agree with @coderabbitai on this - lets just hard-error, rather then conditionally verify.

edelauna (changes requested):
Looks good - has some suggestions, and requesting changes to switch from conditional checksum verification to hard fail on unsupported architectures.

src/services/code-index/semble/provider.ts
navedmerchant: Its a known tradeoff. semble seems to be designed with topK of 10 so i kept it as that. Semble indexes on the path it has been provided so if we index on subdirectories it keeps building its index even if there is an existing index cached on the disk for a parent subdirectory. it also takes a lot more space. This way the entire workspace directory is indexed by semble. If any file changes semble just invalidates and overwrites to old index by rebuilding it. It feels inefficient but semble creators argue that since it is so fast it makes sense to do it to keep index fresh

src/services/code-index/service-factory.ts
navedmerchant: You're right that semble isn't a true embedder — it's a self-contained search binary. The EmbedderProvider type is really acting as a "code index provider selector" here rather than strictly meaning "thing that produces embeddings."

The manager already branches before the factory is involved (_recreateServices checks currentEmbedderProvider === "semble" and creates SembleProvider directly), so the factory throw is purely a defensive guard that can't fire in normal operation. I've added a comment clarifying this.

Splitting into a proper discriminated union would be more type-correct but touches the types package, config schema, webview UI, and serialization — more scope than I want to add to this PR. If we end up adding more self-contained search providers in the future, that'd be the right time to introduce the abstraction. For now semble is the only one and the branching in the manager keeps it clean.

src/services/code-index/semble/provider.ts
navedmerchant: There's no debug-gated OutputChannel in the code-index subsystem — all other providers (OpenAI, Ollama, OpenRouter, scanner, cache-manager) use plain console.log/console.error directly. I've removed the raw query from the log message (now just logs the workspace path and result count) so there's no user content leaking into logs. Happy to introduce a shared debug-gated logger for the code-index module as a follow-up, but didn't want to scope-creep this PR with new logging infrastructure that no other provider uses yet.

src/services/code-index/service-factory.ts
edelauna: Tracked in #479 for a follow-up refactor.

src/services/code-index/semble/semble-downloader.ts
edelauna: Tracked in #481.

src/services/code-index/semble/provider.ts
edelauna: Addressed — `_initPromise` guard is in place (provider.ts lines 64–73). Concurrent callers await the same promise rather than racing to a second download.

src/services/code-index/semble/semble-cli.ts
edelauna: Addressed — `_spawn` now tracks `stdoutBytes` / `stderrBytes` and stops accumulating after 10 MB each (semble-cli.ts lines 104–121).

src/services/code-index/semble/provider.ts
edelauna: Fair enough — understood. Accepting the tradeoff: full-workspace indexing keeps a single Semble cache and the post-hoc filter is the deliberate design. Worth revisiting if we get reports of empty results on large monorepos with scoped queries.

src/services/code-index/semble/semble-downloader.ts
edelauna: Addressed — `TRUSTED_DOWNLOAD_DOMAINS` allowlist and `isTrustedDownloadUrl()` check are in place (semble-downloader.ts lines 323–367). Redirects outside github.com / objects.githubusercontent.com / release-assets.githubusercontent.com are rejected with an error.

src/services/code-index/semble/semble-downloader.ts
edelauna: Addressed — Linux uses `--no-absolute-filenames` and the comment explains macOS bsdtar strips absolute paths by default (semble-downloader.ts lines 244–248).

src/services/code-index/semble/provider.ts
edelauna: Good call removing the raw query.

src/services/code-index/semble/semble-downloader.ts
```diff
@@ -0,0 +1,403 @@
...
+ */
+const TRUSTED_DOWNLOAD_DOMAINS = ["github.com", "objects.githubusercontent.com", "release-assets.githubusercontent.com"]
+
+/**
+ * Validates that a URL belongs to a trusted domain.
+ */
+function isTrustedDownloadUrl(url: string): boolean {
+	try {
+		const parsed = new URL(url)
+		return parsed.protocol === "https:" && TRUSTED_DOWNLOAD_DOMAINS.some((d) => parsed.hostname.endsWith(d))
```
edelauna: `endsWith` has no domain-boundary awareness — `"evilgithub.com".endsWith("github.com")` is `true`. Could replace with `h === d || h.endsWith("." + d)`?

src/services/code-index/semble/semble-downloader.ts
```diff
@@ -0,0 +1,403 @@
...
+
+	// Version mismatch — remove old installation before downloading new one
+	if (installedVersion && installedVersion !== SEMBLE_VERSION) {
+		console.log(`[SembleDownloader] Version changed from ${installedVersion} to ${SEMBLE_VERSION}, updating...`)
+		try {
+			await fs.rm(extractDir, { recursive: true, force: true })
+		} catch {
+			// ignore cleanup errors
+		}
+	}
```
edelauna: If the download or checksum fails after line 153, the user is left with no binary until they retry — the old installation is gone before the new one is verified. Would a staging directory (`extractDir + ".new"`) + atomic rename after verification be feasible here?

src/services/code-index/semble/semble-cli.ts
```diff
@@ -0,0 +1,186 @@
...
+			let stdout = ""
+			let stderr = ""
+			let stdoutBytes = 0
+			let stderrBytes = 0
+
+			child.stdout?.on("data", (data: Buffer) => {
+				stdoutBytes += data.length
+				if (stdoutBytes <= MAX_BUFFER_BYTES) {
+					stdout += data.toString()
+				}
```
edelauna, line 131: When the cap is exceeded, chunks are silently dropped and `_parseOutput` receives truncated JSON → returns `[]` with only a `console.warn`. In the extension host this surfaces as empty search results with no user-visible explanation. Worth logging a warning or rejecting when `stdoutBytes > MAX_BUFFER_BYTES` is first crossed?

src/services/code-index/semble/semble-downloader.ts
```diff
@@ -0,0 +1,403 @@
...
+ * and strips absolute paths to prevent path traversal attacks.
+ */
+function extractTarGz(archivePath: string, destDir: string): Promise<void> {
+	return new Promise((resolve, reject) => {
+		const args = ["-xzf", archivePath, "-C", destDir, "--no-same-owner"]
+		// GNU tar supports --absolute-names / -P to refuse absolute paths.
+		// macOS (bsdtar) strips absolute paths by default and doesn't need this flag.
+		if (process.platform === "linux") {
+			args.push("--no-absolute-filenames")
+		}
```
edelauna, line 266: `--no-absolute-filenames` blocks leading-slash entries but not `../`-based relative traversal or symlink-based zip-slip on either platform. Is the checksum verification considered the primary (and sufficient) guard here, or should we add `--no-overwrite-dir` as belt-and-suspenders?

src/services/code-index/semble/provider.ts
```diff
@@ -0,0 +1,260 @@
...
+	 * Results with missing file paths are excluded.
+	 */
+	private _convertResults(results: SembleSearchResult[], basePath: string): VectorStoreSearchResult[] {
+		return results
+			.filter((r) => r.chunk?.file_path) // Exclude results with no file path
+			.map((r, index) => ({
+				id: `semble-${index}`,
+				score: r.score,
+				payload: {
+					filePath: path.join(basePath, r.chunk.file_path).replace(/\\/g, "/"),
```
edelauna: If the semble binary (or a tampered one) emits a `file_path` like `../../.ssh/id_rsa`, `path.join` normalises the `..` and the resolved path escapes the workspace. The `directoryPrefix` filter below catches this only when a prefix is provided. Should we assert `filePath.startsWith(basePath)` before including a result?

src/services/code-index/semble/__tests__/semble-downloader.spec.ts
```diff
@@ -0,0 +1,690 @@
...
+			;(fs.readFile as any).mockRejectedValue(new Error("ENOENT"))
+
+			// First call returns a redirect, second call returns 200
+			let callCount = 0
+			;(https.get as any).mockImplementation((_url: string, callback: (res: any) => void) => {
+				callCount++
+				const res = new EventEmitter() as any
+				if (callCount === 1) {
+					res.statusCode = 302
+					res.headers = {
```
edelauna, line 344: The untrusted-redirect test only blocks `evil.example.com`. Does `evilgithub.com` (which satisfies `endsWith("github.com")`) also get blocked? Worth adding a case to pin that boundary.

edelauna (commented):
Thanks for addressing my previous request changes, just some additional security comments this time around.

edelauna (approved):
🚀

DScoNOIZ:
This is a really cool and important solution, in my opinion! Thanks for implementing it. Sometimes a lightweight local indexer is really needed for a heavy codebase. It's so you don't have to load ollama and the heavy embedding model.

PR #401, opened 2026-05-29
Handle per-key failures during settings import

taltas:
## Summary
- make settings import resilient to invalid values on individual keys instead of failing the entire import
- continue importing valid settings while collecting/reporting per-key failures
- add coverage for resilient partial import behavior in both import/export and auto-import paths

## Why
RooCode settings import could fail entirely when a single persisted setting was invalid for the current schema. One concrete case is [`imageGenerationProvider`](packages/types/src/image-generation.ts:35) receiving the invalid value `roo`, which causes the import path to reject the whole payload.

This change makes import behavior more resilient by skipping only the invalid keys, preserving valid settings from the same import, and covering the behavior with regression tests.

## Summary by CodeRabbit

* **New Features**
  * Enhanced validation of global settings during import with improved handling of unsupported values
  * Refined warning messages to provide clearer feedback on import issues

* **Tests**
  * Expanded test coverage for global settings import behavior and warning reporting

[Review Change Stack](https://app.coderabbit.ai/change-stack/Zoo-Code-Org/Zoo-Code/pull/401?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/core/config/__tests__/importExport.spec.ts
+++ b/src/core/config/__tests__/importExport.spec.ts
@@ -757,7 +757,7 @@ describe("importExport", () => {
 
 				// Should show warning message with short summary (not full details)
 				expect(showWarningMessageSpy).toHaveBeenCalledWith(
-					expect.stringContaining("1 profile had issues during import."),
+					expect.stringContaining("1 item had issues during import."),
 				)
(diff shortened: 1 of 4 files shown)


edelauna (approved):
Thanks for doing this - sorry it took so long to review.

PR #419, opened 2026-05-31
🔴 URGENT: fix(i18n) resolve stale translations in memoized dialog components

DScoNOIZ:
Fixes issue with stale translations in memoized dialog components.


DScoNOIZ:
@coderabbitai The regression test has already been added in this PR — see `webview-ui/src/i18n/__tests__/TranslationContext.spec.tsx`. It covers:

1. **Normal consumer** re-renders on language change
2. **`React.memo`-wrapped consumer** re-renders on language change
3. **Delete dialog** translations update correctly after language switch

All 4 tests pass ✅

The fix ensures `i18n.language` is in the `useCallback` dependency array so the `translate` function reference updates when language changes, triggering proper re-renders across all `useAppTranslation()` consumers — including memoized dialog components.

navedmerchant:
Please do not bump the version, it will be done when we do a weekly release

navedmerchant:
Also please check the tests are failing

DScoNOIZ:
> Please do not bump the version, it will be done when we do a weekly release
I updated the version only because VISUAL STUDIO CODE doesn't apply updates to the installed .vsix if the package versions match. When merging, simply ignore this file. Best regards. I needed to test the package.

DScoNOIZ:
> Also please check the tests are failing
All my tests pass perfectly. Is it the same for you?

PR #421, opened 2026-05-31
feat(bailian): add Bailian (Alibaba Cloud) as an AI provider

### LINKED ISSUE #420: [MODELS] add Bailian (Alibaba Cloud) as an AI provider
It would be useful to have Bailian (Alibaba Cloud) available as an AI provider in Zoo Code. Bailian is the official platform for the Qwen model series, and offers multiple models from multiple creators.

Models I'd like to use:

- qwen3.7-max, by Qwen
- qwen3.6-plus, by Qwen
- qwen3.6-flash, by Qwen
- deepseek-v4-pro, by DeepSeek
- deepseek-v4-flash, by DeepSeek
- glm-5.1, by Zhipu AI
- kimi-k2.6, by Moonshot AI
- MiniMax-M2.5, by MiniMax

Custom model IDs beyond the above list would also be useful, for models Bailian may add in the future.

One thing to note: Bailian has multiple region endpoints with different pricing, so configuring a region and resolving pricing accordingly would be necessary.

Reference: https://www.alibabacloud.com/help/en/model-studio/models

F915:
### Related GitHub Issue

Closes: https://github.com/Zoo-Code-Org/Zoo-Code/issues/420

### Description

This PR adds **Bailian (Alibaba Cloud)** as a first-class AI Provider.

**Provider Handler** (`src/api/providers/bailian.ts`) extends `BaseOpenAiCompatibleProvider`. Model resolution looks up static presets first, then falls back to the API-fetched cache and canonical matching (handling versioned IDs like `qwen3.7-max-2026-05-17` and namespace-prefixed IDs like `kimi/kimi-k2.6`). 

**Region Endpoints** (`src/api/providers/bailian-region.ts`) support all 8 regions: Beijing, Singapore, Virginia, Frankfurt, Hong Kong, Coding Plan CN, Token Plan CN, and Token Plan SGP.
**Dynamic Model Fetching** (`src/api/providers/fetchers/bailian.ts`) pulls the model list from the `/models` endpoint and filters out non-text models by keyword (image/embed/speech/asr/tts/video/ocr/wanx). Models matching a static preset inherit its metadata and pricing; unmatched models are included with conservative defaults and can be customized. `findMatchingPreset` is exported as a standalone function reused by both the handler's `getModel()` and the webview's `useSelectedModel`.

**Pricing** (`packages/types/src/providers/bailian-pricing.ts`) resolves per region across 4 tables (CN, SGP, EU, US), with international regions preferring their specific table.

**Static Preset Models** — 12 total, covering Qwen (6: 3.7-max/plus, 3.6-plus/flash, 3.5-plus/flash), DeepSeek V4 (pro/flash), Zhipu GLM-5.1, Moonshot Kimi K2.6, MiniMax M2.5, and Xiaomi MiMo V2.5 Pro. Qwen 3.7-plus is the default model.

**UI** (`webview-ui/src/components/settings/providers/Bailian.tsx`): region dropdown, conditional Workspace ID input (Frankfurt/Hong Kong only), API Key input, model picker (static presets pinned to top, API-fetched models appended), and automatic custom model detection with an expandable configuration panel (maxTokens, contextWindow, image support, prompt caching, input/output/cache pricing).

**Infrastructure Changes**: modelCache dedup key changed from `provider` to `provider:baseUrl` to prevent race conditions across concurrent region endpoint refreshes; telemetry added to all catch blocks in modelCache and Task constructor / `updateApiConfiguration`; Task constructor fails fast when `buildApiHandler` throws instead of leaving `this.api` as undefined; `updateApiConfiguration` builds the new handler before updating config to keep the pair atomic; `getModelMaxOutputTokens` guards against negative values (e.g. `-1` meaning server default). Three optional fields added to `ModelInfo`: `supportsReasoningBinary`, `supportsReasoningEffort`, and `maxThinkingTokens`.

**Tests** cover 6 files: handler, fetcher, pricing, Task constructor failure, UI component, and E2E (fetch interceptor pattern covering basic streaming, reasoning params, workspaceId URLs, and related scenarios).

### Test Procedure

```sh
# Backend unit tests
cd src && npx vitest run api/providers/__tests__/bailian.spec.ts
cd src && npx vitest run api/providers/fetchers/__tests__/bailian.spec.ts
cd src && npx vitest run core/task/__tests__/Task.spec.ts

# Pricing unit tests
cd packages/types && npx vitest run src/providers/__tests__/bailian-pricing.spec.ts

# Frontend component tests
cd webview-ui && npx vitest run src/components/settings/providers/__tests__/Bailian.spec.tsx

# E2E tests (mock mode, no API key needed)
pnpm --filter @roo-code/vscode-e2e test:ci:mock -- --file bailian.test.ts

# Full type check
pnpm check-types
```

### Pre-Submission Checklist

### Screenshots / Videos

Bailian provider settings UI

### Documentation Updates

- [x] Yes, documentation updates are required: a Bailian provider docs page covering dynamic model fetching, region selection, Workspace ID configuration, and reasoning mode setup.

### Get in Touch

Discord: Yon (yon.sinc)

## Summary by CodeRabbit

* **New Features**
  * Added Bailian (Alibaba Cloud) provider with regional endpoints, workspace support, model catalog, streaming completions, thinking/reasoning controls, and runtime pricing visibility.
* **UI**
  * New Bailian settings panel, region/workspace/API-key inputs, model picker integration, and custom-model override controls.
* **Behavior**
  * Prompt-caching for compatible models, improved model selection/validation, and default reasoning toggles.
* **Documentation**
  * New Bailian provider configuration guide.
* **Localizations**
  * Bailian strings added across 20+ locales.
* **Tests**
  * Extensive unit, integration, and e2e coverage for Bailian flows.


F915:
@coderabbitai review

src/api/providers/fetchers/modelCache.ts
```diff
@@ -107,6 +118,33 @@ async function fetchModelsFromProvider(options: GetModelsOptions): Promise<Model
 		case "deepseek":
 			models = await getDeepSeekModels(options.baseUrl, options.apiKey)
 			break
+		case "bailian":
```
edelauna, line 121: The `main` branch has `URL_SCOPED_PROVIDERS` / `getCacheKey()` (added in #647) which scopes the memory-cache key by `provider:baseUrl` for litellm, poe, deepseek, ollama, lmstudio, and requesty. Since this branch diverged before that commit, both `memoryCache.set(provider, models)` calls still use a bare provider key. Merging as-is would regress URL-scoped cache isolation for those providers — a user switching litellm base URLs would see the previous URL's models until TTL expires.

Could you rebase onto `main` to pick up the `getCacheKey` fix, then reconcile it with the new `inFlightKey` approach?

src/api/providers/bailian.ts
```diff
@@ -0,0 +1,349 @@
...
+		const params: Record<string, any> = {}
+
+		const isEffortReasoningModel = Array.isArray(modelInfo.supportsReasoningEffort)
+		const isBinaryReasoningModel = modelInfo.supportsReasoningBinary === true && !isEffortReasoningModel
+
+		if (isBinaryReasoningModel) {
+			if (this.options.enableReasoningEffort !== false) {
+				params.enable_thinking = true
+				if (this.options.modelMaxThinkingTokens != null) {
+					params.thinking_budget = this.options.modelMaxThinkingTokens
```
edelauna, line 270: DashScope requires `thinking_budget < max_tokens`, but `thinking_budget` is set directly from `options.modelMaxThinkingTokens` with no check against the `max_tokens` value computed above (e.g. for `qwen3.7-max`, `maxThinkingTokens = 262144` but `max_tokens = 65536` after the 20% cap). This will cause a 400 API error whenever binary reasoning is active.

Would something like `Math.min(this.options.modelMaxThinkingTokens, maxTokens - 1)` work here? (`maxTokens` would need to be threaded into this method.)

src/api/providers/bailian.ts
```diff
@@ -0,0 +1,349 @@
...
+		const { id: modelId, info: modelInfo } = this.getModel()
+
+		const params: Record<string, any> = {
+			model: modelId,
+			messages: [{ role: "user", content: prompt }],
+			stream: false,
+		}
+
+		// Delegate reasoning params to shared helper
+		Object.assign(params, this.getReasoningParams(modelInfo))
```
edelauna, line 218: Same `thinking_budget` concern as in `createStream` — but here `max_tokens` is never set at all. DashScope requires `thinking_budget < max_tokens`; without `max_tokens` in the request the server falls back to its own ceiling and rejects if the budget exceeds it.

Should `completePrompt` set `max_tokens` before spreading `getReasoningParams`?

src/api/providers/bailian.ts
```diff
@@ -0,0 +1,349 @@
...
+
+		try {
+			return this.client.chat.completions.create(
+				params as unknown as OpenAI.Chat.Completions.ChatCompletionCreateParamsStreaming,
+				requestOptions,
+			)
+		} catch (error) {
+			throw handleOpenAIError(error, "Bailian")
```
edelauna, line 196: Peer providers (e.g. `deepseek.ts:136`, `openrouter.ts:366`) call `TelemetryService.instance.captureException` in their catch blocks before re-throwing. The same gap exists in `completePrompt` at line 238. Is there a reason to skip telemetry on Bailian API errors?

packages/types/src/providers/bailian.ts
```diff
@@ -0,0 +1,182 @@
...
+export const bailianModels = {
+	"qwen3.7-max": {
+		maxTokens: 65_536,
+		maxThinkingTokens: 262_144,
+		contextWindow: 1_000_000,
+		supportsImages: false,
+		supportsPromptCache: true,
+		supportsReasoningBinary: true,
+		supportsTemperature: true,
+		inputPrice: 1.65,
```
edelauna, line 31: The comment in `bailian-pricing.ts:31` says these model definitions **MUST NOT duplicate pricing** and that `getBailianPrice()` is the single source of truth — but every model here carries `inputPrice`/`outputPrice`/cache pricing that exactly matches `CN_GLOBAL_PRICES`. Since `getBailianPrice()` always overwrites these at runtime, could the pricing fields be dropped here to avoid the dual-maintenance hazard?

src/api/providers/__tests__/bailian.spec.ts
```diff
@@ -0,0 +1,621 @@
...
+
+		it("versioned variant gets canonical pricing via findMatchingPreset", () => {
+			const h = new BailianHandler({
+				bailianApiKey: "test-key",
+				bailianRegion: "singapore",
+				apiModelId: "qwen3.7-max-2026-05-17",
+			})
+			const result = h.getModel()
+			expect(result.info.maxTokens).toBe(65536)
+			expect(result.info.inputPrice).toBeDefined()
```
edelauna, line 548: The non-versioned test at line 58 asserts the exact preset price with `.toBe(bailianModels[testModelId].inputPrice)`. Could this versioned-ID test do the same — e.g. `expect(result.info.inputPrice).toBe(1.65)` for `qwen3.7-max`? `toBeDefined()` would pass even if canonicalization returned the wrong model's price.

src/api/providers/__tests__/bailian.spec.ts
```diff
@@ -0,0 +1,621 @@
...
+		new BailianHandler({ bailianApiKey: "test-key", bailianRegion: "hongkong", bailianWorkspaceId: "hk-456" })
+		expect(OpenAI).toHaveBeenCalledWith(
+			expect.objectContaining({ baseURL: "https://hk-456.cn-hongkong.maas.aliyuncs.com/compatible-mode/v1" }),
+		)
+	})
+
+	// --- Constructor error tests (previously outside describe block) ---
+
+	it("should throw when Frankfurt region is used without workspaceId", () => {
+		expect(() => new BailianHandler({ bailianApiKey: "test-key", bailianRegion: "frankfurt" })).toThrow()
```
edelauna, line 115: The `invalid-region` test at line 122 uses `.toThrow(/Unknown Bailian region/)` to pin the error message. Could the Frankfurt and Hong Kong tests do the same — e.g. `.toThrow(/workspaceId/)` — so an unrelated crash doesn't accidentally pass?

src/api/providers/__tests__/bailian.spec.ts
```diff
@@ -0,0 +1,621 @@
...
+	})
+
+	it("should throw when Hong Kong region is used without workspaceId", () => {
+		expect(() => new BailianHandler({ bailianApiKey: "test-key", bailianRegion: "hongkong" })).toThrow()
```
edelauna, line 119: Same as the Frankfurt case above — `.toThrow()` with no argument would also pass on an unrelated crash. `.toThrow(/workspaceId/)` would make this more precise.

src/api/providers/__tests__/bailian.spec.ts
```diff
@@ -0,0 +1,621 @@
...
+		expect(result.info.inputPrice).toBe(bailianModels[testModelId].inputPrice)
+		expect(result.info.outputPrice).toBe(bailianModels[testModelId].outputPrice)
+		expect(result.info.contextWindow).toBe(bailianModels[testModelId].contextWindow)
+	})
+
+	// --- getModel() returns getModelParams-computed fields ---
+
+	it("getModel returns maxTokens from getModelParams", () => {
+		const result = handler.getModel()
+		expect(result).toHaveProperty("maxTokens")
```
edelauna, line 67: `getModelParams` always spreads `{ maxTokens, temperature }` into the returned object (even when the values are `undefined`), so `toHaveProperty` can never fail here regardless of what `getModel` actually computes. Would asserting a concrete value — e.g. `expect(result.maxTokens).toBe(expectedValue)` — be more useful?

src/api/providers/fetchers/__tests__/modelCache.spec.ts
```diff
@@ -5,6 +5,7 @@ vi.mock("@roo-code/telemetry", () => ({
 	TelemetryService: {
 		instance: {
 			captureEvent: vi.fn(),
+			captureException: vi.fn(),
```
edelauna, line 8: Is there a test for two concurrent `refreshModels` calls with the same provider but different `baseUrl`s (e.g. two Bailian regions)? With `memoryCache.set(provider, models)` using a bare provider key, the second write would overwrite the first — worth confirming that's the intended behavior and that there's a test for it.

src/api/providers/__tests__/bailian.spec.ts
```diff
@@ -0,0 +1,621 @@
...
+			expect.unreachable("should have thrown")
+		} catch (e: any) {
+			expect(e.message).toContain("beijing")
+			expect(e.message).toContain("singapore")
+		}
+	})
+
+	// --- Streaming ---
+
+	it("createMessage should yield text content from stream", async () => {
```
edelauna, line 140: I don't see a test where `mockCreate` rejects (e.g. a 429 or network error) in the `createMessage` path. Could the streaming error path be covered — confirming the error is re-thrown rather than swallowed?

webview-ui/src/components/settings/providers/__tests__/Bailian.spec.tsx
```diff
@@ -0,0 +1,179 @@
+import React from "react"
+import { render, screen } from "@/utils/test-utils"
+import { Bailian } from "../Bailian"
+import type { ProviderSettings } from "@roo-code/types"
+
+// Mock vscrui Checkbox
+vi.mock("vscrui", () => ({
+	Checkbox: ({ children, checked, onChange }: any) => (
+		<label>
+			<input
```
edelauna, line 10: `Bailian.tsx:166` conditionally renders the workspace ID input when region is `frankfurt` or `hongkong`. All tests here use the default `beijing` config — could a test be added to confirm the field appears for `frankfurt` and is absent for `beijing`?

packages/types/src/provider-settings.ts
```diff
@@ -349,6 +353,17 @@ const mimoSchema = apiModelIdProviderModelSchema.extend({
 	mimoApiKey: z.string().optional(),
 })
 
+const bailianSchema = apiModelIdProviderModelSchema.extend({
+	bailianApiKey: z.string().optional(),
+	bailianRegion: z.enum(BAILIAN_REGIONS).optional(),
+	bailianWorkspaceId: z.string()
+		.refine((val) => /^[a-zA-Z0-9][a-zA-Z0-9_-]*$/.test(val), {
+			message: "Workspace ID must contain only alphanumeric characters, hyphens, and underscores",
```
edelauna, line 361: The regex `/^[a-zA-Z0-9][a-zA-Z0-9_-]*$/` requires the first character to be alphanumeric, but the error message only says "alphanumeric characters, hyphens, and underscores" — implying `_foo` is valid when it isn't. The same message appears in `bailian-region.ts:49`. Should it say "must start with an alphanumeric character and contain only..."?

packages/types/src/providers/bailian.ts
```diff
@@ -0,0 +1,182 @@
...
+		contextWindow: 262_144,
+		supportsImages: false,
+		supportsPromptCache: true,
+		supportsReasoningBinary: true,
+		supportsTemperature: true,
+		inputPrice: 0.8939,
+		outputPrice: 3.7131,
+		cacheWritesPrice: 1.1174,
+		cacheReadsPrice: 0.0894,
+		description: "Moonshot Kimi K2.6 with thinking mode disabled by default. Supports Token Plan and Coding Plan.",
```
edelauna, line 159: The description says "thinking mode disabled by default", but `getReasoningParams` sends `enable_thinking: true` when `enableReasoningEffort !== false` — which is true when the field is `undefined` (user hasn't touched the setting). Should the description say "enabled by default unless explicitly disabled", or should `kimi-k2.6` opt out of the default-on behavior?

packages/types/src/providers/bailian.ts
```diff
@@ -0,0 +1,182 @@
...
+	},
+	"mimo-v2.5-pro": {
```
edelauna, line 161: `bailianModels` and `mimoModels` both define `mimo-v2.5-pro` with different specs: this copy has `contextWindow: 1_000_000` and no `preserveReasoning`, while `mimo.ts` has `contextWindow: 1_048_576` and `preserveReasoning: true`. Is MiMo V2.5 Pro actually available via DashScope Token Plan? If not, removing it here would avoid the silent capability mismatch.

src/api/providers/fetchers/bailian.ts
```diff
@@ -0,0 +1,198 @@
...
+			console.error(`[getBailianModels] HTTP error:`, {
+				status: response.status,
+				statusText: response.statusText,
+				url,
+				body: errorBody.substring(0, 500),
+			})
+
+			TelemetryService.instance.captureException(
+				new Error(`[getBailianModels] HTTP ${response.status}: ${response.statusText}`),
+				{ extra: { url, status: response.status, body: errorBody.substring(0, 500) } },
```
edelauna, line 150: For Frankfurt/Hong Kong regions, `url` here includes the workspace ID as a subdomain (e.g. `https://{workspaceId}.eu-central-1.maas.aliyuncs.com/...`). Should the workspace ID be redacted before it's sent to PostHog — e.g. replacing the subdomain with `<workspace>`?

src/api/providers/bailian.ts
```diff
@@ -0,0 +1,349 @@
...
+ *
+ * Extends {@linkcode BaseOpenAiCompatibleProvider} with Bailian-specific
+ * reasoning parameters (enable_thinking, reasoning_effort), prompt caching
+ * via cache_control injection, and region-aware pricing.
+ *
+ * Supports 8 static preset models and dynamic model fetching from the
+ * DashScope /models endpoint for the full 200+ model catalog.
+ */
+export class BailianHandler extends BaseOpenAiCompatibleProvider<BailianModelId> {
+	private readonly bailianOptions: ApiHandlerOptions
```
edelauna, line 37: The base class already exposes `protected readonly options: ApiHandlerOptions` with the same value. Could `bailianOptions` be removed and the three accesses replaced with `this.options`?

edelauna (changes requested):
Thanks, as it stands this PR needs a bit of work before it can be re-reviewed and merged.

Issue #392, opened 2026-05-29
feat(security): migrate remaining tools to WorkspaceFileAccess — write, edit, list, search, image (#169)


edelauna:
Part of #169. Sub-issue 4 of 4. Depends on #391.

## Context

After `ReadFileTool` is migrated (#391), the remaining tools still use the old `isPathOutsideWorkspace` / `resolveIsOutsideWorkspace` pattern. These are lower urgency than read (write tools already gated by explicit user approval flows) but must be migrated to close #169 fully and to make `WorkspaceFileAccess` the only call site for boundary enforcement.

Tools to migrate in this PR:
- `WriteToFileTool`
- `EditTool`
- `EditFileTool`
- `SearchReplaceTool`
- `ApplyPatchTool`
- `ApplyDiffTool` ← **currently has no boundary check at all** (missed in PR #241); this migration closes that gap
- `ListFilesTool`
- `SearchFilesTool`
- `GenerateImageTool` (both input image read and output write paths)

Also: add the `allowSymlinksOutsideWorkspace` opt-in setting to the settings UI and i18n (deferred from earlier PRs). Wire the global setting through to `WorkspaceFileAccess` as the default when provider state is unavailable. This closes #246 as well.

## Developer Note

For each tool, the migration is the same pattern established in #391:
1. Replace `isPathOutsideWorkspace` / `resolveIsOutsideWorkspace` with `workspaceFileAccess.authorizeRead` or `authorizeWrite` as appropriate.
2. On `!access.ok`, surface `access.message` as a tool error and return before any `fs` operation.
3. Use `access.resolvedPath` for the actual filesystem call.
4. Hoist `getState()` once per `execute()` invocation and pass the `allowSymlinksOutsideWorkspace` flag down — do not fire a separate `getState()` per file in loops.

**`ApplyDiffTool` special note:** this tool currently has zero boundary check. Add `authorizeRead` before the file read in the `UpdateFile`/`DeleteFile` hunk handler, and `authorizeWrite` before the file write. The read-before-check race (content read before boundary check fires) identified in review is resolved naturally by using `access.resolvedPath` — the authorization must succeed before any `fs` call.

**`ListFilesTool` / `SearchFilesTool`:** `authorizeRead` on the root directory is sufficient for the tool-level check. The ripgrep `--follow` flag causing enumeration of out-of-workspace symlink targets is a separate issue and out of scope here.

**Settings UI (closes #246):**
[...]

Issue #391, opened 2026-05-29
feat(security): migrate ReadFileTool and webview readFileContent to WorkspaceFileAccess (#169)


edelauna:
Part of #169. Sub-issue 3 of 4. Depends on #390.

## Context

`ReadFileTool` is the highest-risk path for the symlink escape identified in #169: it reads arbitrary file content and returns it directly to the LLM. It is also the most complex tool to migrate (batch reads, line ranges, streaming partial display) so proving the pattern here gives high confidence before migrating write/edit tools.

The webview `readFileContent` handler in `webviewMessageHandler.ts` is included in this PR because it shares the same risk profile and is small enough to bundle without inflating scope.

PR #241 can be used as a reference for the call sites to migrate — it correctly identified all the `isPathOutsideWorkspace` usages in this tool. The difference here is that instead of calling `this.resolveIsOutsideWorkspace(...)` and then proceeding with raw `fs`, the tool calls `workspaceFileAccess.authorizeRead(...)` and uses `access.resolvedPath` for the actual read.

## Developer Note

Migrate `src/core/tools/ReadFileTool.ts`:
- Replace every call to `isPathOutsideWorkspace` / `resolveIsOutsideWorkspace` with `workspaceFileAccess.authorizeRead({ task, requestedPath, source: "read_file" })`.
- On `!access.ok`, surface `access.message` as a tool error and return early — before any `fs` read.
- Use `access.resolvedPath` (the canonicalized path) for the actual `fs.promises.readFile` call, not the raw input path.
- The batch approval path (`filesToApprove.length > 1`) should call `authorizeRead` per file inside the `Promise.all`, passing the pre-fetched provider state via an optional parameter to avoid N `getState()` calls.

Migrate `src/core/webview/webviewMessageHandler.ts` (the `readFileContent` message handler):
- Same pattern: `authorizeRead` before the `fs` read, early return with `{ content: null, error: access.message }` on failure.

Remove the `allowSymlinksOutsideWorkspace` field from `ExtensionState` / `globalSettingsSchema` **only if** no other tool still references it — it will likely need to stay until sub-issue 4 completes.

**Tests:**
- Normal in-workspace read: content returned correctly.
- Path outside workspace: tool error surfaced, no `fs` call made.
- Symlink inside workspace pointing outside: blocked (uses `WorkspaceFileAccess` real implementation, not a mock of the boundary check itself).
[...]

Issue #390, opened 2026-05-29
feat(security): introduce WorkspaceFileAccess — central authorization layer for all file operations


edelauna:
Part of #169. Sub-issue 2 of 4. Depends on #389.

## Context

Once `WorkspacePathResolver` (#389) provides safe async path canonicalization, we need a policy layer that tools can call instead of performing raw `fs` operations after a boolean check. The current pattern — tool calls `isPathOutsideWorkspace()`, gets a boolean, then proceeds with its own `fs` read/write — is structurally easy to get wrong: the check and the operation are decoupled, so one missed call site (like `ApplyDiffTool` in PR #241) leaves a hole.

The goal of this layer is to make the boundary check **structurally impossible to bypass**: tools request an authorized operation and get back either a resolved path they are permitted to use, or a structured error they must surface.

## Developer Note

Create `src/core/workspace/WorkspaceFileAccess.ts` (or equivalent location consistent with project conventions).

**`authorizeRead(options): Promise<AuthorizeResult>`**

```ts
type AuthorizeResult =
  | { ok: true; resolvedPath: string }
  | { ok: false; reason: "outside_workspace" | "symlink_escapes_workspace" | "realpath_failed" | "permission_denied"; message: string }
```

Options:
- `task: Task` — for reading `allowSymlinksOutsideWorkspace` from provider state (default `false` when state is unavailable — fail closed).
- `requestedPath: string` — absolute path as supplied by the tool.
- `source: string` — tool name, for error messages.

Internally:
1. Call `resolveRealPath` from #389. On error, return `{ ok: false, reason: "realpath_failed" | "permission_denied", message }`.
2. Compare resolved path against each workspace folder (also resolved via `resolveRealPath`). On error resolving a folder, skip it (cannot prove containment).
3. If outside all workspace folders and `allowSymlinksOutsideWorkspace` is `false`, return `{ ok: false, reason: … }`.
4. If `allowSymlinksOutsideWorkspace` is `true`, compare lexical paths (no realpath) — restoring pre-fix behavior for users who opt in.

Expose `authorizeWrite` with the same signature for write operations. (Both may share an internal `authorize` helper.)

Read `allowSymlinksOutsideWorkspace` from `task.providerRef.deref()?.getState()` with a try/catch — provider may have been torn down. Default to `false`.

**Tests** (`src/core/workspace/__tests__/WorkspaceFileAccess.spec.ts`):
[...]

PR #428, opened 2026-06-01
feat(security): introduce WorkspacePathResolver — async symlink-aware path canonicalization (#389)

### LINKED ISSUE #389: feat(security): introduce WorkspacePathResolver — safe async symlink-aware path canonicalization
Part of #169. Sub-issue 1 of 4.

## Context

Issue #169 identified that `isPathOutsideWorkspace()` only resolves `.`/`..` via `path.resolve`, so a symlink inside the workspace that points outside it is treated as inside and slips past the boundary check. PR #241 attempted a fix but spreads enforcement across many tool call sites — meaning every review becomes a hunt for the one path that was missed (`ApplyDiffTool` had no boundary check at all after the migration). The root cause is that raw `fs` operations and path resolution are interleaved throughout individual tools, with no central choke point.

This is the first of four sub-issues that introduce a central `WorkspaceFileAccess` layer to make the boundary check structurally impossible to forget.

## Developer Note

Create `src/utils/WorkspacePathResolver.ts` (or equivalent location consistent with project conventions). This module owns **only** path canonicalization — no policy, no tool logic.

**Requirements:**
- Async (`fs.promises.realpath` / `lstat`) — no synchronous filesystem calls; the extension host event loop must not be blocked.
- Walk up to the nearest existing ancestor on `ENOENT` and re-append the trailing segments, so not-yet-created files under a symlinked ancestor are still resolved correctly.
- Re-throw non-`ENOENT` errors (e.g. `EACCES`, `ELOOP`) so callers can fail closed — do not swallow and fall back to a lexical path.
- Resolve workspace folder paths too (they may themselves be reached via a symlink).
- Case-normalize the result on `darwin` and `win32` (lowercase) before returning, so comparisons against VS Code's `uri.fsPath` are reliable on case-insensitive filesystems.
- Export a single async function with a clear verb-noun name (e.g. `resolveRealPath(target: string): Promise<string>`).

**Tests** (`src/utils/__tests__/WorkspacePathResolver.spec.ts`):
- Symlink inside workspace pointing to a file outside → resolved path is outside.
- Symlink inside workspace pointing to a directory outside → resolved path is outside.
- Not-yet-created file under a symlinked ancestor → ancestor resolved, trailing segment re-appended.
- `EACCES` on symlink target → error re-thrown (not swallowed).
- `ELOOP` (circular symlink) → error re-thrown.
- No workspace folders → function still resolves correctly (policy is not its concern).
- Case normalization on macOS path with mixed case.

Use **real symlinks in a temp directory** for all tests — do not mock `fs`. Clean up in `afterEach`.

**Does not change any tool behavior.** No settings, no UI, no tool migrations in this PR.

proyectoauraorg:
Closes #389. First of four sub-issues under #169 toward a central `WorkspaceFileAccess` layer that makes the workspace boundary check structurally impossible to forget.

Adds `src/utils/WorkspacePathResolver.ts` — a single async `resolveRealPath(target)` that canonicalizes a path by following symlinks. It owns **only** path resolution: no workspace policy, no settings, no tool changes (those land in #390 / #391 / #392).

### Behavior
Carried over from the proven logic in #241, made async per the issue:

- **Async** `fs.promises.realpath` — never blocks the extension host event loop.
- **ENOENT walk-up**: resolves the nearest existing ancestor and re-appends the trailing segments, so a not-yet-created file under a symlinked ancestor still resolves correctly.
- **Fail-closed**: re-throws non-`ENOENT` errors (`EACCES`, `ELOOP`) so a caller performing a security check can fail closed instead of falling back to a lexical path (#169).
- **Case-normalizes** the result on macOS/Windows for reliable comparison against `uri.fsPath`.

### Tests
`src/utils/__tests__/WorkspacePathResolver.spec.ts` — real symlinks in a temp directory, no `fs` mocking:

- symlink → file outside, symlink → directory outside
- not-yet-created file under a symlinked ancestor (walk-up + re-append)
- `EACCES` re-thrown, `ELOOP` (circular symlink) re-thrown
- resolves without any workspace context (it owns no policy)
- case normalization on a case-insensitive platform

Symlink/permission cases skip cleanly on hosts that can't reproduce them (Windows without privileges, root).

Does not change any tool behavior. `tsc`, `eslint`, and the spec (7 tests) pass locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace path resolution to correctly follow symbolic links and prevent path-related issues across platforms.
  * Enhanced case normalization on macOS and Windows to ensure consistent path handling.
  * Added robust error handling for permission failures and circular symlink scenarios.

* **Tests**
  * Added end-to-end tests covering symlink resolution, non-existent-path behavior, permission errors, loop detection, and case-normalization.

### DIFF
--- a/src/utils/WorkspacePathResolver.ts
+++ b/src/utils/WorkspacePathResolver.ts
@@ -1,0 +1,66 @@
+import * as path from "path"
+import * as fs from "fs/promises"
+
+/** Narrow an unknown error to a Node errno exception with the given `code`. */
+function isErrnoException(err: unknown, code: string): boolean {
+	return err instanceof Error && (err as NodeJS.ErrnoException).code === code
+}
+
+// macOS APFS/HFS+ and Windows are case-insensitive: `realpath` can return a different case
+// than the one VS Code registered, so we lowercase the result before returning to keep later
+// comparisons (e.g. against `uri.fsPath`) reliable on those platforms only. Platform is read at
+// call time (not cached) so the behavior stays correct and testable.
+function normalizeCase(p: string): string {
+	const caseInsensitive = process.platform === "darwin" || process.platform === "win32"
+	return caseInsensitive ? p.toLowerCase() : p
+}
+
+/**
+ * Resolve a filesystem path to its canonical, symlink-followed form.
+ *
+ * This is the canonicalization primitive for the workspace boundary check (issue #169). It owns
+ * **only** path resolution — no workspace policy, no settings, no tool logic. The authorization
+ * decision (and the `allowSymlinksOutsideWorkspace` opt-in) lives in `WorkspaceFileAccess`.
+ *
+ * Behavior:
+ * - **Async only** (`fs.promises.realpath`); never blocks the extension host event loop.
+ * - If `target` does not exist yet (e.g. a file about to be created), the realpath of the nearest
+ *   existing ancestor is resolved and the remaining segments are re-appended, so a symlink
+ *   anywhere along the path is still followed while not-yet-created paths can still be evaluated.
+ * - Only `ENOENT` triggers the walk-up. Any other error (e.g. `EACCES`, `ELOOP`) is **re-thrown**
+ *   so a caller performing a security check can fail closed. Silently walking up would mask the
+ *   symlink and could make an out-of-workspace target look "inside" (#169).
+ * - The result is case-normalized on case-insensitive filesystems (macOS, Windows).
+ *
+ * Workspace folder paths should be resolved through this same function by callers, since a folder
+ * may itself be reached via a symlink.
+ */
+export async function resolveRealPath(target: string): Promise<string> {
+	let current = path.resolve(target)
+	const trailing: string[] = []
+
+	// Walk up until an existing path can be resolved, bounded by the filesystem root.
+	while (true) {
+		try {
+			const resolved = await fs.realpath(current)
+			const joined = trailing.length > 0 ? path.join(resolved, ...trailing.reverse()) : resolved
+			return normalizeCase(joined)
+		} catch (err) {
+			if (!isErrnoException(err, "ENOENT")) {
+				// Non-ENOENT (e.g. EACCES, ELOOP): propagate so the caller's security check can
+				// fail closed instead of falling through to the lexical path.
+				throw err
+			}
+
+			const parent = path.dirname(current)
+			if (parent === current) {
+				// Reached the filesystem root without finding an existing path; fall back to the
+				// lexically resolved path (still case-normalized for consistent comparisons).
+				return normalizeCase(path.resolve(target))
+			}
+
+			trailing.push(path.basename(current))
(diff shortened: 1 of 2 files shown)


src/utils/WorkspacePathResolver.ts
```diff
@@ -0,0 +1,66 @@
...
+	// Walk up until an existing path can be resolved, bounded by the filesystem root.
+	while (true) {
+		try {
+			const resolved = await fs.realpath(current)
+			const joined = trailing.length > 0 ? path.join(resolved, ...trailing.reverse()) : resolved
+			return normalizeCase(joined)
+		} catch (err) {
+			if (!isErrnoException(err, "ENOENT")) {
+				// Non-ENOENT (e.g. EACCES, ELOOP): propagate so the caller's security check can
+				// fail closed instead of falling through to the lexical path.
```
edelauna: `realpath` throws `ENOTDIR` (not `ENOENT`) when a path component is a regular file — e.g. `realpath("/workspace/readme.txt/child")`. Should the #390 caller expect to handle `ENOTDIR` alongside `EACCES`/`ELOOP`, or should the walk-up also treat it as "not found"?

```suggestion
				// Non-ENOENT (e.g. EACCES, ELOOP, ENOTDIR): propagate so the caller's
				// security check can fail closed instead of falling through to the lexical path.
```

src/utils/WorkspacePathResolver.ts
```diff
@@ -0,0 +1,66 @@
...
+ * - If `target` does not exist yet (e.g. a file about to be created), the realpath of the nearest
+ *   existing ancestor is resolved and the remaining segments are re-appended, so a symlink
+ *   anywhere along the path is still followed while not-yet-created paths can still be evaluated.
+ * - Only `ENOENT` triggers the walk-up. Any other error (e.g. `EACCES`, `ELOOP`) is **re-thrown**
+ *   so a caller performing a security check can fail closed. Silently walking up would mask the
+ *   symlink and could make an out-of-workspace target look "inside" (#169).
+ * - The result is case-normalized on case-insensitive filesystems (macOS, Windows).
+ *
+ * Workspace folder paths should be resolved through this same function by callers, since a folder
+ * may itself be reached via a symlink.
```
edelauna: `arePathsEqual()` in `path.ts` only case-folds on `win32`, not `darwin` — if #390 compares a `resolveRealPath()` result against a raw `folder.uri.fsPath`, the mismatch could cause false negatives on macOS.

```suggestion
 * Workspace folder paths should be resolved through this same function by callers, since a folder
 * may itself be reached via a symlink. Callers should always compare two `resolveRealPath()` results
 * rather than mixing with raw `uri.fsPath` — `arePathsEqual()` does not case-fold on macOS.
```

edelauna (commented):
🔥 - nice ty for breaking this up!

edelauna (approved):

PR #437, opened 2026-06-01
fix(opencode-go): fetch models unconditionally — the /models endpoint is public

proyectoauraorg:
## Bug

Reported for the Opencode Go provider (added in #172): **the model list never loads** — the picker is empty, a custom model name can't be entered, and the provider falls back to the hardcoded default `glm-5.1`.

## Root cause

In `requestRouterModels` (`webviewMessageHandler.ts`), the Opencode Go fetch was gated behind `if (opencodeGoApiKey)`, with the comment *"its /models endpoint requires auth."* **That premise is wrong.** When `opencodeGoApiKey` isn't in `apiConfiguration` at fetch time (key just typed, stored in Secret Storage and not yet hydrated, or absent), Opencode Go is never added to the fetch candidates, so `routerModels["opencode-go"]` stays `{}` (the seeded default). The `ModelPicker` then renders an empty list and falls back to `opencodeGoDefaultModelId` (`glm-5.1`); since it only offers fetched router models, no custom model can be selected either.

`openrouter` and `vercel-ai-gateway` — also public endpoints — are fetched **unconditionally** in the same handler. Opencode Go should be too.

## Reproduction (deterministic)

Running the **shipped** `getOpencodeGoModels()` against the live endpoint:

```
[NO API KEY]   models loaded: 17
               sample ids: minimax-m3, minimax-m2.7, minimax-m2.5, kimi-k2.6, kimi-k2.5, glm-5.1
[DUMMY KEY]    models loaded: 17
```

```console
$ curl -s -o /dev/null -w "%{http_code}" https://opencode.ai/zen/go/v1/models
200
```

The endpoint returns the full 17-model list with **no Authorization header**, and the existing fetcher parses it correctly. So the fetcher was never the problem — the gate was.

## Fix

Fetch Opencode Go unconditionally, like the other public routers, forwarding the API key when present and still flushing the cache when a new key is supplied (Refresh Models button). One-line behavioral change in the handler + the misleading comment corrected.

## Tests

- Updated the three `requestRouterModels` tests that assumed `"opencode-go": {}` (it's now fetched like the other public routers).
- Added a regression test: **Opencode Go is fetched with `apiKey: undefined` and populates `routerModels` even when no key is configured.**
- ✅ 29/29 in `webviewMessageHandler.spec.ts`, `tsc --noEmit` clean, `eslint --max-warnings=0` clean.

> Note: this fixes the empty picker for all users by loading models without a key. If a region/network blocks `opencode.ai` for a specific user, the fetcher still swallows that error and returns `{}` (logged only) — surfacing that to the UI is a separate, optional follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added an Opencode-Go “refresh models” action in provider settings, including loading, success, and error feedback.
* **Bug Fixes**
  * Opencode-Go router models are now returned for public (no-auth) model fetches even when no API key is configured; the cache refresh happens only when a new key is explicitly provided.
* **Tests**
  * Expanded message-handling and UI test coverage to verify Opencode-Go behavior across missing-key, refresh, and mixed provider outcomes.


proyectoauraorg (commented):
## Code Review: PR #437 — fix(opencode-go): fetch models unconditionally

### Summary

This PR fixes a real and reproducible bug where the Opencode Go model picker stayed empty because the `/models` fetch was gated behind an API key that wasn't always present in `apiConfiguration` at fetch time. The fix is well-scoped, the root cause analysis in the PR description is thorough, and the approach of aligning Opencode Go with the other public routers (`openrouter`, `vercel-ai-gateway`) is the correct one.

### What works well

- **Correct root cause identification.** The PR description clearly demonstrates with `curl` and test output that the `/models` endpoint is public, making the gate unnecessary.
- **Minimal, focused change.** Only 3 files modified, +54/-16 lines. The behavioral change is a single logical operation (removing the `if (opencodeGoApiKey)` gate), which reduces blast radius.
- **Good comment update.** The misleading comment ("its /models endpoint requires auth") is replaced with a detailed explanation of *why* the unconditional fetch is correct and what the previous behavior caused.
- **Cache flush logic preserved.** The `flushModels` call is correctly kept conditional on `message?.values?.opencodeGoApiKey`, so the cache is only invalidated when the user explicitly provides a new key (Refresh Models button). This avoids unnecessary re-fetches.
- **Test coverage is solid.** The regression test in `webviewMessageHandler.spec.ts` explicitly verifies the keyless path (`apiKey: undefined`), and all three existing test suites are updated consistently.

### Issues found

#### Minor — Block-scoping for `const opencodeGoApiKey` in switch case

**File:** `src/core/webview/webviewMessageHandler.ts` line 1021

CodeRabbit already flagged this in review #4404026771 but it has **not been addressed** in the second commit. The `const opencodeGoApiKey` declaration inside the switch case is function-scoped per JavaScript semantics. While it won't cause a runtime error (no other case declares a variable with the same name), it violates the `noSwitchDeclarations` lint rule and is inconsistent with the defensive pattern the other providers use.

The existing providers (`litellm`, `poe`, `deepseek`) avoid this by wrapping their logic in `if` blocks, which implicitly create block scope. Since Opencode Go's unconditional fetch removes the `if` gate, the declaration now sits directly in the switch case body.

**Suggested fix:** Wrap the Opencode Go block in curly braces to create explicit block scope:

```typescript
{
    const opencodeGoApiKey = message?.values?.opencodeGoApiKey ?? apiConfiguration.opencodeGoApiKey

    // Refresh the cache when a new key is explicitly provided (e.g. the Refresh Models button).
    if (message?.values?.opencodeGoApiKey) {
        await flushModels({ provider: "opencode-go", apiKey: opencodeGoApiKey }, true)
    }

    candidates.push({
        key: "opencode-go",
        options: { provider: "opencode-go", apiKey: opencodeGoApiKey },
    })
}
```

#### Suggestion — Use exact assertion in `ClineProvider.spec.ts` line 2469

**File:** `src/core/webview/__tests__/ClineProvider.spec.ts` line 2469

The assertion uses `expect.objectContaining({ provider: "opencode-go" })` while the equivalent assertions for other providers in the same test block use exact object matches. Since the implementation always passes `apiKey` (even when `undefined`), the assertion should be exact for consistency and to catch potential regressions where extra unexpected properties are introduced:

```typescript
// Instead of:
expect(getModels).toHaveBeenCalledWith(expect.objectContaining({ provider: "opencode-go" }))

// Use:
expect(getModels).toHaveBeenCalledWith({ provider: "opencode-go", apiKey: undefined })
```

This also matches the pattern used in the regression test at `webviewMessageHandler.spec.ts`.

### CI Status

All checks pass except **codecov/patch**, which reports a coverage delta. This is expected for a behavioral change that shifts code from "not reached" to "reached" — the new regression test should cover the keyless path. Worth verifying the uncovered lines are the `flushModels` conditional (which is covered by the existing "Refresh Models" test path).

### Edge cases verified

- **No key configured:** Fetch proceeds with `apiKey: undefined` — correct for a public endpoint.
- **Empty string key:** The `??` operator only catches `null`/`undefined`, so an empty string would be passed as-is. This is harmless since the endpoint ignores the header.
- **Network failure:** The `safeGetModels` wrapper catches errors and returns `{}`, so the UI gracefully falls back to the default model. Documented in the PR description as a separate follow-up concern.
- **Performance:** The additional fetch runs in parallel via `Promise.allSettled` with all other providers — no sequential bottleneck introduced.

### Verdict

**Approve with minor comments.** The core fix is correct, well-tested, and low-risk. The two issues above are non-blocking but should be addressed before merge for lint compliance and test consistency.

---
*Reviewed by: DevOps Orquestador Maestro — zSys*

edelauna:
Need to bump up coverage.

navedmerchant:
Skipping minor coderabbit comments and approving

navedmerchant (approved):

PR #449, opened 2026-06-02
feat(litellm): handle reasoning_content and reasoning fields in streaming

### LINKED ISSUE #447: [BUG] LiteLLM provider does not handle reasoning fields
**Describe the bug**
LiteLLMHandler does not process reasoning_content or reasoning fields from the streaming delta. When using
reasoning models (e.g. DeepSeek-R1, QwQ) routed through a LiteLLM proxy, the thinking/reasoning output is
silently dropped and never displayed to the user.

**To Reproduce**
1. Set up a LiteLLM proxy server pointing to a reasoning model (e.g. DeepSeek-R1 or QwQ-32B)
2. Configure Zoo Code to use the LiteLLM provider with that model
3. Send any message and observe the response

**Expected behavior**
The reasoning/thinking content should be displayed in the UI before the final response, as it is when using
DeepSeek or other reasoning-capable providers directly.

**Screenshots**
N/A

**Video**
N/A

3.56.0

**Additional context**
BaseOpenAiCompatibleProvider (used by DeepSeek, Requesty, Unbound, etc.) already handles this correctly by
checking delta.reasoning_content and delta.reasoning in the stream loop. However, LiteLLMHandler overrides
createMessage independently and does not include this logic, causing the field to be silently ignored.

daewoongoh:
### Related GitHub Issue

Closes: #447

### Description

`LiteLLMHandler.createMessage` overrides the streaming loop independently from `BaseOpenAiCompatibleProvider`, but was missing the logic to extract `reasoning_content` / `reasoning` fields from stream deltas. As a result, thinking output from reasoning models (e.g. DeepSeek-R1, QwQ-32B) routed through a LiteLLM proxy was silently dropped and never surfaced in the UI.

**How it was fixed:**
In the streaming loop of `lite-llm.ts`, added the same reasoning extraction pattern already present in  `BaseOpenAiCompatibleProvider`:
- Check `delta.reasoning_content` first, fall back to `delta.reasoning`
- Skip missing, non-string, or empty-string values
- Preserve whitespace-only chunks so streamed word/paragraph boundaries remain intact
- Yield a `{ type: "reasoning", text }` chunk when a non-empty value is found

No changes to types, configuration, or other providers were needed.

### Test Procedure

**Unit tests** (`src/api/providers/__tests__/lite-llm.spec.ts`):

Added a `"reasoning field handling"` describe block with 4 cases:
1. `reasoning_content` delta yields a `reasoning` chunk
2. `reasoning` delta yields a `reasoning` chunk (fallback)
3. When both fields are present, `reasoning_content` takes precedence (only one chunk emitted)
4. Empty string does not yield a chunk, while whitespace-only chunks are preserved

Run locally with:
```bash
node_modules/.bin/vitest run --root . \
  api/providers/__tests__/lite-llm.spec.ts \
  api/providers/__tests__/base-openai-compatible-provider.spec.ts \
  api/providers/utils/__tests__/extract-reasoning.spec.ts
```

### Pre-Submission Checklist

### Screenshots / Videos
N/A

### Documentation Updates

### Additional Notes
Implementation now adds a shared `extractReasoningFromDelta` helper and uses it from both LiteLLMHandler and BaseOpenAiCompatibleProvider so fallback and whitespace handling stay consistent across providers.

### Get in Touch
hehegwk_23849

## Summary by CodeRabbit

* **New Features**
  * Streaming now emits distinct "reasoning" chunks when the model provides non-blank reasoning; extraction logic centralized and applied across providers.
* **Tests**
  * Added tests covering reasoning extraction and streaming: prefer reasoning_content, fall back to reasoning when reasoning_content is null, suppress whitespace/empty values, and ensure single reasoning emission per delta.


daewoongoh:
Hi @edelauna, sorry for the direct ping.

I opened this PR for #449  and have followed the contribution process, including posting in Discord. 
Could you please take a look when you have a chance?

I’d especially appreciate feedback on whether the fallback handling for `reasoning_content` / `reasoning` is implemented in the expected way.

Thanks!

src/api/providers/lite-llm.ts
```diff
@@ -235,6 +235,18 @@ export class LiteLLMHandler extends RouterProvider implements SingleCompletionHa
...
+					for (const key of ["reasoning_content", "reasoning"] as const) {
+						if (key in delta) {
+							const reasoningText = ((delta as any)[key] as string | undefined) || ""
+							if (reasoningText?.trim()) {
+								yield { type: "reasoning", text: reasoningText }
+							}
+							break
+						}
+					}
+				}
```
edelauna: This block is an exact copy of `base-openai-compatible-provider.ts:150-160`. If the break/fallback bug gets fixed in one place, the other will silently stay broken. Would it make sense to extract a shared helper (something like `extractReasoningFromDelta(delta)`) so both providers stay in sync automatically?

src/api/providers/lite-llm.ts
```diff
@@ -235,6 +235,18 @@ export class LiteLLMHandler extends RouterProvider implements SingleCompletionHa
 					yield { type: "text", text: delta.content }
 				}
 
+				if (delta) {
+					for (const key of ["reasoning_content", "reasoning"] as const) {
```
edelauna: I noticed `openai.ts`, `requesty.ts`, `unbound.ts`, and `opencode-go.ts` all only check `reasoning_content` — they don't have the `reasoning` fallback this adds. Was that a deliberate scope choice for this PR, or worth a follow-up to bring those in line too?

src/api/providers/lite-llm.ts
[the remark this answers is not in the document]
edelauna: Could you address this feedback, or alternatively provide rational for resolving it?

daewoongoh:
@coderabbitai review

src/api/providers/lite-llm.ts
daewoongoh: I agreed with the helper suggestion and updated the PR. `lite-llm.ts` and `base-openai-compatible-provider.ts` now both use a shared `extractReasoningFromDelta(delta)` helper, so the fallback behavior stays in sync.

The helper also fixes the fallback bug: `reasoning_content: null`, non-string, or `""` now falls through to `reasoning`, and I added tests for that case. I also preserved whitespace-only chunks like `" "` and `"\n\n"` because streamed reasoning deltas can use those as word/paragraph boundaries.

src/api/providers/lite-llm.ts
daewoongoh: For `openai.ts`, `requesty.ts`, `unbound.ts`, and `opencode-go.ts`, I intentionally kept them out of this PR’s scope. This PR is focused on the LiteLLM issue from #447, with the base provider included only because it had the same duplicated logic. I agree those standalone providers should probably be aligned with the helper too, but I’d prefer to handle that in a follow-up PR to keep this change focused.

daewoongoh:
@edelauna

I’ve addressed the feedback from the review thread and updated the implementation accordingly.

This PR originally focused on fixing the LiteLLM streaming path, where `reasoning_content` / `reasoning` deltas were not being surfaced as reasoning chunks. During review, we also found that the reasoning extraction logic was duplicated between `lite-llm.ts` and `BaseOpenAiCompatibleProvider`, and that the previous loop/break pattern could prevent fallback to `reasoning` when `reasoning_content` was present but `null`, non-string, or an empty string.

To keep the behavior consistent across both paths, I extracted the shared logic into `extractReasoningFromDelta(delta)` and updated both `LiteLLMHandler` and `BaseOpenAiCompatibleProvider` to use it. The helper now:

- prefers `reasoning_content` when it contains a non-empty string
- falls back to `reasoning` when `reasoning_content` is missing, non-string, `null`, or `""`
- preserves whitespace-only streamed chunks such as `" "` and `"\n\n"` so word and paragraph boundaries are not collapsed when reasoning chunks are concatenated downstream

I also updated and added tests for the helper itself, LiteLLM streaming behavior, the null/empty fallback case, and whitespace-preserving behavior in the base OpenAI-compatible provider tests.

For `openai.ts`, `requesty.ts`, `unbound.ts`, `opencode-go.ts`, and any other providers that currently use similar `reasoning_content` / `reasoning` handling, I’m intentionally keeping those out of this PR’s scope so this change remains focused on the LiteLLM issue and the directly duplicated base-provider logic. After this PR, I’ll create a follow-up issue to align those providers with the shared helper as a separate cleanup/improvement task.

edelauna (approved):
Thanks for adding this, created issue #570 to track adding this to the other providers.

Issue #452, opened 2026-06-02
[BUG] Edit Unsuccessful, ??% similar, needs 100%


DannyVarodBlueVine:
**Describe the bug**

I often get "Edit Unsuccessful" (even for 1st prompt in session) and in error_details I see:

No sufficiently similar match found (??% similar, needs 100%)
E.g. 53%, 97%, ...

**To Reproduce**
Could not find a pattern for reproduction, but when this error occurs it continues even on retries.

**Expected behavior**
No issues editing files.

**Screenshots**

**Video**

**What version of zoo are you running**
3.56.0
also in previous versions including in roo

**Additional context**

Sometime these inputs help (not always):
- if edit fails, re-read target file before executing replace tool and update the source text and lines you are trying to replace with what you see in the file
- try smaller edits

nigeldelviero:
Claiming

edelauna:
@nigeldelviero  what’s your plan to address this? I think it might be worth adding back the setting to allow edits with out 100% similarity when auto approve is off

nigeldelviero:
> what’s your plan to address this? I think it might be worth adding back the setting to allow edits with out 100% similarity when auto approve is off

Hi! Yes, that is exactly what I've addressed:
- I have added back a configurable diffFuzzyThreshold setting (ranging from 0.5 to 1.0) with a relaxed default of 0.90 (90% similarity).
- Built a "Diff Match Threshold" slider in the Settings UI (under Context Management) so users can adjust it themselves or lock it back to 1.0 (100% strict match) if they prefer absolute precision.
- When an edit fails to match, the system now logs detailed diagnostics (Levenshtein distance, exact character length differences, and a side-by-side comparison of the search block vs the best matching segment) so the user (or the AI on its next retry) has clear context on why it failed.

Regarding your point on auto-approve, do you want us to enforce/override this threshold to 1.0 (strict) specifically when auto-approve is enabled to prevent silent incorrect edits, or is letting the user control it globally via settings preferred?

andy7d0:
why this bug exists at all? we need just replace text with text and both of them known

nigeldelviero:
> why this bug exists at all? we need just replace text with text and both of them known

The issue exists because the model’s search block is not always an exact byte-for-byte representation of the current file contents.

[...]

PR #470, opened 2026-06-04
fix(edit-unsuccessfull): introduce configurable relaxed diff thresholds and diagnostics

nigeldelviero:
## Problem

Issue: #452 

When file edits are performed, the system previously enforced a strict similarity threshold of 1.0 (exact match) in `MultiSearchReplaceDiffStrategy`. This caused frequent "Edit Unsuccessful" errors even for minor formatting, whitespace, or line-ending discrepancies. 

Additionally, when a match failed, the error feedback was generic (`No sufficiently similar match found`), making it difficult for the AI or the user to diagnose why the edit was rejected.

## Root Cause

The core issue stems from strict matching logic in `MultiSearchReplaceDiffStrategy` which requires search block similarity to be exactly 1.0. This does not tolerate slight formatting changes introduced by pre-commit hooks, formatting extension rules, or AI-generated output variations.

## Solution

1. **Configurable Threshold**: Introduced a configurable `diffFuzzyThreshold` (number, default: 0.90) to allow a relaxed matching similarity.
2. **Settings Webview Integration**: Added a "Diff Match Threshold" slider in the `ContextManagementSettings.tsx` UI under the write-delay settings, with range 50% - 100%.
3. **Core Strategy & Tasks Updates**: 
   - Passed the user's configured `diffFuzzyThreshold` setting from `ExtensionState` to the `Task` and `MultiSearchReplaceDiffStrategy` instance.
   - Enhanced `applyDiff` to report detailed Levenshtein-based diagnostics (Levenshtein distance, character lengths, search range, and similarity scores) when matches fall below the threshold.
4. **Type-safety**: Updated `ExtensionState` definition and initialized the default value (`0.90`) in both `globalSettingsSchema` and `ExtensionStateContextProvider` to prevent TypeScript compilation errors.
5. **Regression Tests**: Added unit tests to `multi-search-replace.spec.ts` to assert default thresholds, success with minor whitespace variations, and the structure of detailed debug outputs when edits fail.

## Verification

All `multi-search-replace` tests pass successfully:
- 1 test files passed | 75 tests passed 
- Successfully verified default 90% threshold 
- Verified diagnostics formatting and distance calculation 
- No compiler issues: pnpm check-types exits with code 0

## Screenshot
image

## Summary by CodeRabbit

* **New Features**
  * Added a configurable fuzzy-match threshold (slider 50%–100%, step 1%, displayed as a percentage, default 90%); value is exposed in settings/webview state and applied to diff tasks (clamped to 50–100%).
  * Failure diagnostics enhanced to include similarity metrics (score, required threshold, Levenshtein distance, search/match lengths).

* **Tests**
  * Added tests validating fuzzy-threshold behavior and diagnostic output.


src/core/diff/strategies/multi-search-replace.ts
```diff
@@ -555,9 +556,13 @@ export class MultiSearchReplaceDiffStrategy implements DiffStrategy {
...
+						error: `No sufficiently similar match found${lineRange} (${Math.floor(bestMatchScore * 100)}% similar, needs ${Math.floor(this.fuzzyThreshold * 100)}%)\n\nDebug Info:\n- Similarity Score: ${Math.floor(bestMatchScore * 100)}%\n- Required Threshold: ${Math.floor(this.fuzzyThreshold * 100)}%\n- Search Range: ${startLine ? `starting at line ${startLine}` : "start to end"}\n- Levenshtein Distance: ${levenDist >= 0 ? `${levenDist} characters` : "N/A"}\n- Search Length: ${searchChunk.length} characters\n- Best Match Length: ${bestMatchContent ? bestMatchContent.length : 0} characters\n- Tried both standard and aggressive line number stripping\n- Tip: Use the read_file tool to get the latest content of the file before attempting to use the apply_diff tool again, as the file content may have changed\n\nSearch Content:\n${searchChunk}${bestMatchSection}${originalContentSection}`,
```
edelauna, line 580: When `startLine` is not set (line 551 branch), `originalContentSection` becomes the entire file content. This ends up in `pushToolResult` and gets sent verbatim to the external LLM API — if the file contains secrets or credentials, a failed diff match would transmit them to a third party. Could we cap this to a bounded window around `bestMatchIndex` even in the no-`startLine` case?

src/core/diff/strategies/multi-search-replace.ts
```diff
@@ -81,10 +81,11 @@ export class MultiSearchReplaceDiffStrategy implements DiffStrategy {
...
 	constructor(fuzzyThreshold?: number, bufferLines?: number) {
-		// Use provided threshold or default to exact matching (1.0)
-		// Note: fuzzyThreshold is inverted in UI (0% = 1.0, 10% = 0.9)
-		// so we use it directly here
-		this.fuzzyThreshold = fuzzyThreshold ?? 1.0
+		// Use provided threshold or default to relaxed matching (0.9)
+		// A value of 0.9 means 90% similarity is required for a match.
+		// This was previously 1.0 (exact match) which caused frequent
+		// "Edit Unsuccessful" errors on minor whitespace differences.
+		this.fuzzyThreshold = fuzzyThreshold ?? DEFAULT_DIFF_FUZZY_THRESHOLD
```
edelauna: The Zod schema enforces `min(0.5)` but the `updateSettings` webview handler stores values via `contextProxy.setValue` without calling `.parse()`, so an out-of-range value (e.g. `0`) can reach here. Worth adding a clamp as defence-in-depth?

```suggestion
		this.fuzzyThreshold = Math.max(0.5, Math.min(1, fuzzyThreshold ?? DEFAULT_DIFF_FUZZY_THRESHOLD))
```

src/core/diff/strategies/__tests__/multi-search-replace.spec.ts
```diff
@@ -1374,4 +1374,97 @@ function sum(a, b) {
...
 		})
 	})
+
+	describe("fuzzyThreshold and diagnostics", () => {
+		const originalContent =
+			"function calculateTotal(price: number, tax: number) {\n\tconst subtotal = price;\n\treturn subtotal + tax;\n}\n"
+
+		it("should use default 0.90 threshold when instantiated without arguments", () => {
+			const strategy = new MultiSearchReplaceDiffStrategy()
+			expect(strategy["fuzzyThreshold"]).toBe(0.9)
```
edelauna: This accesses a private field directly. The next test (`should succeed with near-miss match...`) already proves the default threshold is applied through the public `applyDiff` API — does this assertion add coverage that the behaviour test doesn't already give you?

packages/types/src/vscode-extension-host.ts
```diff
@@ -318,6 +319,7 @@ export type ExtensionState = Pick<
 	taskHistory: HistoryItem[]
 
 	writeDelayMs: number
+	diffFuzzyThreshold?: number
```
edelauna: `diffFuzzyThreshold` now appears both in the `Pick<>` union above (line 307) and here as an explicit field. `writeDelayMs` (line 321) is the established pattern: it only appears in the explicit block, declared as required `number` since `getState()` always coalesces it via `??`. Should this follow the same pattern — remove from `Pick` and use `diffFuzzyThreshold: number` here?

webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -209,6 +209,7 @@ export const ExtensionStateContextProvider: React.FC<{ children: React.ReactNode
 		checkpointTimeout: DEFAULT_CHECKPOINT_TIMEOUT_SECONDS, // Default to 15 seconds
 		language: "en", // Default language code
 		writeDelayMs: 1000,
+		diffFuzzyThreshold: 0.9,
```
edelauna: `DEFAULT_CHECKPOINT_TIMEOUT_SECONDS` is imported and used two lines up for exactly this pattern. Could we import `DEFAULT_DIFF_FUZZY_THRESHOLD` from `@roo-code/types` here instead of the magic literal `0.9`, so the default stays in sync if it ever changes?

webview-ui/src/components/settings/ContextManagementSettings.tsx
```diff
@@ -406,6 +410,31 @@ export const ContextManagementSettings = ({
 					</div>
 				</SearchableSetting>
 
+				<SearchableSetting
```
edelauna, line 413: Is the diff match threshold a diagnostics setting? The other settings in this section (`includeDiagnosticMessages`, `maxDiagnosticMessages`, `writeDelayMs`) all relate to IDE linter/problem feedback. This one controls the patch engine — would it make more sense in a dedicated "File Edits" or `diffStrategy` subsection?

packages/types/src/global-settings.ts
```diff
@@ -102,6 +110,12 @@ export const globalSettingsSchema = z.object({
 	alwaysAllowWriteOutsideWorkspace: z.boolean().optional(),
 	alwaysAllowWriteProtected: z.boolean().optional(),
 	writeDelayMs: z.number().min(0).optional(),
+	/**
+	 * Fuzzy matching threshold for the multi-search-replace diff strategy.
+	 * Range: 0.5 (accept anything) to 1.0 (exact match only).
```
edelauna: "0.5 (accept anything)" — a threshold of 0.5 means 50% minimum similarity, not unconditional acceptance (that would be 0.0). Worth clarifying, e.g. "0.5 (50% minimum similarity)"?

packages/types/src/global-settings.ts
```diff
@@ -23,6 +22,15 @@ import { languagesSchema } from "./vscode.js"
...
 export const DEFAULT_WRITE_DELAY_MS = 1000
 
+/**
+ * Default fuzzy matching threshold for the multi-search-replace diff strategy.
+ * A value of 0.90 means the search content must be at least 90% similar to the
+ * file content for a match to succeed. Lowered from the previous strict 1.0 (exact
+ * match) to reduce "Edit Unsuccessful" errors caused by minor whitespace or
+ * formatting differences.
+ */
+export const DEFAULT_DIFF_FUZZY_THRESHOLD = 0.9
```
edelauna: lets default to 1 - since I think a lot of user's might have auto write on.

edelauna (commented):
🔥 - thank you for taking this on.

Had a couple comments related to my preference for how to implement this. 

Also lets add in the i18n locales so that ci passes.

src/core/diff/strategies/multi-search-replace.ts
nigeldelviero: Great catch, thank you! You're absolutely right, when `startLine` is not provided, the 
`originalContentSection` currently dumps the entire file into the error message, which then 
gets sent to the LLM API via `pushToolResult`. This is a real security concern.

I'll cap the no-`startLine` path to a bounded window around `bestMatchIndex` (using the 
existing `bufferLines` constant), consistent with what the `startLine`-present path already 
does. If there's no `bestMatchIndex` either, I'll omit the original content section entirely 
to avoid any risk of leaking file contents.

Here's the fix I'll apply:

```diff
 const originalContentSection =
     startLine !== undefined && endLine !== undefined
         ? `\n\nOriginal Content:\n${addLineNumbers(
                 resultLines
                     .slice(
                         Math.max(0, startLine - 1 - this.bufferLines),
                         Math.min(resultLines.length, endLine + this.bufferLines),
                     )
                     .join("\n"),
                 Math.max(1, startLine - this.bufferLines),
             )}`
-        : `\n\nOriginal Content:\n${addLineNumbers(resultLines.join("\n"))}`
+        : matchIndex !== -1
+            ? `\n\nOriginal Content (around best match):\n${addLineNumbers(
+                    resultLines
+                        .slice(
+                            Math.max(0, matchIndex - this.bufferLines),
+                            Math.min(resultLines.length, matchIndex + searchLines.length + this.bufferLines),
+                        )
+                        .join("\n"),
+                    Math.max(1, matchIndex + 1 - this.bufferLines),
+                )}`
+            : ""

File: src/core/diff/strategies/__tests__/multi-search-replace.spec.ts (hunk 3 above)
Comment (nigeldelviero) [reply src/core/diff/strategies/__tests__/multi-search-replace.spec.ts]: You're absolutely right. Accessing the private property directly is unnecessary here and goes against best practices. The behavior is already sufficiently covered by the functional tests using the public applyDiff API. I will remove this specific test case

File: packages/types/src/vscode-extension-host.ts (hunk 4 above)
Comment (nigeldelviero) [reply packages/types/src/vscode-extension-host.ts]: Yes, that makes total sense. Since getState() guarantees it will always be coalesced with DEFAULT_DIFF_FUZZY_THRESHOLD, it should indeed be a required number and removed from the Pick union to stay consistent with writeDelayMs. I will apply this cleanup

File: webview-ui/src/components/settings/ContextManagementSettings.tsx (hunk 6 above)
Comment (nigeldelviero) [reply webview-ui/src/components/settings/ContextManagementSettings.tsx:413]: You're absolutely right, it doesn't belong under diagnostics. That section is specifically about IDE linting/problem feedback, while the fuzzy threshold controls the patch engine behavior.

I'll move it into a new dedicated subsection called "File Edits" that sits between diagnostics and condensingThreshold. This keeps related settings grouped together and leaves room for future diff/patch-related settings.

File: packages/types/src/global-settings.ts (hunk 7 above)
Comment (nigeldelviero) [reply packages/types/src/global-settings.ts]: You're absolutely right. Saying "accept anything" for 0.5 is technically inaccurate and misleading, as it still requires a 50% match. 
I'll update the JSDoc comment to "0.5 (50% minimum similarity)" to accurately reflect that this is the lower bound for fuzzy matching, not an unconditional acceptance.

File: packages/types/src/global-settings.ts (hunk 8 above)
Comment (nigeldelviero) [reply packages/types/src/global-settings.ts]: That's a great point. With auto-write enabled, users won't have a chance to review potentially incorrect fuzzy matches before they're applied. Defaulting to 1.0 provides a safer out-of-the-box experience, and users who encounter frequent whitespace-related failures can manually lower the threshold.

I'll change `DEFAULT_DIFF_FUZZY_THRESHOLD` back to 1.0 and update the JSDoc comment to reflect that this is the conservative default for safety with auto-approval workflows.

File: src/core/diff/strategies/multi-search-replace.ts [hunk 9]
```diff
@@ -555,9 +558,13 @@ export class MultiSearchReplaceDiffStrategy implements DiffStrategy {
...
+						error: `No sufficiently similar match found${lineRange} (${Math.floor(bestMatchScore * 100)}% similar, needs ${Math.floor(this.fuzzyThreshold * 100)}%)\n\nDebug Info:\n- Similarity Score: ${Math.floor(bestMatchScore * 100)}%\n- Required Threshold: ${Math.floor(this.fuzzyThreshold * 100)}%\n- Search Range: ${startLine ? `starting at line ${startLine}` : "start to end"}\n- Levenshtein Distance: ${levenDist >= 0 ? `${levenDist} characters` : "N/A"}\n- Search Length: ${searchChunk.length} characters\n- Best Match Length: ${bestMatchContent ? bestMatchContent.length : 0} characters\n- Tried both standard and aggressive line number stripping\n- Tip: Use the read_file tool to get the latest content of the file before attempting to use the apply_diff tool again, as the file content may have changed\n\nSearch Content:\n${searchChunk}${bestMatchSection}${originalContentSection}`,
```
edelauna, line 580: The `startLine`-absent branch (the ternary's else arm, two lines above) still appends the full file via `resultLines.join("\n")`. Could we cap that to a bounded window around `bestMatchIndex` the same way the `startLine`-present branch slices with `bufferLines`? Without it, a failed match on a large file sends the entire content (potentially including secrets) to the LLM API via `pushToolResult`.

src/core/diff/strategies/__tests__/multi-search-replace.spec.ts
```diff
@@ -1374,4 +1374,90 @@ function sum(a, b) {
...
+				"=======\n" +
+				"function calculateTotal(price: number, tax: number) {\n" +
+				"    const subtotal = price;\n" +
+				"    return (subtotal + tax) * 1.1;\n" +
+				"}\n" +
+				">>>>>>> REPLACE"
+
+			const result = await strategy.applyDiff(originalContent, diff)
+			expect(result.success).toBe(false)
+			if (!result.success) {
```
edelauna: The assertions inside `if (!result.success)` are silently skipped if the branch is never entered — the test would still pass green. Could these be moved outside the guard so a missing `failParts` is a hard failure?

```suggestion
			expect(result.success).toBe(false)
			expect(result.failParts).toBeDefined()
			expect(result.failParts!.length).toBeGreaterThan(0)
			const failedPart = result.failParts![0]
			expect(failedPart).toHaveProperty("error")
			expect((failedPart as { error: string }).error).toContain("No sufficiently similar match found")
```

src/core/diff/strategies/__tests__/multi-search-replace.spec.ts
```diff
@@ -1374,4 +1374,90 @@ function sum(a, b) {
...
```
edelauna: Same as above — the diagnostic assertions here are gated behind `if (!result.success)` and will pass silently if the branch is never entered. Moving them outside the guard would make the test fail explicitly when `failParts` is absent.

edelauna (commented):
Thanks for addressing my previous comments, just have a couple more around test setup.

edelauna (changes requested):
ensure codecoverage is met, and also please address previous comments.

src/core/diff/strategies/multi-search-replace.ts
nigeldelviero: Fix applied. The startLine-absent branch in originalContentSection was dumping the full file content via resultLines.join("\n") into the error message sent to the LLM API.

Changed line 553 (the else arm) to use a bounded window around matchIndex, the best match position from the fuzzy search mirroring the same bufferLines ± search block slicing the startLine-present branch already uses.

```typescript
// Before:
: `\n\nOriginal Content:\n${addLineNumbers(resultLines.join("\n"))}`

// After:
: `\n\nOriginal Content:\n${addLineNumbers(
    resultLines.slice(
      Math.max(0, (matchIndex >= 0 ? matchIndex : 0) - this.bufferLines),
      Math.min(resultLines.length, (matchIndex >= 0 ? matchIndex : 0) + searchLines.length + this.bufferLines),
    ).join("\n"),
    Math.max(1, (matchIndex >= 0 ? matchIndex : 0) - this.bufferLines + 1),
  )}`
```

When matchIndex === -1 (no match found at all), falls back to 0 — a window from the start of the file, still bounded. This prevents leaking the full file content (including potential secrets) to the LLM API via pushToolResult.

src/core/diff/strategies/__tests__/multi-search-replace.spec.ts
nigeldelviero: Removed if (!result.success) guards in both test cases. Assertions now run unconditionally, so missing failParts is a hard failure instead of silently passing.

src/core/diff/strategies/__tests__/multi-search-replace.spec.ts
nigeldelviero: Removed the if (!result.success) guard from the "should output enhanced error diagnostics" test. Assertions now run unconditionally, if failParts is ever missing, the test will hard-fail instead of silently passing.

src/core/diff/strategies/multi-search-replace.ts
```diff
@@ -81,10 +81,13 @@ export class MultiSearchReplaceDiffStrategy implements DiffStrategy {
...
 	constructor(fuzzyThreshold?: number, bufferLines?: number) {
-		// Use provided threshold or default to exact matching (1.0)
-		// Note: fuzzyThreshold is inverted in UI (0% = 1.0, 10% = 0.9)
-		// so we use it directly here
-		this.fuzzyThreshold = fuzzyThreshold ?? 1.0
+		// Use provided threshold or default to relaxed matching (0.9)
+		// A value of 0.9 means 90% similarity is required for a match.
+		// This was previously 1.0 (exact match) which caused frequent
+		// "Edit Unsuccessful" errors on minor whitespace differences.
+		// Clamp the threshold to [0.5, 1.0] as a defence-in-depth guard.
```
edelauna: The comment says "default to relaxed matching (0.9)" but `DEFAULT_DIFF_FUZZY_THRESHOLD` is `1.0` — the default is still exact matching. Could you update the comment to reflect the actual default? (The 0.9 value is a user-configurable opt-in, not the default.)

packages/types/src/global-settings.ts
```diff
@@ -102,6 +111,12 @@ export const globalSettingsSchema = z.object({
 	alwaysAllowWriteOutsideWorkspace: z.boolean().optional(),
 	alwaysAllowWriteProtected: z.boolean().optional(),
 	writeDelayMs: z.number().min(0).optional(),
+	/**
+	 * Fuzzy matching threshold for the multi-search-replace diff strategy.
+	 * Range: 0.5 (50% minimum similarity) to 1.0 (exact match only).
+	 * `@default` 1.0
+	 */
+	diffFuzzyThreshold: z.number().min(0).max(1).optional(),
```
edelauna: The JSDoc just above says the range is `0.5–1.0`, and the constructor clamps to `Math.max(0.5, ...)`, but the schema allows values down to `0`. A setting of `0.1` would pass validation, be stored, and then silently become `0.5` at runtime. Should this be `z.number().min(0.5).max(1)` to keep the three layers consistent?
```suggestion
	diffFuzzyThreshold: z.number().min(0.5).max(1).optional(),
```

webview-ui/src/i18n/locales/en/settings.json
```diff
@@ -761,6 +761,18 @@
 				"description": "Time to wait after file writes before proceeding, allowing diagnostic tools to process changes and detect issues."
 			}
 		},
+		"fileEdits": {
+			"diffFuzzyThreshold": {
+				"label": "Diff match threshold",
+				"description": "Lower thresholds make file edits more resilient to formatting and whitespace variations. A threshold of 100% requires an exact match."
+			}
+		},
```
edelauna: This `fileEdits` block is duplicated — the same key appears again at lines 770–775. JSON parsers silently use the last occurrence, so the first block is dead. Mind removing the first copy?

edelauna (changes requested):
Looks good - just some minor nits to clear up before merge.

packages/types/src/global-settings.ts
nigeldelviero: Fixed. Changed z.number().min(0).max(1) to z.number().min(0.5).max(1) so schema validation rejects values below 0.5 at the boundary, matching both the JSDoc range and the Math.max(0.5, ...) runtime clamp.

src/core/diff/strategies/multi-search-replace.ts
nigeldelviero: Fixed. The comment now reads "defaults to exact match (1.0)" with the clarification that 0.9 is user-selected via the diffFuzzyThreshold setting. I also updated the clamp range to [0.5, 1.0] as a layered defense.

edelauna (approved):
Nice! Thanks for adding this back.

PR #473, opened 2026-06-04
fix: prevent parent task from hanging when subtask delegation returns after per-mode API profile switch

### LINKED ISSUE #457: Subtask never returns to parent after a per-mode model/profile switch (attempt_completion silently waits for human)
**Zoo Code version:** 3.57.100116

**API provider/model:** Reproduces with per-mode profiles where the parent (orchestrator) and the spawned subtask use *different* model profiles (any providers).

**Related:** Part of #357 (Epic 2: Task Lifecycle Fixes). This is a **distinct** defect from the step 4/5 atomicity race — it is a return-**guard** that rejects the `active` parent state outright, so the parent never reaches those writes on the cross-profile hop.

### What happens

When an orchestrator/parent task delegates to a subtask whose mode is bound to a **different** API-configuration profile than the parent, the subtask completes its work and calls `attempt_completion`, but control **never returns to the parent**. The run silently waits for a manual click instead of auto-resuming the parent. Delegating to subtasks that share the parent's profile works fine; the first delegation to a *different*-profile mode is what breaks. It is order-independent — whichever profile is used first works, and the first switch to a different one breaks the next return.

### Steps to reproduce

1. Enable per-mode API configs (`modeApiConfigs`) so two modes use two different model profiles — e.g. an orchestrator on profile A and a sub-mode (e.g. `verifier`) on profile B.
2. From the orchestrator, delegate to a **same-profile** subtask. It returns to the parent correctly.
3. From the orchestrator, delegate (`new_task`) to a subtask on the **different** profile.
4. Let that subtask finish and call `attempt_completion`.

### Expected

Control returns to the parent task automatically (the parent resumes with the subtask result), the same as for same-profile subtasks.

### Actual

The subtask emits its completion result, but the parent is never re-opened; the flow stalls waiting for manual input.

### Root cause (verified)

The per-mode profile switch runs `handleModeSwitch` then `activateProviderProfile`, which leaves the parent task record as `status: "active"` while `awaitingChildId` is **still set to the child**. Two separate return-guards then reject because they require the parent status to be exactly `"delegated"`:

1. In the `attempt_completion` tool handler — the pre-check that decides whether to call `delegateToParent` requires `parent.status === "delegated"`. When it is `"active"`, it skips delegation and falls through to the "wait for user" path.
2. In `ClineProvider.reopenParentFromDelegation()` — the guard rejects unless `status === "delegated"`, logging `[reopenParentFromDelegation] Aborting: ... status=active`.

Because `awaitingChildId === childTaskId` already proves the parent is genuinely waiting for this exact child, both guards can safely also accept `status === "active"`.

### Suggested fix

Relax both guards to accept `"active"` in addition to `"delegated"`, only while `awaitingChildId` still points at the returning child:

- `attempt_completion` pre-check: `(parent.status === "delegated" || parent.status === "active") && parent.awaitingChildId === child`
- `reopenParentFromDelegation` guard: abort only when `status !== "delegated" && status !== "active"`

The unrelated cleanup paths (stack removal on `removeClineFromStack`, and `cancelTask`) should keep the strict `"delegated"`-only check.

### Workaround

Enabling **Lock the API configuration across all modes** (`lockApiConfigAcrossModes`) avoids the profile switch, so returns work — but every mode then runs on the same model, losing per-mode model routing.

**Roadmap alignment:** Reliability First — fixes a silent stall that blocks reliable multi-mode/agent workflows.

DScoNOIZ:
## Description

Fixes #457 — a critical bug where subtask delegation never returns to the parent when two modes use different API configuration profiles (per-mode `modeApiConfigs`).

## Problem

When an orchestrator/parent task delegates to a subtask whose mode uses a **different** `modeApiConfigs` profile than the parent, the flow silently stalls after the subtask completes. The parent is never resumed, requiring manual intervention.

## Root Cause

The per-mode profile switch runs `handleModeSwitch` → `activateProviderProfile`, which leaves the parent task record as `status: "active"` while `awaitingChildId` still points at the child. Two guards then reject because they require `status === "delegated"`:

1. **`reopenParentFromDelegation()`** — an overly strict `cancelledDelegationChildIds.has(childTaskId)` guard blocks reopen permanently when `cancelTask()` fails to detach the parent (race condition in `runDelegationTransition`). The child is added to the blacklist, and the second `attempt_completion` call can never recover.

2. **`AttemptCompletionTool` delegation pre-check** — only accepts `status === "delegated"` for the parent, missing the `"active"` state that occurs after a per-mode profile switch.

### `src/core/webview/ClineProvider.ts`
- **Removed `cancelledDelegationChildIds` check** from `reopenParentFromDelegation()` guard. The blacklist is redundant: `cancelTask()` sets parent status to `"active"` *before* adding to the blacklist. If the parent is still `"delegated"` and awaiting this child, it is safe to reopen. The `status + awaitingChildId` check already handles orphaned children.

### `src/core/tools/AttemptCompletionTool.ts`
- **Added parent verification** before delegation: verify `parentHistory.status === "delegated"` and `parentHistory.awaitingChildId === task.taskId` before calling `delegateToParent()`. If parent detached (cancelled/resumed), skip delegation gracefully.
- **Exempt subtask delegation** from `didToolFailInCurrentTurn` guard (`&& !task.parentTaskId`). Subtasks finishing their assigned work should not be blocked by a parent's error flag.
- **Added console.warn** diagnostics for failed reopen attempts.
- **Fixed error logging** — removed stale `historyLookupTaskId` variable that could point at the wrong task.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task delegation robustness by preventing concurrent delegation operations and adding validation checks.
  * Enhanced error recovery when delegating tasks to parent, with rollback capabilities.
  * Added explicit warnings when parent task cannot be resumed from delegation.

* **Refactor**
  * Refined task completion blocking logic for delegated subtasks.
  * Strengthened delegation state management with improved metadata persistence.

### DIFF
--- a/src/core/tools/AttemptCompletionTool.ts
+++ b/src/core/tools/AttemptCompletionTool.ts
@@ -40,8 +40,10 @@ export class AttemptCompletionTool extends BaseTool<"attempt_completion"> {
 		const { result } = params
 		const { handleError, pushToolResult, askFinishSubTaskApproval } = callbacks
 
-		// Prevent attempt_completion if any tool failed in the current turn
-		if (task.didToolFailInCurrentTurn) {
+		// Prevent attempt_completion if any tool failed in the current turn.
+		// Subtask delegation (parentTaskId) is exempt — the subtask is legitimately
+		// finishing its assigned work, not trying to "escape" from a failure.
+		if (task.didToolFailInCurrentTurn && !task.parentTaskId) {
 			const errorMsg = t("common:errors.attempt_completion_tool_failed")
 
 			await task.say("error", errorMsg)
@@ -86,20 +88,23 @@ export class AttemptCompletionTool extends BaseTool<"attempt_completion"> {
 				// to prevent duplicate tool_results when user revisits from history
 				const provider = task.providerRef.deref() as DelegationProvider | undefined
 				if (provider) {
-					let historyLookupTaskId = task.taskId
 					try {
 						const { historyItem } = await provider.getTaskWithId(task.taskId)
 						const status = historyItem?.status
 
 						if (status === "completed") {
-							// Subtask already completed - skip delegation flow entirely
-							// Fall through to normal completion ask flow below (outside this if block)
-							// This shows the user the completion result and waits for acceptance
-							// without injecting another tool_result to the parent
+							// Subtask already completed - emit task completed and return
+							// without showing completion_result ask again.
+							// This prevents infinite loop when user revisits a completed subtask from history:
+							// the subtask shows completion_result → user accepts → attempt_completion runs again
+							// → delegation flow attempts parent reopen → parent already resumed → loops forever.
+							this.emitTaskCompleted(task)
+							return
 						} else if (status === "active") {
-							historyLookupTaskId = task.parentTaskId
+							// Verify parent still awaits this child before asking the user.
+							// If parent detached (cancelled/resumed), skip delegation to avoid
+							// asking the user to return to a task no longer waiting for us.
 							const { historyItem: parentHistory } = await provider.getTaskWithId(task.parentTaskId)
-
 							if (
 								parentHistory?.status === "delegated" &&
 								parentHistory?.awaitingChildId === task.taskId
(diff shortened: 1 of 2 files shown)


edelauna:
I'm going to close this in favour of https://github.com/Zoo-Code-Org/Zoo-Code/pull/510 - if the issue is reopened we can re-investigate this PR.

PR #483, opened 2026-06-05
fix(commands): correct multi-line quoted command parsing, auto-approval, and malformed-command error surfacing

### LINKED ISSUE #484: [BUG] Multi-line script in a quoted argument is misclassified as multiple commands, defeating auto-approval
### Problem

A single command that wraps a multi-line script in a quoted argument (e.g. `sh -c '...'` with a real line break inside the quotes) is misclassified as a multi-statement sequence. The embedded newline is treated as a command separator before quote handling, so the quoted script is shredded into bogus sub-commands. This defeats allowlist auto-approval (the wrapper prompts the user even when allowlisted) and produces a noisy command-pattern breakdown in the UI.

### Context

Affects anyone whose agent runs a wrapper command (`sh -c`, `bash -c`, `python3 -c`, etc.) with a multi-line script passed as a single quoted argument. Even when the wrapper prefix is on the auto-approval allowlist, the command falls through to a manual approval prompt, adding friction to otherwise-approved workflows. This maps to the Reliability First roadmap goal (consistently reliable command execution).

### Reproduction steps

1) Add `sh` to the command auto-approval allowlist.
2) Ask the coding agent to run a wrapper command whose quoted argument contains an actual line break between statements, e.g.:
   ```
   sh -c 'echo 1;
   echo 2'
   ```
   (note the real newline inside the single-quoted argument)
3) Observe the identified command components and the resulting auto-approval decision.

### Expected result

The quoted argument stays intact, so the identified commands are:
```
sh -c 'echo 1; echo 2'
sh
```
and the command auto-approves because `sh` is allowlisted.

### Actual result

The embedded newline splits the command before quoting is handled, yielding bogus components:
```
sh -c echo 1   (incorrect: "echo 1" is a partial fragment of the quoted string)
echo 2         (incorrect: "echo 2" is a string argument, not a separate command)
sh
```
Those fragments match neither the allowlist nor the denylist, so `getCommandDecision` falls through to `ask_user` even though `sh` is allowlisted. The UI also surfaces the stray script-line fragments as separate command patterns.

### Variations tried

The same misclassification occurs whenever a newline is embedded in the quoted argument, across all quote styles:
- Single-quoted multi-line argument (`sh -c '...\n...'`)
- Double-quoted multi-line argument (`sh -c "...\n..."`)
- ANSI-C quoted multi-line argument (`sh -c $'...\n...'`), which additionally leaks an internal placeholder token into the output.

Genuine unquoted multi-statement scripts (newlines outside any quotes) correctly split into separate sub-commands -- that behavior is expected and should be preserved.

### App Version

v3.56.0

awschmeder:
### Related GitHub Issue

Closes: #484

> Issue #484 has been approved and assigned, satisfying the Issue-First policy.

### Description

A single command that wraps a multi-line script in a quoted argument (e.g. `sh -c '...'`) was misclassified as a multi-statement sequence, which defeated allowlist auto-approval and produced a noisy command-pattern breakdown in the UI. Malformed commands (unterminated quotes) were silently passed to the terminal, risking partial execution of the well-formed prefix before the shell aborted.

**Root causes fixed:**

- `parseCommand` split on every newline *before* any quote handling, so newlines inside a quoted argument were treated as command separators. Each script line became a bogus sub-command. Because those fragments matched neither the allowlist nor the denylist, `getCommandDecision` fell through to `ask_user` even when the wrapper command (e.g. `sh -c`) was allowlisted.
- Single-quoted strings were never masked, so their contents leaked out as stray tokens.
- ANSI-C quoting (`$'...'`) leaked an internal placeholder (`SQUOTE_0__`) because `shell-quote` parsed `$__SQUOTE_0__` as a variable expansion and corrupted the marker.
- Commands with an *unterminated* quote were passed to the terminal for execution, resulting in partial execution of the desired command before the shell hit the syntax error -- a hazard that is especially likely when LLMs generate deeply nested quoting.

**How (key implementation details):**

In `src/shared/parse-command.ts`:
- Mask quoted strings (single, double, and ANSI-C `$'...'`) at the top of `parseCommand()` **before** splitting on newlines, so newlines embedded in a quoted argument stay within their command. Genuine unquoted newlines still split into separate sub-commands, preserving per-statement safety checks.
- The masking uses a single shared `scanTopLevelQuotes` state machine (left-to-right alternation, escape-aware, `#`-comment-aware). Both `findUnterminatedQuote` and `maskTopLevelQuotes` delegate to this one scanner -- eliminating the duplicated state machines that existed in the initial implementation. `parseCommandLine` remains separate: it operates on already-masked single lines and uses `shell-quote` for operator tokenization, a different concern.
- `findUnterminatedQuote()` returns `{ quoteType, openIndex, message } | null`. `parseCommand()` returns any input with an unterminated quote as a single opaque token, so a fragment can never be auto-approved in isolation.
- Removed dead `arrayIndexing` / `__ARRAY_N__` restore branch -- it was never populated and silently returned `undefined` on any match. `${...}` patterns are handled by the `parameterExpansions` bucket, which is populated and restored correctly.

In `src/core/auto-approval/` and `src/tools/execute-command/`:
- `getCommandDecision` returns a dedicated `"malformed_command"` result (distinct from `"ask_user"`) when `findUnterminatedQuote` fires, anchoring the safety boundary against regression.
- `ExecuteCommandTool` surfaces a malformed-command as a `toolError` with a human-readable message locating the unterminated quote, so the agent receives structured feedback instead of a silent shell syntax error.

In `webview-ui/`:
- `CommandExecutionStatus` gains an `error` state and `CommandExecution` renders a visible error indicator (e.g. the red dot) for malformed commands, replacing the silent no-op the UI previously showed.

A single fix at the shared `parseCommand` source corrects both the backend auto-approval decision and the webview pattern breakdown, since both consume this function.

**Reviewers should note:** plain POSIX single-quote handling is intentionally not escape-aware (POSIX single quotes treat backslash literally and always terminate on an apostrophe); only the ANSI-C `$'...'` form is escape-aware. The `findUnterminatedQuote` return type carries `quoteType`, `openIndex`, and `message` to support located malformed-command error messages.

**Roadmap alignment:** Reliability First -- ensures command execution and auto-approval are consistently reliable for wrapped multi-line scripts, and provides clear, actionable feedback for malformed input rather than silent partial execution.

### Test Procedure

Automated (all affected suites pass):
- `cd src && npx vitest run shared/__tests__/parse-command.spec.ts core/auto-approval/__tests__/commands.spec.ts`
- `cd webview-ui && npx vitest run src/components/chat/__tests__/CommandExecution.spec.tsx`

Coverage added:
- `src/shared/__tests__/parse-command.spec.ts`: single/double quoted multi-line payloads, escaped inner quotes, mixed quote styles, ANSI-C `$'...'` (no placeholder leak), ANSI-C escaped-apostrophe + newline, shell-quote parse-failure fallback restoration, newline preservation, genuine multi-statement splitting, malformed unterminated-quote safe path (single/double/ANSI-C and an embedded-newline case), full `findUnterminatedQuote` suite (balanced input, each unterminated style with reported index, backslash-escaped quotes outside strings, `#` comments, mixed quote styles, ANSI-C/double-quote escape handling, apostrophe inside a `#` comment after a closed quoted argument).
- `src/core/auto-approval/__tests__/commands.spec.ts`: wrapped multi-line scripts auto-approve when the wrapper prefix is allowlisted; genuine multi-statement scripts still require every statement to be allowlisted; unterminated-quote commands return `"malformed_command"` regardless of allowlist (including `["*"]`).
- `webview-ui/src/components/chat/__tests__/CommandExecution.spec.tsx`: pattern selector no longer surfaces stray script-line fragments; error card renders for malformed commands.

Manual: verified in the Extension Development Host that a well-formed `sh -c '...'` multi-line wrapper auto-approves, a malformed unterminated-quote command shows an error card and a `toolError` (no standalone fragment, no silent partial execution), and genuine multi-statement input still requires each statement allowlisted.

### Pre-Submission Checklist

- [x] **Issue Linked**: This PR is linked to an approved GitHub Issue (see "Related GitHub Issue" above). _(Issue #484 is approved and assigned.)_

### Screenshots / Videos

A video of the manual test confirmation is provided in the attached comments.

### Documentation Updates

### Additional Notes

The fix is a single change at the shared `parseCommand` source, so both the backend auto-approval decision and the webview pattern breakdown are corrected together. The malformed-command error card and `toolError` surfacing are additive: they give the agent and user explicit, located feedback instead of a silent shell abort.


src/shared/__tests__/parse-command.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+			expect(parseCommand(input)).toEqual(["sh -c 'echo a && echo b | grep x'"])
+		})
+
+		it("preserves the embedded newline in the restored command", () => {
+			const input = "sh -c 'echo 1\necho 2'"
+			const result = parseCommand(input)
+			expect(result).toEqual(["sh -c 'echo 1\necho 2'"])
+			expect(result[0]).toContain("\n")
+		})
+	})
```
edelauna, line 75: What does `parseCommand` return for an unclosed quote like `sh -c 'echo test` (no closing quote)? The top-level regex `'[^']*'` requires a closing quote to match, so this input passes through unmasked and an embedded newline would still split — is that the intended contract? Worth a test to pin the behavior.

edelauna (commented):
Nice! I'm ok with this - if someone's allow listing commands which start with `sh -c` then I can accept that as a proxy for essentially YOLO mode.

One comment to increase test coverage, but otherwise this looks good.

src/shared/__tests__/parse-command.spec.ts
awschmeder: Valid concern - regex based parser is not adequate to catch malformed inputs (which LLMs tend to generate when complex quoting is involved). I am reviewing and will re-submit this with a more robust parser.

awschmeder: FYI - related `yolo -c` utility https://github.com/awschmeder/yolo-cli

awschmeder: @edelauna I updated the PR adding a quote-syntax checker that detects if the command contains unterminated quotes and prevents it from reaching the regex, returning them as a single token for user approval.

```
	if (findUnterminatedQuote(command) !== null) {
		return [command]
	}
```
	
A future PR should consider blocking commands that have an unterminated quote with an error message sent back to the agent for correction. 

LLMs occasionally generate improper quote syntax in deeply nested quote sequences. In singular commands the shell will catch the parse error, but in a multipart command it can result in partial execution of the command sequence before the shell hits a parse error, which could have unintended side effects and creates more work for the agent to recover from.

awschmeder:
## Regex parsing: current state after this change

To clarify scope in response to the review note about cross-line regex masking: this PR converts the single hazardous cross-line quote-masking step to a state machine, but does **not** eliminate all regex parsing. Summary below.

### Converted to the state-machine parser

- The top-level quote masker in `parseCommand()` (previously a single comment-blind `command.replace(/.../g)` that ran across the whole multi-line string) is now `maskTopLevelQuotes()`, which walks the input left-to-right using the same rules as `findUnterminatedQuote()`. It respects `#` comments, so a quote inside a comment can no longer pair with a quote on a later line and hide a real newline separator.

This was the only quote-aware, cross-line regex, and it was the source of the reported defect.

### Regex still in use (intentional, unchanged by this PR)

These all run **after** the cross-line newline split (i.e. on a single line at a time) or operate on non-shell content, so the comment-blind cross-line hazard does not apply to them:

1. **Trivial / non-parsing**
   - `/\s/.test(...)` single-char whitespace checks inside the two state machines.
   - `masked.split(/\r\n|\r|\n/)` newline split, which runs only after quotes are masked.
   - The `__PLACEHOLDER_(\d+)__` restore/match patterns (internal sentinel tokens, not shell syntax).

2. **Per-line tokenization in `parseCommandLine()`** -- still regex-based:
   - PowerShell redirections, `$((...))` and `$[...]` arithmetic, `${...}` parameter expansion, `<(...)`/`>(...)` process substitution, ANSI-C `$'...'`, `$var` / special bash variables, `$(...)` and backtick subshells, single/double quote masking, and the shell-quote fallback operator split.

### Out of scope

Converting the per-line tokenizers in `parseCommandLine()` to a unified state machine would be a substantially larger refactor. If desired, it should be tracked as a separate issue/PR rather than expanding this fix.

awschmeder:
# Shell Quoting Syntax Error -- Manual Test Results

Tests were run using `sh -c` to observe how the shell handles quoting syntax errors in various
contexts. The goal was to identify whether a quoting error prevents all execution (parse-time
rejection) or allows partial execution (runtime failure).

---

## Tests 1-9: Parse-Time Rejection

In all single-invocation cases, `zsh` parses the entire command expression before executing
anything. A quoting error anywhere in the expression blocks all execution.

| # | Description | Command tested | Result |
|---|-------------|---------------|--------|
| 1 | POSIX single quote, unmatched double inside | `sh -c 'echo "hello world'` | `zsh:1: unmatched "` |
| 2 | ANSI-C `$'...'` outer, unmatched double inside | `sh -c $'echo "hello world'` | `zsh:1: unmatched "` |
| 3 | Double quote outer, unmatched single inside | `sh -c "echo 'hello world"` | `zsh:1: unmatched '` |
| 4 | Single quote + literal newline + unmatched double | `sh -c 'echo "hello` + newline + `world'` | `zsh:2: unmatched "` |
| 5 | Double quote + literal newline + unmatched single | `sh -c "echo 'hello` + newline + `world"` | `zsh:2: unmatched '` |
| 6 | ANSI-C `\n` escape + unmatched double | `sh -c $'echo "hello\nworld'` | `zsh:2: unmatched "` |
| 7 | Compound `&&`, unmatched quote on second command | `sh -c 'echo hello && echo "world'` | `zsh:1: unmatched "` |
| 8 | Compound `;`, unmatched quote on second command | `sh -c 'echo hello; echo "world'` | `zsh:1: unmatched "` |
| 9 | Compound `\|`, unmatched quote on second command | `sh -c 'echo hello \| echo "world'` | `zsh:1: unmatched "` |

**Observation:** Regardless of quoting style or compound operator used, `zsh` rejects the
entire expression at parse time. Nothing executes.

---

## Tests 10-11: Partial Execution

Two patterns were found that allow the first command to execute before the quoting error is
encountered.

### Test 10 -- `eval` deferral via ANSI-C quoting

**Command tested:**
```zsh
sh -c $'echo hello && eval \'echo "world\''
```

**Output:**
```
hello
(eval):1: unmatched "
```

**Explanation:** The ANSI-C `$'...'` outer quoting makes the full expression syntactically
valid from `zsh`'s perspective. `echo hello` runs successfully. `eval` then receives
`echo "world` as a string and attempts to parse it at runtime, producing a deferred parse
error from `eval`'s own parser rather than `zsh`'s pre-execution parser.

---

### Test 11 -- Two separate `sh -c` invocations separated by `;`

**Command tested:**
```zsh
sh -c 'echo hello'; sh -c 'echo "world'
```

**Output:**
```
hello
zsh:1: unmatched "
```

**Explanation:** The `;` separator causes `zsh` to parse and execute each statement
independently in sequence. The first invocation (`sh -c 'echo hello'`) completes successfully.
`zsh` then attempts to parse the second invocation and fails on the unmatched `"` before that
command can run.

---

## Summary

| Pattern | Execution behavior |
|---------|-------------------|
| Single `sh -c` with any quoting style | Parse-time rejection -- nothing executes |
| Compound operators (`&&`, `;`, `\|`) inside single `sh -c` | Parse-time rejection -- nothing executes |
| `eval` with deferred broken string (ANSI-C outer quoting) | Partial execution -- first command runs, `eval` fails at runtime |
| Two separate `sh -c` calls joined by `;` | Partial execution -- first call completes, second fails at parse time |

The only reliable way to get partial execution with a quoting syntax error is to either defer
the broken string to a runtime evaluator (`eval`, `bash -c`, etc.) or split the commands into
separate shell invocations.

All sequences were correctly parsed and matched when 'sh -c' is on the Execute auto-approve list.

awschmeder:
**Prompt for ZooCode: Paste this into the chat for manual testing**

# Manual Test Procedure -- Multiline Command Parsing

Run the following shell commands one at a time using your shell execution tool.
After each command, observe the actual output and confirm it matches the
expected result listed. Report PASS or FAIL for each before moving to the next.

All commands use `sh` and `echo` only. Newlines are represented as literal
newlines in heredoc-style inputs below.

---

## Part 1: Unquoted newlines (should split)

### 1. Three echo commands on separate lines split into three entries
```sh
echo a
echo b
echo c
```
Expected: each line executes independently, producing output:
```
a
b
c
```

---

## Part 2: POSIX single-quoted multi-line argument (should NOT split)

### 2. Newline inside single quotes is part of the argument
```sh
sh -c 'echo a
echo b'
```
Expected: both `echo a` and `echo b` run as part of the single `sh -c` call:
```
a
b
```

### 3. Operators inside single quotes are literal
```sh
sh -c 'echo hello && echo world'
```
Expected: the `&&` is passed as a literal string to `sh`, which interprets it:
```
hello
world
```

---

## Part 3: ANSI-C quoting ($'...') (should NOT split)

### 4. ANSI-C quoted multi-line argument stays as one sh -c call
```sh
sh -c $'echo a\necho b'
```
Expected:
```
a
b
```

### 5. ANSI-C with escaped apostrophe and embedded newline
```sh
sh -c $'echo it\'s ok\necho done'
```
Expected: **syntax error, no output.**

Why: the ANSI-C string expands to the two-line literal script
`echo it's ok` + newline + `echo done`. When `sh -c` parses that script the
apostrophe in `it's` opens a POSIX single-quoted region that is never closed,
so `sh` reports an unmatched quote error before executing anything.

---

## Part 4: Double-quoted multi-line argument (should NOT split)

### 6. Newline inside double quotes is part of the argument
```sh
sh -c "echo a
echo b"
```
Expected:
```
a
b
```

---

## Part 5: Heredoc -- unquoted delimiter (should NOT split body)

### 7. Unquoted delimiter heredoc
```sh
sh << EOF
echo hello
echo world
EOF
```
Expected:
```
hello
world
```

### 8. Single-quoted delimiter heredoc (body is literal, no expansion)
```sh
sh << 'EOF'
echo hello
echo world
EOF
```
Expected:
```
hello
world
```

### 9. Double-quoted delimiter heredoc (body allows expansion)
```sh
sh << "EOF"
echo hello
echo world
EOF
```
Expected:
```
hello
world
```

### 10. <<- heredoc strips leading tabs from body and terminator
```sh
sh <<-EOF
	echo hello
	echo world
	EOF
```
Expected:
```
hello
world
```

### 11. Multi-line heredoc body -- all body lines execute under one sh call
```sh
sh << 'EOF'
echo line1
echo line2
echo line3
EOF
```
Expected:
```
line1
line2
line3
```

### 12. Command after heredoc terminator runs as a separate command
```sh
sh << EOF
echo inside
EOF
echo outside
```
Expected:
```
inside
outside
```
(`inside` comes from the heredoc; `outside` from the trailing `echo outside`.)

### 13. # comment inside heredoc body is preserved as literal text
```sh
sh << 'EOF'
# this is a comment
echo hello
EOF
```
Expected: the comment line is passed to `sh` which ignores it, then `echo hello` runs:
```
hello
```

---

## Part 6: Locale quoting ($"...") (should NOT split, $ prefix preserved)

Note: `$"..."` is a bash locale-quoting construct. When no gettext translation
catalog is present (the common case), the string is passed through unchanged
and the `$` is consumed by bash. This behavior is consistent across bash 3.2
(macOS) and bash 4+ (Linux) when no catalog matches.

### 14. Locale-quoted string -- no leading $ in output
```sh
echo $"hello world"
```
Expected:
```
hello world
```
The `$` is consumed by bash as part of the `$"..."` token; only the string
content is passed to `echo`.

### 15. Operators inside locale-quoted string are not shell operators
```sh
sh -c 'echo $"hello && world"'
```
Expected:
```
hello && world
```
The `&&` inside `$"..."` must not split the command -- `echo` receives the
whole string as one argument. The `$` is consumed by bash.

---

## Part 7: Malformed / unterminated quotes (should not execute partial commands)

### 16. Unterminated single quote -- shell should report a syntax error, not execute partial content
```sh
sh -c 'echo hello
```
Expected: the shell reports an unmatched quote error. No partial output from `echo hello` should appear before the error.

### 17. Unterminated heredoc -- bash executes the body at EOF (no error)
```sh
sh << EOF
echo hello
```
(Submit the input without an `EOF` terminator line and observe the result.)

Expected on **bash (all versions, non-interactive)**:
```
hello
```
Why: bash in non-interactive script execution treats end-of-file as an implicit
heredoc terminator and silently executes the body -- it does NOT report an
error. This is consistent behavior across bash 3.2 through 5.x on all
platforms.

Note -- intentional parser divergence: the Zoo Code `parseCommand` function
treats a missing heredoc terminator as malformed and returns the whole input as
a single opaque token (same as an unterminated quote). This is deliberately
more conservative than bash: when the LLM generates a heredoc without a
terminator, the user is prompted for approval rather than each body line being
evaluated independently for auto-approval. The body still executes correctly at
runtime via bash -- the conservative behavior only affects the auto-approval
decision.

---

## Summary

| # | Scenario | Expected Result |
|---|----------|----------------|
| 1 | Unquoted newlines | Each line executes separately |
| 2 | Single-quoted newline | Both echos run under one sh -c |
| 3 | Operators in single quotes | && passes through to sh |
| 4 | ANSI-C $'...' with \n | Both echos run under one sh -c |
| 5 | ANSI-C with escaped apostrophe | Syntax error (apostrophe opens unclosed quote in expanded script) |
| 6 | Double-quoted newline | Both echos run under one sh -c |
| 7 | Heredoc unquoted delimiter | Body lines run under sh |
| 8 | Heredoc single-quoted delimiter | Body literal, lines run under sh |
| 9 | Heredoc double-quoted delimiter | Body expandable, lines run under sh |
| 10 | <<- heredoc | Tab-indented body and terminator work |
| 11 | Multi-line heredoc | All body lines run under one sh |
| 12 | Command after heredoc | Heredoc body + trailing command both run |
| 13 | # comment in heredoc body | Comment ignored by sh, echo runs |
| 14 | Locale quoting $"..." | `hello world` ($ consumed by bash; no leading $ in output) |
| 15 | Operators in locale quotes | && not split; `hello && world` printed as one string |
| 16 | Unterminated single quote | Syntax error, no partial execution |
| 17 | Unterminated heredoc | bash executes body at EOF; parser conservatively returns opaque token |

awschmeder:
## Part 8: Herestring (<<<) -- manual test procedure

Added in the latest commit: herestring (`<<<`) support and fix. A herestring feeds a single word as stdin -- it has no body or terminator and must NOT be treated as a heredoc by the parser.

**Bug that was fixed:** both `findUnterminatedQuote` and `maskTopLevelQuotes` matched `<<` as a heredoc opener without guarding against a third `<`. For `cmd <<< word`, the first `<` triggered the heredoc branch at position `i`, the guard prevented it, fell through to `result += '<'; i++`, then the second `<` at `i+1` re-triggered the branch and consumed the rest of input as an unterminated heredoc body.

**Fix:** explicit `<<<` passthrough before the `<<` heredoc branch in both functions -- emit three `<` chars and advance `i` by 3 via `continue`.

### Test 18: Simple herestring -- treated as one command

```sh
sh <<< 'echo hello'
```

Expected:
```
hello
```

(`sh` receives `echo hello` as its stdin script and runs it.)

---

### Test 19: Herestring followed by a second command -- must split into two commands

```sh
sh <<< 'echo hello'
echo done
```

Expected: two separate commands execute:
```
hello
done
```

The parser must NOT consume `echo done` as part of the herestring body. Without the fix, the second `<` re-triggered the heredoc branch and consumed `echo done` as a body line, returning the entire input as one opaque token.

---

### Test 20: Herestring with a single-quoted multiline word -- one command

```sh
sh <<< 'echo line1
echo line2'
```

Expected: the quoted string contains a literal newline; `sh` receives the two-line script and runs both echos:
```
line1
line2
```

---

### Test 21: Herestring with an ANSI-C quoted multiline word -- one command

```sh
sh <<< $'echo line1\necho line2'
```

Expected:
```
line1
line2
```

awschmeder:
# Shell Quoting Test Results

## Summary
All 21 tests PASSED ✓

---

## Part 1: Unquoted newlines (should split)

### Test 1: Three echo commands on separate lines
**Status: PASS** ✓

Command:
```sh
echo a
echo b
echo c
```

Expected Output:
```
a
b
c
```

Actual Output:
```
a
b
c
```

---

## Part 2: POSIX single-quoted multi-line argument (should NOT split)

### Test 2: Newline inside single quotes is part of the argument
**Status: PASS** ✓

Command:
```sh
sh -c 'echo a
echo b'
```

Expected Output:
```
a
b
```

Actual Output:
```
a
b
```

### Test 3: Operators inside single quotes are literal
**Status: PASS** ✓

Command:
```sh
sh -c 'echo hello && echo world'
```

Expected Output:
```
hello
world
```

Actual Output:
```
hello
world
```

---

## Part 3: ANSI-C quoting ($'...') (should NOT split)

### Test 4: ANSI-C quoted multi-line argument stays as one sh -c call
**Status: PASS** ✓

Command:
```sh
sh -c $'echo a\necho b'
```

Expected Output:
```
a
b
```

Actual Output:
```
a
b
```

### Test 5: ANSI-C with escaped apostrophe and embedded newline
**Status: PASS** ✓

Command:
```sh
sh -c $'echo it\'s ok\necho done'
```

Expected Output: **syntax error, no output**

Actual Output:
```
sh: -c: line 0: unexpected EOF while looking for matching `''
sh: -c: line 2: syntax error: unexpected end of file
```

---

## Part 4: Double-quoted multi-line argument (should NOT split)

### Test 6: Newline inside double quotes is part of the argument
**Status: PASS** ✓

Command:
```sh
sh -c "echo a
echo b"
```

Expected Output:
```
a
b
```

Actual Output:
```
a
b
```

---

## Part 5: Heredoc -- unquoted delimiter (should NOT split body)

### Test 7: Unquoted delimiter heredoc
**Status: PASS** ✓

Command:
```sh
sh << EOF
echo hello
echo world
EOF
```

Expected Output:
```
hello
world
```

Actual Output:
```
hello
world
```

### Test 8: Single-quoted delimiter heredoc (body is literal, no expansion)
**Status: PASS** ✓

Command:
```sh
sh << 'EOF'
echo hello
echo world
EOF
```

Expected Output:
```
hello
world
```

Actual Output:
```
hello
world
```

### Test 9: Double-quoted delimiter heredoc (body allows expansion)
**Status: PASS** ✓

Command:
```sh
sh << "EOF"
echo hello
echo world
EOF
```

Expected Output:
```
hello
world
```

Actual Output:
```
hello
world
```

### Test 10: <<- heredoc strips leading tabs from body and terminator
**Status: PASS** ✓

Command:
```sh
sh <<-EOF
	echo hello
	echo world
	EOF
```

Expected Output:
```
hello
world
```

Actual Output:
```
hello
world
```

### Test 11: Multi-line heredoc body -- all body lines execute under one sh call
**Status: PASS** ✓

Command:
```sh
sh << 'EOF'
echo line1
echo line2
echo line3
EOF
```

Expected Output:
```
line1
line2
line3
```

Actual Output:
```
line1
line2
line3
```

### Test 12: Command after heredoc terminator runs as a separate command
**Status: PASS** ✓

Command:
```sh
sh << EOF
echo inside
EOF
echo outside
```

Expected Output:
```
inside
outside
```

Actual Output:
```
inside
outside
```

### Test 13: # comment inside heredoc body is preserved as literal text
**Status: PASS** ✓

Command:
```sh
sh << 'EOF'
# this is a comment
echo hello
EOF
```

Expected Output:
```
hello
```

Actual Output:
```
hello
```

---

## Part 6: Locale quoting ($"...") (should NOT split, $ prefix preserved)

### Test 14: Locale-quoted string -- no leading $ in output
**Status: PASS** ✓

Command:
```sh
echo $"hello world"
```

Expected Output:
```
hello world
```

Actual Output:
```
hello world
```

### Test 15: Operators inside locale-quoted string are not shell operators
**Status: PASS** ✓

Command:
```sh
sh -c 'echo $"hello && world"'
```

Expected Output:
```
hello && world
```

Actual Output:
```
hello && world
```

---

## Part 7: Malformed / unterminated quotes (should not execute partial commands)

### Test 16: Unterminated single quote -- shell should report a syntax error
**Status: PASS** ✓

Command:
```sh
sh -c 'echo hello
```

Expected Output: Syntax error, no partial output

Actual Output:
```
/bin/sh: -c: line 0: unexpected EOF while looking for matching `''
/bin/sh: -c: line 1: syntax error: unexpected end of file
```

### Test 17: Unterminated heredoc -- bash executes the body at EOF (no error)
**Status: PASS** ✓

Command:
```sh
sh << EOF
echo hello
```
(submitted without EOF terminator line)

Expected Output:
```
hello
```

Actual Output:
```
hello
```

---

## Part 8: Herestring (<<<) -- single-line stdin redirect

### Test 18: Simple herestring -- treated as one command
**Status: PASS** ✓

Command:
```sh
sh <<< 'echo hello'
```

Expected Output:
```
hello
```

Actual Output:
```
hello
```

### Test 19: Herestring followed by a second command
**Status: PASS** ✓

Command:
```sh
sh <<< 'echo hello'
echo done
```

Expected Output:
```
hello
done
```

Actual Output:
```
hello
done
```

### Test 20: Herestring with a single-quoted multiline word
**Status: PASS** ✓

Command:
```sh
sh <<< 'echo line1
echo line2'
```

Expected Output:
```
line1
line2
```

Actual Output:
```
line1
line2
```

### Test 21: Herestring with an ANSI-C quoted multiline word
**Status: PASS** ✓

Command:
```sh
sh <<< $'echo line1\necho line2'
```

Expected Output:
```
line1
line2
```

Actual Output:
```
line1
line2
```

---

## Final Test Summary

| # | Scenario | Status |
|---|----------|--------|
| 1 | Unquoted newlines | PASS ✓ |
| 2 | Single-quoted newline | PASS ✓ |
| 3 | Operators in single quotes | PASS ✓ |
| 4 | ANSI-C $'...' with \n | PASS ✓ |
| 5 | ANSI-C with escaped apostrophe | PASS ✓ |
| 6 | Double-quoted newline | PASS ✓ |
| 7 | Heredoc unquoted delimiter | PASS ✓ |
| 8 | Heredoc single-quoted delimiter | PASS ✓ |
| 9 | Heredoc double-quoted delimiter | PASS ✓ |
| 10 | <<- heredoc with tabs | PASS ✓ |
| 11 | Multi-line heredoc | PASS ✓ |
| 12 | Command after heredoc | PASS ✓ |
| 13 | Comment in heredoc body | PASS ✓ |
| 14 | Locale quoting $"..." | PASS ✓ |
| 15 | Operators in locale quotes | PASS ✓ |
| 16 | Unterminated single quote | PASS ✓ |
| 17 | Unterminated heredoc | PASS ✓ |
| 18 | Simple herestring | PASS ✓ |
| 19 | Herestring + trailing command | PASS ✓ |
| 20 | Herestring with single-quoted multiline word | PASS ✓ |
| 21 | Herestring with ANSI-C quoted multiline word | PASS ✓ |

**Total: 21/21 PASSED ✓**

awschmeder:
@coderabbitai review

src/core/auto-approval/__tests__/commands.spec.ts
```diff
@@ -99,3 +99,51 @@ describe("getCommandDecision — integration with dangerous substitution checks"
...
+
+	it("auto-approves a genuine multi-statement script when every statement is allowed", () => {
+		const multiStatement = "echo hello\nls -la"
+		expect(getCommandDecision(multiStatement, ["echo", "ls"])).toBe("auto_approve")
+	})
+
+	it("auto-approves an ANSI-C quoted ($'...') multi-line argument when the wrapper prefix is allowed", () => {
+		const ansiC = "sh -c $'echo 1\necho 2'"
+		expect(getCommandDecision(ansiC, ["sh"])).toBe("auto_approve")
+	})
```
edelauna, line 148: The new suite covers well-formed multi-line commands but does not test the core safety boundary this PR introduces: that an unterminated-quote command is NOT auto-approved even when its prefix is allowlisted. Could we add a case like `getCommandDecision("sh -c 'echo a\nrm -rf /", ["sh"])` asserting `"ask_user"`, and one with `["*"]`? That would anchor the protection against regression.

src/shared/__tests__/parse-command.spec.ts
```diff
@@ -0,0 +1,471 @@
...
+
+		it("treats a backslash-escaped quote outside a string as literal", () => {
+			// `echo \'` is a literal quote, not an unterminated region.
+			expect(findUnterminatedQuote("echo \\'")).toBeNull()
+			expect(findUnterminatedQuote('echo \\"')).toBeNull()
+		})
+
+		it("ignores a quote that appears inside a comment", () => {
+			expect(findUnterminatedQuote("echo hi # it's fine")).toBeNull()
+		})
```
edelauna, line 454: Is there a test for a properly closed quote followed by a `#` comment containing an unmatched quote — e.g. `sh -c 'cmd' #it's here`? The comment-handling path in `scanTopLevelQuotes` was written for this case but I don't see it exercised in the `findUnterminatedQuote` suite.

src/shared/parse-command.ts
```diff
@@ -2,35 +2,511 @@ import { parse } from "shell-quote"
...
+			}
+			const match = command.slice(start, i)
+			quotes.push(match)
+			result += `__TOPLEVEL_QUOTE_${quotes.length - 1}__`
+			continue
+		}
+
+		if (char === '"') {
+			// Double quote: escape-aware, ends at the next unescaped ".
+			const start = i
```
edelauna, line 336: Both `scanTopLevelQuotes` and `parseCommandLine` implement the shell quoting state machine independently (and a third time in the regex chain below). Any fix to escape handling or a new quote style needs to land in multiple places — is a consolidation being tracked, or is there a reason they need to stay separate?

src/shared/parse-command.ts
```diff
@@ -108,8 +606,22 @@ function parseCommandLine(command: string): string[] {
 			return `__SUBSH_${subshells.length - 1}__`
 		})
 
-	// Then handle quoted strings
-	processedCommand = processedCommand.replace(/"[^"]*"/g, (match) => {
+	// Mask quoted strings (single and double) so their contents -- including
+	// operators like &&, |, ; and any embedded newlines -- are not treated as
+	// command separators. A single left-to-right scan with an alternation is used
+	// so that whichever quote opens first wins, preventing a quote of one style
```
edelauna, line 598: Is `arrayIndexing` ever populated? I can't find anything that pushes into it or emits an `__ARRAY_N__` placeholder — if that pattern never appears in the masked string, this restore branch is dead. And if a command literally contained `__ARRAY_0__` the substitution would silently return `undefined`.

edelauna (commented):
Seems like a lot was added since last review, some additional comments.

awschmeder:
Addressing @edelauna's review threads:

**Re: [Safety boundary tests](https://github.com/Zoo-Code-Org/Zoo-Code/pull/483#discussion_r3373458764)**
Done. Added `"returns malformed_command for a command with an unterminated quote regardless of allowlist"` which asserts `getCommandDecision("sh -c 'echo a\necho b", ["sh"])`, `["*"]`, and `[malformed]` all return `"malformed_command"`. The decision is a distinct value rather than `"ask_user"` so the protection is explicitly anchored against regression.

---

**Re: [Comment + apostrophe test](https://github.com/Zoo-Code-Org/Zoo-Code/pull/483#discussion_r3373458773)**
Added: `"ignores an apostrophe inside a # comment that follows a closed quoted argument"` in the `findUnterminatedQuote` suite asserts that `"echo 'hello' # it's a comment"` and `'echo "hello" # it\'s a comment'` both return `null`. This specifically exercises the comment path in `scanTopLevelQuotes` for the case you described.

---

**Re: [DRY -- duplicated state machines](https://github.com/Zoo-Code-Org/Zoo-Code/pull/483#discussion_r3373458780)**
Consolidated. `findUnterminatedQuote` and `maskTopLevelQuotes` now both delegate to a single shared `scanTopLevelQuotes` state machine -- one pass covers span detection, unterminated-quote detection, and comment handling. `parseCommandLine` remains separate intentionally: it operates on already-masked single lines and uses `shell-quote` for operator tokenization, which is a different concern. The two scanners introduced by this PR are now one.

---

**Re: [Dead `arrayIndexing` code](https://github.com/Zoo-Code-Org/Zoo-Code/pull/483#discussion_r3373458784)**
Correct -- `arrayIndexing` was never populated and the `__ARRAY_N__` restore was dead code (returning `undefined` on any match). Removed. `${...}` patterns are now handled by the `parameterExpansions` bucket, which is actually populated and restored correctly.

src/shared/parse-command.ts
awschmeder: Looks like it was just dead code and never implemented.  Latest commit removes it.

src/shared/parse-command.ts
awschmeder: Agreed, not DRY... new commit consolidates parsing into one function and improves error detection.

src/shared/__tests__/parse-command.spec.ts
awschmeder: Added to test coverage.

src/core/auto-approval/__tests__/commands.spec.ts
awschmeder: I widened the PR scope to include triggering a toolError on malformed commands so they will no longer execute.

awschmeder:
https://github.com/user-attachments/assets/5d43a032-131f-4f9a-b270-cd8c4b926b9d

(re-uploaded with better resolution)...

src/shared/parse-command.ts
awschmeder: Actually...  good point about hthe plaseholder substition - its worse than that as parseCommand will break if the input contains any of the placeholder tokens.  The latest commit closes the gap by pre-escaping any token text and de-escaping it after parsing.

awschmeder:
@coderabbitai review

edelauna (approved):
Thanks for this contribution!

PR #333, opened 2026-05-26
fix(terminal): use detected shell for execa and vscode terminal creation (#321)

F915:
### Related GitHub Issue

Closes: #321

### Description

Zoo Code correctly detects the user's configured default shell (e.g., PowerShell) and tells the AI to generate commands using that shell's syntax, but the Inline Terminal actually executed commands in `cmd.exe`. This caused:

- **PowerShell-specific commands** (like `Write-Host`) to fail with "not recognized" errors
- **Non-ASCII output garbling** on systems with non-ASCII locales — `cmd.exe` encodes output using the system codepage (e.g., CP936 for zh-CN) rather than UTF-8

This PR connects shell detection to both terminal execution paths so the user's configured shell is used consistently.

---

The work evolved in two phases:

**Phase 1 (original PR)** passed `getShell()` to both execution paths — `execa({ shell: getShell() })` for inline terminal and `createTerminal({ shellPath: getShell() })` for external terminal.

**Phase 2 (follow-up)** was prompted by [review feedback from @edelauna](https://github.com/Zoo-Code-Org/Zoo-Code/pull/333#discussion_r3319361333), who pointed out that passing `getShell()` as `shellPath` on Windows with WSL configured was problematic — the function could return a Unix path (`/bin/bash`) that isn't valid as a Windows `shellPath`. Further investigation revealed that the root issue wasn't specific to WSL: the existing shell detection in `shell.ts` used manual VS Code config parsing with a ~95-entry allowlist, an approach that was inherently fragile and could diverge from VS Code's own internal shell resolution.

Phase 2 addressed this at the root by replacing the manual allowlist-based detection with `vscode.env.shell` — the API VS Code itself uses and has provided since 1.37. This eliminated the mismatch entirely and allowed several downstream simplifications.

---

#### Phase 1 — Connect shell detection to execution paths

1. **ExecaTerminalProcess.ts** — Changed `shell: true` to `shell: getShell()`. The detected shell is now passed to execa subprocess execution.
2. **Terminal.ts** — Added `shellPath: getShell()` to `vscode.window.createTerminal()`.
3. **ExecaTerminalProcess.spec.ts** — Assertions updated from `shell: true` to `shell: "/mock/fallback-shell"`.
4. **TerminalRegistry.spec.ts** — Assertions updated to include the new `shellPath` parameter.

#### Phase 2 — Consolidate shell detection and fix WSL support

[...]

Issue #321, opened 2026-05-25
[BUG] Windows: shell mismatch between AI prompt generation and Inline Terminal execution


F915:
### Problem (one or two sentences)

Zoo Code correctly detects the user's configured default shell (e.g., PowerShell 7) and tells the AI to generate commands using that shell's syntax, but the Inline Terminal actually executes commands in `cmd.exe`. This causes PowerShell-specific commands to fail with garbled error text, and all non-ASCII output appears garbled on non-UTF-8 locales.

### Context (who is affected and when)

**Who is affected:** All Windows users who have configured a non-cmd default shell in VSCode (PowerShell 7, git-bash, WSL, etc.), particularly those in non-UTF-8 locales (Chinese/936, Japanese/932, Korean/949).

**When it happens:** Whenever Zoo Code's Inline Terminal executes a command. This is the default behavior since `terminalShellIntegrationDisabled` defaults to `true`.

**Not affected:** macOS / Linux users (native UTF-8 shells). Windows users who have not changed their default shell from cmd.exe.

### Reproduction steps

1. Windows 11 (zh-CN locale), VSCode with Zoo Code v3.55.0
2. Set `terminal.integrated.defaultProfile.windows` = `PowerShell 7 (UTF-8)`
3. In Zoo Code Inline Terminal, run `Write-Host "test"`
4. Observe: the command is not executed. Instead, garbled Chinese text appears where the cmd.exe error message should be
5. Run `echo 中文测试`
6. Observe: output shows garbled text instead of the expected Chinese characters

### Expected result

Commands should execute in the shell that Zoo Code detects (PowerShell 7). `Write-Host "test"` should execute as a valid PowerShell cmdlet. Non-ASCII output should display correctly in UTF-8.

### Actual result

Commands execute in `cmd.exe`. PowerShell cmdlets are not recognized. Non-ASCII output is decoded using the system code page (CP936) instead of UTF-8, producing garbled text.

### Variations tried (optional)

- **Shell detection confirmed:** `getShell()` correctly returns the path to pwsh.exe from VSCode settings.
- **Encoding switch ineffective:** `chcp 65001 && echo 中文` still produces garbled output (cmd.exe parses the command text using CP936 before `chcp` executes).
- **cmd /c wrapper works:** `cmd /c "chcp 65001 > nul && echo 中文测试"` produces correct output (the new subprocess allows the code page to be switched before the internal command is parsed).

### App Version

v3.55.0

[...]

PR #509, opened 2026-06-06
fix: use VS Code default shell for command execution instead of /bin/sh

### LINKED ISSUE #508: [BUG] Inline terminal always uses system shell instead of what shell is configured in VS code
**Describe the bug**
Inline terminal always uses system shell, which is /bin/sh on Linux, even if VS code is configured to always use e.g. bash. This introduces divergence between how user expects inline terminal to work (e.g. that .bashrc would be sourced) compared to what actually happens.

If inline terminal is disabled then all actions are done in VS code terminal. It would be least confusing if inline terminal setting does not change what shell is being used. 

**To Reproduce**
1. Select bash as VS code shell
2. Reload window to get clean state
3. Give zoo code this prompt "please run `sleep 60`"
4. Run `ps aux | grep sleep` to observe how Zoo code launched the command
5. See that /bin/sh is always used.

**Expected behavior**
VS code configured shell should be used.

**Screenshots**
N/A

**Video**
N/A

Latest master as of week ago - before 3.56.0 got released.

**Additional context**
None.

p12tic:
### Related GitHub Issue

Closes: #508 

### Description

When terminalShellIntegrationDisabled is true (the default), commands are executed via execa. Previously, execa used shell: true which defaults to /bin/sh on Linux, ignoring the user's VS Code terminal shell settings.

Now execa falls back to getShell() which respects:
- VS Code terminal.integrated.defaultProfile.linux
- os.userInfo().shell
- $SHELL environment variable
- Platform-specific defaults (/bin/bash on Linux, /bin/zsh on macOS)

### Test Procedure

- Select bash as VS code shell
- Reload window to get clean state
- Give zoo code this prompt "please run sleep 60"
- Run ps aux | grep sleep to observe how Zoo code launched the command
- See that /bin/bash being used as expected.

### Pre-Submission Checklist

### Screenshots / Videos

n/a

### Documentation Updates

not needed

### Additional Notes

n/a

### Get in Touch

povilas_78868

## Summary by CodeRabbit

* **Refactor**
  * Improved shell selection mechanism for terminal processes.
* **Tests**
  * Updated test suite to verify shell selection behavior.

### DIFF
--- a/src/integrations/terminal/ExecaTerminalProcess.ts
+++ b/src/integrations/terminal/ExecaTerminalProcess.ts
@@ -3,6 +3,7 @@ import psTree from "ps-tree"
 import process from "process"
 
 import type { RooTerminal } from "./types"
+import { getShell } from "../../utils/shell"
 import { BaseTerminal } from "./BaseTerminal"
 import { BaseTerminalProcess } from "./BaseTerminalProcess"
 
@@ -40,7 +41,7 @@ export class ExecaTerminalProcess extends BaseTerminalProcess {
 			this.isHot = true
 
 			this.subprocess = execa({
-				shell: BaseTerminal.getExecaShellPath() || true,
+					shell: BaseTerminal.getExecaShellPath() || getShell(),
 				cwd: this.terminal.getCurrentWorkingDirectory(),
 				all: true,
 				// Ignore stdin to ensure non-interactive mode and prevent hanging
--- a/src/integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts
+++ b/src/integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts
@@ -1,6 +1,7 @@
 // npx vitest run integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts
 
 const mockPid = 12345
+const mockGetShell = vitest.fn(() => "/bin/bash")
 
 vitest.mock("execa", () => {
 	const mockKill = vitest.fn()
@@ -19,6 +20,10 @@ vitest.mock("execa", () => {
 
 vitest.mock("ps-tree", () => ({
 	default: vitest.fn((_: number, cb: any) => cb(null, [])),
+}))
+
+vitest.mock("../../../utils/shell", () => ({
+	getShell: () => mockGetShell(),
 }))
 
 import { execa } from "execa"
@@ -63,7 +68,7 @@ describe("ExecaTerminalProcess", () => {
 			const execaMock = vitest.mocked(execa)
 			expect(execaMock).toHaveBeenCalledWith(
 				expect.objectContaining({
-					shell: true,
+					shell: "/bin/bash",
 					cwd: "/test/cwd",
 					all: true,
 					env: expect.objectContaining({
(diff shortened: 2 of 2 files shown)


edelauna:
Thanks for the contribution! We have a broader PR open at #333 that covers the same underlying issue (#321) and goes further — it fixes shell selection and includes handling for Windows/WSL edge cases that were surfaced during review.

Could you coordinate with @F915 to see if there's anything here worth folding in

edelauna:
After review I'm going to reopen this PR since the other one is still in draft, if you'd like to review #333 to incorporate any existing feedback into your PR that would help with future review.

F915:
I have an open PR (#333) that relates to the same issue (#321 / #508).

Noted in [a review comment](https://github.com/Zoo-Code-Org/Zoo-Code/pull/333#discussion_r3319361333) on #333, changing shell: true to shell: getShell() has a known limitation on Windows with WSL — getShell() can return /bin/bash, which is not a valid Windows shellPath.

The current version of #333 takes a different approach, using vscode.env.shell for shell resolution instead of modifying shell: true alone. It's still in draft — testing across platforms and terminal modes is not complete yet.

Would love to hear your thoughts if you have any.

p12tic:
I think #333 can go ahead, it covers more edge cases.

p12tic:
Seems like #333 was not merged, so I'm reopening this PR.

src/integrations/terminal/ExecaTerminalProcess.ts
```diff
@@ -40,7 +41,7 @@ export class ExecaTerminalProcess extends BaseTerminalProcess {
 			this.isHot = true
 
 			this.subprocess = execa({
-				shell: BaseTerminal.getExecaShellPath() || true,
+					shell: BaseTerminal.getExecaShellPath() || getShell(),
```
edelauna, line 44: Nit: this line picked up an extra leading tab — it's indented one level deeper than its siblings (`cwd`, `all`, `stdin`, `env`), curious how lint didn't flag this.

src/integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts
```diff
@@ -1,6 +1,7 @@
 // npx vitest run integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts
 
 const mockPid = 12345
+const mockGetShell = vitest.fn(() => "/bin/bash")
```
edelauna, line 4: Per the [Vitest docs](https://vitest.dev/api/vi.html#vi-clearallmocks), `clearAllMocks()` in `afterEach` resets call history but does **not** restore `mockReturnValue` overrides. The fallback test sets `mockGetShell.mockReturnValue("/usr/bin/zsh")`, which will bleed into any subsequent test that relies on the default `"/bin/bash"`. Would `resetAllMocks()` + a `mockGetShell.mockImplementation(() => "/bin/bash")` in `beforeEach` be safer?

src/integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts
```diff
@@ -105,13 +110,14 @@ describe("ExecaTerminalProcess", () => {
 			)
 		})
 
-		it("should fall back to shell=true when execaShellPath is undefined", async () => {
+		it("should fall back to getShell() when execaShellPath is undefined", async () => {
 			BaseTerminal.setExecaShellPath(undefined)
+			mockGetShell.mockReturnValue("/usr/bin/zsh")
```
edelauna, line 115: The "should use execaShellPath when set" test (above) sets `execaShellPath` to `"/bin/bash"` and `mockGetShell` also defaults to `"/bin/bash"`, so the assertion `shell: "/bin/bash"` passes regardless of which branch of the `||` fires. Would adding `expect(mockGetShell).not.toHaveBeenCalled()` there make it actually verify the short-circuit?

edelauna (changes requested):
Thanks for re-opening, had some feedback on test structure, otherwise this is looking good.

PR #516, opened 2026-06-07
chore: gate marketplace publish behind PR approval check

edelauna:
## Summary

- Adds a `check-pr-approval` job to `marketplace-publish.yml` that runs before the `publish-stable` job
- Uses the GitHub REST API to find the PR associated with the triggering commit, then queries the GraphQL API for `reviewDecision`
- If no PR exists, or `reviewDecision` is not `APPROVED`, the check job fails with a non-zero exit — blocking the `marketplace-production` environment gate from surfacing entirely
- `publish-stable` depends on `check-pr-approval` via `needs:`, so a failed check produces a visible red failure rather than a silent green skip

## Why GraphQL over REST

`reviewDecision` is only available on the GraphQL `PullRequest` object — the REST `GET /pulls/{number}` endpoint does not expose it. Using `reviewDecision` means GitHub's own computed effective state is used, which correctly accounts for:
- Reviewer write-permission requirements
- Superseded approvals (CHANGES\_REQUESTED after an approval)
- Blocking reviews and stale approvals

## Security note

The `check-pr-approval` job uses `GITHUB_TOKEN` with `pull-requests: read` only. This is safe because the workflow triggers on `push` to tags — only collaborators with write access to the repo can push tags, so untrusted fork contributors cannot trigger this workflow or leak the token.

## Test plan

- [ ] Push a tag from a branch whose PR has no approvals → `check-pr-approval` fails, no environment approval prompt appears
- [ ] Push a tag from a branch whose PR is approved → `check-pr-approval` passes, `marketplace-production` environment gate fires as normal
- [ ] Push a tag from a branch whose PR had an approval later superseded by CHANGES\_REQUESTED → `check-pr-approval` fails

## Summary by CodeRabbit

* **Chores**
  * Added an approval gate to the stable publish workflow so only commits tied to APPROVED pull requests are published.
* **Documentation**
  * Updated release procedure: require an approved release PR before tagging, retry tagging if approval fails, enqueue the PR for merge after successful deployment, and prevent merging to main until deployment succeeds.

### DIFF
--- a/.github/workflows/marketplace-publish.yml
+++ b/.github/workflows/marketplace-publish.yml
@@ -7,7 +7,51 @@ on:
   workflow_dispatch:
 
 jobs:
+  check-pr-approval:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    steps:
+      - name: Check PR approval status
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          SHA: ${{ github.sha }}
+        run: |
+          pr_number=$(gh api "repos/${REPO}/commits/${SHA}/pulls" \
+            --header "Accept: application/vnd.github+json" \
+            --jq '.[0].number // empty')
+
+          if [ -z "$pr_number" ]; then
+            echo "No PR found for commit ${SHA}. Deployment requires an approved PR."
+            exit 1
+          fi
+
+          review_decision=$(gh api graphql \
+            -f owner="${REPO%%/*}" \
+            -f name="${REPO#*/}" \
+            -F number="$pr_number" \
+            -f query='
+              query($owner: String!, $name: String!, $number: Int!) {
+                repository(owner: $owner, name: $name) {
+                  pullRequest(number: $number) {
+                    reviewDecision
+                  }
+                }
+              }' \
+            --jq '.data.repository.pullRequest.reviewDecision // "NONE"')
+
+          echo "PR #${pr_number} review_decision: ${review_decision}"
+
+          if [ "$review_decision" != "APPROVED" ]; then
+            echo "PR #${pr_number} is not approved (state: ${review_decision}). Deployment blocked."
+            exit 1
+          fi
+
+          echo "PR #${pr_number} is approved. Proceeding."
+
   publish-stable:
+    needs: [check-pr-approval]
     runs-on: ubuntu-latest
     environment: marketplace-production
     permissions:
--- a/.roo/commands/release.md
+++ b/.roo/commands/release.md
@@ -82,32 +82,31 @@ mode: code
     - If the release includes translated README or package-localization updates, include those files in the same PR.
     - Let the release validation workflow and normal PR checks run before merge.
 
-11. After the release PR is merged, stop for a release review on the resulting `main` commit.
+11. Once the release PR is open and passing checks, get it approved by a reviewer before proceeding.
+
+    - Do not create the tag until the PR has at least one approval — the publish workflow enforces this automatically and will fail if no approved PR is found for the tagged commit.
+
+12. After the PR is approved, create the release tag on the release branch tip and push it:
 
     ```bash
-    git switch main
-    git pull origin main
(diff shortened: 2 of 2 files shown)


.github/workflows/marketplace-publish.yml
```diff
@@ -7,7 +7,51 @@ on:
...
+                    reviewDecision
+                  }
+                }
+              }' \
+            --jq '.data.repository.pullRequest.reviewDecision // "NONE"')
+
+          echo "PR #${pr_number} review_decision: ${review_decision}"
+
+          if [ "$review_decision" != "APPROVED" ]; then
+            echo "PR #${pr_number} is not approved (state: ${review_decision}). Deployment blocked."
```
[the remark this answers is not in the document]
edelauna: This is exactly the point, we'd like to publish from the branch. I'll update `release.md` to better document.

.github/workflows/marketplace-publish.yml
```diff
@@ -7,7 +7,51 @@ on:
...
+                  pullRequest(number: $number) {
+                    reviewDecision
+                  }
+                }
+              }' \
+            --jq '.data.repository.pullRequest.reviewDecision // "NONE"')
+
+          echo "PR #${pr_number} review_decision: ${review_decision}"
+
+          if [ "$review_decision" != "APPROVED" ]; then
```
taltas, line 46: One thing here, should we be checking if the base branch is actually main? Otherwise we can publish from a different base right, I'm not sure we want that.

Something like this:

```suggestion
          pr_info=$(gh api graphql \
            -f owner="${REPO%%/*}" \
            -f name="${REPO#*/}" \
            -F number="$pr_number" \
            -f query='
              query($owner: String!, $name: String!, $number: Int!) {
                repository(owner: $owner, name: $name) {
                  pullRequest(number: $number) {
                    reviewDecision
                    baseRefName
                  }
                }
              }' \
            --jq '.data.repository.pullRequest')
            
          review_decision=$(jq -r '.reviewDecision // "NONE"' <<<"$pr_info")
          base_ref=$(jq -r '.baseRefName // empty' <<<"$pr_info")
          
          echo "PR #${pr_number} review_decision: ${review_decision}"

         if [ "$review_decision" != "APPROVED" ] || [ "$base_ref" != "main" ]; then
```

.roo/commands/release.md
```diff
@@ -82,32 +82,31 @@ mode: code
...
+    - A human approver must then approve the deployment before the extension is published to VS Code Marketplace and Open VSX.
+
+14. After a successful deployment, add the release PR to the merge queue.
+
+    ```bash
+    gh pr merge [pr-number] --auto --squash
+    ```
+
+    - Do not merge before the deployment succeeds — merging first and then discovering a publish failure leaves `main` ahead of what was actually shipped.
+    - The merge queue runs all required checks against the release branch before merging to `main`.
```
taltas, line 112: Is this actually happening with the merge queue running all the required checks?

under `.github/workflows/release-validation.yml` I dont any checks_request for the merge_group:

```
    merge_group:
        types: [checks_requested]
```

taltas (changes requested):
Couple of questions/suggestions

.github/workflows/marketplace-publish.yml
navedmerchant: I think the intention is that we merge from a release branch instead of main. The person doing the release needs to make sure they do not rebase from main before doing the release

edelauna: Exactly, we want to release a snapshot of main + any changes. It’s possible that development on main continues, so releasing hot fixes becomes a bit trickier, but I think it could still be done via CLI,  just also require the ability to tag main directly. 

By using the branch I think it’s a bit clearer to reviewers what’s being released. Whereas if we are pushing tags directly to main we’d need to review the git graph and process would be a bit more opaque.

navedmerchant (approved):

.roo/commands/release.md
edelauna: This ties into the fact that releases are cut from release branches, when a tag gets pushed.

taltas (approved):

PR #525, opened 2026-06-07
fix(terminal): kill entire process group on Stop so child commands are not orphaned

### LINKED ISSUE #524: [BUG] Clicking Stop does not kill child processes spawned by the terminal -- shell exits but children keep running
## Problem

When Zoo Code runs a command via the inline terminal (Execa) and the user clicks **Stop**, killing the shell process does not propagate to the commands the shell spawned. The shell sits waiting for its children to finish -- so processes like `sleep 30` keep running until they complete naturally, even though the user clicked Stop.

## Context

Affects any user running a long-running command through the inline terminal (Execa provider path). The command appears stopped from Zoo Code's perspective but the underlying process keeps consuming resources and holding locks until it finishes on its own.

## Reproduction steps

1. Extension version v3.56.0, macOS
2. Termina lettings: Integrated terminal or inline terminal enabled (Settings > Terminal > uncheck "Use VSCode's built-in terminal")
3. Ask Zoo Code to run: `sleep 30`
4. Click **Stop** before the command finishes
5. In a separate terminal: `ps aux | grep sleep | grep -v grep`
6. Observe: the `sleep` process is still alive and counting down

Same behavior with nested children: `sh -c 'sh -c "sleep 30"'` or commands that spawn background jobs.

## Expected result

Clicking Stop kills the command and all processes it spawned, immediately.

## Actual result

The shell wrapper is killed but its child processes (the actual commands) keep running. The shell was waiting for them to finish and the kill signal was not propagated to the children.

## Root cause

The Execa subprocess is spawned with `shell: true`, which means Node spawns a shell, and the shell in turn spawns the actual command. Killing the shell PID leaves its children orphaned -- the shell doesn't propagate the signal, it just exits and the children get reparented.

The previous attempt to fix this used `ps-tree` to enumerate children after a 100 ms delay, but this was racy (children spawned after the snapshot survived) and missed grandchildren.

## Fix (in linked PR)

Spawn the subprocess with `detached: true` so the shell and all of its descendants share a single process group, then on abort send `process.kill(-pid, "SIGKILL")` (negative PID targets the entire process group). This kills the shell and every process it spawned atomically, with no race condition and no external dependency on `ps-tree`.

## App Version

v3.56.0

## Notes

- The VSCode integrated terminal path (`TerminalProcess`) has a related but separate limitation: it sends Ctrl+C via `sendText`, which only reaches the foreground process. Background jobs (`cmd &`) survive until PTY teardown. The inline terminal is preferred and is fully addressed by the linked PR.
- The `ps-tree` dependency is removed as part of this fix.

awschmeder:
### Related GitHub Issue

Closes: #524

> **Note:** Issue #524 is awaiting maintainer approval/assignment. The "Issue Linked" checklist item is left unchecked per project policy until the issue is approved.

### Description

When Execa spawns a command with `shell: true`, Node forks a shell process which in turn forks the actual command. Killing only the shell PID leaves its children running -- the shell exits but does not propagate the signal to its children, so processes like `sleep 30` keep going until they finish naturally.

**How the fix works:**

- On POSIX, add `detached: true` to the Execa options. This causes the OS to place the shell and all of its descendants into a new process group, with the shell as the group leader. On Windows, `detached: true` is omitted -- it causes a console-window flash that `windowsHide` cannot reliably suppress, and it is not needed since `taskkill /F /T` handles tree-kill without a detached process.
- On `abort()`, branch on platform:
  - **POSIX (macOS/Linux):** send `process.kill(-pid, "SIGKILL")` (negative PID = kill the entire process group). This atomically signals every process in the group -- the shell, its children, grandchildren, and any background jobs -- with no race condition.
  - **Windows:** call `spawnSync("taskkill", ["/F", "/T", "/PID", pid])`. The `/T` flag kills the full process tree (shell + all descendants). `process.kill(-pid)` is POSIX-only and not available on Win32.
- On abort, emit `{ exitCode: 137, signalName: "SIGKILL" }` instead of `exitCode: 0`, so the UI running-indicator dot correctly turns red (consistent with the integrated terminal path).
- Remove the `ps-tree` dependency and the associated 100 ms delayed PID-tracking logic, which was the previous (incomplete) attempt at this fix.

**Reviewers should note:**

- The `detached: true` option is required on POSIX for `process.kill(-pid, ...)` to work. Without it the subprocess is in the parent's process group and a negative-PID kill would affect unrelated processes.
- A fallback to `subprocess.kill("SIGKILL")` is retained for the edge case where the primary kill throws (e.g. ESRCH -- process already exited).
- The VSCode integrated terminal path (`TerminalProcess`) is **not** changed by this PR. It sends Ctrl+C via `sendText` and cannot reach background jobs inside the PTY. This is a known limitation documented in the test results file.
- **Windows `taskkill` path has not been manually tested by the author** -- unit tests covering the mocked `spawnSync` call are present (platform-guarded, skipped on non-win32). Community verification welcome.

### Test Procedure

**Automated tests (18 passing, 2 skipped on non-win32):**
```
cd src && npx vitest run integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts
```

New tests cover:
- Process group kill via negative PID (POSIX)
- Windows process tree kill via `taskkill /F /T /PID` (skipped on non-win32)
- Fallback to `subprocess.kill` when primary kill throws (both platforms)
- No-op when PID is not yet set
- `exitCode: 137` emitted on abort
- Race condition: stream throws before `aborted` flag is checked

**Manual test (macOS/Linux):**
1. Ask Zoo Code to run: `sleep 30`
2. Click **Stop** before the command finishes
3. In a separate terminal: `ps aux | grep sleep | grep -v grep`
4. Expected: empty output (process is dead)

Tested successfully: `sleep 30`, nested `sh -c 'sh -c "sleep 30"'`, loop with sleeps.

Note: background jobs via `sh -c 'sleep 15 & sleep 15 & sleep 15 & wait'` are killed correctly when using the inline terminal, but are not killed when using integrated terminal due to fundamental limitations with the integrated terminal path.

### Pre-Submission Checklist

- [ ] **Issue Linked**: This PR is linked to an approved GitHub Issue -- issue #524 is filed but awaiting maintainer approval/assignment.

### Screenshots / Videos

No UI changes beyond the running-indicator dot color correction (green -> red on abort).

### Documentation Updates

### Additional Notes

- `ps-tree` and `@types/ps-tree` are removed from `src/package.json` as they are no longer used.
- `detached: true` is set only on POSIX (where it is required for `process.kill(-pid, "SIGKILL")` to target only the spawned process group, not the Node.js parent process). On Windows it is skipped: `detached: true` triggers a console-window flash that `windowsHide` cannot reliably suppress due to a race in the Windows `CreateProcess` API.
- On Windows, `taskkill /F /T /PID` achieves the equivalent tree-kill. `taskkill.exe` is a Win32 system binary available on all Windows versions since XP, invoked directly via `spawnSync` (bypasses the user's configured shell). Note: `taskkill /T` walks the PPID tree rather than targeting a kernel-tracked group, so it is susceptible to orphaned descendants if an intermediate process exits before Stop is clicked. This is strictly less racy than the previous `ps-tree`-based approach, which was identically vulnerable but introduced an extra 100 ms delay that widened the race window. A fully race-free Windows solution would require a helper utility to assign processes to a Windows Job Object (which the kernel tracks independently of the PPID chain), but that is out of scope for this PR.


.changeset/kill-orphaned-child-processes.md
```diff
@@ -0,0 +1,5 @@
+---
+"zoo-code": patch
+---
+
```
navedmerchant, line 4: No need to commit a changeset file

navedmerchant (changes requested):
detached: true would briefly flash a terminal in windows causing a jarring user experience. I am not sure if there is a good solution to this problem.

awschmeder:
**Note on Windows behavior:** The  path added in the latest commit has not been manually tested on a Windows machine by the author. The unit tests for it are present but skipped on non-win32 (they exercise the mocked  call, not the real  binary).  I will test it on Windows and confirm.

PR #533, opened 2026-06-08
refactor(terminal): redesign profile settings UX — unified dropdown, consistent layout, and styling (#119)(#321)

### LINKED ISSUE #119: Support choosing terminal profile for inline terminal executaion

### LINKED ISSUE #321: [BUG] Windows: shell mismatch between AI prompt generation and Inline Terminal execution

F915:
### Related GitHub Issue

Closes: #119
Closes: #321

### Description

This PR builds on PR #277, which laid the groundwork for the VS Code terminal profile override feature. We've made a few refinements to the settings UI to improve the user experience:

**1. Reordered the Advanced Settings layout.** The "Use Inline Terminal" checkbox is now placed at the top of the Advanced Settings section, with all dependent options (profile dropdown, timeout, delay, zsh options) appearing below it. This follows the natural workflow: choose terminal type first, then configure it.

**2. Simplified profile selection into a single dropdown.** The two radio buttons — "Use VS Code default profile" and "Override shell for Zoo Code" — have been consolidated into one `Select` component. The first option is "Follow VS Code profile (recommended)", followed by the available path-based profiles. The selection state is encoded in the dropdown value directly, with no intermediate toggle.

**3. Aligned the Configure button with the project's design system.** The "Choose default profile in VS Code" button now uses the shadcn/ui `Button` component (`variant="secondary"`) with a `Terminal` icon, matching the style used elsewhere in settings (e.g., "Add Slash Command" in `SlashCommandsSettings`). The button auto-stretches with window width via a flex container.

- The Configure button is only shown when "Follow VS Code profile" is selected — when a specific profile is already chosen, the button is not needed.
- English text refined and all 17 locales synchronized.
- 15 unit tests added for dropdown selection behavior, conditional button visibility, profile clearing on stale values, layout order, icon rendering, and i18n key coverage.

### Test Procedure

**Unit tests:**

```bash
cd webview-ui
npx vitest run src/components/settings/__tests__/TerminalSettings.profile.spec.tsx
```

**Type check:**

```bash
pnpm check-types
```

**i18n validation:**

```bash
node scripts/find-missing-translations.js --area webview
```

### Pre-Submission Checklist

#### Before
before

#### After
after1
after2

### Documentation Updates

### Additional Notes

Together with PR #277, this PR completes the terminal profile override feature for issues #119 and #321.

### Get in Touch

Discord: Yon (yon.sinc)

## Summary by CodeRabbit

* **New Features**
  * Unified terminal shell profile dropdown replaces separate default/override controls
  * "Follow VS Code profile" option added; selecting it clears any explicit profile
  * Configure button shown only when no explicit profile selected; shows "no profiles" hint when none available
  * Profile selector rendered below the inline-terminal checkbox to preserve ordering

* **Localization**
  * Updated terminal profile labels, options and descriptions across all locales

* **Tests**
  * Expanded tests for dropdown behavior, Configure button, ordering and i18n keys


webview-ui/src/components/settings/TerminalSettings.tsx
```diff
@@ -300,6 +193,77 @@ export const TerminalSettings = ({
...
+									<Select
+										value={terminalProfile || DEFAULT_PROFILE_VALUE}
+										onValueChange={(value) =>
+											setCachedStateField(
+												"terminalProfile",
+												value === DEFAULT_PROFILE_VALUE ? undefined : value,
+											)
+										}
+										data-testid="terminal-profile-dropdown">
+										<SelectTrigger className="w-full">
```
edelauna: The `data-testid` is on `<Select>` (line above), which is a Radix context provider with no DOM node — the attribute is silently dropped in production. Could this move to `<SelectTrigger>` instead (consistent with how `terminal-output-preview-size-dropdown` is handled at line 132)?

```suggestion
							<SelectTrigger className="w-full" data-testid="terminal-profile-dropdown">
```

webview-ui/src/components/settings/TerminalSettings.tsx
```diff
@@ -300,6 +193,77 @@ export const TerminalSettings = ({
...
+								<SearchableSetting
+									settingId="terminal-profile"
+									section="terminal"
+									label={t("settings:terminal.profile.label")}>
+									<label className="block font-medium mb-1">
+										{t("settings:terminal.profile.label")}
+									</label>
+
+									<Select
+										value={terminalProfile || DEFAULT_PROFILE_VALUE}
```
edelauna: Could this use `??` instead of `||`? If `terminalProfile` ever arrives as `""` (e.g. from a future serialization path — `SettingsView` already converts `undefined → ""` before saving), `||` would silently swap it for the sentinel.

```suggestion
							value={terminalProfile ?? DEFAULT_PROFILE_VALUE}
```

webview-ui/src/components/settings/__tests__/TerminalSettings.profile.spec.tsx
```diff
@@ -96,136 +95,147 @@ describe("TerminalSettings VS Code terminal profile (#277)", () => {
...
 
-	it("shows the default radio selected and no dropdown when no profile is set", () => {
+	it("shows the dropdown with 'Following VS Code profile' as default when no profile is set", () => {
 		setup()
-		const defaultRadio = screen.getByTestId("terminal-profile-default-radio")
-		expect(defaultRadio).toBeChecked()
-		expect(screen.queryByTestId("terminal-profile-dropdown")).not.toBeInTheDocument()
+		const dropdown = screen.getByTestId("terminal-profile-dropdown")
+		expect(dropdown).toBeInTheDocument()
+		expect(dropdown.getAttribute("data-value")).toBe("__zoo_code_follow_vscode_sentinel__")
```
edelauna: This string literal will silently drift if `DEFAULT_PROFILE_VALUE` is ever renamed. Would it make sense to export the constant from `TerminalSettings.tsx` and import it here? (Same applies to line 147.)

webview-ui/src/components/settings/__tests__/TerminalSettings.profile.spec.tsx
```diff
@@ -96,136 +95,147 @@ describe("TerminalSettings VS Code terminal profile (#277)", () => {
...
-		expect(setCachedStateField).not.toHaveBeenCalled()
+		// Re-render with profile selected (cleanup first to avoid stale DOM from first render)
+		cleanup()
+		render(
+			<TerminalSettings
+				terminalShellIntegrationDisabled={false}
+				terminalProfile="Git Bash"
+				setCachedStateField={setCachedStateField}
+			/>,
+		)
```
edelauna: Is there a reason for `cleanup()` + a fresh `render()` here rather than using RTL's `rerender()`? The fresh mount also silently drops the `onTerminalProfilePickerOpened` prop, so a regression where the button reappears after re-render wouldn't be caught.

docs/after1.png
edelauna, line 1: lets remove these `.png` files for now, since they relate more to the PR description then over docs for the app.

edelauna (commented):
Nice! Thanks for doing this. Had a couple nits before merge.

F915:
> Nice! Thanks for doing this. Had a couple nits before merge.

@edelauna  Thanks for the review. All five items addressed in [67455004](https://github.com/Zoo-Code-Org/Zoo-Code/pull/533/commits/6745500490a53bac9990a78c32ba640a3fbe4ba6) — moved the `data-testid` to `SelectTrigger`, switched to `??`, exported the sentinel constant, switched to `rerender()`, and removed the screenshots.

There's one design question I raised in [PR #333](https://github.com/Zoo-Code-Org/Zoo-Code/pull/333#issuecomment-4661646899) that's relevant here too — whether the profile override should affect only when Inline Terminal is off, or also when it's on. The current implementation follows the existing Zoo Code pattern: override applies only when Inline Terminal is off. That said, I'm considering whether we should eventually let the override affect both modes — since the UI and docs already clearly recommend Inline Terminal and "Follow VS Code profile" as defaults, that seems like enough guidance for most users, and those with custom shell needs tend to understand the trade-offs.

edelauna (approved):
Much nicer - thank you.

I currently like the idea of only displaying this when the Use Inline Terminal options is un-selected. 

There's a separate issue tracking how to set the profile used by execa.

PR #538, opened 2026-06-08
feat(workspace): add rootResolution setting for multi-root workspaces

simurg79:
## Summary

Ports the **workspace root resolution** feature from `simurg79/Roo-Code` (commit `d89c1ff`) into Zoo-Code.

Introduces a new `zoo-code.workspace.rootResolution` VS Code setting that controls how Zoo resolves the "project root" in a multi-root workspace (used to locate `.roomodes`, `.roo/mcp.json`, `.roo/rules/`, and other project-scoped configuration):

| Value | Behavior |
| --- | --- |
| `activeEditor` (default) | Prefer the workspace folder containing the active editor; fall back to `workspaceFolders[0]`. **Preserves legacy behavior.** |
| `firstFolder` | Always use `workspaceFolders[0]`. **Deterministic** — independent of which file is focused. |

The default preserves today's behavior exactly, so this is a non-breaking, opt-in change.

## Motivation

The historic active-editor heuristic is convenient but non-deterministic: switching the focused file between roots silently changes which project config is picked up, and it can disagree with the marketplace installers (which always target `workspaceFolders[0]`). `firstFolder` gives a stable root for CI/automation and multi-root setups where only one folder is the real project.

## Changes

- [`src/utils/path.ts`](src/utils/path.ts): add `WorkspaceRootResolution` type + exception-safe `getRootResolutionStrategy()` helper; `getWorkspacePath()` and `getWorkspacePathForContext()` now honor the setting.
- [`src/package.json`](src/package.json): register `zoo-code.workspace.rootResolution`.
- [`src/package.nls.json`](src/package.nls.json): localized setting strings.
- [`src/utils/__tests__/path.spec.ts`](src/utils/__tests__/path.spec.ts): expanded coverage for both strategies, exception safety, and `getWorkspacePathForContext`.
- [`docs/design/workspace-root-resolution.md`](docs/design/workspace-root-resolution.md): design doc.

## Port notes

- Settings namespace adapted from the fork's `roo-cline.*` to Zoo-Code's `zoo-code.*`.
- The test reads the shared `vscode` mock via `resolve.alias` (mutation-based) to match Zoo-Code's `src/vitest.config.ts`.

## Testing

`cd src && npx vitest run utils/__tests__/path.spec.ts` → **30 passed**. Both modified JSON files validated as parseable. Pre-commit lint hooks passed.

## Summary by CodeRabbit

* **New Features**
  * Workspace root resolution setting for multi-root workspaces: Users can now choose how the extension determines the workspace root. Two options: (1) active editor—determines root based on the currently active file (default), or (2) first folder—always uses the first workspace folder. Supported in 17 languages.


src/package.json
```diff
@@ -414,6 +414,20 @@
...
+						"firstFolder"
+					],
+					"enumDescriptions": [
+						"%settings.workspace.rootResolution.activeEditor.description%",
+						"%settings.workspace.rootResolution.firstFolder.description%"
+					],
+					"default": "activeEditor",
+					"description": "%settings.workspace.rootResolution.description%",
+					"markdownDescription": "%settings.workspace.rootResolution.description%"
 				}
```
edelauna, line 432: Without an explicit `"scope"` this defaults to `"window"`, which means a committed `.vscode/settings.json` can override it. Since this setting controls which workspace folder is used to locate `.roo/mcp.json` (and therefore which MCP servers get registered), could a malicious workspace settings file redirect that to an attacker-controlled folder in a multi-root workspace?

If workspace-level override should be blocked, `"scope": "machine"` would prevent it. If it is intentional, worth documenting the trust decision here.

src/utils/path.ts
```diff
@@ -111,17 +111,70 @@ export const toRelativePath = (filePath: string, cwd: string) => {
...
+ *   2. The path utilities are imported very early during extension activation,
+ *      so we must never throw from a config read.
+ */
+function getRootResolutionStrategy(): WorkspaceRootResolution {
+	try {
+		// Read directly from the `zoo-code` section to avoid importing
+		// `Package` here (path.ts is a low-level utility and we want to keep
+		// it free of circular dependencies on shared/* and package.json).
+		const value = vscode.workspace
+			.getConfiguration("zoo-code")
```
edelauna: Every other `getConfiguration` call in the codebase uses `Package.name` to respect the `PKG_NAME` build override. If this extension is ever built under a different package name, `getConfiguration("zoo-code")` will silently return only the supplied defaults — no error, no warning. Is the circular-dependency concern in the comment above actually verified? `shared/package.ts` only imports from `../package.json` and doesn't look like it touches `path.ts`.

src/utils/__tests__/path.spec.ts
```diff
@@ -63,6 +108,155 @@ describe("Path Utilities", () => {
 		})
 
 		it("should return undefined when outside a workspace", () => {})
```
edelauna: This test body is empty and the description looks wrong — `getWorkspacePath` always returns a `string`, never `undefined`. Was this meant to be filled in?

src/utils/__tests__/path.spec.ts
```diff
@@ -63,6 +108,155 @@ describe("Path Utilities", () => {
...
+				try {
+					expect(getWorkspacePath("/fallback")).toBe("/fallback")
+				} finally {
+					restore()
+				}
+			})
+		})
+
+		it("falls back to default behavior when reading the setting throws", () => {
+			mockWorkspace.getConfiguration = () => {
```
edelauna: `getConfiguration` is mutated here directly but `withWorkspaceMock`'s cleanup (lines 66–70) only restores `workspaceFolders`, `getWorkspaceFolder`, and `activeTextEditor`. Recovery depends on the outer `beforeEach` running `setRootResolution(undefined)` next. If this test is ever moved or the `beforeEach` changes, subsequent tests will throw on any config read. Worth saving/restoring `getConfiguration` in the `finally` block here, or extending `withWorkspaceMock` to cover it.

src/utils/__tests__/path.spec.ts
```diff
@@ -63,6 +108,155 @@ describe("Path Utilities", () => {
...
+				getWorkspaceFolder: spy,
+			})
+			try {
+				expect(getWorkspacePathForContext("/some/other/workspace/file.ts")).toBe("/test/workspace")
+				// Should not have looked up the context path's workspace folder.
+				expect(spy).not.toHaveBeenCalled()
+			} finally {
+				restore()
+			}
+		})
```
edelauna, line 281: What happens in `activeEditor` mode when `contextPath` is provided but `getWorkspaceFolder` returns `null`? That fallback path (the `console.debug` branch in `path.ts:183–186`) doesn't appear to be covered here.

docs/design/workspace-root-resolution.md
edelauna, line 1: lets not commit this, this is a working file for you.

edelauna (changes requested):
Nice! Thanks for adding this, had a couple edge case comments.

src/package.json
simurg79: Good catch — this setting feeds MCP-server discovery, so I've set `"scope": "machine"` to prevent a committed workspace `settings.json` from redirecting the resolved root. This intentionally disallows workspace-level overrides (the security tradeoff is deliberate).

src/utils/path.ts
simurg79: You're right — `shared/package.ts` only imports `package.json`, so there's no cycle. Switched to `getConfiguration(Package.name)` so it honors the `PKG_NAME` build override, and removed the inaccurate comment.

src/utils/__tests__/path.spec.ts
simurg79: Removed — it was a leftover stub with an incorrect description (`getWorkspacePath` always returns a string). The no-workspace path is already covered by the `falls back to defaultCwdPath` cases.

src/utils/__tests__/path.spec.ts
simurg79: Done — `withWorkspaceMock` now snapshots and restores `getConfiguration` too, and the throwing-config test routes its mock through it, so a broken mock can't leak into sibling tests regardless of `beforeEach`.

src/utils/__tests__/path.spec.ts
simurg79: Added a test for that case (`activeEditor` + `contextPath` set + `getWorkspaceFolder` → null); it covers the `console.debug` fallback branch and asserts it falls back to `workspaceFolders[0]`.

docs/design/workspace-root-resolution.md
simurg79: Removed from the PR — that was a working scratch doc, not intended to ship.

edelauna (approved):
Thanks for the contribution, this is a cool feature for multiple folders within a workspace.

PR #555, opened 2026-06-10
Add Fable 5 support across Anthropic providers

taltas:
## Summary
- add Claude Fable 5 model support across Anthropic-related provider type definitions and runtime handlers
- update fetchers and model parameter handling for Fable 5 request behavior, including provider-specific temperature and reasoning handling
- add and expand provider/fetcher/shared tests covering Anthropic, Vertex, Bedrock, Requesty, OpenRouter, and Vercel AI Gateway Fable 5 behavior

## Testing
- Prior implementation subtask reported the relevant provider tests passed before this review-prep workflow.
- Updated regression coverage is included in the provider and transform test files in this branch.

## Summary by CodeRabbit

* **New Features**
  * Added support for Claude Fable 5 across Anthropic, Bedrock, OpenRouter, Vercel AI Gateway, and Vertex.
  * Enabled adaptive thinking / reasoning (budget & binary) and prompt-caching support where applicable.
  * Model-specific temperature handling: temperature is omitted when unsupported.

* **Tests**
  * Added unit and integration tests covering Claude Fable 5 behavior and parsing across providers.


packages/types/src/providers/bedrock.ts
```diff
@@ -217,6 +217,24 @@ export const bedrockModels = {
 			},
 		],
 	},
+	"anthropic.claude-fable-5": {
```
JamesRobert20, line 220: This entry declares contextWindow 1M, but `anthropic.claude-fable-5` isn't in `BEDROCK_1M_CONTEXT_MODEL_IDS`. Every other Claude model that advertises 1M keeps a 200k base plus a tier that gets unlocked by the `context-1m-2025-08-07` beta header, and that header is gated on this list. Fable 5 skips that pattern on all three paths, including direct Anthropic (the beta push in `anthropic.ts` doesn't list it either). If the model needs the beta header to serve 1M, requests past 200k will fail and the window advertised here is misleading. If Fable 5 serves 1M natively without the header, this is correct as written. Which is it?

packages/types/src/providers/vertex.ts
```diff
@@ -434,6 +434,21 @@ export const vertexModels = {
 			},
 		],
 	},
+	"claude-fable-5": {
```
JamesRobert20, line 437: Same question as the Bedrock entry: `claude-fable-5` declares 1M context but isn't in `VERTEX_1M_CONTEXT_MODEL_IDS`, so the `context-1m-2025-08-07` beta header is never sent on this path. Fine if the model serves 1M without the header, broken if it needs it.

packages/types/src/providers/openrouter.ts
```diff
@@ -43,6 +43,7 @@ export const OPEN_ROUTER_PROMPT_CACHING_MODELS = new Set([
 	"anthropic/claude-opus-4.1",
 	"anthropic/claude-opus-4.5",
 	"anthropic/claude-opus-4.6",
+	"anthropic/claude-fable-5",
```
JamesRobert20, line 46: Fable 5 is in `OPEN_ROUTER_PROMPT_CACHING_MODELS` but not `OPEN_ROUTER_REASONING_BUDGET_MODELS`. The fetcher only sets `supportsReasoningBudget` from that second set, and the fable-5 override block in `fetchers/openrouter.ts` doesn't set it either, so on OpenRouter the reasoning toggle won't show up even though every other provider path has `supportsReasoningBudget: true`. The fetcher test only checks maxTokens, contextWindow, and supportsTemperature, so this gap isn't caught.

src/api/providers/__tests__/bedrock.spec.ts
```diff
@@ -1524,13 +1542,15 @@ describe("AwsBedrockHandler", () => {
...
 				expect(isAdaptiveThinkingModel("anthropic.claude-opus-4-7")).toBe(true)
 				expect(isAdaptiveThinkingModel("anthropic.claude-opus-4-8")).toBe(true)
+				expect(isAdaptiveThinkingModel("anthropic.claude-fable-5")).toBe(true)
 				// Future-proof Sonnet patterns — guarded even before a registry entry exists.
 				expect(isAdaptiveThinkingModel("anthropic.claude-sonnet-4-7")).toBe(true)
 				expect(isAdaptiveThinkingModel("anthropic.claude-sonnet-4-8")).toBe(true)
 			})
 
 			it("returns true when the id carries a cross-region or global prefix", () => {
 				expect(isAdaptiveThinkingModel("us.anthropic.claude-opus-4-8")).toBe(true)
```
JamesRobert20, line 1565: This asserts the `global.` prefix on Fable 5, but `anthropic.claude-fable-5` isn't in `BEDROCK_GLOBAL_INFERENCE_MODEL_IDS`, so `getModel()` never produces `global.anthropic.claude-fable-5` at runtime (the prefix is only applied for ids in that list). The test passes because `isAdaptiveThinkingModel` matches on substring, but it covers an id that can't occur in production. If Fable 5 should be reachable through global inference, add it to the list.

src/api/providers/fetchers/requesty.ts
```diff
@@ -45,6 +45,12 @@ export async function getRequestyModels(baseUrl?: string, apiKey?: string): Prom
 				cacheReadsPrice: parseApiPrice(rawModel.cached_price),
 			}
 
+			if (rawModel.id === "anthropic/claude-fable-5") {
```
JamesRobert20, line 48: This override has no direct coverage. The requesty handler tests mock `../fetchers/modelCache` with a hardcoded fable-5 entry, so `getRequestyModels` parsing this branch never runs in the suite. A small test that feeds a raw `anthropic/claude-fable-5` model through `getRequestyModels` and checks these three flags would close that.

JamesRobert20 (commented):
Reviewed the Fable 5 registry wiring. A few of the model-info additions don't have the matching allowlist entry, which is where the behavior actually gets gated. Notes inline.

One smaller thing not inline: maxTokens is 8192 on Bedrock/Vertex vs 128k on direct Anthropic. That matches the opus-4-5 pattern in the same files so it's likely intentional, but it does cap long outputs on those paths. Just confirming that's deliberate.

JamesRobert20 (changes requested):

packages/types/src/providers/bedrock.ts
edelauna: Resolved. Confirmed from the AWS Bedrock console that Fable 5 can only be used through a cross-region inference profile — so the 1M context window is native and the `context-1m-2025-08-07` beta header is not needed (correctly omitted from `BEDROCK_1M_CONTEXT_MODEL_IDS`). Added `anthropic.claude-fable-5` to `BEDROCK_GLOBAL_INFERENCE_MODEL_IDS` since the inference profile is mandatory for this model.

packages/types/src/providers/vertex.ts
edelauna: Resolved. Confirmed that Fable 5 has a native 1M context window on Vertex AI — it does not require the `context-1m-2025-08-07` opt-in beta header. Correctly omitted from `VERTEX_1M_CONTEXT_MODEL_IDS`; the `contextWindow: 1_000_000` in the registry entry is accurate as-is.

packages/types/src/providers/openrouter.ts
edelauna: Resolved. Added `anthropic/claude-fable-5` to `OPEN_ROUTER_REASONING_BUDGET_MODELS` so `supportsReasoningBudget` is set on the OpenRouter path. Also added `supportsReasoningBinary = true` to the Fable 5 override block in `fetchers/openrouter.ts` to match the other provider paths. The OpenRouter fetcher test now asserts both flags.

src/api/providers/__tests__/bedrock.spec.ts
edelauna: Resolved. Added `anthropic.claude-fable-5` to `BEDROCK_GLOBAL_INFERENCE_MODEL_IDS` — confirmed from the AWS Bedrock console that cross-region inference is mandatory for Fable 5 (not optional like the other models in this list). The existing test at this line now covers a real production path, and a new test was added to verify `getModel()` produces `global.anthropic.claude-fable-5` when `awsUseGlobalInference: true`.

src/api/providers/fetchers/requesty.ts
edelauna: Resolved. Added `src/api/providers/fetchers/__tests__/requesty.spec.ts` with direct unit tests for `getRequestyModels` that exercise the Fable 5 override branch (lines 48–52) without mocking the model cache. Covers both that the three flags are applied for `anthropic/claude-fable-5` and that they are not applied to other models.

packages/types/src/providers/anthropic.ts
```diff
@@ -126,6 +126,23 @@ export const anthropicModels = {
...
+		cacheWritesPrice: 12.5, // $12.50 per million tokens
+		cacheReadsPrice: 1.0, // $1.00 per million tokens
+		// Fable 5 uses the same adaptive-thinking / binary-toggle convention as
+		// Opus 4.7+ on the direct Anthropic provider path.
+		supportsReasoningBudget: true,
+		supportsReasoningBinary: true,
+		supportsTemperature: false,
+		description:
+			"Claude Fable 5 is Anthropic's most capable widely released model for the most demanding reasoning and long-horizon agentic work.",
+	},
```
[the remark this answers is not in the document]
edelauna: The `maxTokens` discrepancy is intentional and already documented inline: `anthropic.ts` sets `maxTokens: 128_000` with the comment `// Overridden to 8k if enableReasoningEffort is false`, while Bedrock and Vertex use `8192` as the base (matching the same pattern as opus-4-5 and opus-4-8 on those paths). No change needed.

navedmerchant (approved):

JamesRobert20 (approved):
Re-checked the follow-up commit against the earlier review threads. The gaps are closed:

- OpenRouter: `anthropic/claude-fable-5` is in `OPEN_ROUTER_REASONING_BUDGET_MODELS`, the fetcher override sets `supportsReasoningBinary`, and the test now asserts both reasoning flags.
- Bedrock global inference: model is in `BEDROCK_GLOBAL_INFERENCE_MODEL_IDS` with a test that `getModel()` emits `global.anthropic.claude-fable-5` when the toggle is on.
- Requesty: new `fetchers/__tests__/requesty.spec.ts` hits the Fable 5 override branch directly.
- 1M context on Bedrock/Vertex: omission from the beta-header allowlists is consistent with native 1M, per the thread replies.
- maxTokens split across provider registries matches the existing opus pattern.

Looks good from here.

PR #567, opened 2026-06-11
fix: apply apiRequestTimeout consistently across providers

### LINKED ISSUE #565: [BUG] apiRequestTimeout is not applied consistently across providers

Changing `zoo-code.apiRequestTimeout` does not consistently affect all API providers. Some provider clients pass `timeout: getApiRequestTimeout()` to the underlying SDK/client, while others do not, so the global timeout setting can behave differently depending on the selected provider.

This affects users running long-running model requests, slow local models, router-based providers, or provider backends where responses can take longer than the default timeout. Users expect `apiRequestTimeout` to be a global setting, but switching providers may silently change whether that setting is honored.

Related: #101 reported a similar timeout symptom for long LM Studio / OpenAI Compatible workflows. This issue is broader: it tracks provider-level consistency of the timeout configuration.

### Reproduction steps

1. Set a custom timeout in VS Code settings, for example:

   ```json
   {
     "zoo-code.apiRequestTimeout": 3600
   }
   ```
2. Run or inspect a provider where the client is initialized with timeout: getApiRequestTimeout(), for example:
    - src/api/providers/openai.ts
    - src/api/providers/base-openai-compatible-provider.ts
    - src/api/providers/lm-studio.ts
3. Compare with providers/client wrappers that initialize API clients without passing the same timeout setting, for example:
    - src/api/providers/router-provider.ts
    - src/api/providers/openai-native.ts
    - src/api/providers/unbound.ts
    - src/api/providers/anthropic.ts
    - src/api/providers/openai-compatible.ts
4. Use one of the providers that does not pass the configured timeout and trigger a long-running request, or audit the provider constructor/request path.
5. Observe that the configured global timeout may not be applied in the same way as with providers that explicitly pass timeout: getApiRequestTimeout().

### Expected result

`zoo-code.apiRequestTimeout` should be honored consistently by all providers that perform API/network requests, unless a provider cannot support it. In unsupported cases, the limitation should be documented or handled explicitly.

### Actual result

Only some providers appear to wire `getApiRequestTimeout()` into their SDK/client configuration. Other providers initialize clients or request wrappers without passing the configured timeout, so changing the global timeout setting can affect some providers but not others.

This was found by source inspection rather than a single runtime-only repro.

Providers/classes that appear to honor the setting:
- `OpenAiHandler`
- `BaseOpenAiCompatibleProvider`
- `LmStudioHandler`

Providers/classes that may need review:
- `RouterProvider` and providers that extend it, such as LiteLLM / Vercel AI Gateway paths
- `OpenAiNativeHandler`
- `UnboundHandler`
- `AnthropicHandler`
- `OpenAICompatibleHandler` using the AI SDK wrapper

There may be additional provider-specific request paths that need the same audit.

### App Version

v3.58.1

### API Provider (optional)

Not Applicable / Other

### Model Used (optional)

_No response_

_No response_

```shell

```

daewoongoh:
### Related GitHub Issue

Closes: #565

### Description

This PR makes the `zoo-code.apiRequestTimeout` setting behave consistently across API providers.

Previously, only a subset of providers explicitly propagated `getApiRequestTimeout()` into their underlying SDK/client configuration, while others relied on default client behavior. As a result, changing the global timeout setting could produce different runtime behavior depending on the selected provider.

### Test Procedure
```bash
cd src && npx vitest run api/providers/__tests__/
```

### Pre-Submission Checklist

### Screenshots / Videos

N/A

### Documentation Updates

### Additional Notes

This change is focused on consistency rather than introducing new timeout behavior. The goal is to ensure that the global `apiRequestTimeout` setting is applied uniformly across provider implementations so users receive predictable behavior regardless of the selected provider.

### Providers without client-level timeout support
| Provider (UI label) | File | SDK / Transport | Reason |
| --- | --- | --- | --- |
| Amazon Bedrock | `src/api/providers/bedrock.ts` | `@aws-sdk/client-bedrock-runtime` | AWS SDK uses its own option keys; already configured with `requestTimeout: 0`. Streaming semantics differ from OpenAI/Anthropic and are handled separately. |
| Google Gemini | `src/api/providers/gemini.ts` | `@google/genai` | Google GenAI SDK has no client-level `timeout`; only per-request `httpOptions.timeout` is supported. Would require threading the value into every call site. |
| GCP Vertex AI | `src/api/providers/vertex.ts` | `@google/genai` (extends `GeminiHandler`) | Same constraint as Google Gemini. |
| Mistral | `src/api/providers/mistral.ts` | `@mistralai/mistralai` | SDK does not expose a `timeout` option on the client. |
| Ollama | `src/api/providers/native-ollama.ts` | `ollama` npm package | Package handles timeouts internally (see in-file comment). |
| VS Code LM API | `src/api/providers/vscode-lm.ts` | VS Code Language Model API | Built-in API has no timeout option. |
| Poe | `src/api/providers/poe.ts` | `ai-sdk-provider-poe` (Vercel AI SDK) | `createPoe` has no client-level `timeout`; only per-call `abortSignal` is available. |

> The internal `OpenAI Compatible (AI SDK)` base class (`src/api/providers/openai-compatible.ts`, used by Moonshot) shares the Vercel AI SDK constraint above. The user-facing **OpenAI Compatible** option in the settings UI maps to a different handler (`src/api/providers/openai.ts`) and **does** honor `apiRequestTimeout`.
>
> `src/api/providers/fake-ai.ts` is a test fixture and is intentionally excluded.

### Get in Touch
hehegwk_23849

## Summary by CodeRabbit

* **Bug Fixes**
  * Apply a computed API request timeout consistently across AI providers, ensuring requests always use a bounded timeout.

* **Documentation**
  * Updated `apiRequestTimeout` descriptions (including translations) to show default/range (600, 1–3600), remove “0 disables timeout” wording, and explicitly list providers where the setting has no effect.

* **Tests**
  * Updated unit tests and mocks to always expect a numeric timeout and to provide stable configuration reads during initialization.

* **Chores**
  * Tightened extension setting schema (`type` to integer, minimum to 1).


src/package.nls.json
```diff
@@ -38,7 +38,7 @@
...
 	"settings.maximumIndexedFilesForFileSearch.description": "Maximum number of files to index for the @ file search feature. Higher values provide better search results in large projects but may use more memory. Default: 10,000.",
 	"settings.useAgentRules.description": "Enable loading of AGENTS.md files for agent-specific rules (see https://agent-rules.org/)",
-	"settings.apiRequestTimeout.description": "Maximum time in seconds to wait for API responses (0 = no timeout, 1-3600s, default: 600s). Higher values are recommended for local providers like LM Studio and Ollama that may need more processing time.",
+	"settings.apiRequestTimeout.description": "Maximum time in seconds to wait for API responses (0 = no timeout, 1-3600s, default: 600s). Higher values are recommended for local providers that may need more processing time. Unsupported providers (setting has no effect): Amazon Bedrock, Google Gemini, GCP Vertex AI, Mistral, Ollama, VS Code LM API, Poe.",
```
edelauna: Two issues with this unsupported-providers list:

1. **"GCP Vertex AI" is now partially incorrect** — `apiRequestTimeout` works for Claude models on Vertex (routed to `AnthropicVertexHandler`, which this PR wires). It only has no effect for Gemini models on Vertex (routed to `VertexHandler`). Could this be scoped to `"GCP Vertex AI (Gemini models)"` to avoid confusing users on the Anthropic/Claude Vertex path?

2. **Moonshot is missing** — the PR description notes that `MoonshotHandler` extends `OpenAICompatibleHandler` which uses the Vercel AI SDK and has no client-level timeout support. Should Moonshot appear in this list too?

src/api/providers/__tests__/anthropic-vertex.spec.ts
```diff
@@ -64,6 +64,7 @@ describe("VertexHandler", () => {
 			expect(AnthropicVertex).toHaveBeenCalledWith({
 				projectId: "test-project",
 				region: "us-central1",
+				timeout: expect.any(Number),
```
edelauna: This only tests the bare construction path (`{ projectId, region, timeout }`). `anthropic-vertex.ts` has two other branches — `parsedVertexCredentials` and `vertexKeyFile` — that also pass `timeout`. Are those covered somewhere, or worth adding here?

src/api/providers/anthropic.ts
```diff
@@ -44,6 +45,7 @@ export class AnthropicHandler extends BaseProvider implements SingleCompletionHa
 		this.client = new Anthropic({
 			baseURL: this.options.anthropicBaseUrl || undefined,
 			[apiKeyFieldName]: this.options.apiKey,
+			timeout: getApiRequestTimeout(),
```
edelauna: The `apiRequestTimeout` setting has `"type": "number"` in `package.json`, so users can enter decimal values like `16.1`. `16.1 * 1000` produces a non-integer float, and the Anthropic SDK calls `validatePositiveInteger("timeout", ...)` in its constructor — which would throw for any decimal input. Would `Math.round` in `getApiRequestTimeout()` (or tightening the schema to `"type": "integer"`) be the right fix?

edelauna (commented):
Nice! Thanks for this contribution.

Do you think it would be worth filing an issue to ultimately update the base class to enforce timeout going forward?

daewoongoh:
> Nice! Thanks for this contribution.
> 
> Do you think it would be worth filing an issue to ultimately update the base class to enforce timeout going forward?

Good point. I agree that enforcing timeouts in a common base class would be preferable in the long run.

I looked into that direction, but the current implementations don't all share the same base class and some rely on different SDKs/providers with provider-specific timeout handling. Because of that, I scoped this PR to the immediate fix rather than a broader architectural refactor.

I'd be supportive of opening a follow-up issue to explore a more unified approach.

src/package.nls.json
daewoongoh: I updated the unsupported-provider wording across the NLS files so Vertex is scoped to Gemini models only, and added Moonshot to the unsupported list. The text now uses `GCP Vertex AI (Gemini)` rather than broadly listing all of GCP Vertex AI, so the Claude-on-Vertex path remains correctly represented as supported via `AnthropicVertexHandler`.

src/api/providers/__tests__/anthropic-vertex.spec.ts
daewoongoh: I added explicit coverage for both additional Anthropic Vertex constructor branches: `vertexJsonCredentials` and `vertexKeyFile`.

The tests now assert that `GoogleAuth` receives the expected credentials/keyFile options and that the resulting `AnthropicVertex` constructor config includes `timeout`. I also adjusted the `GoogleAuth` mock to use a function expression so it can be invoked with `new`, matching the handler path.

src/api/providers/anthropic.ts
daewoongoh: I fixed this in two places: the configuration schema now uses `type: "integer"`, and `getApiRequestTimeout()` rounds the computed milliseconds before passing the value to SDK clients. I also centralized the timeout bounds/default handling around the documented 1–3600s range, so invalid or out-of-range values fall back to the default instead of propagating an SDK-invalid timeout value.

daewoongoh:
@edelauna 

Thanks for the review. I pushed follow-up commits addressing the timeout-related comments:

- Scoped the unsupported-provider wording to `GCP Vertex AI (Gemini)` and added Moonshot.
- Removed the local-provider examples that conflicted with the unsupported-provider list.
- Changed the timeout setting schema to `integer`, centralized the documented 1–3600s bounds, and rounded milliseconds before passing values to SDK clients.
- Added Anthropic Vertex tests for both `vertexJsonCredentials` and `vertexKeyFile` paths to verify timeout passthrough.
- Updated the `GoogleAuth` mock so it can be constructed with `new` in those tests.

I also removed the previous `0 = no timeout` wording from the docs/validation. The SDKs we pass this value to, including OpenAI and Anthropic, treat `timeout` as a millisecond request timeout value and do not expose `0` as a supported cross-provider “disable timeout” mode. Passing `0` would therefore not reliably mean “no timeout” and could result in immediate or otherwise invalid timeout behavior depending on the client. To keep the setting predictable across providers, this PR now documents and validates `apiRequestTimeout` as an integer in the 1–3600 second range.

src/api/providers/utils/timeout-config.ts
```diff
@@ -1,26 +1,19 @@
...
-	if (configTimeout <= 0) {
-		return undefined
-	}
+export function getApiRequestTimeout(): number {
+	const configTimeout = vscode.workspace
+		.getConfiguration(Package.name)
+		.get<number>("apiRequestTimeout", DEFAULT_TIMEOUT_SECONDS)
 
-	return configTimeout * 1000 // Convert to milliseconds
+	const seconds = isValidTimeout(configTimeout) ? configTimeout : DEFAULT_TIMEOUT_SECONDS
```
edelauna, line 17: Previously, setting `apiRequestTimeout: 0` returned `undefined` and the SDK used its own default (no explicit cap) — this was explicitly documented as `"0 = no timeout"` in the old i18n string. Now it silently falls back to 600s. Worth a CHANGELOG note so users who relied on that escape hatch know what changed.

src/api/providers/openai-codex.ts
```diff
@@ -371,6 +372,7 @@ export class OpenAiCodexHandler extends BaseProvider implements SingleCompletion
 						apiKey: accessToken,
 						baseURL: CODEX_API_BASE_URL,
 						defaultHeaders: codexHeaders,
+						timeout: getApiRequestTimeout(),
```
edelauna: Every other provider calls `getApiRequestTimeout()` once in the constructor and stores it in the client. Here it's re-read from VS Code config on every request (because the client is created lazily). Is the per-request re-evaluation intentional? If not, a `private readonly timeoutMs = getApiRequestTimeout()` field set at construction would align this with the other providers.

src/api/providers/__tests__/openrouter.spec.ts
```diff
@@ -108,6 +114,7 @@ describe("OpenRouterHandler", () => {
 				"X-Title": "Zoo Code",
 				"User-Agent": `ZooCode/${Package.version}`,
 			},
+			timeout: expect.any(Number),
```
edelauna: Could this use a concrete value instead of `expect.any(Number)`? As written, a mutation replacing `getApiRequestTimeout()` with `0` (which the OpenAI SDK treats as "abort immediately") would still pass. The pattern in `openai-timeout.spec.ts` — mocking `getApiRequestTimeout` to return a sentinel and asserting the exact value — catches that class of bug.

src/api/providers/__tests__/requesty.spec.ts
```diff
@@ -82,6 +82,7 @@ describe("RequestyHandler", () => {
 				"X-Title": "Zoo Code",
 				"User-Agent": `ZooCode/${Package.version}`,
 			},
+			timeout: expect.any(Number),
```
edelauna: Same question as the openrouter test — `expect.any(Number)` accepts `0` or any hardcoded literal. Worth pinning to a specific sentinel so mutations are detectable.

src/api/providers/__tests__/anthropic-vertex.spec.ts
```diff
@@ -66,7 +78,58 @@ describe("VertexHandler", () => {
...
```
edelauna: Same `expect.any(Number)` concern here (and lines 108 and 130). If `timeout: getApiRequestTimeout()` were replaced with `timeout: 0`, all three assertions would still pass.

src/package.nls.json
```diff
@@ -39,7 +39,7 @@
 	"settings.autoImportSettingsPath.description": "Path to a ZooCode configuration file to automatically import on extension startup. Supports absolute paths and paths relative to the home directory (e.g. '~/Documents/roo-code-settings.json'). Leave empty to disable auto-import.",
 	"settings.maximumIndexedFilesForFileSearch.description": "Maximum number of files to index for the @ file search feature. Higher values provide better search results in large projects but may use more memory. Default: 10,000.",
 	"settings.useAgentRules.description": "Enable loading of AGENTS.md files for agent-specific rules (see https://agent-rules.org/)",
-	"settings.apiRequestTimeout.description": "Maximum time in seconds to wait for API responses (0 = no timeout, 1-3600s, default: 600s). Higher values are recommended for local providers like LM Studio and Ollama that may need more processing time.",
+	"settings.apiRequestTimeout.description": "API response timeout (seconds, default: 600, range: 1–3600). Higher values are recommended for local providers. Unsupported providers: Amazon Bedrock, GCP Vertex AI (Gemini), Google Gemini, Mistral, Moonshot, Ollama, Poe, VS Code LM API.",
```
edelauna: "GCP Vertex AI (Gemini)" in the unsupported list might confuse users running **Anthropic Claude** on Vertex — `AnthropicVertexHandler` does wire `getApiRequestTimeout()` now (see `anthropic-vertex.ts:47`), so the timeout _is_ effective for that path. Something like "Google Gemini (including Vertex AI)" might make the distinction clearer.

edelauna (changes requested):
Hey @daewoongoh thanks again for addressing the feedback, after some testing I have a few more comments.

src/api/providers/utils/timeout-config.ts
daewoongoh: I added a changeset note documenting the behavior change: Updated `apiRequestTimeout` validation. Values must be integers between 1 and 3600 seconds; invalid or out-of-range values, including `0`, now fall back to 600 seconds. This aligns with the SDK's default timeout value.

src/api/providers/openai-codex.ts
daewoongoh: Not intentional. I updated OpenAI Codex to cache the timeout at handler construction time and pass that cached value when the lazy client is created, so it now matches the behavior of the other providers.

src/api/providers/__tests__/openrouter.spec.ts
daewoongoh: Updated. I mocked `getApiRequestTimeout()` with a concrete sentinel value and now assert that OpenRouter passes that exact value to the OpenAI client instead of only checking `expect.any(Number)`.

src/api/providers/__tests__/requesty.spec.ts
daewoongoh: Updated Requesty as well. Both client-construction assertions now verify the exact mocked timeout value rather than accepting any number.

src/api/providers/__tests__/anthropic-vertex.spec.ts
daewoongoh: Updated the Anthropic Vertex tests to use the same concrete timeout sentinel across the bare, JSON credentials, and key file credentials paths.

src/package.nls.json
daewoongoh: Agreed. I updated the unsupported-provider wording to `Google Gemini(directly or through the Vertex AI platform)` across the NLS files so Claude-on-Vertex users do not read this as applying to Anthropic Vertex.

edelauna (approved):

PR #569, opened 2026-06-11
fix(gateway): surface in-stream errors from Zoo and Vercel AI gateways

### LINKED ISSUE #587: [BUG] Provider errors raised mid-stream are not propagated to the extension

Errors raised by the upstream model provider after streaming has started are not propagated to the extension. Instead of the real provider error, the user sees a generic `the model did not provide a response`.

Affects users on the Zoo managed gateway and the Vercel AI Gateway provider whenever the provider fails mid-response (rate limits, upstream server errors, content filtering, timeouts, etc.). The failure is logged on the gateway side but never reaches the extension UI.

### Reproduction steps

1) Environment: Zoo Code extension on the Zoo or Vercel AI gateway, any streaming-capable model.
2) Send a request that triggers an upstream provider error after the response has begun streaming (e.g. exhaust a provider rate limit so the provider returns `429 Too many requests`).
3) Observe the extension shows only a generic "no response" error, while the gateway logs show the actual provider error message.

### Expected result

The extension surfaces the provider's actual error message, following the same path as HTTP-thrown errors (including the Zoo gateway sign-in / add-credits / budget prompts where applicable).

### Actual result

The extension shows `error, the model did not provide a response` and the real provider error is silently dropped.

### App Version

v3.58.1

### API Provider (optional)

OpenAI Compatible

```
Too many requests, please wait before trying again
```

---

**Technical root cause (for the linked PR #569):** Once a gateway response begins streaming, the HTTP status is already committed as `200`, so the provider reports failures as an in-band error chunk (`{ error: { message, status?, code? } }`) rather than a thrown HTTP error. `zoo-gateway.ts` and `vercel-ai-gateway.ts` only read `chunk.choices[0]?.delta` and ignored chunks containing an `error`, so the stream ended empty. `openrouter.ts` already guards this with an `if ("error" in chunk)` check; these two handlers did not.

JamesRobert20:
### Related GitHub Issue

Closes: #587

### Description

When a gateway response starts streaming, the HTTP status is already committed as `200`. Errors raised by the upstream model provider mid-stream are therefore delivered as an **in-band error chunk** on the SSE stream rather than as a thrown HTTP error.

Both the Zoo Gateway and the Vercel AI Gateway handlers iterated chunks and only read `chunk.choices[0]?.delta`, silently ignoring any chunk shaped like `{ error: ... }`. The stream then ended with no content, so the extension surfaced a generic "the model did not provide a response" message even though the real provider error was visible in the gateway logs. This affects any mid-stream provider error (server errors, content filtering, upstream timeouts, rate limits, etc.) — a `429 Too Many Requests` is just the example that surfaced it. `openrouter.ts` already guards against this with an `if ("error" in chunk)` check — these two handlers did not.

**How this is fixed:**

- **`zoo-gateway.ts`** — detect the in-stream error chunk and rebuild it via a new `toGatewayStreamError` helper into an `Error` carrying `status`/`code`. Because it is thrown from inside the existing `try` block, the current `surfaceGatewayApiError` / `classifyGatewayApiError` path handles it unchanged (sign-in on 401, add-credits on 402, budget prompt on budget 429, contact-support on 403), and the upstream message reaches the user for everything else.
- **`vercel-ai-gateway.ts`** — detect the in-stream error chunk and throw the upstream message, or a default fallback when the chunk carries no message (this handler has no Zoo-specific auth/credit UX).

Note: plain rate-limit `429`s deliberately do **not** raise a toast (`classifyGatewayApiError` returns `none`); the real message now surfaces inline in the chat, which matches the existing tested behavior for non-budget 429s.

### Test Procedure

Unit tests, scoped:

```
pnpm exec vitest run api/providers/__tests__/zoo-gateway.spec.ts api/providers/__tests__/vercel-ai-gateway.spec.ts
```

Added coverage:

- Zoo Gateway throws the upstream reason when an in-stream `error` chunk is received.
- Zoo Gateway still surfaces the add-credits prompt when the in-stream error carries a budget code.
- `toGatewayStreamError` maps `message`/`status`/`code` and falls back to a default message.
- Vercel AI Gateway throws the upstream reason from an in-stream `error` chunk, and a default message when the chunk has no message.

All specs pass; `check-types` and `lint` are green across the workspace.

### Pre-Submission Checklist

- [x] **Scope**: Changes are focused on the gateway in-stream error bug.
- [x] **Self-Review**: Performed a thorough self-review.
- [x] **Testing**: New unit tests added for both handlers.
- [x] **Documentation Impact**: No user-facing docs needed.
- [x] **Contribution Guidelines**: Read and followed.

### Documentation Updates

### Additional Notes

This PR pairs with a server-side change in the website repo that emits the in-band `error` part on the SSE stream (the AI SDK reports upstream failures as an `error` part on `fullStream` instead of throwing). This extension change makes the client honor those chunks.


.changeset/gateway-stream-error-propagation.md
```diff
@@ -0,0 +1,5 @@
+---
+"zoo-code": patch
```
navedmerchant: We do not need a changeset file

navedmerchant (changes requested):
Please link an issue number and remove changeset, otherwise looks good to merge

navedmerchant (approved):

JamesRobert20:
Both points addressed:

- Removed the changeset file.
- Opened and linked #587 (filed against the bug template) — Closes: #587 is now in the description.

Issue #570, opened 2026-06-11
[ENHANCEMENT] Align openai, requesty, unbound, and opencode-go providers to use shared extractReasoningFromDelta helper


edelauna:
### Problem (one or two sentences)

Providers `openai.ts`, `requesty.ts`, `unbound.ts`, and `opencode-go.ts` each have their own inline `reasoning_content` check with no fallback to `reasoning`, and none use the shared `extractReasoningFromDelta` helper introduced in PR #449. If the fallback logic or whitespace-preservation behavior ever changes, these providers will silently drift out of sync.

### Context (who is affected and when)

Affects users of OpenAI, Requesty, Unbound, and OpenCode-Go providers when a model streams reasoning via the `reasoning` field (rather than `reasoning_content`). The inconsistency was identified during review of #449 and intentionally deferred to keep that PR focused.

### Desired behavior (conceptual, not technical)

All providers that handle streaming reasoning deltas should use `extractReasoningFromDelta` from `src/api/providers/utils/extract-reasoning.ts`, so fallback behavior (`reasoning_content` → `reasoning`) and whitespace preservation are consistent everywhere.

### Constraints / preferences (optional)

Changes should be limited to swapping the inline checks for the shared helper — no behavioral changes beyond gaining the `reasoning` fallback and whitespace preservation that `extractReasoningFromDelta` already provides.

### Request checklist

- [x] I've searched existing Issues and Discussions for duplicates
- [x] This describes a specific problem with clear context and impact

### Acceptance criteria (optional)

Given any provider that streams reasoning deltas,
When a model sends `reasoning_content: null` and a populated `reasoning` field,
Then a `{ type: "reasoning", text: ... }` chunk is yielded (same as LiteLLM and BaseOpenAiCompatibleProvider).
And whitespace-only chunks like `" "` and `"\n\n"` are preserved.

### Proposed approach (optional)

For each of `openai.ts`, `requesty.ts`, `unbound.ts`, `opencode-go.ts`: import `extractReasoningFromDelta` and replace the existing inline `reasoning_content` check with a call to the helper, matching the pattern in `lite-llm.ts:239-241`.

### Trade-offs / risks (optional)

Low risk — the helper is already tested and in use. The only behavioral delta is that these providers gain the `reasoning` fallback they were previously missing; existing `reasoning_content` handling is unchanged.

_Tracked from PR #449 review._

daewoongoh:
Claiming

PR #588, opened 2026-06-12
refactor(providers): use extractReasoningFromDelta helper for reasoni…

daewoongoh:
### Related GitHub Issue

#570

### Description

This PR replaces duplicated inline reasoning-delta extraction logic across seven providers with the shared `extractReasoningFromDelta` helper.

Affected providers:

- `deepseek`
- `mimo`
- `openai`
- `opencode-go`
- `qwen-code`
- `requesty`
- `unbound`

Previously, each provider had a near-identical streaming check for `delta.reasoning_content` and emitted reasoning chunks inline. Since `src/api/providers/utils/extract-reasoning.ts` already provides the canonical extraction behavior and is already used by other OpenAI-compatible providers, this PR brings the remaining duplicated call sites in line with the shared implementation.

### Test Procedure

```bash
cd src && npx vitest run api/providers/__tests__/
```

### Pre-Submission Checklist

  _Existing tests cover the changed provider behavior, and the shared helper is already test-covered._

### Screenshots / Videos

N/A

### Documentation Updates

### Additional Notes

This is a focused provider refactor. It removes duplicated reasoning extraction logic and routes all seven affected providers through the shared `extractReasoningFromDelta` helper.

The intended behavior for existing `reasoning_content` chunks is unchanged for normal string payloads. The practical improvement is that these providers now also handle the `reasoning` fallback supported by the helper, so future or model-specific deltas using that field will stream reasoning text instead of being ignored.

### Get in Touch

hehegwk_23849

## Summary by CodeRabbit

* **Refactor**
  * Streamed “reasoning” text extraction was unified across multiple provider integrations, ensuring consistent emission of `reasoning` chunks using a shared approach (including correct field precedence and fallback).
* **Tests**
  * Added/extended unit tests for each provider to validate streaming reasoning parsing, covering emission, fallback when the primary field is missing, and preference when both reasoning fields are present.


src/api/providers/__tests__/openai.spec.ts
```diff
@@ -219,6 +219,45 @@ describe("OpenAiHandler", () => {
...
+				},
+			}))
+
+			const chunks: any[] = []
+			for await (const chunk of handler.createMessage(systemPrompt, messages)) {
+				chunks.push(chunk)
+			}
+
+			expect(chunks).toContainEqual({ type: "reasoning", text: "router-style thought" })
+		})
```
edelauna, line 261: Does the helper priority rule get exercised here? If a delta has *both* `reasoning_content` and `reasoning` populated, the expected winner is `reasoning_content` — but that case isn't covered at the provider level. The unit tests for `extractReasoningFromDelta` cover it, but a provider-level check would catch a future regression where the helper call gets bypassed or re-inlined.

edelauna (commented):
Thanks for adding this, had one question related to test coverage

src/api/providers/__tests__/openai.spec.ts
daewoongoh: Good catch. I added a provider-level priority test to cover the case where both `delta.reasoning_content` and `delta.reasoning` are present. The test asserts that only `reasoning_content` is emitted as the reasoning chunk, so it should catch regressions where the helper is bypassed or the logic is re-inlined incorrectly.

edelauna (approved):
Thank you for taking on this cleanup.

PR #589, opened 2026-06-12
fix(diff-view): restore scroll position and fix tab handling on save/deny

### LINKED ISSUE #586: [BUG] Diff view scroll position not preserved -- viewport jumps on open and after Save/Deny
# [BUG] Diff view scroll position not preserved -- viewport jumps on open and after Save/Deny

When Zoo Code applies a file edit and opens the diff editor, the viewport is not reliably
positioned at the changed region. After the user clicks "Save" or "Deny", the viewport
always jumps to the top of the file, destroying their scroll position.

---

Affects all users working on files longer than a screen height. The disruption compounds
in long files and in sessions with many edits. Two distinct failure points:

1. The diff editor opens at the wrong scroll position -- behavior depends inconsistently
   on whether the edit is a pure addition, pure deletion, or mixed.
2. After clicking "Save" or "Deny", the viewport unconditionally resets to the top of
   the file, regardless of where the user was before the diff appeared.

Additionally, files that were not already open before the diff are left as open tabs
after Save/Deny, polluting the user's tab bar.

---

## Reproduction steps

**Note**: When testing manually, instruct Zoo to use the diff tool so that it does not attempt to use shell commands to apply the edits.

### Scenario A -- file is ALREADY OPEN in the editor

| Operation | Diff scroll behavior |
|-----------|----------------------|
| Add 1 line at end | Does NOT scroll to change |
| Remove 1 line at end | Does NOT scroll to change |
| Remove 1 line + add 1 line | DOES scroll to change |

Steps:
1. Open a file of several hundred lines in VS Code.
2. Ask Zoo Code to append a line near the end of the file (pure addition).
3. Observe: diff editor opens with viewport at the top, not the changed line.
4. Click "Save" or "Deny".
5. Observe: viewport jumps to the top of the file.
6. Repeat steps 2-5 with a pure deletion and then a mixed remove+add edit.

### Scenario B -- file is NOT already open

| Operation | Diff scroll behavior |
|-----------|----------------------|
| Remove 1 line at end | Opens file -- does NOT scroll to change |
| Add 1 line at end | Opens file -- scrolls to end (correct) |
| Remove 1 line + add 1 line | Opens file -- does NOT scroll to change |

Steps:
1. Close the target file if it is open.
2. Ask Zoo Code to edit a region near the end of the file.
3. Observe: diff editor opens; viewport position depends on operation type (see table).
4. Click "Save" or "Deny".
5. Observe: viewport jumps to the top. The file tab remains open even though it was not
   open before Zoo Code intervened.

---

## Expected result

- The diff editor always opens with the viewport positioned at the first changed line,
  regardless of whether the change is a pure addition, pure deletion, or mixed.
- After "Save": if the file was open before the diff, the viewport is restored to its
  pre-diff scroll position. If the file was not open before, the tab is closed.
- After "Deny": same behavior as "Save".

---

## Actual result

- The diff editor inconsistently positions the viewport. Pure additions and pure
  deletions do not scroll to the change; only mixed (remove + add) edits sometimes do.
- After "Save" or "Deny", the viewport always resets to the top of the file.
- Files that were not previously open remain as open tabs after Save/Deny, polluting the
  tab bar.

---

## Variations tried

Occurs regardless of API provider or model. Observed with both `write_to_file` and
`apply_diff` tool calls.

---

## App Version

_(fill in current Zoo Code version before filing)_

---

## API Provider

Not Applicable -- reproducible with any provider and model.

---

## Upstream prior art

This is a long-standing unresolved bug inherited from the Roo-Code codebase. Related
upstream issues (none resolved, none fixed):

### Scroll / viewport

| Issue | Title | Status |
|-------|-------|--------|
| [RooCodeInc/Roo-Code#8112](https://github.com/RooCodeInc/Roo-Code/issues/8112) | Preserve viewport when applying diffs to the active editor (avoid re-open) | OPEN |
| [RooCodeInc/Roo-Code#12084](https://github.com/RooCodeInc/Roo-Code/issues/12084) | After each modification is completed, the mouse/focus jumps to the top of the code | OPEN |
| [RooCodeInc/Roo-Code#10142](https://github.com/RooCodeInc/Roo-Code/issues/10142) | Diff view does not auto-scroll to bottom when streaming long content | OPEN |
| [RooCodeInc/Roo-Code#9282](https://github.com/RooCodeInc/Roo-Code/issues/9282) | Restore editor scroll position after closing diff view | CLOSED (no fix) |
| [RooCodeInc/Roo-Code#6853](https://github.com/RooCodeInc/Roo-Code/issues/6853) | Diffs do not return to cursor position | CLOSED (no fix) |
| [RooCodeInc/Roo-Code#7996](https://github.com/RooCodeInc/Roo-Code/issues/7996) | File scroll to edit location after "User Edits" in diff | CLOSED (narrow scope) |

### Tab management (related UX degradation)

| Issue | Title | Status |
|-------|-------|--------|
| [RooCodeInc/Roo-Code#6003](https://github.com/RooCodeInc/Roo-Code/issues/6003) | Auto-close Roo Tabs | CLOSED (no fix) |
| [RooCodeInc/Roo-Code#11326](https://github.com/RooCodeInc/Roo-Code/issues/11326) | Pinned tabs are unpinned when edited by RooCode | OPEN |
| [RooCodeInc/Roo-Code#6010](https://github.com/RooCodeInc/Roo-Code/issues/6010) | Add option to disable/enable focus on diff view | CLOSED (no fix) |
| [RooCodeInc/Roo-Code#6005](https://github.com/RooCodeInc/Roo-Code/issues/6005) | Open Tabs in Correct Tab Group if Auto Focus Disabled | CLOSED (no fix) |

### Root cause note (from #8112)

> Re-opening the document via the VS Code diff API resets the viewport to line 1.
> When the target file is already the active editor, re-opening it resets the scroll
> position unconditionally. The prescribed fix: when the target file is the currently
> active editor, avoid re-opening it; apply the change in-place to preserve the viewport.
> When the file is not already open, continue using the diff UI but scroll the diff
> editor to the first changed line before displaying it.

The upstream proposal for tab cleanup (#6003) introduced a `PostEditBehaviorUtils` class
inside `DiffViewProvider` to centralize post-edit tab tracking and closure, with two
user settings (`autoCloseRooTabs`, `autoCloseAllRooTabs`) -- but this was never merged.

awschmeder:
### Related GitHub Issue

Closes: #586

### Description

Fixes the diff-view scroll and tab-handling problems described in #586.
The core logic changes are confined to [`DiffViewProvider.ts`](../src/integrations/editor/DiffViewProvider.ts);
no tool files needed to change because the default diff path already calls
`open()`/`update()`/`scrollToFirstDiff()`/`saveChanges()` in the right order.

The "how", broken down by the symptoms in the issue:

1. **Diff opens at the wrong scroll position.**
   - Removed the unconditional `scrollEditorToLine(0)` in `open()`. It fired after the
     diff rendered and overrode whatever position should have been shown, leaving
     off-screen edits out of view.
   - `scrollToFirstDiff()` is refactored into three private helpers:
     - `findFirstDiffLine()` walks `diff.diffLines()` to compute the first-change line and
       clamps it to a valid index so pure end-of-file removals (where the computed index
       lands past `document.lineCount - 1`) are still brought into view.
     - `resolveLiveEditor()` re-resolves the modified-side editor from
       `vscode.window.visibleTextEditors` at scroll time, replacing any stale/detached
       `activeDiffEditor` reference whose `visibleRanges` no longer reflect the on-screen
       diff widget. A `revealRange` on a stale reference is a silent no-op; this was the
       root cause of "scrolls to top" when the file was already open.
     - `revealDiffLine()` anchors the editor selection on the target line before calling
       `revealRange`. The selection nudge is necessary when the target is already inside
       the current viewport -- a bare `revealRange` on a visible line is a no-op.
   - A deferred re-reveal fires 100 ms after the synchronous one. The `vscode.diff`
     command re-lays-out the diff widget after a short delay; the deferred reveal ensures
     the scroll sticks after that layout pass. It is gated on `activeDiffEditor === editor`
     so a stale timer from a previous edit never acts on a later edit's diff widget.

2. **Viewport jumped to the top after Save/Deny.**
   - `open()` now captures `preEditScrollLine` (the first visible line) BEFORE the
     tab-closing loop -- after that loop the editor is no longer in `visibleTextEditors`
     and the capture would always be `undefined`.
   - Both accept and deny paths route through `showEditedFileWithoutDisruptingFocus()`,
     which calls `revealRange(..., AtTop)` to restore the pre-edit scroll position.
   - `cancelDeferredScroll()` is called at the start of both `saveChanges()` and
     `revertChanges()` before any programmatic editor activation. With auto-approve writes,
     `saveChanges()` runs in the tight window right after `scrollToFirstDiff()`'s deferred
     timer is set; the cancellation prevents the late reveal from firing after the viewport
     is restored and yanking the user back to the diff target.

3. **Tabs left open for files that were not previously open.**
   - `saveChanges()` previously re-showed the file unconditionally; `closeAllDiffViews()`
     only closes diff-scheme tabs, so the plain file tab leaked. The re-show is now gated.
   - Added `closeFileTab()` to close the transiently opened plain file tab when the file
     was not open before the edit AND the user did not interact with it.
   - Added a `userTouchedDocument` signal: while the diff is open for a
     not-previously-open file, an `onDidChangeActiveTextEditor` listener marks the file as
     "touched" if the user activates its own (non-diff) editor tab. Touched files are kept
     open; untouched ones are closed. The listener is disposed before any programmatic
     `showTextDocument` so our own re-show never counts as a touch, and is cleared in
     `reset()`.

**Additional fixes discovered during implementation:**

4. **Focus disruption on keep-open.** After Save/Deny, showing the edited file previously
   stole the foreground even when the user had navigated to a different file.
   `showEditedFileWithoutDisruptingFocus()` captures the user's active editor before the
   re-show, uses `preserveFocus: true`, and re-activates the user's prior editor afterward
   when they were viewing something else.

5. **Pin state lost.** Closing the target file's tab to open the diff dropped VS Code's pin
   state. `open()` now records `documentWasPinned`; after the diff closes the pin is
   restored via `workbench.action.pinEditor`.

6. **Preview-tab eviction.** Opening the diff reuses the editor group's single preview slot
   and permanently evicts unrelated preview tabs. `open()` now snapshots all unrelated
   preview tabs (with their scroll positions) before the diff opens; `restorePreviewTabs()`
   re-opens any that were evicted after the diff closes. Files that no longer exist on disk
   are skipped gracefully.

7. **Diff pane interaction keeps target tab open.**
   If the user clicks or edits inside the diff pane itself, the target file's tab is kept
   open after Save/Deny -- even if the file was not previously open. This extends the
   existing `userTouchedDocument` concept to cover diff-pane interaction via
   `onDidChangeTextEditorSelection` (filtered to `kind = Mouse | Keyboard` to exclude
   programmatic selection changes such as those from `scrollToFirstDiff()`). The listener
   compares by `document` identity rather than editor object identity to handle stale
   editor references that VS Code creates when the diff widget refreshes. If the user only
   scrolls in the diff pane, the existing close behavior is preserved.

8. **Scroll position mirrors the most-recently-viewed pane.**
   When the target file is re-revealed after a diff closes, it scrolls to whichever
   position the user last viewed:
   - If the user last scrolled in the **diff pane** -> the target file mirrors that line.
   - If the user last scrolled in the **target file's own editor** -> that position is used.
   - Otherwise -> the pre-edit scroll position is restored (existing behavior).
   Both the diff and target file editors are tracked via `onDidChangeTextEditorVisibleRanges`;
   `lastScrolledSource` records which fired most recently so the correct position wins.

9. **User-configurable auto-close settings (UI tab in Settings).**
   The auto-close decisions from item 3 above are now user-adjustable via three new settings
   exposed in **Settings > UI**:
   - **Auto-close Zoo-opened files** (`autoCloseZooOpenedFiles`, default: on) -- close
     transient file tabs Zoo opened once the diff is accepted/denied.
   - **Auto-close even after user interaction** (`autoCloseZooOpenedFilesAfterUserEdited`,
     default: off) -- override the "keep if touched" guard; only effective when the base
     toggle is also on.
   - **Auto-close newly created files** (`autoCloseZooOpenedNewFiles`, default: off) -- also
     close the tab when Zoo created the file (not just edited an existing one).
   All three are wired from the global settings schema through `SettingsView` cached state
   to `UISettings` checkboxes, and read in `DiffViewProvider.keepOrCloseEditedFile()`.
   i18n keys added under the `ui` namespace in all 18 locale files.

Reviewers should pay attention to: the ordering of `disposeActiveEditorListener()` ->
`cancelDeferredScroll()` -> `closeAllDiffViews()` -> `keepOrCloseEditedFile()` ->
`restorePreviewTabs()` in both `saveChanges()` and `revertChanges()`, the clamp logic in
`findFirstDiffLine()` and the second clamp in `revealDiffLine()`, and the stale-editor
guard in `resolveLiveEditor()`.

**Incidental change (discovered during this work):** the diff tab label constant
`DIFF_VIEW_LABEL_CHANGES` still read "Original <-> Roo's Changes". Since this is a
user-visible label in the rebranded fork, it was corrected to "Original <-> Zoo's
Changes". It is a one-line string change referenced only via the constant (tests use the
constant, not the literal), so it is behavior-neutral for tab matching. Called out here for
transparency since it is outside the strict scope of the scroll fix.

### Test Procedure

Automated:

```
cd src && npx vitest run integrations/editor/__tests__/DiffViewProvider.spec.ts
cd webview-ui && npx vitest run src/components/settings/__tests__/UISettings.spec.tsx
```

DiffViewProvider: 67 tests pass, covering:

- `scrollToFirstDiff()`: first-change line for addition-only, deletion-only, mixed, and
  end-of-file removal (clamped); selection anchor before reveal (in-viewport fix); deferred
  re-reveal fires at 100 ms; live editor resolved at reveal time (stale reference fix);
  stale timer from a previous edit does not act on a new diff editor; no-op when no editor.
- `preEditScrollLine`: captured from `visibleRanges[0].start.line` in `open()`; `undefined`
  when no `visibleRanges`; `saveChanges()` and `revertChanges()` call `revealRange(AtTop)`
  when set; no `revealRange` when `undefined`; deferred scroll cancelled so it cannot fight
  the scroll restore (auto-approve race).
- `userTouchedDocument`: `saveChanges()` and `revertChanges()` close the tab when untouched;
  keep it open when touched; activation listener fires correctly; listener ignored for diff
  tab activations.
- `userTouchedDiffEditor`: `saveChanges()` keeps the tab open when the user clicked/edited
  in the diff pane; closes it when the user only scrolled; listener registered in `open()`;
  programmatic selection changes (kind=Command or undefined) do not set the flag; stale
  editor reference resolved by document identity; unrelated editor events ignored.
- `showEditedFileWithoutDisruptingFocus()`: re-activates user's prior editor when they
  navigated away; does not re-activate when already on the edited file; re-pins and focuses
  correctly; does not pin when not originally pinned.
- Preview tab snapshot/restore: captures preview tabs with scroll positions, excludes the
  diff target; restores evicted tabs in preview mode with scroll; skips still-open tabs;
  skips files that no longer exist.
- Scroll position precedence: uses `diffScrollLine` when `lastScrolledSource` is `'diff'`;
  uses `targetFileScrollLine` when `lastScrolledSource` is `'targetFile'`; falls back to
  `preEditScrollLine` when source is undefined; no `revealRange` when all sources are
  undefined; target-file scroll overrides diff scroll regardless of values.
- `diffScrollListener` wiring: records `diffScrollLine` and sets `lastScrolledSource` to
  `'diff'` on diff editor scroll; records `targetFileScrollLine` and sets source to
  `'targetFile'` on target file scroll; ignores unrelated editors; `lastScrolledSource`
  reflects the most recent scroll between the two editors.
- Auto-close settings decision table: already-open file never auto-closed; transient tab
  kept when `autoCloseZooOpenedFiles` is false; closed when true (default); touched tab kept
  by default; closed when `autoCloseZooOpenedFilesAfterUserEdited` is true (and base is on);
  new-file tab closed when `autoCloseZooOpenedNewFiles` is true; all unset settings preserve
  existing behavior.

UISettings: auto-close checkbox tests -- renders all three checkboxes; each reflects its
prop value (true/false); each calls `setCachedStateField` with the correct key on toggle.

Manual (Extension Development Host, F5): unit tests cannot confirm the composite diff
widget visibly scrolls or that focus/tabs behave correctly across real VS Code sessions.
Full manual matrix is documented in a comment attached to this PR.

Summary: for a 100+ line file, verify the diff opens on the first changed line
and the viewport is restored / the tab is closed appropriately after Save and after Deny,
for both already-open and not-open files, across pure-add, pure-delete, and mixed edits,
including diffs at the top, midpoint, and end of the file. Confirm no focus steal, no
preview-tab loss, and no stuck-scroll across sequential edits. Also verify: clicking or
editing inside the diff pane keeps the target tab open; scrolling only in the diff pane
still closes the tab; after Save the target file scrolls to the last-viewed position in
the diff or target file editor, whichever was scrolled most recently. Verify the three
auto-close checkboxes in Settings > UI toggle the behavior as documented.

### Pre-Submission Checklist

### Screenshots / Videos

Attached in a follow-up comments.

### Documentation Updates

### Additional Notes

The fix targets the DEFAULT (diff-visible) path only. When the
`EXPERIMENT_IDS.PREVENT_FOCUS_DISRUPTION` experiment is enabled, tools call
`saveDirectly()` instead, which intentionally avoids any diff view, scroll-to-change, or
viewport restoration -- none of the code in this PR runs on that path.

### Commentary / Opinion

Automatically closing an un-touched tab after applying an edit or creating a new file could be somewhat confusing to novice coders.

On the other hand, not closing un-touched files after edits creates significant tab pollution in long sessions, leading to dozens of tabs left open.

The unclosed-tab pollution problem also burdens the context window in an uncontrolled manner and drives up inference cost, as Zoo's default includes 20 open tabs (probably set too high, by the way... 1 is a better default).

On balance, I think the best solution is to automatically close files that the user did not either explicitly open or have some affirmative interaction with by viewing / focusing on or editing. The three new settings in Settings > UI give users full control to opt out of any part of this behavior.

## Summary by CodeRabbit

* **New Features**
  * Added three new settings to control whether editor tabs opened by Zoo are automatically closed after accept/reject, optionally after user interaction, and for newly created files.
* **Bug Fixes**
  * Improved diff view behavior to open on the first changed line (with end-of-file clamping) and better preserve/restore tab focus and scroll positions when accepting/rejecting.
  * Updated the diff view label to “Original ↔ Zoo’s Changes”.
* **Documentation**
  * Added release-note metadata and UI localization for the new auto-close settings across supported languages.
* **Tests**
  * Expanded DiffViewProvider test coverage for scroll, focus, and tab restore/auto-close scenarios.


awschmeder:
# Manual Test Plan: Diff View Scroll Position + Tab/Focus Handling

Covers the fixes in [`DiffViewProvider.ts`](../src/integrations/editor/DiffViewProvider.ts)
for issue #586 and the follow-on bugs found during review. Unit tests verify which line is
targeted and which branch runs, but they cannot confirm that the composite VS Code diff
widget visibly scrolls, or that tab/focus behavior is correct in a live session. 

Run this matrix manually in the Extension Development Host.

## Bugs this plan must catch

These are the concrete defects found and fixed. Each has dedicated coverage below.

1. **In-viewport no-scroll** -- file already open and scrolled to the middle; a diff
   targeting a line already inside the viewport did not scroll (the captured editor
   reported a stale viewport, so revealRange was a silent no-op).
2. **Scroll-to-top on open** -- file already open and scrolled to the middle; the diff
   jumped to the top and never showed the change (stale/detached editor reference).
3. **End-of-file no-scroll** -- a diff targeting the last line failed to scroll.
4. **Stuck at line 0** -- once the scroll broke, every subsequent edit also scrolled to
   the top until the file was closed and reopened (leaking deferred-scroll timers acting
   on a later edit's diff editor).
5. **Auto-approve late-timer race** -- with auto-approve write enabled, a deferred scroll
   could fire after the post-save viewport restore and yank the file back to the change.
6. **Preview-tab eviction** -- an unrelated file open as a preview tab (italic,
   not-yet-edited) was permanently lost when the diff reused the editor group's single
   preview slot.
7. **Focus disruption on keep-open** -- after Save/Deny, the edited file was force-shown
   and stole the foreground even when the user had navigated to a different file.

## Prerequisites

1. Launch the Extension Development Host: open the repo in VS Code and press `F5`
   (Run Extension). A second VS Code window opens with the dev build of Zoo Code.
2. In the dev window, open a workspace containing a **long file** -- several hundred lines,
   well beyond one screen height. A file of ~400-600 lines is ideal so that "top",
   "midpoint", and "end" are all clearly off-screen from one another. Have at least two
   such files available (referred to below as `file-1` and `file-2`).
3. Confirm the `preventFocusDisruption` experiment is **OFF** (Settings ->
   Experimental). This fix only applies to the default diff-visible path; with the
   experiment on, `saveDirectly()` runs and none of this behavior applies.
4. When prompting Zoo, instruct it to **use the diff/apply_diff tool** so it does not apply
   edits via shell commands. Example: "Using the apply_diff tool, remove the last line of
   `long.ts`." This keeps the edit on the path under test.
5. Know how to toggle **auto-approve writes** -- Matrix 4 requires it ON, all other
   matrices assume it OFF (so you can observe the diff and click Save/Deny).

## What "correct" looks like

- **On diff open**: the diff editor is scrolled so the first changed line is visible
  (centered), regardless of whether the change is a pure addition, pure deletion, or mixed,
  regardless of where in the file the change is, and **regardless of where the file was
  already scrolled** (including when the change is already inside the current viewport).
- **After Save**:
  - File was already open before the edit -> the file is shown and the viewport is
    restored to the pre-edit scroll position (the saved line at the top of the viewport).
  - File was NOT open before, and the user did not interact with it -> the transient tab
    is closed (no tab pollution).
  - File was NOT open before, but the user clicked into its tab during the diff -> the tab
    stays open.
- **After Deny**: same tab keep/close and scroll-restore behavior as Save. A
  newly-created (non-existent) file is removed and its tab closed.
- **Focus**: focus is never stolen to the edited file. If you navigate to a different file
  while the diff is up, you remain on that file after Save/Deny -- you are neither pulled
  back to the edited file nor left on a stale diff tab.
- **Preview tabs**: an unrelated file open as a preview tab is not permanently lost; after
  the diff session ends it is restored as a preview tab at its prior scroll position.
- **Sequential edits**: scrolling stays correct edit after edit; a broken scroll never
  "sticks" across edits.

---

## Matrix 1 -- Diff opens on the first change (scroll-to-first-diff)

For each combination below, trigger the edit and observe ONLY the diff-open scroll
position (do not Save/Deny yet -- that is Matrix 2/3).

Edit positions to test: **line 0 / top**, **midpoint of the file**, **end of the file**.
Edit shapes to test: **pure addition**, **pure deletion**, **mixed (remove + add)**.

For the "Already open" rows, first **scroll the file to the middle** before prompting, so
the in-viewport and stale-viewport cases are exercised (bugs 1-3).

| # | File state | Position | Shape | Expected on open |
|---|-----------|----------|-------|------------------|
| 1.1 | Already open (scrolled to top) | Top | Add | Diff shows the added line near the top |
| 1.2 | Already open (scrolled to MIDDLE) | Midpoint | Add | Diff scrolls to the midpoint change (NOT the top) |
| 1.3 | Already open (scrolled to MIDDLE) | End | Add | Diff scrolls to the end change |
| 1.4 | Already open (scrolled to MIDDLE) | Midpoint | Delete | Diff scrolls to the deletion |
| 1.5 | Already open (scrolled to MIDDLE) | End | Delete | Diff scrolls to the deletion (clamped to last line, widget moves) |
| 1.6 | Already open (scrolled to MIDDLE) | Midpoint | Mixed | Diff scrolls to the first changed line |
| 1.7 | Already open (scrolled to MIDDLE) | End | Mixed | Diff scrolls to the first changed line |
| 1.8 | NOT open | End | Delete | File opens; diff scrolls to the deletion |
| 1.9 | NOT open | End | Add | File opens; diff scrolls to the added region |
| 1.10 | NOT open | Midpoint | Mixed | File opens; diff scrolls to the first changed line |
| 1.11 | Already open (scrolled to MIDDLE) | **Change INSIDE the current viewport** | Mixed | Diff scrolls/recenters on the change; viewport does NOT jump to the top (bug 1) |
| 1.12 | Already open (scrolled to MIDDLE) | End (last line) | Mixed | Diff scrolls to the last line; the widget actually moves (bug 3) |

Pay special attention to **1.2-1.7, 1.11, 1.12** -- these are the regressions where an
already-open, mid-scrolled file either jumped to the top or refused to scroll.

### 1.S -- "Stuck scroll" sequential check (bug 4)

1. Open `file-1`, scroll to the **middle**.
2. Ask Zoo to make an **end-of-file** edit to `file-1` via apply_diff. Observe the
   diff-open scroll (it should reach the end, not the top).
3. WITHOUT closing the file, ask Zoo to make a **midpoint** edit to `file-1`.
4. Expected: the second edit scrolls to the midpoint change. It must NOT be "stuck"
   scrolling to line 0. Repeat for several edits in a row.

---

## Matrix 2 -- Save (accept) behavior

Set up a known pre-edit scroll position before each run by manually scrolling the file, so
restoration is observable.

| # | File state before edit | User action during diff | After clicking Save |
|---|------------------------|-------------------------|---------------------|
| 2.1 | Open, scrolled to midpoint | none | File shown; viewport restored to the midpoint scroll position |
| 2.2 | Open, scrolled to end | none | File shown; viewport restored to the end scroll position |
| 2.3 | NOT open | none | Transient file tab is CLOSED; no leftover tab |
| 2.4 | NOT open | click into the file's editor tab during the diff | File tab STAYS open after Save |
| 2.5 | Open; then navigate to a different file during the diff | viewing another file | Focus stays on the other file; you are not pulled back |
| 2.6 | New file (did not exist) | none | File is created and saved; no stray diff tab; viewport sane |

---

## Matrix 3 -- Deny (reject) behavior

| # | File state before edit | User action during diff | After clicking Deny |
|---|------------------------|-------------------------|---------------------|
| 3.1 | Open, scrolled to midpoint | none | Edit reverted; viewport restored to the midpoint scroll position |
| 3.2 | NOT open | none | Edit reverted; transient file tab is CLOSED |
| 3.3 | NOT open | click into the file's editor tab during the diff | Edit reverted; file tab STAYS open |
| 3.4 | New file (did not exist) | none | File is deleted; its tab is closed; created dirs cleaned up |
| 3.5 | Open; navigate elsewhere during the diff | viewing another file | Focus stays on the other file |

---

## Matrix 4 -- Auto-approve late-timer race (bug 5)

Enable **auto-approve writes** for these rows so save runs immediately after the diff
opens (the tight window where a deferred scroll could fire after the viewport is restored).

| # | File state before edit | Position | Expected after the auto-approved save settles |
|---|------------------------|----------|-----------------------------------------------|
| 4.1 | Open, scrolled to midpoint | Midpoint | File ends restored to the pre-edit midpoint scroll; it does NOT settle on the change line a beat later |
| 4.2 | Open, scrolled to midpoint | End | File ends restored to the pre-edit scroll; no late jump to the end |
| 4.3 | Open, scrolled to top | Midpoint | File ends at the pre-edit top; no delayed scroll-to-change |
| 4.4 | Sequence: auto-approve several edits in a row | mixed positions | No accumulating drift; each settles at the restored position; no tab accumulation |

Watch for ~100ms-late movement: the buggy behavior was a brief correct restore followed
by the viewport snapping to the diff target.

---

## Matrix 5 -- Preview-tab eviction (bug 6)

A preview tab is the italicized, single reusable tab VS Code shows when you open a file
but have not edited it.

| # | Setup | Action | Expected |
|---|-------|--------|----------|
| 5.1 | Open `file-1` as a **preview** tab (single click in Explorer, do not edit). `file-2` is closed. | Ask Zoo to apply_diff `file-2`, then Save. | `file-1` is NOT permanently lost: after the diff session it is restored as a preview tab at its prior scroll position. `file-2` handled per Matrix 2. |
| 5.2 | Same as 5.1 | Click Deny instead of Save | Same restore of `file-1`'s preview tab. |
| 5.3 | Open `file-1` as a preview tab, scrolled to the middle. | apply_diff `file-2` + Save | `file-1` restored AND scrolled back to its prior (middle) position. |
| 5.4 | Open `file-1` as a **permanent** (non-preview, e.g. double-click or edited) tab. | apply_diff `file-2` + Save | `file-1` is untouched throughout (permanent tabs are never evicted); no duplicate `file-1` tab appears. |
| 5.5 | Open `file-1` as a preview tab. | During the diff, the preview file is deleted/renamed on disk, then Save | Restore is skipped gracefully -- no error dialog, no broken tab. |

---

## Matrix 6 -- Focus preservation on keep-open (bug 7)

| # | Setup | Action sequence | Expected after Save/Deny |
|---|-------|-----------------|--------------------------|
| 6.1 | `file-1` open (non-preview). `file-2` closed. | apply_diff `file-2`; click `file-2`'s plain tab; then click `file-1`; then Save | You remain on `file-1` (where you were actually looking). Focus is NOT yanked to `file-2`. `file-2` stays open in the background. |
| 6.2 | Same as 6.1 | Same but click Deny | Same: you remain on `file-1`; edit reverted; `file-2` stays open in the background. |
| 6.3 | `file-2` already open and active. | apply_diff `file-2`; stay on `file-2`; Save | You remain on `file-2` at its restored scroll position (no redundant focus round-trip). |
| 6.4 | Two editor groups (split). `file-1` active in group 1. | apply_diff `file-2` (opens in a group); navigate around; Save | You end on whatever editor you were last viewing; no surprise foreground change. |

---

## Regression smoke checks

- **Pinned tab**: pin the file's tab, run an edit + Save. Confirm the tab is not unpinned
  or duplicated.
- **Dirty unrelated tab**: have another file with unsaved changes open; run an edit + Save.
  Confirm the unrelated dirty tab is untouched (`closeFileTab` filters on `!isDirty` and
  matches only the target path).
- **Multiple sequential edits**: ask Zoo to make several edits in a row to the same long
  file; confirm the viewport does not drift to the top between edits, the scroll never gets
  "stuck", and tabs do not accumulate.
- **Diff tab label**: confirm the diff tab reads "Original <-> Zoo's Changes".
- **Interrupted session**: start an edit, then reload the dev window (Cmd+Shift+P ->
  "Developer: Reload Window") before accepting. Confirm no preview file is left
  permanently promoted and no stray diff tab remains.

awschmeder:
**Videos from Manual Testing:**

Target file open - scrolled to top - diff scrolls to edit at top, middle, bottom:
https://github.com/user-attachments/assets/b290570c-373b-4cb7-b640-966a038ce0aa

Target file open - scrolled to middle - diff scrolls to edit and restores viewport after accept:
https://github.com/user-attachments/assets/f2e8a29f-0faf-4d9c-b12a-115b2c1af94d

Target file not open - diff scrolls to edit - tab is cleaned up after accept or reject:
https://github.com/user-attachments/assets/db8eee31-64f5-4c91-a352-59b3d9d81a16

Stress test multiple sequential writes & scrolling with apply_diff - no race conditions or errors:
https://github.com/user-attachments/assets/02d95019-9826-4f7e-8fa3-178a95ce02dc

Preview tab open with file 1, diff targets unopened file 2, after accept, preview tab is restored:
https://github.com/user-attachments/assets/a89d501b-b456-4694-8446-c1321abb71a9

Permanent tab open with file 1, diff targets unopened file 2, user changes tab then accepts - no jump back:
https://github.com/user-attachments/assets/197ddd37-2f38-4811-b97f-44058d6a72b6

Pinned file targeted with diff - stays pinned correctly:
https://github.com/user-attachments/assets/a016eb27-6f54-4759-b3f5-d2dec3041bb6

awschmeder:
**Follow-up commit summary**

A second commit (`6ea564c`) was added after the initial PR push to address a gap in the preview-tab restore logic.

**Fix: restore preview tabs to their original editor group**

The initial implementation of `captureUnrelatedPreviewTabs()` / `restorePreviewTabs()` captured each evicted preview tab's URI and scroll position but did not record which editor group (column) the tab belonged to. As a result, `showTextDocument` during restore omitted `viewColumn`, causing VS Code to reopen the tab in the active group rather than its original group.

Changes:
- `snapshotPreviewTabs` type extended to include `viewColumn: vscode.ViewColumn`
- `captureUnrelatedPreviewTabs()` now records `group.viewColumn` alongside the URI and scroll line
- `restorePreviewTabs()` passes `viewColumn: snapshot.viewColumn` to `showTextDocument`
- Tests updated: `setTabs()` helper accepts a `viewColumn` argument; snapshot/restore assertions verify the correct column is captured and passed through

awschmeder:
**Update -- commit `003a31e`**

Two additional behaviors have been implemented on top of the original scroll/tab fix:

**Keep target tab open when user interacts with the diff pane**
Previously, the target file tab was only kept open if the user activated the file's own editor tab. This commit extends that: if the user clicks or edits anywhere inside the diff pane itself (`onDidChangeTextEditorSelection` with `kind = Mouse | Keyboard`), the tab is also kept open after Save/Deny. Programmatic selection changes (e.g. from `scrollToFirstDiff()`) are filtered out via the `kind` check. A stale-editor guard compares by `document` identity rather than editor object identity, so the listener works correctly when VS Code creates a fresh wrapper around the same document.

**Target file scrolls to most-recently-viewed position**
When the target file is re-revealed, it now scrolls to whichever position the user last viewed:
- If the user last scrolled in the **diff pane** -- target file mirrors that line.
- If the user last scrolled in the **target file's own editor** -- that position is used.
- Otherwise -- pre-edit scroll position is restored (existing behavior).

**Tests:** 59 tests pass (was 44). New coverage: `userTouchedDiffEditor` keep/close logic, programmatic-kind filter, stale-editor regression, scroll precedence (diff vs. target file vs. fallback), and `diffScrollListener` wiring.

src/integrations/editor/DiffViewProvider.ts
```diff
@@ -400,16 +538,15 @@ export class DiffViewProvider {
...
-			if (this.documentWasOpen) {
-				await vscode.window.showTextDocument(vscode.Uri.file(absolutePath), {
-					preview: false,
-					preserveFocus: true,
-				})
-			}
-
 			await this.closeAllDiffViews()
+
+			await this.keepOrCloseEditedFile(absolutePath)
```
edelauna: Is there a test for the case where `userTouchedDiffEditor` is `true` but `revertChanges` is called? The intent (deny ignores diff-touch; only save respects it) is documented in the comment on `keepOrCloseEditedFile`, but without a test this asymmetry is easy to break accidentally.

src/integrations/editor/DiffViewProvider.ts
```diff
@@ -448,6 +585,203 @@ export class DiffViewProvider {
...
+			this.deferredScrollTimer = undefined
+		}
+	}
+
+	// Shared accept/deny cleanup: keep the edited file open if it was already open
+	// before the edit or the user interacted with it during the diff; otherwise
+	// close the tab that was opened transiently for the diff.
+	// Pass keepIfTouchedDiff=true (from saveChanges) to also keep the file when
+	// the user clicked or edited inside the diff editor itself.
+	private async keepOrCloseEditedFile(absolutePath: string, keepIfTouchedDiff = false): Promise<void> {
```
edelauna: The `keepIfTouchedDiff` parameter exists only so `saveChanges` can pass its own `this.userTouchedDiffEditor` field through this helper, while `revertChanges` passes nothing (default `false`). Would it be simpler to inline the diff-touch check directly in `saveChanges` before calling this method, so the shared helper only handles the `documentWasOpen`/`userTouchedDocument` conditions?

src/integrations/editor/DiffViewProvider.ts
```diff
@@ -623,11 +1024,21 @@ export class DiffViewProvider {
...
 		this.streamedLines = []
 		this.preDiagnostics = []
+		this.preEditScrollLine = undefined
+		this.diffScrollLine = undefined
+		this.targetFileScrollLine = undefined
+		this.lastScrolledSource = undefined
+		this.userTouchedDocument = false
+		this.userTouchedDiffEditor = false
+		this.snapshotPreviewTabs = []
+		this.disposeActiveEditorListener()
```
edelauna: `disposeActiveEditorListener()` runs after `closeAllDiffViews()` has already awaited async tab-close operations. Both `saveChanges` and `revertChanges` dispose listeners *before* any async editor manipulation — should `reset()` follow the same pattern and move the dispose + `cancelDeferredScroll()` calls to before `closeAllDiffViews()`?

edelauna (commented):
thanks for this, had some comments related to the implementation.

.changeset/fix-diff-scroll-position.md
```diff
@@ -0,0 +1,7 @@
+---
+"zoo-code": patch
+---
+
```
navedmerchant, line 4: we need to investigate why so many agents create these changeset files lol, I saw it in other PRs too. I think it might be some instruction in the zoo repository

navedmerchant:
Taking a look, @awschmeder can you see why CI is failing?

.changeset/fix-diff-scroll-position.md
navedmerchant: figured it out, looks like they look at changesets in the repo and assume we need to generate one. We need to add instructions about it in agents.md, or just allow changesets in general

navedmerchant:
Resolved @edelaunas comments in my commit, merging this, really improves the UX, but files almost flash before your eyes lol. We should promote the experimental feature of not opening files further

navedmerchant (approved):

PR #628, opened 2026-06-14
refactor(Task): extract RateLimitClock from Task static state (#361)

### LINKED ISSUE #361: [Story 1.1] Extract `RateLimitClock` singleton
Part of #356 (Epic 1: Foundation).

**Depends on:** Nothing — additive, no behavior change.

## Context

`Task.lastGlobalApiRequestTime` is `private static` (line 287, `Task.ts`) — shared across all `Task` instances in the Node.js module. When two subtasks run concurrently they stomp on each other's rate-limit clock, causing both to see stale timestamps and either over-delay or skip the delay entirely. This must be fixed before any concurrent task execution is safe.

## Developer Notes

Create `src/core/task/RateLimitClock.ts`:
- `RateLimitClock` class with `getLastRequestTime(): number | undefined` and `recordRequest(): void`.
- Export `createRateLimitClock(): RateLimitClock`.

In `src/core/task/Task.ts`:
- Remove `private static lastGlobalApiRequestTime?: number` (line 287) and `static resetGlobalApiRequestTime()` (line 294).
- Add `private rateLimitClock: RateLimitClock` as a constructor-injected field with a default factory call so existing callsites need no change.
- Replace **6 call sites** with `this.rateLimitClock.*` (lines 2440, 3849, 3854, 3902, 4277–4278).
- Add optional `rateLimitClock?: RateLimitClock` to `TaskOptions`.

In `src/core/webview/ClineProvider.ts`:
- Construct one `RateLimitClock` per provider; pass it to `createTask()` so all tasks under one provider share the clock, preserving the original rate-limit intent across subtasks.
- **Behavioral note:** The current `private static` field is global across ALL providers in the Node.js module. Per-provider scoping means two providers with different API keys no longer share rate-limit state. This is intentionally better, but verify existing rate-limit tests don't assume cross-provider sharing.
- **Multi-window analysis:** Each VS Code window runs its own extension host (separate Node.js process), so per-provider scoping does not introduce a regression for multi-window setups.

**Files:** `src/core/task/RateLimitClock.ts` (new), `src/core/task/Task.ts`, `src/core/webview/ClineProvider.ts`

**Tests** (`src/core/task/__tests__/RateLimitClock.spec.ts`):
- Two `Task` instances with the same clock share rate-limit waits (second waits for first to clear).
- Two `Task` instances with different clocks are fully independent (validates per-provider scoping).
- First call (no prior time) returns no delay.
- Time elapsed < limit → delay applied; time elapsed > limit → no delay.

## Acceptance Criteria

- `Task.resetGlobalApiRequestTime()` is deleted; `grep -n "Task.lastGlobalApiRequestTime" src/core/task/Task.ts` returns zero matches.
- All existing `Task.spec.ts` rate-limiting tests pass without modification.
- No static state remains on `Task` for rate limiting.

edelauna:
### Related GitHub Issue

Closes: #361

### Description

Replaces `Task.lastGlobalApiRequestTime` (`private static`) with an injectable `RateLimitClock` instance, scoped per `ClineProvider`.

**Problem:** The static field was shared across all `Task` instances in the Node.js module. When two subtasks ran concurrently, they stomped on each other's rate-limit timestamps — causing both to see stale values and either over-delay or skip the delay entirely.

**Solution:**
- New `src/core/task/RateLimitClock.ts` — small class with `getLastRequestTime()` and `recordRequest()`
- `Task` accepts an optional `rateLimitClock` via `TaskOptions` (defaults to a new instance if omitted)
- `ClineProvider` creates one `RateLimitClock` per provider instance and passes it to all tasks it creates, so parent and child tasks share rate-limit state within a provider
- Removed `Task.lastGlobalApiRequestTime` and `Task.resetGlobalApiRequestTime()`

**Behavioral change:** Rate-limit state is now per-provider instead of global. Two providers with different API keys no longer share rate-limit timestamps. This is intentionally better — the old global sharing was overly broad.

### Test Procedure

- New `RateLimitClock.spec.ts` — covers initial undefined state, recording, monotonic updates, and cross-instance isolation
- Updated existing `Task.spec.ts` rate-limit tests to use explicit shared clocks instead of static resets — all pass
- `grep -rn "lastGlobalApiRequestTime\|resetGlobalApiRequestTime" src/` returns zero matches
- Full test suite: 398 files pass, 6192 tests pass (1 pre-existing unrelated failure in `ShadowCheckpointService.spec.ts`)

### Pre-Submission Checklist

### Documentation Updates

### Additional Notes

**Files changed:**
- `src/core/task/RateLimitClock.ts` (new)
- `src/core/task/__tests__/RateLimitClock.spec.ts` (new)
- `src/core/task/Task.ts` (modified — 6 call sites updated)
- `src/core/task/__tests__/Task.spec.ts` (modified — tests use shared clocks)
- `src/core/webview/ClineProvider.ts` (modified — creates and injects clock)

## Summary by CodeRabbit

* **Improvements**
  * Refactored API rate limiting to use per-task rate-limit clocks instead of a shared global timestamp, improving control across concurrent operations.
  * Enhanced throttling behavior by tracking the most recent request time per clock instance, reducing unintended waiting between independent tasks.
  * Improved coordination when restoring or creating tasks so related operations share the same rate-limit tracking.

### DIFF
--- a/src/core/task/__tests__/Task.spec.ts
+++ b/src/core/task/__tests__/Task.spec.ts
@@ -10,6 +10,7 @@ import type { GlobalState, ProviderSettings, ModelInfo } from "@roo-code/types"
 import { TelemetryService } from "@roo-code/telemetry"
 
 import { Task } from "../Task"
+import { createRateLimitClock } from "../RateLimitClock"
 import { ClineProvider } from "../../webview/ClineProvider"
 import { ApiStreamChunk } from "../../../api/transform/stream"
 import { ContextProxy } from "../../config/ContextProxy"
@@ -400,14 +401,6 @@ describe("Cline", () => {
 
 	describe("getEnvironmentDetails", () => {
 		describe("API conversation handling", () => {
-			beforeEach(() => {
-				Task.resetGlobalApiRequestTime()
-			})
-
-			afterEach(() => {
-				Task.resetGlobalApiRequestTime()
-			})
-
 			it("should strip non-protocol fields from API conversation history before sending to the API", async () => {
 				const cline = new Task({
 					provider: mockProvider,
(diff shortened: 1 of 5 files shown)


src/core/task/__tests__/Task.spec.ts
```diff
@@ -669,6 +662,91 @@ describe("Cline", () => {
...
+
+				const iterator = cline.attemptApiRequest(0)
+				await iterator.next()
+
+				// The retry should have used the rate limit window.
+				// recordRequest() was called during attemptApiRequest, so the
+				// clock has a timestamp, and rateLimitSeconds=10 means the
+				// backoff should account for the rate limit window (10s) being
+				// larger than the base exponential delay (3s).
+				const retryMessages = saySpy.mock.calls.filter((call) => call[0] === "api_req_retry_delayed")
```
navedmerchant: Nit (test strength): This test is named "should respect rate limit window in
  retry backoff" and the inline comment explains that with rateLimitSeconds=10
  and base backoff requestDelaySeconds=3, the 10s rate-limit window should
  dominate the delay. But the assertions only check that a api_req_retry_delayed
   message was emitted and that clock.getLastRequestTime() is defined — neither
  confirms the delay actually reflected the rate-limit window. The test would
  pass even if the backoff used the 3s base value (or any non-zero delay), so it
  doesn't really prove the stated behavior.
 Consider asserting on the computed delay so a regression in the
  window-vs-backoff logic would fail here. Since delay is already mocked as
  mockDelay, something like:
  
  > // The rate-limit window (10s) should dominate the base exponential backoff
  (3s)
  > expect(mockDelay).toHaveBeenCalledWith(10000)
  > 
   or assert the countdown ran for ~10 iterations / parse the seconds out of
  the api_req_retry_delayed payload. That ties the test to its intent. Not
  blocking — current coverage is adequate, just weaker than the name implies.
  
  If you want it tighter/less wordy for an inline review thread, the one-liner
  version:
  
  Assertions here only check that a retry-delay message fired and the clock is
  set — they don't verify the delay reflected the 10s window vs the 3s base
  backoff, so the test would pass even if the window logic regressed. Since
  delay is mocked, consider expect(mockDelay).toHaveBeenCalledWith(10000) (or
  asserting the countdown length) to match the test's stated intent.
  Non-blocking.

navedmerchant (dismissed):
A small nit comment, but non blocking, approved

navedmerchant (approved):

PR #633, opened 2026-06-15
Add completion change actions

### LINKED ISSUE #661: [ENHANCEMENT] Add completion change review and restore actions
## Type
Enhancement

## Problem / Value
Users need a simple way to review and undo the changes produced by the latest prompt after an attempt completes. This helps users validate agent output and recover quickly before starting a new task.

## Context
This issue was created for PR #633 to provide the required linked issue for the completion change actions feature.

The requested improvement is to show inline "See New Changes" and "Restore Changes" actions on completion results when a latest-prompt checkpoint is available, while keeping the existing "Start New Task" action.

## Constraints/Preferences
Keep the actions scoped to changes from the latest prompt, ask for confirmation before restoring, reuse the existing checkpoint diff and restore behavior, and preserve the existing completion flow.

Related PR: #633

ivanarifin:
### Related GitHub Issue

Closes: #661

### Description

This PR adds a Kilo Code-inspired completion experience that helps users view and undo the changes made for the latest prompt after an `attempt_completion` result.

Key implementation details:
- Adds inline `See New Changes` and `Restore Changes` buttons to the completion result row, following the Kilo Code legacy placement/pattern while preserving the existing `Start New Task` completion action.
- Scopes the actions to the checkpoint created after the latest user prompt so users can inspect only the new changes from that prompt.
- Uses existing checkpoint diff and restore behavior so `Restore Changes` restores files and rewinds the task state to the last prompt checkpoint.
- Adds typed webview messages and shared checkpoint lookup logic in `@roo-code/types`.
- Adds localized labels for the new actions.

Reviewers should pay attention to the checkpoint selection logic and the restore flow, since those define exactly which changes are shown and undone.

### Test Procedure

Validated with:

- `git diff --check`
- `pnpm --filter @roo-code/types build`
- `cd src && npx vitest run core/webview/__tests__/completionCheckpoint.spec.ts core/webview/__tests__/webviewMessageHandler.checkpoint.spec.ts`
- `cd webview-ui && npx vitest run src/components/chat/__tests__/ChatView.clear-approval-buttons.spec.tsx src/components/chat/__tests__/SeeNewChangesButtons.spec.tsx`
- `cd webview-ui && pnpm exec tsc --noEmit`
- Commit hook ran `prettier --write` and repository lint successfully.

Manual behavior verified locally:
- Completion result shows Kilo-style inline `See New Changes` and `Restore Changes` buttons when a latest-prompt checkpoint exists.
- `See New Changes` opens the checkpoint diff scoped to the latest prompt.
- `Restore Changes` asks for confirmation, then restores files and rewinds task state to the latest prompt checkpoint.
- Bottom completion action remains `Start New Task`.

### Pre-Submission Checklist

### Screenshots / Videos

Not included. This is a small chat-row UI addition modeled after the existing Kilo Code legacy completion action pattern.

### Documentation Updates

- [x] Yes, documentation updates are required. User-facing docs may want to mention that completion results now provide prompt-scoped change review and restore actions when checkpoints are available.

### Additional Notes

This is intentionally inspired by Kilo Code legacy: the completion result itself carries the user-facing actions to review or undo the new changes, making it easier to validate an agent's work before starting a new task.

### Get in Touch

ivanarifin

## Summary by CodeRabbit

* **New Features**
  * Added “See New Changes” and “Restore Changes” actions on completion results, enabling users to preview diffs and restore files/task state from the relevant checkpoint.
* **Bug Fixes**
  * Improved completion checkpoint handling with safer restore behavior, including timeout handling and clearer error reporting when checkpoint operations fail.
* **Documentation**
  * Added localized chat UI strings for the new actions across supported languages.
* **Tests**
  * Expanded unit/integration coverage for checkpoint diff/restore flows, UI button behavior, and timeout/failure scenarios.


webview-ui/src/components/chat/SeeNewChangesButtons.tsx
```diff
@@ -0,0 +1,60 @@
...
+	const restoreChangesCallback = useCallback(() => setRestoringChanges(true), [])
+
+	const confirmRestoreChangesCallback = useCallback(() => {
+		vscode.postMessage({ type: "completionCheckpointRestore", checkpointTs: checkpoint.ts })
+	}, [checkpoint.ts])
+
+	const cancelRestoreChangesCallback = useCallback(() => setRestoringChanges(false), [])
+
+	return (
+		<div className="flex flex-horizontal gap-2 w-full">
```
edelauna: `flex-horizontal` is not a Tailwind utility — it will be silently ignored, so the buttons stack vertically instead of sitting side-by-side. Should this be `flex-row`?

```suggestion
		<div className="flex flex-row gap-2 w-full">
```

packages/types/src/message.ts
```diff
@@ -275,6 +275,76 @@ export const clineMessageSchema = z.object({
 
 export type ClineMessage = z.infer<typeof clineMessageSchema>
 
+export interface CompletionCheckpoint {
+	ts: number
+	commitHash: string
+}
+
+const isInitialTaskMessage = (message: ClineMessage | undefined): boolean => {
+	return message?.type === "say" && (message.say === "text" || (message as { say?: string }).say === "task")
```
edelauna: The cast `(message as { say?: string }).say === "task"` bypasses the `ClineSay` union check — `"task"` is not in `clineSays`. Is this a legacy runtime value that needs to be added to the union, or is this branch dead code that can be removed?

src/core/webview/webviewMessageHandler.ts
```diff
@@ -1288,6 +1308,51 @@ export const webviewMessageHandler = async (
...
+
+				try {
+					await provider.getCurrentTask()?.checkpointRestore({
+						ts: checkpoint.ts,
+						commitHash: checkpoint.commitHash,
+						mode: "restore",
+					})
+				} catch (error) {
+					vscode.window.showErrorMessage(t("common:errors.checkpoint_failed"))
+				}
```
edelauna, line 1432: The caught `error` is dropped here — sibling handlers at lines 365 and 536 call `console.error("Error in ...", error)` before showing the user message. Without logging, this failure is invisible in the extension host output.

```suggestion
				} catch (error) {
					console.error("[completionCheckpointRestore] checkpointRestore failed:", error)
					vscode.window.showErrorMessage(t("common:errors.checkpoint_failed"))
				}
```

src/core/webview/__tests__/webviewMessageHandler.checkpoint.spec.ts
```diff
@@ -134,4 +145,73 @@ describe("webviewMessageHandler - checkpoint operations", () => {
...
+
+		it("restores files and task state to the checkpoint created after the latest prompt", async () => {
+			await webviewMessageHandler(mockProvider, { type: "completionCheckpointRestore", checkpointTs: 4 })
+
+			expect(mockProvider.cancelTask).toHaveBeenCalled()
+			expect(mockCline.checkpointRestore).toHaveBeenCalledWith({
+				ts: 4,
+				commitHash: "latest-prompt-checkpoint",
+				mode: "restore",
+			})
```
edelauna, line 183: This asserts both calls happened but not that `cancelTask` came before `checkpointRestore`. Would a mutation that reversed the order still pass? Restoring before cancelling would apply the checkpoint to a live task. Worth adding an ordering check — e.g., a shared `callOrder` array that both mocks push to.

src/core/webview/__tests__/webviewMessageHandler.checkpoint.spec.ts
```diff
@@ -134,4 +145,73 @@ describe("webviewMessageHandler - checkpoint operations", () => {
...
+
+			await webviewMessageHandler(mockProvider, { type: "completionCheckpointRestore", checkpointTs: 4 })
+
+			expect(mockProvider.cancelTask).toHaveBeenCalled()
+			expect(mockCline.checkpointRestore).not.toHaveBeenCalled()
+			const vscode = await import("vscode")
+			expect(vscode.window.showErrorMessage).toHaveBeenCalledWith("errors.checkpoint_timeout")
+		})
+	})
 })
```
edelauna, line 253: Is there a test for the path where `checkpointRestore` itself throws (the `errors.checkpoint_failed` branch at lines 1349–1351)? The timeout path is covered, but I don't see one for the restore error.

src/core/webview/webviewMessageHandler.ts
```diff
@@ -112,6 +113,25 @@ export const webviewMessageHandler = async (
 		vscode.window.showInformationMessage(getRouterUnavailableSignInMessage())
 	}
 
+	const getCompletionCheckpointForMessage = (
```
edelauna: The `ForMessage` suffix implies a single-message filter, but this function actually either does a timestamp lookup or falls back to a full-conversation traversal. Would a name like `resolveCompletionCheckpoint` convey the dual strategy more clearly?

src/core/webview/webviewMessageHandler.ts
```diff
@@ -1288,6 +1308,51 @@ export const webviewMessageHandler = async (
...
+				: undefined
+
+			if (checkpoint) {
+				await provider.cancelTask()
+
+				try {
+					await pWaitFor(() => provider.getCurrentTask()?.isInitialized === true, { timeout: 3_000 })
+				} catch (error) {
+					vscode.window.showErrorMessage(t("common:errors.checkpoint_timeout"))
+					return
```
edelauna, line 1413: This `return` on timeout correctly prevents the restore attempt on an uninitialized task — but the sibling `"checkpointRestore"` case (~line 1299) doesn't `return` on timeout and would proceed. Worth a comment here explaining the difference, or fixing the sibling to match.

src/core/webview/__tests__/webviewMessageHandler.checkpoint.spec.ts
```diff
@@ -134,4 +145,73 @@ describe("webviewMessageHandler - checkpoint operations", () => {
...
+
+			await webviewMessageHandler(mockProvider, { type: "completionCheckpointDiff" })
+			await webviewMessageHandler(mockProvider, { type: "completionCheckpointRestore" })
+
+			expect(mockCline.checkpointDiff).not.toHaveBeenCalled()
+			expect(mockCline.checkpointRestore).not.toHaveBeenCalled()
+			expect(mockProvider.cancelTask).not.toHaveBeenCalled()
+		})
+
+		it("does not re-derive a different checkpoint when an explicit checkpoint timestamp is missing", async () => {
```
edelauna: The test name says "timestamp is missing" but it passes `checkpointTs: 999` (provided but unmatched). The `undefined` case is actually different — it falls back to `getCompletionCheckpoint()` and would return a checkpoint. That fallback path seems untested. Worth renaming this test and adding a separate one for `checkpointTs: undefined`.

packages/types/src/__tests__/message.test.ts
```diff
@@ -12,3 +20,46 @@ describe("ask messages", () => {
...
+			commitHash: "latest-prompt-checkpoint",
+		})
+	})
+
+	it("returns the first checkpoint after an initial task row before completion", () => {
+		const messages = [
+			{ type: "say", say: "task", ts: 1, text: "Initial task" },
+			{ type: "say", say: "checkpoint_saved", ts: 2, text: "checkpoint-after-initial-task" },
+			{ type: "ask", ask: "completion_result", ts: 3, text: "Task complete", partial: false },
+		] as ClineMessage[]
```
edelauna: The `as ClineMessage[]` cast hides that `say: "task"` isn't in the `ClineSay` union. If `"task"` is a real runtime value (e.g. from persisted Roo Code history), it probably belongs in `clineSays` so the type and the test agree without casting.

webview-ui/src/components/chat/__tests__/SeeNewChangesButtons.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+	it("wires tooltip copy into the primary actions", () => {
+		const { getByText } = render(
+			<SeeNewChangesButtons checkpoint={{ ts: 2, commitHash: "checkpoint-after-user-prompt" }} />,
+		)
+
+		expect(getByText("chat:seeNewChanges.title")).toHaveAttribute("title", "chat:seeNewChanges.tooltip")
+		expect(getByText("chat:restoreChanges.title")).toHaveAttribute("title", "chat:restoreChanges.tooltip")
+	})
+})
```
edelauna, line 71: All four tests pass a non-empty `commitHash`. What happens when `commitHash` is `""` — which `isCheckpointMessage` allows since it only checks `typeof text === "string"`? The restore button would be shown as disabled; worth a test to lock that state down.

edelauna (changes requested):
Nice! Thanks for adding this, had a couple comments.

Could you also add a screen recording showing this change in action. 

Thank you.

ivanarifin:
> Nice! Thanks for adding this, had a couple comments.
> 
> Could you also add a screen recording showing this change in action.
> 
> Thank you.


https://github.com/user-attachments/assets/a828522f-d717-4a0a-b3e0-55c075f88e76

src/core/webview/webviewMessageHandler.ts
```diff
@@ -1378,6 +1396,52 @@ export const webviewMessageHandler = async (
...
+				? resolveCompletionCheckpoint(currentCline, message.checkpointTs)
+				: undefined
+
+			if (currentCline && checkpoint) {
+				await currentCline.checkpointDiff({
+					ts: checkpoint.ts,
+					commitHash: checkpoint.commitHash,
+					mode: "to-current",
+				})
+			}
```
edelauna, line 1397: test comment

src/core/webview/webviewMessageHandler.ts
```diff
@@ -1378,6 +1396,52 @@ export const webviewMessageHandler = async (
...
+				try {
+					await pWaitFor(() => provider.getCurrentTask()?.isInitialized === true, { timeout: 3_000 })
+				} catch (error) {
+					vscode.window.showErrorMessage(t("common:errors.checkpoint_timeout"))
+					return
+				}
+
+				try {
+					await provider.getCurrentTask()?.checkpointRestore({
+						ts: checkpoint.ts,
```
edelauna, line 1425: Two questions here: (1) `checkpoint` was resolved from `currentCline` captured before `cancelTask()`, but `checkpointRestore` is called on a fresh `getCurrentTask()` after re-init — could those be different task instances? (2) If `getCurrentTask()` returns `undefined` here the optional chain silently skips the restore with no user feedback — should this be an explicit guard?

webview-ui/src/components/chat/SeeNewChangesButtons.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { VSCodeButton } from "@vscode/webview-ui-toolkit/react"
+
+import type { CompletionCheckpoint } from "@roo-code/types"
+
+import { vscode } from "@src/utils/vscode"
+
+export const SeeNewChangesButtons = ({ checkpoint }: { checkpoint: CompletionCheckpoint }) => {
+	const { t } = useTranslation()
+	const [restoringChanges, setRestoringChanges] = useState(false)
+
```
edelauna, line 10: If `completionCheckpoint` changes while `restoringChanges` is `true` (e.g. a new turn produces a different checkpoint without unmounting this row), the confirm dialog would stay open against the old checkpoint. Would `key={checkpoint.ts}` at the call site in ChatRow be the right fix?

webview-ui/src/components/chat/ChatView.tsx
```diff
@@ -129,6 +129,23 @@ const ChatViewComponent: React.ForwardRefRenderFunction<ChatViewRef, ChatViewPro
...
+		for (let i = messages.length - 1; i >= 0; i--) {
+			const message = messages[i]
+
+			if (message?.type === "say" && message.say === "completion_result") {
+				return message.ts
+			}
+
+			// Handle ask-only completion_result rows that carry completion text.
+			if (message?.type === "ask" && message.ask === "completion_result" && (message.text ?? "") !== "") {
+				return message.ts
```
edelauna, line 145: `getCompletionCheckpoint` in `@roo-code/types` matches `ask:completion_result` unconditionally, but this scan adds `(message.text ?? "") !== ""`. Could a zero-text `ask:completion_result` be the checkpoint anchor per `getCompletionCheckpoint` but not receive the buttons here?

src/core/webview/webviewMessageHandler.ts
ivanarifin: I'm sorry, are you testing something or want me to change something?

ivanarifin: > test comment

or can i just mark this as resolved?

packages/types/src/vscode-extension-host.ts
```diff
@@ -657,6 +659,7 @@ export interface WebviewMessage {
 	ids?: string[]
 	terminalOperation?: "continue" | "abort"
 	messageTs?: number
+	checkpointTs?: number
 	restoreCheckpoint?: boolean
```
edelauna, line 663: Broader design question: should `checkpointTs` exist at all here? The VS Code security model treats the webview as an untrusted boundary ([docs](https://code.visualstudio.com/api/extension-guides/webview), [Trail of Bits research](https://blog.trailofbits.com/2023/02/21/vscode-extension-escape-vulnerability/)). By accepting a `checkpointTs` hint from the webview, the host lets an untrusted caller nominate *which* workspace state to restore — a privileged action. The alternative: drop this field entirely and have the handler always call `getCompletionCheckpoint(currentTask.clineMessages)` itself. The webview requests the *action*, the host decides the *target* from its own server-side state.

src/core/webview/webviewMessageHandler.ts
```diff
@@ -1378,6 +1396,52 @@ export const webviewMessageHandler = async (
...
+			break
+		}
+		case "completionCheckpointRestore": {
+			const currentCline = provider.getCurrentTask()
+			const checkpoint = currentCline
+				? resolveCompletionCheckpoint(currentCline, message.checkpointTs)
+				: undefined
+
+			if (checkpoint) {
+				await provider.cancelTask()
```
edelauna, line 1407: This block has a TOCTOU window: `checkpoint` is resolved from `currentCline` (pre-cancel), but `checkpointRestore` is called on whatever `getCurrentTask()` returns after re-init — which could be a different task instance. The VSCode extensibility docs recommend capturing an identity token before a cancellation and asserting it after ([patterns & principles](https://vscode-docs.readthedocs.io/en/latest/extensions/patterns-and-principles/); see also [vscode #265847](https://github.com/microsoft/vscode/pull/265847) for a core fix of the same pattern). Suggestion:
```typescript
const originalTaskId = currentCline.taskId
await provider.cancelTask()
// ... pWaitFor ...
const restoredTask = provider.getCurrentTask()
if (!restoredTask || restoredTask.taskId !== originalTaskId) {
  vscode.window.showErrorMessage(t('common:errors.checkpoint_failed'))
  return
}
await restoredTask.checkpointRestore({ ... })
```

src/core/webview/webviewMessageHandler.ts
edelauna: I'll mark this resolved, I'll post a more appropriate comment.

edelauna (changes requested):
Was reviewing this more in depth and wanted to also request some architectural changes to the feature to not allow the webview to pass a `checkpointTs`.

packages/types/src/vscode-extension-host.ts
edelauna: Done — `checkpointTs` is dropped entirely. The webview now posts bare `{ type: "completionCheckpointDiff" }` / `{ type: "completionCheckpointRestore" }`, and the host derives the target itself via `resolveCompletionCheckpoint()` → `getCompletionCheckpoint(currentCline.clineMessages)` (`webviewMessageHandler.ts` ~L1389/L1403). Webview requests the action, host decides the target from authoritative task state.

src/core/webview/webviewMessageHandler.ts
edelauna: Fixed exactly along those lines. We capture `originalTaskId = currentCline.taskId` before `cancelTask()`, and after re-init assert `restoredTask.taskId === originalTaskId` — if it changed (or `getCurrentTask()` is undefined) we surface `errors.checkpoint_failed` and bail (`webviewMessageHandler.ts` ~L1407-L1414).

src/core/webview/webviewMessageHandler.ts
edelauna: Both addressed by the identity check added in this round: `getCurrentTask()` returning undefined (or a different task) now explicitly errors with `checkpoint_failed` instead of silently no-op'ing — see the guard right after the `pWaitFor` re-init.

src/core/webview/webviewMessageHandler.ts
edelauna: Added: `console.error("[completionCheckpointRestore] checkpointRestore failed:", error)` before the toast now, matching the sibling handlers.

webview-ui/src/components/chat/SeeNewChangesButtons.tsx
edelauna: Good catch — switched to `flex flex-row gap-2 w-full` (`SeeNewChangesButtons.tsx`).

packages/types/src/message.ts
edelauna: It's a real runtime value, so added `"task"` to `clineSays` (`packages/types/src/message.ts`). `isInitialTaskMessage` no longer needs the cast.

packages/types/src/__tests__/message.test.ts
edelauna: Cast removed — with `"task"` now in `clineSays` the test fixture type-checks without it.

src/core/webview/webviewMessageHandler.ts
edelauna: Renamed to `resolveCompletionCheckpoint` (no `ForMessage` suffix).

webview-ui/src/components/chat/ChatView.tsx
edelauna: Added a comment in `completionResultTs` explaining the intentional divergence: zero-text `ask:completion_result` rows are hidden by `visibleMessages`, so we attach the actions to the latest *renderable* completion row while the host still derives the checkpoint target from authoritative task state via `getCompletionCheckpoint`.

webview-ui/src/components/chat/SeeNewChangesButtons.tsx
edelauna: Worth flagging honestly: I did **not** apply the `key={checkpoint.ts}` fix here. The checkpoint is wired through to the row via the `completionCheckpoint` prop (ChatView → ChatRow), but the row itself isn't keyed by the checkpoint ts, so the confirm-dialog-stale-against-old-checkpoint edge case you describe is still theoretically open. Leaving this unresolved — happy to add the key if you'd like it locked down.

edelauna: Update: applied. Added `key={completionCheckpoint.ts}` to both `SeeNewChangesButtons` render sites in `ChatRow.tsx` (say + ask completion_result paths), so the component remounts — and `restoringChanges` resets to `false` — whenever the checkpoint identity changes. Targeted tests (`SeeNewChangesButtons.spec.tsx`, `ChatView.clear-approval-buttons.spec.tsx`) and `check-types` pass.

webview-ui/src/components/chat/__tests__/SeeNewChangesButtons.spec.tsx
edelauna: This is now moot — when the component was simplified to drop the untrusted `checkpointTs`, the `checkpoint` prop (and both `disabled={!checkpoint.commitHash}` lines) were removed entirely (`SeeNewChangesButtons` takes no props now). The restore button is no longer conditionally disabled, so there's no empty-`commitHash` state to lock down. The host-side `resolveCompletionCheckpoint` returns `undefined` when there's no valid checkpoint, in which case the buttons aren't rendered at all (`completionCheckpoint ?` guard in `ChatRow`).

edelauna (approved):
Cool feature, thanks for adding it.

PR #645, opened 2026-06-17
fix(terminal): deliver completion signal when end event wins the race against setActiveStream

### LINKED ISSUE #489: [BUG] simply hang after command execution

Happens after clicking "Run" for a given command:

Image

We are not given the "Cancel/Stop" button, the request doesn't seem to be sent yet, new messages are just "queued"

Image

Image

This **never** happened to me with Roo-Code (could be a Zoo specific regression).

Image

User relying on command-execution ? (Maybe specific to Zoo-Code + Claude ?)

### Reproduction steps

1. Confirm a command
2. Command runs inside the terminal
3. Zoo-Code consider it being running
4. No way to stop it (hang)

(except reloading the window from developer console)

### Expected result

knowing exactly in which state is currently the extension and act upon it

### Actual result

no button, no information, no timeout

_No response_

### App Version

00fc24734f13a30aa

### API Provider (optional)

Anthropic

### Model Used (optional)

Claude Opus 4.6

_No response_

```shell

```

### LINKED ISSUE #622: [BUG]  zoo code stuck after that run process
**Describe the bug**
i have a issue, when i use zoo code, it's stuck often after that agent run process!

**Expected behavior**
expect to not stuck and continue to work after that running process is closed!

**Screenshots**
Image

3.60.0

edelauna:
## Related GitHub Issue

Closes: #489
Closes: #622

## Description

There is a race condition between two paths that both try to deliver stream and completion events to `TerminalProcess`:

- **Path A** (`TerminalProcess.run()`): calls `executeCommand()` → `setActiveStream(stream)` → sets `terminal.running = true`
- **Path B** (global `onDidEndTerminalShellExecution` in `TerminalRegistry`): fires when VS Code signals the command finished

For fast-completing commands (~25% of the time), VS Code fires the end event **before** Path A's `setActiveStream()` sets `running = true`. The end handler's `!terminal.running` guard then bails out with just `terminal.busy = false` — never calling `shellExecutionComplete()`. The `await shellExecutionComplete` promise in `TerminalProcess.run()` hangs forever, and Zoo appears stuck.

### Fix

When the end event arrives while `!terminal.running` but a `process` is attached and waiting, we now call `terminal.shellExecutionComplete(exitDetails)` instead of silently dropping the signal. When no process exists (spurious end event for a non-Zoo command), the original `busy = false` fallback is preserved.

Also:
- Added `console.info` diagnostic log for the race path (replaces removed `console.error` that misclassified it as an error)
- Removed pre-existing redundant `terminal.busy = false` on the normal completion path (already handled by `shellExecutionComplete()`)

### Key files
- `src/integrations/terminal/TerminalRegistry.ts` — the fix (end handler guard)
- `src/integrations/terminal/BaseTerminal.ts` — `shellExecutionComplete()` and `setActiveStream()` (context)
- `src/integrations/terminal/TerminalProcess.ts` — `run()` with the hanging `await` (context)

## Test Procedure

- Two new unit tests in `TerminalRegistry.spec.ts` simulate the race by firing the captured `onDidEndTerminalShellExecution` handler while `terminal.running === false`:
  1. **Process exists (race):** verifies `shell_execution_complete` is emitted exactly once and terminal returns to idle (`busy=false`, `running=false`)
  2. **No process (spurious):** verifies `busy` is cleared and `shellExecutionComplete` is NOT called
- All 16 tests pass, lint clean

## Pre-Submission Checklist

- [x] **Issue Linked**: Closes #489 and #622
- [x] **Scope**: Focused on the race condition fix
- [x] **Self-Review**: Yes
- [x] **Testing**: Two new unit tests covering both branches
- [x] **Contribution Guidelines**: Reviewed

## Documentation Updates

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal process handling to prevent hangs when shell execution end events arrive during startup.
  * Terminal now correctly manages idle state in edge case scenarios.

* **Tests**
  * Added test coverage for terminal shell execution race condition scenarios.

### DIFF
--- a/src/integrations/terminal/TerminalRegistry.ts
+++ b/src/integrations/terminal/TerminalRegistry.ts
@@ -103,12 +103,21 @@ export class TerminalRegistry {
 					}
 
 					if (!terminal.running) {
-						console.error(
-							"[TerminalRegistry] Shell execution end event received, but process is not running for terminal:",
-							{ terminalId: terminal?.id, command: process?.command, exitCode: e.exitCode },
-						)
+						// The end event can arrive before setActiveStream() has set
+						// running=true (race between the global VS Code event and the
+						// synchronous call in TerminalProcess.run). If a process is
+						// waiting for completion, deliver the signal so it doesn't
+						// hang forever. See #489 / #622.
+						if (process) {
+							console.info(
+								"[TerminalRegistry] End event arrived before running=true (race); delivering completion signal",
+								{ terminalId: terminal.id, exitCode: e.exitCode },
+							)
+							terminal.shellExecutionComplete(exitDetails)
+						} else {
+							terminal.busy = false
+						}
 
-						terminal.busy = false
 						return
 					}
 
@@ -123,7 +132,6 @@ export class TerminalRegistry {
 
 					// Signal completion to any waiting processes.
 					terminal.shellExecutionComplete(exitDetails)
-					terminal.busy = false // Mark terminal as not busy when shell execution ends
 				},
 			)
 
--- a/src/integrations/terminal/__tests__/TerminalRegistry.spec.ts
+++ b/src/integrations/terminal/__tests__/TerminalRegistry.spec.ts
@@ -205,6 +205,88 @@ describe("TerminalRegistry", () => {
 		})
 	})
 
+	describe("onDidEndTerminalShellExecution race condition (#489, #622)", () => {
+		let endHandler: (e: any) => Promise<void>
+
+		beforeEach(() => {
+			// Reset the initialized flag so we can call initialize() in this block.
+			TerminalRegistry["isInitialized"] = false
+
+			// The global vscode mock doesn't define shell execution event
+			// methods, so add them before spying.
+			;(vscode.window as any).onDidStartTerminalShellExecution ??= () => ({ dispose: () => {} })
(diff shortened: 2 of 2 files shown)


navedmerchant (approved):

drzraf:
Did you identified the commit introducing this regression ? Because it never happened to me with Roo-Code.

edelauna:
Dug into this to answer directly: no, this wasn't introduced by a Zoo Code change. The buggy code path (the `!terminal.running` branch that just logged a `console.error` and silently dropped the completion signal instead of calling `shellExecutionComplete()`) is upstream Roo-Code code, unchanged since [`185365af5d`](https://github.com/RooCodeInc/Roo-Code/commit/185365af5d) (Aug 2025) — the last commit to touch `TerminalRegistry.ts` before this file's history was carried over into this fork. The commit SHAs for this file are identical between the two repos, so it's shared history, not a coincidence.

I also pulled `TerminalRegistry.ts` directly from `RooCodeInc/Roo-Code`'s current default branch via the API, and the same unpatched dead-end is still there. That repo is now archived, so it was never going to get fixed upstream either way.

The race is timing-dependent (fires for roughly ~25% of fast-completing commands per the PR description), so it's plausible it just never happened to trigger in your Roo-Code usage rather than the code being different there. Nothing in the history points to a Zoo-specific regression — this fix (#645) is closing a latent bug this fork inherited.

Issue #17, opened 2026-05-05
[Refactor] Extract ClineMessagesStore from Task.ts


doctarock:
## Parent Issue
Part of #8 — Code Refactor on Monolith Files

## Summary
Extract the UI chat message persistence cluster from `src/core/task/Task.ts` into a dedicated `ClineMessagesStore` module.

## Scope
**Methods to extract:**
- `getSavedClineMessages()`
- `addToClineMessages()`
- `overwriteClineMessages()`
- `updateClineMessage()`
- `saveClineMessages()`
- `findMessageByTimestamp()`

**State owned:**
- `clineMessages: ClineMessage[]`
- `cloudSyncedMessageTimestamps: Set<number>`
- `lastMessageTs?: number`

**Estimated size:** ~200–250 lines

## Notes
This is a symmetric pattern to the already-extracted `apiConversationHistory` module, making it a low-risk, straightforward extraction. `Task.ts` becomes a thin consumer of both stores.

## Constraints (from #8)
- Preserve method names and call signatures
- Maintain event emission and side-effect order
- Run type checks after extraction
- One extraction per commit

PR #647, opened 2026-06-17
fix: LiteLLM cache key collision and silent fallback to non-existent default model

### LINKED ISSUE #638: [Bug] LiteLLM provider loses model selection and falls back to non-existent model after "Sync Models"
## Summary

Two root bugs cause LiteLLM provider users to lose their selected model:

1. **Cache key collision**: All LiteLLM servers share one cache entry keyed only on `"litellm"`. Switching between profiles backed by different LiteLLM servers silently serves the wrong model list, and the stale list persists across VS Code restarts via the disk cache. The same collision affects Ollama, LM Studio, Poe, DeepSeek, and Requesty. Additionally, for providers that enforce per-key model allowlists (LiteLLM, Poe, Requesty), two different API keys on the same server can see different model lists -- but share the same cache entry.

2. **Silent fallback to hardcoded default**: When the model list is empty (due to the collision above, a failed sync, or a transient error), `useSelectedModel` silently resets the configured model ID to `claude-3-7-sonnet-20250219` -- a model that typically does not exist on user LiteLLM servers -- causing all requests to fail.

## Steps to Reproduce

**Bug 1a (URL cache key collision):**

1. Configure Profile A: LiteLLM provider with `baseUrl=http://server-a:4000`. Select a model, e.g. `my-custom-model`.
2. Configure Profile B: LiteLLM provider with `baseUrl=http://server-b:4000`. The model list from server A is served for Profile B (wrong list).
3. Restart VS Code. Both profiles still see the wrong list -- the stale disk cache (`litellm_models.json`) persists server A's list.

**Bug 1b (API key cache key collision):**

1. Configure Profile A: LiteLLM provider with `baseUrl=http://server:4000`, `apiKey=key-admin` (full model list).
2. Configure Profile B: same `baseUrl`, `apiKey=key-restricted` (limited model list per LiteLLM allowlist).
3. Profile B sees Profile A's (admin) model list due to cache collision.

**Bug 2 (silent fallback):**

1. Configure LiteLLM with a valid server. Select `my-custom-model`.
2. Cause the model list to become empty (switch profiles, trigger a failed sync, etc.).
3. The active model silently resets to `claude-3-7-sonnet-20250219`; requests fail.
4. Clicking "Sync Models" shows success but the selection is not restored.

### Bug 1 (Primary): Cache key collision for URL-scoped and key-scoped providers

**Files:** `src/api/providers/fetchers/modelCache.ts`, `src/api/providers/router-provider.ts`

The backend model cache uses the provider name alone as the cache key for all three layers:

```ts
memoryCache.set(provider, models)    // key: "litellm"
writeModels(provider, data)          // filename: "litellm_models.json"
inFlightRefresh.get(provider)        // key: "litellm"
```

Two categories of collision:

- **URL collision**: LiteLLM, Poe, DeepSeek, Ollama, LM Studio, and Requesty are URL-scoped providers. Two different servers (`http://server-a:4000` and `http://server-b:4000`) collide on the same cache entry.
- **Key collision**: LiteLLM, Poe, and Requesty support per-key model allowlists. Two different API keys on the same server can see different model lists but share the same cache entry.

Additionally, `router-provider.ts` `getModel()` falls back to `getModelsFromCache(this.name)` -- passing only the plain provider name -- which misses the compound cache entry written by `fetchModel()`. The cold-start fallback never finds a cached URL-scoped result.

### Bug 2 (Primary): Silent fallback to hardcoded default when model list is empty

**Files:** `webview-ui/src/components/ui/hooks/useSelectedModel.ts`, `webview-ui/src/components/settings/providers/LiteLLM.tsx`, `webview-ui/src/components/settings/ApiOptions.tsx`, `src/core/webview/webviewMessageHandler.ts`

Four sub-causes interact:

**2a.** `getValidatedModelId` returns `defaultModelId` when `availableModels` is `{}` -- indistinguishable from "model genuinely not found." Any configured model ID silently falls back to `"claude-3-7-sonnet-20250219"`.

**2b.** The "Sync Models" button updates `ExtensionStateContext` (ModelPicker dropdown) but does NOT invalidate the React Query cache that `useSelectedModel` reads. React Query's stale (empty) litellm map persists after a successful sync.

**2c.** The debounced `requestRouterModels` in `ApiOptions.tsx` fires without credentials for `litellm`, causing the backend to fetch with stale saved credentials during credential editing.

**2d.** Even when `ApiOptions.tsx` sends current credentials in `message.values`, `webviewMessageHandler.ts` resolved them with `||` (saved config first), ignoring current field values whenever any saved key existed:

```ts
// Before fix -- saved config always wins
const litellmApiKey = apiConfiguration.litellmApiKey || message?.values?.litellmApiKey
```

This is opposite to the `??` pattern used by DeepSeek in the same function.

## Impact

- Any user with two LiteLLM profiles pointing to different servers sees the wrong model list.
- Users with per-key model allowlists on a shared LiteLLM server see the wrong model list when switching API keys.
- The same URL collision affects Ollama, LM Studio, Poe, DeepSeek, and Requesty.
- A failed or stale fetch resets the selected model to a non-existent default, causing all requests to fail.
- The "Sync Models" button does not restore the correct selection.
- The issue persists across VS Code restarts.

## Affected Files

- `src/api/providers/fetchers/modelCache.ts`
- `src/api/providers/router-provider.ts`
- `src/core/webview/webviewMessageHandler.ts`
- `webview-ui/src/components/ui/hooks/useSelectedModel.ts`
- `webview-ui/src/components/settings/providers/LiteLLM.tsx`
- `webview-ui/src/components/settings/ApiOptions.tsx`

awschmeder:
### Related GitHub Issue

Closes: #638

### Description

Fixes two root bugs causing LiteLLM provider users to lose their selected model:

**Bug 1 -- Cache key collision** (`src/api/providers/fetchers/modelCache.ts`, `src/api/providers/router-provider.ts`):

All URL-scoped providers (LiteLLM, Ollama, LM Studio, Poe, DeepSeek, Requesty) previously shared one cache entry keyed only on the provider name. Switching between profiles backed by different servers silently served the wrong model list, and the stale list persisted across VS Code restarts via the disk cache.

Fix: compound cache key -- URL-scoped providers use `provider:baseUrl`; key-scoped providers (LiteLLM, Poe, Requesty) additionally include a short, irreversible discriminator derived from the API key (`provider:baseUrl:<discriminator>`) so that two different API keys on the same server never share a cache entry. The discriminator is a 32-bit value derived via PBKDF2 and truncated, so the value written to the on-disk cache filename cannot be reversed to identify the API key; this also keeps the derivation off CodeQLs js/insufficient-password-hash sink (Nodes crypto.pbkdf2Sync is not a modeled weak-hash operation), resolving that finding at the source rather than dismissing it. The `RouterProvider.getModel()` cold-start fallback is corrected to pass the full options so it resolves the same compound key that `fetchModel()` wrote under.

**Bug 2 -- Silent fallback to hardcoded default** (`webview-ui/src/components/ui/hooks/useSelectedModel.ts`, `src/core/webview/webviewMessageHandler.ts`):

When the model list was empty, `useSelectedModel` reset the configured model ID to `claude-3-7-sonnet-20250219` -- a model that typically does not exist on user LiteLLM servers.

Four sub-fixes:
- Preserve the configured model ID when the list is empty; return `""` (no selection) rather than a phantom default when nothing is configured
- Invalidate the React Query router-models cache after a successful "Sync Models" click so `useSelectedModel` picks up the refreshed list
- Pass current LiteLLM credentials in the debounced `requestRouterModels` message
- Correct credential priority in `webviewMessageHandler.ts` -- `message.values` (current unsaved field state) now takes precedence over saved config via `??` instead of `||`, matching the pattern already used for DeepSeek

**Bug 3 -- TypeScript check-types failures** (`src/shared/api.ts`, `src/api/providers/fetchers/modelCache.ts`, `src/api/providers/fetchers/__tests__/modelCache.spec.ts`):

Three compile errors introduced by the compound cache key work:
- `dynamicProviderExtras.litellm` declared `apiKey: string` (required), but the cache key logic supports the no-key case for LiteLLM (URL-only scoping when no key is supplied). Changed to `apiKey?: string` and updated `fetchModelsFromProvider` to pass `options.apiKey ?? ""` to the underlying fetcher.
- `mockCache.get` and `mockCache.set` in the compound-cache-key test were typed as the original NodeCache overloads rather than `Mock`; fixed with explicit casts.

### Test Procedure

**Automated:**
- Updated `useSelectedModel.spec.ts`: the existing test that asserted fallback to `claude-3-7-sonnet-20250219` now asserts preservation of the configured model ID; a new test covers the empty-ID case when no model is configured.
- Run: `cd webview-ui && npx vitest run src/components/ui/hooks/__tests__/useSelectedModel.spec.ts`

**Manual (Bug 1 -- cache key collision):**
1. Configure Profile A: LiteLLM with `baseUrl=http://server-a:4000`, select a model.
2. Configure Profile B: LiteLLM with `baseUrl=http://server-b:4000`.
3. Verify Profile B shows its own model list, not server A's.
4. Restart VS Code; verify both profiles still show the correct lists.

**Manual (Bug 2 -- silent fallback):**
1. Configure LiteLLM with a valid server, select `my-custom-model`.
2. Trigger a failed sync or switch profiles to empty the model list.
3. Verify the active model remains `my-custom-model` rather than resetting to `claude-3-7-sonnet-20250219`.
4. Click "Sync Models"; verify the model list repopulates and the selection is preserved.

### Pre-Submission Checklist

### Screenshots / Videos

N/A -- backend cache and hook logic changes; no UI visual changes.

### Documentation Updates

### Additional Notes

The compound cache key changes the on-disk filename for URL-scoped providers (e.g. `litellm_models.json` becomes `litellm_<hash>.json`). Existing single-key cache files are simply ignored on first load and will be replaced on the next successful fetch -- no migration needed.

### Get in Touch

Discord: awschmeder

### DIFF
--- a/src/api/providers/fetchers/__tests__/modelCache.spec.ts
+++ b/src/api/providers/fetchers/__tests__/modelCache.spec.ts
@@ -504,3 +504,69 @@ describe("key-scoped cache key derivation", () => {
 		expect(discriminator).toMatch(/^[0-9a-f]{8}$/)
 	})
 })
+
+describe("compound cache key derivation across scoping dimensions", () => {
+	// Exercises every branch of getCacheKey via the public getModels() entry point.
+	// litellm is url-scoped AND key-scoped; openrouter is neither, so it hits the bare
+	// provider fallback. The fetcher mocks let us observe the cache key the result is
+	// written under (first arg of the matching memoryCache.set call).
+	const mockModels = {
+		"compound/model": {
+			maxTokens: 4096,
+			contextWindow: 200000,
+			supportsPromptCache: false,
+			description: "Compound cache key model",
+		},
+	}
+
+	let mockSet: Mock
+
+	beforeEach(() => {
+		vi.clearAllMocks()
+		const MockedNodeCache = vi.mocked(NodeCache)
+		const mockCache = new MockedNodeCache()
+		mockCache.get.mockReturnValue(undefined)
+		mockSet = mockCache.set
+		mockGetLiteLLMModels.mockResolvedValue(mockModels)
+		mockGetOpenRouterModels.mockResolvedValue(mockModels)
+	})
+
+	const writtenCacheKey = (): string => {
+		const call = mockSet.mock.calls.find((c) => c[1] === mockModels)
+		return call?.[0] as string
+	}
+
+	it("includes both the server URL and the key discriminator for url+key-scoped providers", async () => {
+		await getModels({ provider: "litellm", apiKey: "compound-key", baseUrl: "http://host:4000" })
+		const cacheKey = writtenCacheKey()
+
+		// Expected shape: provider:url:keyDiscriminator
+		expect(cacheKey).toMatch(/^litellm:http:\/\/host:4000:[0-9a-f]{8}$/)
+	})
+
+	it("normalizes trailing slashes in the server URL so equivalent URLs share a cache key", async () => {
+		await getModels({ provider: "litellm", apiKey: "compound-key", baseUrl: "http://host:4000/" })
+		const withSlash = writtenCacheKey()
+
+		mockSet.mockClear()
+		await getModels({ provider: "litellm", apiKey: "compound-key", baseUrl: "http://host:4000" })
+		const withoutSlash = writtenCacheKey()
+
+		expect(withSlash).toEqual(withoutSlash)
+	})
+
+	it("includes only the server URL when a url-scoped provider has no API key", async () => {
+		await getModels({ provider: "litellm", baseUrl: "http://host:4000" })
+		const cacheKey = writtenCacheKey()
+
+		// No trailing key discriminator when apiKey is absent.
+		expect(cacheKey).toBe("litellm:http://host:4000")
+	})
+
+	it("falls back to the bare provider name for providers that are neither url- nor key-scoped", async () => {
+		await getModels({ provider: "openrouter", apiKey: "ignored-key", baseUrl: "http://ignored:4000" })
+		const cacheKey = writtenCacheKey()
+
+		expect(cacheKey).toBe("openrouter")
+	})
+})
--- a/src/api/providers/fetchers/__tests__/modelCache.spec.ts
+++ b/src/api/providers/fetchers/__tests__/modelCache.spec.ts
@@ -434,3 +434,73 @@ describe("empty cache protection", () => {
 		})
 	})
 })
+
+describe("key-scoped cache key derivation", () => {
+	// Exercises the per-API-key cache discriminator that all KEY_SCOPED_PROVIDERS share.
+	// Requesty is used only because it is a key-scoped provider with a mocked fetcher; the
+	// behavior under test is provider-agnostic.
+	const keyScopedProvider = "requesty" as const
+
+	let mockCache: any
+	let mockSet: Mock
+
+	const mockModels = {
+		"key-scoped/model": {
+			maxTokens: 4096,
+			contextWindow: 200000,
+			supportsPromptCache: false,
+			description: "Key-scoped provider model",
+		},
+	}
+
+	beforeEach(() => {
+		vi.clearAllMocks()
+		const MockedNodeCache = vi.mocked(NodeCache)
+		mockCache = new MockedNodeCache()
+		mockCache.get.mockReturnValue(undefined)
+		mockSet = mockCache.set
+		mockGetRequestyModels.mockResolvedValue(mockModels)
+	})
+
+	// Returns the cache key the result was written under (first arg of the matching set call).
+	const writtenCacheKey = (): string => {
+		const call = mockSet.mock.calls.find((c) => c[1] === mockModels)
+		return call?.[0] as string
+	}
+
+	it("writes different cache keys for different API keys", async () => {
+		await getModels({ provider: keyScopedProvider, apiKey: "key-one" })
+		const firstKey = writtenCacheKey()
+
+		mockSet.mockClear()
+		await getModels({ provider: keyScopedProvider, apiKey: "key-two" })
+		const secondKey = writtenCacheKey()
+
+		expect(firstKey).toBeDefined()
+		expect(secondKey).toBeDefined()
+		expect(firstKey).not.toEqual(secondKey)
+	})
+
+	it("writes the same cache key for repeated calls with the same API key", async () => {
+		await getModels({ provider: keyScopedProvider, apiKey: "stable-key" })
+		const firstKey = writtenCacheKey()
+
+		mockSet.mockClear()
+		await getModels({ provider: keyScopedProvider, apiKey: "stable-key" })
+		const secondKey = writtenCacheKey()
+
+		expect(firstKey).toEqual(secondKey)
+	})
+
+	it("does not embed the raw API key in the cache key and truncates the discriminator", async () => {
+		const apiKey = "super-secret-api-key-value"
+		await getModels({ provider: keyScopedProvider, apiKey })
+		const cacheKey = writtenCacheKey()
+
+		// The raw secret must never appear in the on-disk-bound cache key.
+		expect(cacheKey).not.toContain(apiKey)
+		// The discriminator is the trailing key-component: an 8-char (32-bit) hex string.
+		const discriminator = cacheKey.split(":").pop() as string
+		expect(discriminator).toMatch(/^[0-9a-f]{8}$/)
+	})
+})
--- a/src/api/providers/fetchers/__tests__/modelCache.spec.ts
+++ b/src/api/providers/fetchers/__tests__/modelCache.spec.ts
@@ -432,6 +432,52 @@ describe("empty cache protection", () => {
 			expect(result1).toEqual(mockModels)
 			expect(result2).toEqual(mockModels)
 		})
+
+		it("scopes in-flight dedup by API key for key-scoped providers", async () => {
+			// In-flight dedup is keyed on the compound cache key, so concurrent refreshes for a
+			// key-scoped provider must dedup only when the API key matches. Two different keys
+			// (different compound keys) each trigger their own fetch; the same key shares one.
+			const mockModels = {
+				"requesty/model": {
+					maxTokens: 4096,
+					contextWindow: 200000,
+					supportsPromptCache: false,
+					description: "Requesty model",
+				},
+			}
+			mockGetRequestyModels.mockResolvedValue(mockModels)
+
+			const { refreshModels } = await import("../modelCache")
+
+			// Different keys -> separate compound keys -> two distinct fetches.
+			const [a, b] = await Promise.all([
+				refreshModels({ provider: "requesty", apiKey: "key-one" }),
+				refreshModels({ provider: "requesty", apiKey: "key-two" }),
+			])
+			expect(mockGetRequestyModels).toHaveBeenCalledTimes(2)
+			expect(a).toEqual(mockModels)
+			expect(b).toEqual(mockModels)
+
+			mockGetRequestyModels.mockClear()
+
+			// Same key -> same compound key -> a single shared in-flight fetch.
+			let resolveShared: (value: typeof mockModels) => void
+			mockGetRequestyModels.mockReturnValue(
+				new Promise<typeof mockModels>((resolve) => {
+					resolveShared = resolve
+				}),
+			)
+
+			const shared1 = refreshModels({ provider: "requesty", apiKey: "same-key" })
+			const shared2 = refreshModels({ provider: "requesty", apiKey: "same-key" })
+
+			expect(mockGetRequestyModels).toHaveBeenCalledTimes(1)
+
+			resolveShared!(mockModels)
+			const [s1, s2] = await Promise.all([shared1, shared2])
+			expect(s1).toEqual(mockModels)
+			expect(s2).toEqual(mockModels)
+		})
 	})
 })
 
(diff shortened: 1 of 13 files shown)


awschmeder:
> The CodeQL `js/insufficient-password-hash` alert (alert #17) is a false positive. The two `createHash("sha256")` calls in `modelCache.ts` are **not** password hashing -- they are used purely as cache key discriminators:
> 
> 1. In `getCacheKey()`: a 16-char SHA-256 prefix of the API key is included in the compound cache key so that two different API keys on the same server get separate cache entries (relevant for LiteLLM, Poe, and Requesty which enforce per-key model allowlists). The hash is never stored as a credential or used for authentication.
> 
> 2. In `cacheKeyToFilename()`: SHA-256 of the full compound cache key produces a collision-free, filesystem-safe filename component. Again, no authentication involved.
> 
> CodeQL's taint-flow analysis traces the API key variables (`litellmApiKey`, `poeApiKey`, etc.) through to the `createHash` sink and flags it as insufficient password hashing. The inline `// codeql[...]` suppression does not work here because the alert sources are in other files. A maintainer dismissal of alert #17 as "False positive" in the GitHub Security tab would resolve the CodeQL check.

Replaced with new implementation to avoid the advanced security flagging.

awschmeder:
**Resolved the CodeQL `js/insufficient-password-hash` finding at the source.**

The key-scoped cache discriminator no longer hashes the API key with SHA-256. It now uses a memoized PBKDF2 derivation truncated to 32 bits (`deriveApiKeyDiscriminator` in `modelCache.ts`). This:

- **Removes the CodeQL sink entirely** -- Node's `crypto.pbkdf2Sync` is not modeled as a weak-hash `CryptographicOperation`, so the source-to-sink path no longer exists and the alert is genuinely resolved rather than dismissed as a false positive.
- **Keeps the on-disk value non-reversible to the key** -- the heavy truncation means any preimage search against the cache filename yields an astronomically large set of candidate keys, so the API key cannot be identified; PBKDF2 adds a per-guess cost on top.
- **Preserves behavior** -- distinct keys map to distinct cache entries; the derivation is deterministic and memoized (runs at most once per key per session).

Added provider-agnostic tests covering per-key separation, determinism, and the non-identifying discriminator shape. `modelCache.spec.ts` is 21/21 green. The changeset and PR description are updated to describe the PBKDF2/truncation derivation.

awschmeder:
@coderabbitai review

src/api/providers/fetchers/modelCache.ts
```diff
@@ -311,17 +424,28 @@ export const flushModels = async (options: GetModelsOptions, refresh: boolean =
  * @param provider - The provider to get models for.
  * @returns Models from memory cache, disk cache, or undefined if not cached.
  */
-export function getModelsFromCache(provider: ProviderName): ModelRecord | undefined {
+export function getModelsFromCache(
+	options: GetModelsOptions | ProviderName,
```
edelauna, line 434: The `| ProviderName` string branch sets `cacheKey = options` verbatim, bypassing `getCacheKey`. For any URL/key-scoped provider (`poe`, `lmstudio`) this produces a bare key that never matches what `getModels`/`refreshModels` wrote. The existing callers in `poe.ts:41` and `lm-studio.ts:173` both pass a bare string today, so those providers always miss the cache despite being listed as fixed in the changeset. Could this overload be dropped and those two call sites updated to pass `GetModelsOptions`?

src/api/providers/fetchers/modelCache.ts
```diff
@@ -224,34 +336,34 @@ export const refreshModels = async (options: GetModelsOptions): Promise<ModelRec
...
-				inFlightRefresh.delete(provider)
+				inFlightRefresh.delete(cacheKey)
 			}
 		}
 	})()
 
 	// Track the in-flight request (auth-scoped providers are excluded; see above).
 	if (!shouldSkipCache) {
-		inFlightRefresh.set(provider, refreshPromise)
+		inFlightRefresh.set(cacheKey, refreshPromise)
```
edelauna, line 372: The IIFE runs synchronously up to its first `await`. If `fetchModelsFromProvider` resolves via a microtask before this line executes, the `finally` at line 359 deletes `cacheKey` from the map before it has ever been inserted — then this line inserts an already-resolved promise that is never evicted. Could `inFlightRefresh.set(cacheKey, refreshPromise)` be placed just before the IIFE is invoked rather than after?

src/api/providers/router-provider.ts
```diff
@@ -69,17 +69,25 @@ export abstract class RouterProvider extends BaseProvider {
...
 			return { id, info: cachedModels[id] }
 		}
 
-		// Last resort: return default model
-		return { id: this.defaultModelId, info: this.defaultModelInfo }
+		// Last resort: preserve the configured model ID (falling back to the default
+		// only when none is configured) so an as-yet-unfetched model isn't silently
+		// swapped for the hardcoded default. info still comes from defaults since we
+		// have no fetched or cached metadata for the configured model at this point.
+		return { id, info: this.defaultModelInfo }
```
edelauna, line 96: The previous code returned `{ id: this.defaultModelId, info: this.defaultModelInfo }` here, guaranteeing a non-empty ID. Now it returns the computed `id` from line 65, which is `this.modelId ?? this.defaultModelId`. If `this.modelId` is `""` (empty string), `??` doesn't fall through — `id` stays `""` and gets forwarded to the API. Should line 65 use `||` instead of `??` to also guard against an empty string?

webview-ui/src/components/ui/hooks/useSelectedModel.ts
```diff
@@ -167,7 +167,15 @@ function getSelectedModel({
...
 		}
 		case "litellm": {
-			const id = getValidatedModelId(apiConfiguration.litellmModelId, routerModels.litellm, defaultModelId)
+			// When the model list is empty (not yet loaded or still loading),
+			// preserve the configured model ID. LiteLLM is a proxy with no inherent
+			// default model, so we never substitute a hardcoded default here -- when
+			// nothing is configured we return an empty ID so the picker shows "no
+			// selection" rather than a phantom model that does not exist on the server.
+			const hasModels = routerModels.litellm && Object.keys(routerModels.litellm).length > 0
+			const id = hasModels
```
edelauna, line 176: When `hasModels` is true but the configured model ID is absent from the list, `getValidatedModelId` falls back to `defaultModelId`. Is there a test covering this case? A mutation that removes the `hasModels` branch (always returning `apiConfiguration.litellmModelId ?? ""`) would pass the current tests but silently break this validated-fallback path.

webview-ui/src/components/settings/providers/LiteLLM.tsx
```diff
@@ -55,6 +57,9 @@ export const LiteLLM = ({
 				if (refreshStatus === "loading") {
 					if (!litellmErrorJustReceived.current) {
 						setRefreshStatus("success")
+						// Invalidate the react-query router models cache so
+						// useSelectedModel picks up the refreshed list.
+						queryClient.invalidateQueries({ queryKey: ["routerModels"] })
```
edelauna: The query key in `useRouterModels` is `["routerModels", provider || "all"]` (two elements). Passing `["routerModels"]` here uses React Query v5 prefix matching, which invalidates *all* provider subkeys — not just LiteLLM. Should this target `["routerModels", "all"]` instead?

webview-ui/src/components/ui/hooks/__tests__/useSelectedModel.spec.ts
```diff
@@ -571,12 +571,37 @@ describe("useSelectedModel", () => {
...
+
+			expect(result.current.provider).toBe("litellm")
+			// LiteLLM has no inherent default; with nothing configured the ID is empty rather than a phantom model
+			expect(result.current.id).toBe("")
+			expect(result.current.info).toEqual(litellmDefaultModelInfo)
+		})
+
 		it("should use litellmDefaultModelInfo when selected model not found in routerModels", () => {
 			mockUseRouterModels.mockReturnValue({
 				data: {
```
edelauna, line 655: Is there a test for the dynamic transition — render with a non-empty `litellm` map and a selected model, then re-render with an empty map and assert the selection is preserved? That seems like the primary user-visible scenario the PR is fixing (selection held during a sync that momentarily empties the list).

src/api/providers/fetchers/__tests__/modelCache.spec.ts
```diff
@@ -434,3 +434,139 @@ describe("empty cache protection", () => {
 		})
```
edelauna, line 434: The in-flight dedup test uses `openrouter`, whose cache key is the bare provider name. Could a companion test with a key-scoped provider (e.g. `requesty`) verify that two concurrent calls with *different* API keys each trigger their own fetch, while two calls with the *same* key share a single in-flight promise?

edelauna (changes requested):
Nice! Thanks for adding caching

Just a note on poe and lm-studio providers: 

**Two providers listed as fixed still have broken cache lookups after this PR.**

`poe.ts:41` and `lm-studio.ts:173` both call `getModelsFromCache` with a bare string (`"poe"` and `"lmstudio"`). Since both are now declared in `URL_SCOPED_PROVIDERS` (and Poe also in `KEY_SCOPED_PROVIDERS`), their cache entries are written under compound keys like `poe:<url>:<discriminator>` — but the bare-string lookup in `getModelsFromCache` bypasses `getCacheKey` and will always miss. These `getModel()` implementations silently fall back to hardcoded defaults on every call.

edelauna (approved):
Thanks for addressing the feedback

PR #652, opened 2026-06-19
feat(opencode-go): native model params, anthropic-format routing, and context-token fix

### LINKED ISSUE #646: [BUG] Opencode Go provider model parameters are incorrect
**Describe the bug**
Opencode Go provider does not display correct model parameters, including Context Window, Max output, Supports prompt caching, price, etc. Currently, all models are using the same incorrect default values.

**To Reproduce**
1. Go to 'Settings-Providers'
2. Select Opencode Go as the API Provider
3. Select any model and check the model information displayed below.

**Expected behavior**
The model pricing information should match the content on the official website (https://opencode.ai/docs/go/). The context window and other parameters may refer to the official settings of each respective model.

**Screenshots**
Image

3.60

navedmerchant:
### Related GitHub Issue

Closes: #646

### Description

The Opencode Go provider previously treated every model as a generic OpenAI-compatible chat-completions endpoint. The Go gateway, however, routes a subset of its curated models (Qwen, MiniMax) exclusively through the Anthropic Messages wire format (`/v1/messages`), and the `/v1/models` endpoint only reliably returns `id` plus optional `context_window`/`max_tokens` — it advertises none of the capability flags or pricing the extension needs. This PR closes those gaps so Go-plan models behave correctly in the model picker, reasoning controls, max-output slider, and cost reporting.

Key implementation details:

- **Native model registry** (`packages/types/src/providers/opencode-go.ts`): Added `opencodeGoModels` — a curated `Record<string, ModelInfo>` encoding per-model context windows, max tokens, capability flags (`supportsReasoningEffort`, `preserveReasoning`, `supportsMaxTokens`, `supportsPromptCache`), tool include/exclude lists, and pricing sourced from vendor specs and the Go pricing table. Added `getOpencodeGoModelInfo()` for lookup with fallback to `opencodeGoDefaultModelInfo`.
- **Wire-format routing** (`packages/types/src/providers/opencode-go.ts`): Added `OPENCODE_GO_ANTHROPIC_FORMAT_MODELS` and `isOpencodeGoAnthropicFormatModel()` to identify models that must use `/v1/messages`. Sending these to the OpenAI endpoint is rejected with `401 Model <id> is not supported for format oa-compat`.
- **Protocol selection** (`packages/types/src/provider-settings.ts`): `getApiProtocol()` now returns `"anthropic"` for Anthropic-format Go models so token/cost aggregation re-adds cache tokens into `contextTokens` (Anthropic usage reports `input_tokens` excluding cache, with separate `cache_creation_input_tokens`/`cache_read_input_tokens` fields — without this the cached prefix is dropped and context-window usage is undercounted).
- **Handler dual-path** (`src/api/providers/opencode-go.ts`): Added a dedicated Anthropic SDK client and `resolveModel()` that branches `getModelParams` on a literal `format` ("anthropic" vs "openai"). Implemented an Anthropic streaming path that maps `message_start`/`content_block_*` events to the internal stream chunk shape, plus `addAnthropicCacheControl()` for server-side prompt caching. `completePrompt` and the OpenAI path now honour `includeMaxTokens`, `modelMaxTokens`, and `reasoning_effort`.
- **Fetcher merge** (`src/api/providers/fetchers/opencode-go.ts`): `parseOpencodeGoModel()` now starts from the native registry, then overrides `contextWindow`/`maxTokens`/`supportsImages` with live `/models` values when present (gateway stays source of truth for volatile fields), falling back to defaults for unknown models.
- **Default model**: Bumped `opencodeGoDefaultModelId` from `glm-5.1` to `glm-5.2`.

Reviewers should pay attention to: the format-routing set must stay in sync with the Go model table, and the Anthropic usage/cost aggregation path (the `contextTokens` fix is the core correctness change).

### Test Procedure

- Unit tests added/updated across three layers:
  - `packages/types/src/__tests__/provider-settings.test.ts` — asserts `getApiProtocol` returns `"anthropic"` for the Qwen/MiniMax Go models and `"openai"` for the rest.
  - `src/api/providers/__tests__/opencode-go.spec.ts` — covers `resolveModel` format branching, the Anthropic streaming event mapping, cache-control injection, `completePrompt` dual-path, and `reasoning_effort`/`max_completion_tokens` handling.
  - `src/api/providers/fetchers/__tests__/opencode-go.spec.ts` — covers registry-merge precedence (live values override registry; registry supplies capabilities/pricing; unknown models fall back to defaults).
- Run from the correct workspaces:
  - `cd packages/types && npx vitest run src/__tests__/provider-settings.test.ts`
  - `cd src && npx vitest run api/providers/__tests__/opencode-go.spec.ts api/providers/fetchers/__tests__/opencode-go.spec.ts`
- Manual verification: select a Qwen/MiniMax model in the Go provider and confirm requests hit `/v1/messages` (no `401 ... not supported for format oa-compat`); confirm `contextTokens` reflects the cached prefix and cost is reported.

### Pre-Submission Checklist

### Screenshots / Videos

N/A — no UI changes; behaviour is in provider routing, model metadata, and cost/context calculation.

### Documentation Updates

## Summary by CodeRabbit

* **New Features**
  * Updated Opencode Go default model from **glm-5.1** to **glm-5.2**.
  * Expanded the Opencode Go model registry with richer native capabilities (reasoning, images, prompt caching) and pricing details.
  * Added routing for select Opencode Go models to use **Anthropic `/v1/messages`** (including cache-control behavior and improved token/cost handling).

* **Bug Fixes**
  * Improved model resolution and parameter generation (token limits, reasoning effort behavior) for both Anthropic- and OpenAI-compatible paths.

* **Tests**
  * Significantly expanded coverage for model classification, registry merging, streaming/complete flows, and Anthropic integration.


src/api/providers/opencode-go.ts
```diff
@@ -39,36 +87,113 @@ export class OpencodeGoHandler extends RouterProvider implements SingleCompletio
...
+					settings: this.options,
+					defaultTemperature: OPENCODE_GO_DEFAULT_TEMPERATURE,
+				})
+			: getModelParams({
+					format: "openai",
+					modelId: id,
+					model: info,
+					settings: this.options,
+					defaultTemperature: OPENCODE_GO_DEFAULT_TEMPERATURE,
+				})
```
edelauna, line 135: For non-GPT-5 models with an explicit `maxTokens`, `getModelMaxOutputTokens` clamps to `Math.min(maxTokens, ceil(contextWindow × 0.2))`. For GLM-5.1 that resolves to `Math.min(131072, 40960)` = **40,960** — a significant reduction from the old code which used `info.maxTokens` directly. Is the 20% clamp the intended default when no user slider override is set, or should models that advertise `supportsMaxTokens: true` skip the clamp?

src/api/providers/opencode-go.ts
```diff
@@ -110,15 +235,246 @@ export class OpencodeGoHandler extends RouterProvider implements SingleCompletio
...
+	 * MiniMax handler: `message_start`/`message_delta` carry usage, content
+	 * blocks carry text/thinking/tool_use, and a final cost chunk is emitted
+	 * from the accumulated token counts. Prompt-cache breakpoints are applied
+	 * to the system prompt and last two user messages when the model advertises
+	 * `supportsPromptCache`, since the Go gateway honours server-side caching
+	 * and reports cache tokens in usage.
+	 */
+	private async *streamAnthropicMessage(
+		modelId: string,
+		info: { supportsPromptCache?: boolean },
```
edelauna: Should this be `ModelInfo` rather than `{ supportsPromptCache?: boolean }`? The `calculateApiCostAnthropic` call below force-casts `info` — a caller passing only `{ supportsPromptCache: false }` would silently get `totalCost: 0` because `inputPrice`/`outputPrice` would be `undefined`.

src/api/providers/opencode-go.ts
```diff
@@ -110,15 +235,246 @@ export class OpencodeGoHandler extends RouterProvider implements SingleCompletio
...
+			system: systemBlocks,
+			messages: supportsPromptCache
+				? this.addAnthropicCacheControl(sanitizedMessages, cacheControl)
+				: sanitizedMessages,
+			stream: true,
+			tools: convertOpenAIToolsToAnthropic(metadata?.tools ?? []),
+			tool_choice: convertOpenAIToolChoiceToAnthropic(metadata?.tool_choice, metadata?.parallelToolCalls),
+		}
+
+		const stream = await this.anthropicClient.messages.create(requestParams)
```
edelauna: Pre-stream errors (401, 429, network) propagate unwrapped here, but the `completePrompt` Anthropic branch wraps them with `"Opencode Go completion error: ..."`. Should this also have a try/catch for consistency?

packages/types/src/providers/opencode-go.ts
```diff
@@ -20,3 +20,318 @@ export const opencodeGoDefaultModelInfo: ModelInfo = {
...
+	// --- Xiaomi MiMo ---
+	"mimo-v2.5": {
+		maxTokens: 131_072,
+		contextWindow: 1_048_576,
+		supportsImages: true,
+		supportsPromptCache: false,
+		preserveReasoning: true,
+		inputPrice: 0.14,
+		outputPrice: 0.28,
+		cacheReadsPrice: 0.0028,
```
edelauna, line 143: `supportsPromptCache: false` means no cache-control breakpoints are ever injected for MiMo, so `cacheReadsPrice` here can never be applied — the extension won't report cache reads for a model that doesn't advertise prompt cache support. Does the Go gateway actually bill and report `cached_tokens` for MiMo on the OA-compat path without client-side breakpoints?

packages/types/src/providers/opencode-go.ts
```diff
@@ -20,3 +20,318 @@ export const opencodeGoDefaultModelInfo: ModelInfo = {
...
+		maxTokens: 131_072,
+		contextWindow: 1_000_000,
+		supportsImages: true,
+		supportsPromptCache: true,
+		includedTools: ["search_and_replace"],
+		excludedTools: ["apply_diff"],
+		preserveReasoning: true,
+		inputPrice: 0.3,
+		outputPrice: 1.2,
+		cacheReadsPrice: 0.06,
```
edelauna, line 219: M2.5 and M2.7 both declare `cacheWritesPrice: 0.375`, and M3 has `supportsPromptCache: true` with cache injection active. Is M3 exempt from write charges, or is this a gap?

packages/types/src/providers/opencode-go.ts
```diff
@@ -20,3 +20,318 @@ export const opencodeGoDefaultModelInfo: ModelInfo = {
...
+		cacheReadsPrice: 0.5,
+		cacheWritesPrice: 3.125,
+		description:
+			"Qwen3.7 Max - Alibaba's flagship text-only reasoning agent model with a 1M context window, designed for long-horizon agent workflows. Available via the Opencode Go plan.",
+	},
+
+	// --- DeepSeek ---
+	"deepseek-v4-pro": {
+		maxTokens: 384_000,
+		contextWindow: 1_000_000,
```
edelauna, line 282: Without `supportsMaxTokens: true`, `getModelMaxOutputTokens` clamps the effective default to `Math.min(384000, ceil(1000000 × 0.2))` = 200,000 and the UI slider is hidden. Should DeepSeek expose the slider like GLM does?

packages/types/src/providers/opencode-go.ts
```diff
@@ -20,3 +20,318 @@ export const opencodeGoDefaultModelInfo: ModelInfo = {
...
+export const OPENCODE_GO_ANTHROPIC_FORMAT_MODELS = new Set<string>([
+	// --- Alibaba Qwen ---
+	"qwen3.7-max",
+	"qwen3.7-plus",
+	"qwen3.6-plus",
+	// --- MiniMax ---
+	"minimax-m3",
+	"minimax-m2.7",
+	"minimax-m2.5",
+])
```
edelauna, line 342: Is there a test asserting that every ID in this set also exists as a key in `opencodeGoModels`? A model added to the registry but omitted here will silently hit the OA-compat endpoint and get a 401 at runtime.

src/api/providers/__tests__/opencode-go.spec.ts
```diff
@@ -258,4 +334,174 @@ describe("OpencodeGoHandler", () => {
...
+				type: "usage",
+				inputTokens: 10,
+				outputTokens: 0,
+				cacheWriteTokens: 2,
+				cacheReadTokens: 3,
+			})
+			// ... message_delta output tokens ...
+			expect(chunks).toContainEqual({ type: "usage", inputTokens: 0, outputTokens: 5 })
+			// ... and a final cost chunk.
+			expect(chunks.some((c) => c.type === "usage" && c.totalCost !== undefined)).toBe(true)
```
edelauna: This passes even when `totalCost` is computed from `outputTokens: 0` (the accumulation bug CodeRabbit flagged). Could this assert `c.totalCost > 0` so CI would catch that regression?

edelauna (changes requested):

edelauna (approved):

PR #662, opened 2026-06-20
fix(ask_followup_question): report non-array follow_up as a type error

### LINKED ISSUE #511: [BUG] Ollama model infinite loop with "ask_followup_question"
Error Log: 

```
Extension version: 3.58.0 
Provider: ollama (proxy) 
Model: qwen3-coder-next:latest Roo tried to use ask_followup_question without value for required parameter 'follow_up'. Retrying... assist with code
```

Reproduction Steps:
 - Setup Ollama as provider
 - Choose Qwen model
 - Ask questio in any mode (architect, code, ask, etc.)
 - Infinite loop because it never completes the follow_up question

I'm not exactly sure what to file this as and it could simply be a bug in the way the model is trained vs. how Zoo-Code interacts with it. Just trying a new extension setup and running into an issue where the model never finishes because Zoo keeps attempting to make API requests, despite seemingly the model being done with its response and waiting on a next prompt from me. 

Is this an issue with the integration model (Qwen) or that Zoo is expecting it to ask a question when it was simply done and waiting for the next prompt so it thought to ask a follow up question the model didn't intend to make? 

If this is user error on my part (i.e. bad model setup) then I can close. I know there's a warning that says Zoo is expected to work with Claude models, so it's quite possibly an issue there.

nh2:
### Related GitHub Issue

Closes: #511.

### Description

A native tool call for `ask_followup_question` with a present-but-non-array `follow_up` (e.g. a keyed object produced by incremental JSON parsing) was forwarded unchecked and then rejected by the tool via `recordMissingParamError`, which always reports the misleading message

    Missing value for required parameter follow_up

The value was the wrong type, not missing, which sent the model into a retry loop with the same payload.

Fix:

AskFollowupQuestionTool.execute now distinguishes the cases: `null`/`undefined` `follow_up` still reports a missing parameter, while a present-but-non-array value reports a clear type/shape error (surfaced to both the user via `say(error)` and the model via `toolError`) instructing it to retry with a JSON array.

`NativeToolCallParser` keeps forwarding the raw follow_up value on the finalize path (with an explanatory comment) rather than guarding with `Array.isArray`, so the tool can emit the precise error instead of the call being silently dropped to null by the generic parser guard.

Adds test coverage for string and keyed-object non-array follow_up values, plus a parser finalize test confirming the value is forwarded rather than dropped.

Assisted-By: Claude Opus 4.8 in Zoo Code.

### Test Procedure

I let Zoo add some tests which look sane and pass, and I'm test-triving the built `.vsix` currently.

### Pre-Submission Checklist

### Get in Touch

I'm not really using Discord but I read mentions here on Github.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation and error handling for the ask follow-up question tool. Invalid parameter types now trigger precise error messages that clearly specify the expected array format, enabling users to quickly identify and resolve configuration mistakes.


nh2:
Hmm, after some time I got another such failure again even with my patch in:

> Roo tried to use ask_followup_question without value for required parameter 'follow_up'.

Not sure yet how that can be.

nh2:
I found that with my new problem, the model generates `<parameter name="follow_up">`.

That is even though `src/core/prompts/sections/tool-use.ts` says explicitly:

> Use the provider-native tool-calling mechanism. Do not include XML markup

which seems to suggest a training bias to emit XML (I'm using Claude Opus 4.8 here).

edelauna (approved):
Thanks for addressing this. I'm wondering if we should file a ticket for the xml findings you mention, and at the very least emit a retry signal with updated instructions?

nh2:
> I'm wondering if we should file a ticket for the xml findings

Filed now:

* https://github.com/Zoo-Code-Org/Zoo-Code/issues/1034

PR #668, opened 2026-06-20
fix: auto-close files setting cannot be unchecked — always reverts to checked

### LINKED ISSUE #667: [BUG] Auto close files setting cannot be unchecked — always reverts to checked
## Describe the bug

The "Auto close Zoo opened files" checkbox in Settings → UI cannot be unchecked. After unchecking and clicking Save, the setting does not persist — the checkbox always reverts to checked when the settings view is reopened.

Additionally, the `autoCloseZooOpenedFilesAfterUserEdited` and `autoCloseZooOpenedNewFiles` settings suffer from the same root cause: they are saved to global state but never sent back to the webview.

## To Reproduce

1. Open Zoo Code Settings → UI section
2. Uncheck the "Auto close Zoo opened files" checkbox
3. Click **Save**
4. Navigate away from Settings (e.g., back to chat)
5. Reopen Settings → UI section
6. Observe the checkbox is checked again

## Expected behavior

The "Auto close Zoo opened files" checkbox should remain unchecked after saving and reopening settings. The saved value (`false`) should persist across sessions.

## Root Cause

The three auto-close settings (`autoCloseZooOpenedFiles`, `autoCloseZooOpenedFilesAfterUserEdited`, `autoCloseZooOpenedNewFiles`) are **missing from both `getState()` and `getStateToPostToWebview()`** in `ClineProvider.ts`. The settings are correctly saved to global state via `contextProxy.setValue()` in the `updateSettings` handler, but:

1. `getStateToPostToWebview()` never includes these fields in the state posted to the webview, so the webview always receives `undefined` for them. When the SettingsView re-mounts, `cachedState` is initialized from `extensionState` (which has `undefined`), and the checkbox renders `checked={undefined ?? true}` → always checked.

2. `getState()` never returns these fields, so the `DiffViewProvider` (which reads `saveState?.autoCloseZooOpenedFiles ?? true`) always gets `undefined` and falls back to the default `true` — meaning the auto-close behavior ignores the user's saved preference entirely.

## What version of zoo are you running

Latest (main branch)

## Additional context

The fields are properly defined in the `ExtensionState` type (`packages/types/src/vscode-extension-host.ts`) and the `globalSettingsSchema` (`packages/types/src/global-settings.ts`), and the webview correctly sends them in the `updateSettings` message. The gap is purely in the backend state serialization — the fields were never added to the `getState()` and `getStateToPostToWebview()` return objects.

navedmerchant:
### Related GitHub Issue

Closes: #667

### Description

The three auto-close settings (`autoCloseZooOpenedFiles`, `autoCloseZooOpenedFilesAfterUserEdited`, `autoCloseZooOpenedNewFiles`) were saved to global state via `contextProxy.setValue()` but never included in the state posted back to the webview or returned by `getState()`. This had two effects:

1. **UI symptom**: `getStateToPostToWebview()` omitted these fields, so the webview always received `undefined`. When the SettingsView re-mounted, `cachedState` was initialized from `extensionState` (with `undefined`), and the checkbox rendered `checked={undefined ?? true}` → always checked. The user could uncheck and save, but the value was never sent back, so it appeared unsaved on reopen.

2. **Functional symptom**: `getState()` omitted these fields, so the `DiffViewProvider` (which reads `saveState?.autoCloseZooOpenedFiles ?? true`) always got `undefined` and fell back to the default `true` — the auto-close behavior ignored the user's saved preference entirely.

**Fix**: Added the three fields to both `getState()` (reading from `stateValues`) and `getStateToPostToWebview()` (destructuring from `getState()` and including in the return object with their documented defaults: `true`, `false`, `false`).

Also added a changeset instruction to `AGENTS.md` per maintainer request.

### Test Procedure

**Unit tests** (added to `src/core/webview/__tests__/ClineProvider.taskHistory.spec.ts`):

1. `getStateToPostToWebview returns saved autoCloseZooOpenedFiles value` — Sets values via `contextProxy.setValue()`, calls `getStateToPostToWebview()`, and asserts the saved values are present in the returned state.
2. `getStateToPostToWebview defaults autoCloseZooOpenedFiles to true when unset` — Verifies unset values default to `true`/`false`/`false` in posted state.
3. `getState returns saved autoCloseZooOpenedFiles value for DiffViewProvider` — Verifies `getState()` returns the saved value so `DiffViewProvider` can read it.

**Run tests**:
```bash
cd src && npx vitest run core/webview/__tests__/ClineProvider.taskHistory.spec.ts
cd src && npx vitest run core/webview/__tests__/ClineProvider.spec.ts integrations/editor/__tests__/DiffViewProvider.spec.ts
```

All 188 tests pass (18 + 170). Lint and type-check also pass.

### Pre-Submission Checklist

- [x] **Issue Linked**: This PR is linked to issue #667.
- [x] **Scope**: My changes are focused on the linked issue (one fix per PR).
- [x] **Testing**: New regression tests have been added to cover the changes.
- [x] **Documentation Impact**: No documentation updates required (internal state serialization fix).

### Screenshots / Videos

N/A — no visual UI changes; the fix corrects data flow so the existing checkbox behaves correctly.

### Documentation Updates

### Additional Notes

The root cause was a straightforward omission: the fields were added to the `ExtensionState` type and `globalSettingsSchema`, the webview correctly sends them in `updateSettings`, and the `DiffViewProvider` correctly reads them from `getState()` — but the `getState()` and `getStateToPostToWebview()` return objects were never updated to include them. Since all fields are `optional` in the type, TypeScript did not flag the missing properties.

### Get in Touch

N/A

## Release Notes

* **New Features**
  * Added three new auto-close settings: auto-close opened files (enabled by default), auto-close opened files after user edits, and auto-close newly created files.

* **Tests**
  * Added test coverage for the new auto-close settings.


src/core/webview/__tests__/ClineProvider.taskHistory.spec.ts
```diff
@@ -685,6 +685,51 @@ describe("ClineProvider Task History Synchronization", () => {
...
+
+		it("getState returns saved autoCloseZooOpenedFiles value for DiffViewProvider", async () => {
+			await provider.resolveWebviewView(mockWebviewView)
+
+			await provider.contextProxy.setValue("autoCloseZooOpenedFiles", false)
+
+			const state = await provider.getState()
+
+			// DiffViewProvider reads from getState(); the field must be present.
+			expect(state.autoCloseZooOpenedFiles).toBe(false)
```
edelauna: Does this test need assertions for `autoCloseZooOpenedFilesAfterUserEdited` and `autoCloseZooOpenedNewFiles` too? Since the bug was about all three fields being missing from `getState()`, a regression that drops those two would pass this test as-is.

```suggestion
			await provider.contextProxy.setValue("autoCloseZooOpenedFiles", false)
			await provider.contextProxy.setValue("autoCloseZooOpenedFilesAfterUserEdited", true)
			await provider.contextProxy.setValue("autoCloseZooOpenedNewFiles", true)

			const state = await provider.getState()

			// DiffViewProvider reads from getState(); the field must be present.
			expect(state.autoCloseZooOpenedFiles).toBe(false)
			expect(state.autoCloseZooOpenedFilesAfterUserEdited).toBe(true)
			expect(state.autoCloseZooOpenedNewFiles).toBe(true)
```

src/core/webview/__tests__/ClineProvider.taskHistory.spec.ts
```diff
@@ -685,6 +685,51 @@ describe("ClineProvider Task History Synchronization", () => {
 		})
 	})
 
+	describe("auto-close settings are included in posted state", () => {
```
edelauna: Is there a reason these tests live in `ClineProvider.taskHistory.spec.ts` rather than `ClineProvider.spec.ts`? The existing `getState` settings coverage (e.g. `autoCondenseContext`, `writeDelayMs`) is in the latter, so a contributor looking for where settings persistence is tested would miss these.

edelauna (changes requested):

edelauna (approved):

PR #434, opened 2026-06-01
feat(api): pass abortSignal to streaming API calls for all providers (#404)

easonLiangWorldedtech:
## Summary

Fixes #404 — When the user clicks Stop during streaming, the HTTP request to the provider continues running because the abort signal wasn't being passed through. This wastes API tokens/compute on the provider side.

### Changes

**Commit 1: `feat(api): add abort signal support for API providers and task`**
- Added `{ signal: metadata?.abortSignal }` to streaming `.chat.completions.create()` calls in **16 providers**:
  - lm-studio.ts, openai.ts, deepseek.ts
  - base-openai-compatible-provider.ts (merged into requestOptions)
  - lite-llm.ts, mimo.ts, opencode-go.ts, openrouter.ts, qwen-code.ts, requesty.ts, unbound.ts, vercel-ai-gateway.ts, zai.ts
  - anthropic.ts, azure-openai.ts, bedrock.ts
- Added `abortSignal?: AbortSignal` to `ApiHandlerCreateMessageMetadata` interface
- Wired abort signal through Task.ts's AbortController

**Commit 2: `feat(api): integrate external abort signal with internal controllers`**
- Previously, Anthropic, Bedrock, and OpenAI Native used their own internal AbortControllers but did **not** listen to the external abort signal from the Task execution chain. Clicking stop would not immediately cancel requests to these providers. Now all providers consistently respect the user's stop action.
- Added optional `metadata` parameter with `abortSignal` to `completePrompt` methods across **22 provider implementations**:
  - Anthropic, Anthropic Vertex, Base OpenAI Compatible
  - Gemini/Vertex, LiteLLM, LM Studio, MiniMax, Mistral
  - Native Ollama, OpenAI, OpenAI Compatible, OpenRouter
  - Qwen Code, Requesty, Unbound, Vercel AI Gateway
  - Opencode Go, xAI, Zoo Gateway, VSCode LM, Poe, Bedrock, OpenAI Codex, OpenAI Native
- Added comment explaining Mistral `completePrompt` does not support non-streaming abort (Mistral SDK limitation)

### Test Coverage
- `task-abort-signal-passing.spec.ts` — Comprehensive tests for abort signal flow through Task execution chain
- `openai-compatible-abort-signal.spec.ts` — Provider-level abort signal verification for OpenAI Compatible handler
- Added abort signal tests for all modified providers (20+ test files updated)

### Architecture
```
User clicks Stop
  → Task.abort() called
  → currentRequestAbortController.abort() triggered
[...]

PR #674, opened 2026-06-20
feat: add abort signal core plumbing

easonLiangWorldedtech:
## Summary

Core plumbing to pass `AbortSignal` through the API layer so HTTP requests can be cancelled mid-stream when the user clicks stop. Prevents wasted API tokens/compute on the provider side.

## Changes

- **src/api/index.ts**: Add `abortSignal?: AbortSignal` to `ApiHandlerCreateMessageMetadata` interface
- **src/core/task/Task.ts**: Wire Task's `AbortController.signal` into createMessage metadata
- **src/core/task/__tests__/Task.spec.ts**: Tests verifying signal is passed and matches currentRequestAbortController

## Related

Closes #434

## Summary by CodeRabbit

* **New Features**
  * Added support for cancelling in-flight streaming responses by forwarding the active abort signal to the underlying streaming request.
* **Bug Fixes**
  * Improved consistency of abort-signal propagation across context condensation and automatic context-management so cancellation is respected in all flows.
* **Tests**
  * Expanded coverage to verify abort-signal forwarding during context condensation, request attempts, and forced context-management, including rejection when cancellation occurs mid-stream.

### DIFF
--- a/src/core/task/Task.ts
+++ b/src/core/task/Task.ts
@@ -4161,6 +4161,7 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
 		// Create an AbortController to allow cancelling the request mid-stream
 		this.currentRequestAbortController = new AbortController()
 		const abortSignal = this.currentRequestAbortController.signal
+		metadata.abortSignal = abortSignal
 		// Reset the flag after using it
 		this.skipPrevResponseIdOnce = false
 
--- a/src/core/task/Task.ts
+++ b/src/core/task/Task.ts
@@ -1555,6 +1555,11 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
 		const metadata: ApiHandlerCreateMessageMetadata = {
 			mode,
 			taskId: this.taskId,
+			...(this.currentRequestAbortController?.signal
+				? {
+						abortSignal: this.currentRequestAbortController.signal,
+					}
+				: {}),
 			...(allTools.length > 0
 				? {
 						tools: allTools,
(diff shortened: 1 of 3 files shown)


src/core/task/Task.ts
```diff
@@ -4161,6 +4161,7 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
 		// Create an AbortController to allow cancelling the request mid-stream
 		this.currentRequestAbortController = new AbortController()
 		const abortSignal = this.currentRequestAbortController.signal
+		metadata.abortSignal = abortSignal
```
edelauna: The `Promise.race` between `firstChunkPromise` and `abortPromise` (lines 4187–4198) is the core behaviour this adds — if the signal fires before the first chunk, `attemptApiRequest` should throw `"Request cancelled by user"`. Is that path covered by a test? The new tests verify the signal is *passed* correctly, but I don't see one that actually calls `abort()` during `iterator.next()` and asserts the error.

edelauna: Minor: `abortSignal` is grafted onto `metadata` after the object literal closes (line 4144), while every other field is set inline. Worth pulling the `AbortController` creation up above the literal and including `abortSignal` there?

edelauna (changes requested):
Thanks for doing this. Would like to see an additional test case, and also could you update Condensing/context-management API calls, three `ApiHandlerCreateMessageMetadata` construction sites in `Task.ts` — at lines 1555, 3763, and 3979 — do not set  `abortSignal`. These are used for long-running condensing and context-management API calls that are also cancellable,  but the "core plumbing" claim is incomplete if these paths are excluded.

Suggested fix: Pass the active `currentRequestAbortController?.signal` to these metadata objects, or document explicitly that condensing calls are intentionally out of scope with a comment at each site.

edelauna (approved):
thank you for breaking this up

PR #687, opened 2026-06-22
fix: shell default profile name type guard

### LINKED ISSUE #686: [BUG] Shell detection crashes when terminal default profile setting is not a string
**Describe the bug**
Zoo can crash with `Provider ended the request: t.toLowerCase is not a function` when the VS Code terminal default profile setting is misconfigured with a non-string value.

**To Reproduce**
1. Open VS Code settings JSON.
2. Set the terminal default profile to a non-string value. For example, on Windows:
   ```json
   {
     "terminal.integrated.defaultProfile.windows": 1
   }
   ````
    Other invalid values such as `true`, `[]`, or `{}` can also reproduce the issue.
3. Run Zoo and trigger a request that causes Zoo to resolve the shell configuration.
4. See error:
   ```text
   Provider ended the request: t.toLowerCase is not a function
   ```

**Expected behavior**
Zoo should not crash when VS Code terminal profile settings contain an invalid type.

**Screenshots**
N/A

**Video**
N/A

3.62.0

**Additional context**
This seems to be a runtime type validation issue rather than a TypeScript typing issue. `config.get<string>()` only affects the static type, but it does not guarantee that the actual value from user settings is a string.

The current code assumes `defaultProfileName` is a string before calling `.toLowerCase()`, which can fail when the VS Code settings JSON is manually edited or corrupted.

daewoongoh:
### Related GitHub Issue

Closes: #686 

### Description

`config.get<string>()` in `src/utils/shell.ts` provides only a compile-time type hint — if a user sets a non-string value (number, boolean, array, object) in `settings.json`, it is returned as-is at runtime. This causes a TypeError when `.toLowerCase()` is subsequently called on `defaultProfileName`.

- Add a `typeof === "string"` runtime type guard to coerce non-string values to `null`
- Consolidate the duplicated `getWindowsTerminalConfig`, `getMacTerminalConfig`, and `getLinuxTerminalConfig` into a single generic `getTerminalConfig<T>(platformKey)`
- Add test cases for non-string input handling (8 cases) and getTerminalConfig behavior (7 cases)

### Test Procedure

- Run `vitest run src/utils/__tests__/shell.spec.ts` — all 53 tests pass
- Verified that non-string `defaultProfileName` scenarios (number, boolean, array, object) return the platform fallback shell without throwing TypeError

### Pre-Submission Checklist

### Screenshots / Videos

N/A

### Documentation Updates

- [X ] Yes, documentation updates are required. (Please describe what needs to be updated or link to a PR in the docs repository).

### Additional Notes

`config.get<T>()` does not guarantee type safety at runtime due to VS Code API behavior — runtime validation at the system boundary is necessary.

### Get in Touch

hehegwk_23849

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for terminal profile selection on Windows, macOS, and Linux, including cases where configuration values are missing, null, or of invalid types, plus fallback behavior when configuration access fails.

* **Refactor**
  * Streamlined terminal configuration retrieval into a single shared flow that selects the correct platform profile data and applies consistent safe fallbacks on errors.


src/utils/shell.ts
```diff
@@ -136,36 +136,19 @@ type LinuxTerminalProfiles = Record<string, LinuxTerminalProfile>
...
-		return { defaultProfileName, profiles }
-	} catch {
-		return { defaultProfileName: null, profiles: {} as MacTerminalProfiles }
-	}
-}
+type TerminalProfiles = WindowsTerminalProfiles | MacTerminalProfiles | LinuxTerminalProfiles
 
-function getLinuxTerminalConfig() {
+function getTerminalConfig<T extends TerminalProfiles>(
+	platformKey: string,
```
edelauna: Could this be narrowed to `"windows" | "osx" | "linux"`? Passing `"darwin"` instead of `"osx"` would silently return empty config today. `Terminal.getPlatformProfileKey()` already returns this union.

Taking it a step further, a mapped-type approach would also eliminate the explicit `<T>` at each call site and prevent mismatches like `getTerminalConfig<MacTerminalProfiles>("windows")`:

```suggestion
	platformKey: "windows" | "osx" | "linux",
```

src/utils/shell.ts
```diff
@@ -136,36 +136,19 @@ type LinuxTerminalProfiles = Record<string, LinuxTerminalProfile>
...
+): { defaultProfileName: string | null; profiles: T } {
 	try {
 		const config = vscode.workspace.getConfiguration("terminal.integrated")
-		const defaultProfileName = config.get<string>("defaultProfile.linux")
-		const profiles = config.get<LinuxTerminalProfiles>("profiles.linux") || {}
+		const rawProfileName = config.get<string>(`defaultProfile.${platformKey}`)
+		const defaultProfileName = typeof rawProfileName === "string" ? rawProfileName : null
+		const profiles = config.get<T>(`profiles.${platformKey}`) || ({} as T)
```
edelauna: Should this be `??` instead of `||`? The rest of the codebase is moving toward nullish coalescing for config reads (e.g. `Task.ts:3696`, `extension.ts:155`), and `??` is more precise — only fires on `null`/`undefined`.

```suggestion
		const profiles = config.get<T>(`profiles.${platformKey}`) ?? ({} as T)
```

src/utils/__tests__/shell.spec.ts
```diff
@@ -370,6 +370,200 @@ describe("Shell Detection Tests", () => {
...
+				get: vi.fn((key: string) => {
+					if (key === "defaultProfile.osx") return {}
+					if (key === "profiles.osx") return {}
+					return undefined
+				}),
+			}
+			vi.mocked(vscode.workspace.getConfiguration).mockReturnValue(mockConfig as any)
+
+			expect(getShell()).toBe("/bin/zsh")
+		})
```
edelauna, line 564: Two things about the macOS/Linux non-string tests:

1. **Asymmetric coverage** — Windows tests all 4 non-string types (numeric, boolean, array, object) but macOS and Linux only test 2. The PR description commits to covering all four.

2. **Mutation resistance** — these tests won't actually catch removal of the `typeof` guard. `getMacShellFromVSCode` never calls `.toLowerCase()` on `defaultProfileName` — it only uses it as an object key. So a non-string value silently misses the key lookup → `undefined` → `normalizeShellPath` returns `null` → falls back to `/bin/zsh`, which is exactly what the test asserts. Could you add a case where the non-string value, if left unguarded, would match a profile key (making the guarded vs. unguarded paths observably different)?

edelauna (changes requested):

src/utils/shell.ts
daewoongoh: Agreed. I changed the profiles fallback from `||` to `??` so we only fall back to an empty profile map for `null`/`undefined`, which is more precise for config reads.

src/utils/__tests__/shell.spec.ts
daewoongoh: I added coverage for this in `9fbc9e1`.

Changes made:
- Added the missing boolean and array `defaultProfileName` cases for macOS and Linux so they now match the Windows non-string coverage.
- Added mutation-resistant macOS/Linux tests where a numeric non-string value would otherwise match a real `"1"` profile key if the `typeof` guard were removed.

src/utils/shell.ts
daewoongoh: I replaced the TerminalProfiles union with a PlatformProfilesMap, so getTerminalConfig() now accepts only "windows" | "osx" | "linux" and infers the matching profile map type from the platform key. 
This removes the explicit generic type argument at each call site and makes mismatches like getTerminalConfig<MacTerminalProfiles>("windows") a compile-time error.

edelauna (approved):
Thanks for adding this.

PR #700, opened 2026-06-23
fix: streaming tool-call argument loss in NativeToolCallParser (#695)

### LINKED ISSUE #695: [BUG] Streaming tool-call argument deltas are silently dropped when arguments arrive before the tool-call id
## Description

When a provider streams a tool call whose first delta(s) carry argument bytes
*before* the tool-call `id` is known, those leading argument bytes are silently
discarded by `NativeToolCallParser.processRawChunk`. The accumulator then begins
mid-JSON, so the finalized arguments are incomplete or unparseable. Downstream
this surfaces as a misleading "missing required parameter" error even though the
model did supply the parameter.

This is the streaming-path root cause behind intermittent tool-call failures. Here the data is lost in transit during
streaming reassembly, so by the time finalization runs the parameter genuinely
is absent from the accumulated JSON.

### Related issues

- #690 -- `ask_followup_question` "missing required parameter follow_up"
- #218 -- "tool call silently dropped when id and name arrive in separate SSE
  chunks" (P1-high). Same class of cross-chunk-identity defect; closest prior art.
- #547 -- DeepSeek 5-defect compatibility report. Its Defect 1 ("Missing Required
  Parameters" / infinite retry) and Defect 4 ("XML Tag Fallback Missing": model
  emits `<read_file>...</read_file>` as plain text under load) overlap on
  symptoms (missing params, XML leaking into content) but attribute them to
  schema strict-mode and XML-fallback detection rather than streaming data loss.
  Defect 4 corroborates that the observed "XML in response content" is model-side
  output, not environment serialization of native calls into XML.

## Root Cause

In `src/core/assistant-message/NativeToolCallParser.ts`, `processRawChunk`
(approx. lines 99-164) keys its `rawChunkTracker` by the stream `index` but only
*initializes* a tracker entry when a chunk carries an `id`:

```ts
let tracked = this.rawChunkTracker.get(index)

// Initialize new tool call tracking when we receive an id
if (id && !tracked) {
    tracked = { id, name: name || "", hasStarted: false, deltaBuffer: [] }
    this.rawChunkTracker.set(index, tracked)
}

if (!tracked) {
    return events // <-- any `arguments` on this chunk are dropped permanently
}
```

If the first chunk for a given `index` contains `arguments` but no `id`, the
early return discards those bytes. The existing `deltaBuffer` only buffers
deltas that arrive *after* a tracker exists (between tracker creation and the
`tool_call_start` emission); it cannot help pre-`id` deltas because no tracker
exists yet to hold them.

### Failure chain

1. Leading `arguments` fragment arrives without `id` -> dropped at the early return.
2. `id` arrives in a later chunk -> tracker created, accumulation starts mid-JSON.
3. `finalizeStreamingToolCall` cannot recover the missing leading keys -> returns
   `null` or an object missing the first parameter(s).
4. `Task.ts` (approx. line 2817) marks the tool complete with missing params;
   the tool guard reports "missing required parameter" -- a misleading symptom.

## Provider Scope

The fix is centralized: every provider funnels tool deltas through the single
`processRawChunk` function, so all of them benefit. The most exposed providers
are the raw-passthrough ones that forward `id`/`name`/`arguments` exactly as
received from upstream:

- `lite-llm.ts` (proxies arbitrary upstream models -- highest exposure)
- the OpenAI-compatible family (`openai.ts`, `openrouter.ts`, `requesty.ts`,
  `unbound.ts`, `deepseek.ts`, `vercel-ai-gateway.ts`, `zoo-gateway.ts`,
  `base-openai-compatible-provider.ts`, etc.)

`openai-native.ts` and `openai-codex.ts` already self-guard by withholding
`tool_call_partial` until a name is known; they are not currently susceptible,
and the parser fix makes that defensive logic redundant (cleanup is optional and
out of scope here).

### This is a known, documented limitation

The weakness is already called out in code comments at the two providers that
work around it -- confirming the parser, not the providers, is the correct place
to fix it:

- `src/api/providers/openai-native.ts` (~line 1235):

  ```ts
  // Avoid emitting incomplete tool_call_partial chunks; the downstream
  // NativeToolCallParser needs a name to start a call.
  ```

- `src/api/providers/openai-codex.ts` (~line 954):

  ```ts
  // Codex/Responses may stream tool-call arguments, but these delta events are not guaranteed
  // to include a stable id/name. Avoid emitting incomplete tool_call_partial chunks because
  // NativeToolCallParser requires a name to start a call.
  ```

These guards prevent the drop for those two providers but were never added to the
raw-passthrough providers (LiteLLM, OpenAI-compatible family), which is precisely
why those remain exposed. Fixing `processRawChunk` resolves the limitation
centrally and renders both guards redundant.

## Affected Files

- `src/core/assistant-message/NativeToolCallParser.ts` (`processRawChunk`, ~99-164)

## Proposed Fix

Initialize the tracker on first sight of an `index`, independent of `id`:

1. Create the `rawChunkTracker` entry when an `index` is first seen, with `id`
   left optional until it arrives.
2. Buffer all `arguments` into `deltaBuffer` while `hasStarted` is false (covers
   both pre-`id` and pre-`name` cases).
3. Record `id` and `name` when they arrive; emit `tool_call_start` and flush the
   buffered deltas only once both `id` and `name` are known.
4. Keep the downstream event sequence identical so no provider or `Task.ts`
   changes are required.

## Test Procedure

1. Add a parser-layer unit test in
   `src/core/assistant-message/__tests__/NativeToolCallParser.spec.ts` that feeds
   `processRawChunk` a leading chunk with `arguments` and no `id`, followed by a
   chunk carrying `id` and `name`, and asserts the full argument string is
   preserved through `finalizeStreamingToolCall`.
2. Verify existing provider specs that drive `processRawChunk` directly still
   pass (`lmstudio-native-tools.spec.ts`, `qwen-code-native-tools.spec.ts`,
   `openrouter.spec.ts`).
3. Manual: reproduce a tool call via LiteLLM in F5 development mode and confirm
   no spurious "missing required parameter" error.

awschmeder:
### Related GitHub Issue

Closes: #695

### Description

When a provider streams a tool call whose first delta(s) arrive *before* the tool-call `id` is known, those leading argument bytes are silently discarded by `NativeToolCallParser.processRawChunk`. This causes downstream "missing required parameter" and other spurious provider-dependent errors even when the model supplied the correct tool use syntax.

This PR fixes the issue by centralizing the tracking of streaming tool calls in `NativeToolCallParser`. The `rawChunkTracker` is now initialized on the first sight of a stream `index`, independent of whether an `id` is present. All `arguments` deltas are buffered until both `id` and `name` are known, ensuring no data loss during streaming reassembly.

#### Scope note: accompanying `Task.ts` change

The fix spans two layers because the parser and its streaming consumer must agree on when a tool call is finalized:

- **`NativeToolCallParser`** (the core fix): buffers pre-`id` argument deltas and only emits `tool_call_end` for started trackers with a non-empty `id`, preventing both data loss and phantom end events.
- **`Task.ts`** (required consumer change): providers emit a `tool_call_end` stream chunk on `finish_reason: "tool_calls"` (either directly, or via `processFinishReason()` for openrouter/lm-studio/qwen-code). `Task.ts`'s stream `switch` had **no** `tool_call_end` case, so those chunks were silently dropped and tool calls only finalized at stream end via `finalizeRawChunks()`. Without this change, the parser's correctly-buffered arguments would still not be presented during streaming. Adding the new case would have created a third copy of the finalize/present logic (the codebase already had two: the per-chunk event loop and the `finalizeRawChunks()` loop), so all three sites are consolidated into a single idempotent helper, `finalizeStreamingToolCallById(id)`. Re-finalizing an already-cleared id is a safe no-op, so the new streaming finalization and the end-of-stream pass cannot double-present.

### Test Procedure

1. Ran the newly added unit test in `src/core/assistant-message/__tests__/NativeToolCallParser.spec.ts` which verifies that leading argument bytes arriving before the `id` are correctly preserved and finalized.
2. Verified that existing provider tests in the same test file pass.
3. Added `src/core/task/__tests__/finalizeStreamingToolCallById.spec.ts`, which exercises the real `Task.finalizeStreamingToolCallById` helper across the success, malformed-JSON, untracked-id, and idempotent re-finalize paths (closes the patch-coverage gap on the consumer change).

### Pre-Submission Checklist

### Screenshots / Videos

N/A

### Documentation Updates

### Additional Notes

N/A

### Get in Touch

@awschmeder

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streamed native tool-call reassembly, allowing tool-call start when `id`/`name` arrive after arguments.
  * Buffered early argument chunks until the tool call is identified.
  * Prevented phantom or duplicate tool-call end events for malformed or double-finished streams.
  * Preserved correct event sequencing and isolation for parallel tool calls by stream index.
* **Tests**
  * Added streaming reassembly and event-ordering coverage for late/split/out-of-order chunks.
  * Added coverage for consistent finalization behavior in `finalizeStreamingToolCallById`.


awschmeder:
@coderabbitai review

src/core/assistant-message/__tests__/NativeToolCallParser.spec.ts
```diff
@@ -343,4 +343,201 @@ describe("NativeToolCallParser", () => {
...
+				{ index: 0, arguments: argsA.slice(8) },
+				{ index: 1, arguments: argsB.slice(8) },
+			])
+
+			const starts = events.filter((e) => e.type === "tool_call_start")
+			expect(starts).toHaveLength(2)
+
+			const resultA = finalized.get("call_a")
+			const resultB = finalized.get("call_b")
+			if (resultA?.type === "tool_use") {
```
edelauna, line 485: If `finalizeStreamingToolCall` returned `null` here (the exact failure mode this PR fixes), this guard silently skips the payload assertion and the test passes green. Could you add `expect(resultA).not.toBeNull()` before this `if`? Same for `resultB` on line 460.

src/core/assistant-message/__tests__/NativeToolCallParser.spec.ts
```diff
@@ -343,4 +343,201 @@ describe("NativeToolCallParser", () => {
...
+
+			const { events, finalized } = drive([
+				{ index: 0, id: "call_single", name: "read_file", arguments: fullArgs },
+			])
+
+			expect(events.map((e) => e.type)).toEqual(["tool_call_start", "tool_call_delta", "tool_call_end"])
+			expect(events.every((e) => e.id === "call_single")).toBe(true)
+
+			const result = finalized.get("call_single")
+			if (result?.type === "tool_use") {
```
edelauna, line 506: Same pattern — if `finalizeStreamingToolCall` returns `null`, the path assertion is silently skipped.

```suggestion
				const result = finalized.get("call_single")
				expect(result).not.toBeNull()
				expect(result?.type).toBe("tool_use")
				if (result?.type === "tool_use") {
					expect((result.nativeArgs as { path: string }).path).toBe("src/single.ts")
				}
```

src/core/assistant-message/__tests__/NativeToolCallParser.spec.ts
```diff
@@ -343,4 +343,201 @@ describe("NativeToolCallParser", () => {
...
+			expect(result).not.toBeNull()
+			expect(result?.type).toBe("tool_use")
+			if (result?.type === "tool_use") {
+				const nativeArgs = result.nativeArgs as { path: string; mode?: string }
+				expect(nativeArgs.path).toBe("src/leading.ts")
+				expect(nativeArgs.mode).toBe("slice")
+			}
+		})
+
+		it("handles id and name arriving in separate chunks (issue #218)", () => {
```
edelauna, line 414: Could we also cover the reverse ordering — name in chunk 1, id in chunk 2, with buffered args in between? That path is valid under the new logic but nothing in the suite currently catches a mutation that drops `tracked.id` from the start-gate condition (`!tracked.hasStarted && tracked.id && tracked.name` at line 132 of the source).

src/core/assistant-message/NativeToolCallParser.ts
```diff
@@ -118,38 +119,39 @@ export class NativeToolCallParser {
...
-		if (!tracked) {
-			return events
+		// Record id and name as they arrive (they may come in separate chunks).
+		if (id) {
+			tracked.id = id
 		}
-
-		// Update name if present in chunk and not yet set
 		if (name) {
 			tracked.name = name
```
edelauna, line 132: If a provider sends `name: ""` (empty string, permitted by the type), the falsy `if (name)` check silently discards it and `tracked.name` stays `""` permanently — `hasStarted` would never fire for that tracker. Is `if (name !== undefined)` closer to the intended semantics?

src/core/assistant-message/NativeToolCallParser.ts
```diff
@@ -172,10 +174,15 @@ export class NativeToolCallParser {
...
 		if (finishReason === "tool_calls" && this.rawChunkTracker.size > 0) {
 			for (const [, tracked] of this.rawChunkTracker.entries()) {
-				events.push({
-					type: "tool_call_end",
-					id: tracked.id,
-				})
+				// Only emit an end for trackers that actually started. A tracker that
+				// never received an id/name (malformed stream) must not emit a phantom
+				// end; since start requires an id, hasStarted implies tracked.id is set.
+				if (tracked.hasStarted && tracked.id) {
```
edelauna, line 186: `processFinishReason` emits `tool_call_end` for all started trackers but doesn't clear `rawChunkTracker`. Task.ts calls `finalizeRawChunks()` unconditionally afterward, which iterates the same map and emits another round of ends. Today the second wave is silently dropped (the id is already deleted from `streamingToolCallIndices` on the first pass), but it's a latent double-fire risk if that dispatch path changes. Would adding `this.rawChunkTracker.clear()` at the end of the `if (finishReason === "tool_calls")` block make `finalizeRawChunks()` a safe no-op for this path?

src/core/prompts/tools/native-tools/ask_followup_question.ts
```diff
@@ -14,7 +14,7 @@ Example: Asking with mode switch
 
 const QUESTION_PARAMETER_DESCRIPTION = `Clear, specific question that captures the missing information you need`
 
-const FOLLOW_UP_PARAMETER_DESCRIPTION = `Required list of 2-4 suggested responses; each suggestion must be a complete, actionable answer and may include a mode switch`
+const FOLLOW_UP_PARAMETER_DESCRIPTION = `Required array of 1-4 suggested responses, always an array even for a single suggestion; each suggestion must be a complete, actionable answer and may include a mode switch`
```
edelauna: This wording change (2→1 minimum, "list"→"array") looks like a follow-up to the prior `minItems` schema fix rather than something related to the streaming argument buffering in this PR. The PR description doesn't mention it — was this intentional to bundle here, or should it live in a separate commit so it's independently bisectable?

prs/fix-toolcall-dropped-leading-deltas.md
edelauna, line 1: lets remove this from diff

edelauna (changes requested):
Thanks! Had a couple comments on testing some edge cases.

awschmeder:
Pushed a follow-up commit (`f2396b9`) extending this fix to the streaming consumer in `Task.ts`.

**Problem:** Providers emit a `tool_call_end` stream chunk on `finish_reason: "tool_calls"` -- either directly (OpenAI-style providers tracking their own `activeToolCallIds`) or via `NativeToolCallParser.processFinishReason()` (openrouter, lm-studio, qwen-code). But `Task.ts`'s stream `switch` had no `tool_call_end` case, so those chunks were silently dropped and tool calls only finalized at stream end via `finalizeRawChunks()`.

**Changes:**
- Added a `tool_call_end` case to the `Task.ts` stream switch so tools finalize and present during streaming.
- Extracted the previously triplicated finalize/present logic (per-chunk event loop, the new case, and the `finalizeRawChunks` loop) into a single idempotent helper `finalizeStreamingToolCallById(id)`. Re-finalizing an already-cleared id is a safe no-op, so the new streaming finalization and the end-of-stream `finalizeRawChunks()` pass cannot double-present.
- Corrected the `NativeToolCallParser` test `drive` helper to emit ends via `finalizeRawChunks()` (matching what `Task.ts` actually does) instead of `processFinishReason()`; its comment previously misdescribed the production wiring. The dedicated double-fire test still exercises `processFinishReason` directly, since that remains a real provider-facing API.

**Verification:**
- `NativeToolCallParser.spec.ts`: 21/21 pass
- `openrouter.spec.ts` + `lmstudio-native-tools.spec.ts` + `base-openai-compatible-provider.spec.ts`: 45/45 pass
- `duplicate-tool-use-ids.spec.ts` + `presentAssistantMessage-custom-tool.spec.ts`: 18/18 pass
- `npx tsc --noEmit` on `src`: clean

src/core/assistant-message/__tests__/NativeToolCallParser.spec.ts
awschmeder: Yes--added expect(...).not.toBeNull() before the payload assertions so a null finalize result fails instead of passing silently.

src/core/assistant-message/__tests__/NativeToolCallParser.spec.ts
awschmeder: Confirmed--added a test feeding name -> buffered args -> id that asserts start fires only once the id arrives and buffered deltas flush in order, exercising the tracked.id part of the start-gate.

src/core/assistant-message/NativeToolCallParser.ts
awschmeder: Yes---switched recording to name !== undefined and added a nameSeen flag used by the start-gate (the truthiness check would otherwise have stranded the tracker). The comment clarifies this is a defensive guard rather than expected provider behavior...

src/core/assistant-message/NativeToolCallParser.ts
awschmeder: Confirmed --- processFinishReason now clears rawChunkTracker, making the subsequent finalizeRawChunks() a safe no-op; added a regression test asserting exactly one tool_call_end across both paths.

This comment also opened up a deeper review of handling at end of stream, which ended up needing to widen the scope slightly to handle the consumer side in Task.ts (PR description was revised above with Scope note):

src/core/prompts/tools/native-tools/ask_followup_question.ts
awschmeder: 😆 Oops!!  Something from another branch I was working on... removed.

prs/fix-toolcall-dropped-leading-deltas.md
awschmeder: 👍

src/core/assistant-message/__tests__/NativeToolCallParser.spec.ts
awschmeder: Confirmed -- the expect(result).not.toBeNull() and expect(result?.type).toBe("tool_use") guards now run before the if, so a null return fails the test rather than being silently skipped.

awschmeder:
@coderabbitai review

PR #710, opened 2026-06-24
fix(vscode-lm): reliable auto context condensing

### LINKED ISSUE #714: vscode-lm: automatic context condensing never triggers (maxTokens -1 + inflated window)

On the **VS Code LM (`vscode-lm`) provider**, automatic context condensing does not trigger reliably. Even when the context gauge in the UI shows the window as effectively full, auto-condense never fires, so the conversation eventually overflows the model's real input limit.

Affects users on the VS Code LM (Copilot) provider who rely on automatic context condensing. It surfaces during long conversations with large-window models (e.g., a Claude/GPT-5 family entry): the UI context gauge climbs to full, but auto-condense never kicks in.

## Root cause

The `vscode-lm` provider reports `maxTokens: -1` (unlimited) and an inflated **live** context window (Copilot's advertised window, far larger than the realistic usable input). Two problems follow:

1. The condense gate computed `contextPercent` against the **full** context window instead of the **available input space** (`contextWindow - reservedForOutput`), so usage was under-reported and the threshold was never reached.
2. A negative `maxTokens` (`-1`) was used directly as the reserved-output value, distorting the window math (`maxTokens || DEFAULT` kept `-1`).

## Reproduction steps

1. Select the VS Code LM (Copilot) provider with a large-window model (e.g., a Claude/GPT-5 family entry).
2. Enable automatic context condensing with a normal threshold (e.g., 70–80%).
3. Drive a long conversation until the UI context gauge shows the window near/over full.
4. Observe that auto-condense does not trigger, despite the gauge indicating the context is effectively full.

## Expected result

Auto-condense should fire in line with the context gauge: usage should be measured against the usable input space, and the gate should use the model's real input ceiling (the curated `maxInputTokens`) rather than the inflated live window. A negative/unlimited `maxTokens` should fall back to a sane default reserve.

## Actual result

Auto-condense never triggers; usage is measured against the inflated full window, so the threshold is never reached and the conversation eventually overflows the model's real input limit.

## Variations tried

Reproduces across large-window vscode-lm models regardless of the configured condense threshold, because the denominator (full live window) is wrong rather than the threshold.

## App Version

N/A (provider-level context-management behavior; not tied to a specific release).

## API Provider

VS Code Language Model API (`vscode-lm` / Copilot).

## Model Used

Large-window vscode-lm models (e.g., a Claude/GPT-5 family entry).

## Fix

Addressed by #710:
- Treat `maxTokens: -1` (unlimited) as the default output reserve in `willManageContext`/`manageContext`.
- Measure `contextPercent` against available input space (`contextWindow - reservedForOutput`), with a safe fallback to the full window when the reserve is unknown.
- Add an optional `getCondenseContextWindow()` `ApiHandler` seam; `vscode-lm` overrides it to use the curated static `maxInputTokens`.
- Refresh the vscode-lm model catalog/default and add UI guards so the context bar and the gate share one source of truth.

simurg79:
### Related GitHub Issue

Closes #714

### Description

Fixes unreliable automatic context condensing on the VS Code LM (`vscode-lm`) provider. The provider reports `maxTokens: -1` (unlimited) and an inflated live context window, so the auto-condense gate computed usage against the wrong denominator and effectively never fired even when the UI context gauge showed the window as full.

- Treat `maxTokens: -1` (unlimited) as the default output reserve in both `willManageContext` and `manageContext` instead of letting a negative value distort the window math.
- Measure usage against the available input space (`contextWindow - reservedForOutput`), matching the UI gauge, with a safe fallback to the full window when the reserve is unknown/unlimited.
- Add an optional `getCondenseContextWindow()` `ApiHandler` seam; only the vscode-lm provider overrides it to drive the gate from the curated static-table `maxInputTokens`. Every other provider falls back to `modelInfo.contextWindow` (behavior unchanged).
- Refresh the VS Code LM model catalog and default model (`claude-sonnet-4.5`).
- UI guards so the context bar treats a negative `maxTokens` as zero reserve and resolves an unlisted family to the default-model window.

### Test Procedure

- `packages/types` → `vscode-llm.spec.ts`: model catalog invariants.
- `src` → `context-management.spec.ts`: reserve guard + available-input denominator, including the `availableInputTokens <= 0 → 100%` fallback and end-to-end `manageContext` summarization.
- `src` → `vscode-lm.spec.ts`: `getCondenseContextWindow()` resolution (static family, live fallback, non-positive guard).
- `webview-ui` → `TaskHeader.spec.tsx`, `useSelectedModel.spec.ts`: UI reserve/window guards.
- Full CI: all checks green; `codecov/patch` at 88.88% (≥ 80% target).

### Pre-Submission Checklist

- [x] **Issue Linked**: Closes #714 (see "Related GitHub Issue" above).
- [x] **Scope**: Changes are scoped to the vscode-lm auto-condense fix (one fix per PR).
- [x] **Testing**: Added/updated unit tests for the changed logic.
- [x] **Documentation Impact**: Considered; no user-facing docs required (behavior aligns auto-condense with the existing context gauge).
- [x] No extension version bump (Zoo-Code release versioning is maintainer-managed); no agent-created changeset.
- [x] All CI checks pass.

### Documentation Updates

- [x] No documentation updates are required; behavior aligns auto-condense with the existing context gauge.

### Additional Notes

Port of simurg79/Roo-Code#11 into Zoo-Code. Applied by context (paths map 1:1); the upstream version bump was intentionally omitted.

## Summary

* **New Features**
  * Refreshed VS Code LLM provider model catalog and updated the default model selection.
  * Improved provider-aware auto-condense context-window behavior.

* **Bug Fixes**
  * Auto-summarization thresholds now use available input space, with safer handling when output is “unlimited”.
  * Fixed context usage percentage display so negative/unlimited output settings don’t skew results.

* **Tests**
  * Expanded coverage for VS Code LLM model selection, context-window derivation, and auto-condense gating scenarios.


.changeset/vscode-lm-condense-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+"zoo-code": patch
+---
+
+Fix unreliable automatic context condensing on the VS Code LM (vscode-lm) provider. The condense gate now treats the provider's `maxTokens: -1` (unlimited) as the default output reserve and measures usage against available input space, and a new `getCondenseContextWindow()` seam makes the gate use the curated model `maxInputTokens` instead of the inflated live window. Also refreshes the VS Code LM model catalog and default model.
```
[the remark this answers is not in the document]
simurg79: Done in e45155d3b: removed .changeset/vscode-lm-condense-fix.md (changesets are maintainer-managed per AGENTS.md).

src/api/providers/__tests__/vscode-lm.spec.ts
```diff
@@ -440,6 +441,84 @@ describe("VsCodeLmHandler", () => {
...
+					vsCodeLmModelSelector: { vendor: "copilot", family },
+				})
+				guardHandler["client"] = mockLanguageModelChat as unknown as vscode.LanguageModelChat
+				expect(guardHandler.getCondenseContextWindow()).toBe(guardHandler.getModel().info.contextWindow)
+				expect(guardHandler.getCondenseContextWindow()).toBe(mockLanguageModelChat.maxInputTokens)
+				guardHandler.dispose()
+			} finally {
+				;(vscodeLlmModels[family] as { maxInputTokens: number }).maxInputTokens = original
+			}
+		})
```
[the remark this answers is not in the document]
simurg79: Fixed in e45155d3b: client is now left unset so the claude-opus-4.8 selector family drives the lookup. The zeroed static row is consulted, the maxInputTokens > 0 guard fails, and it falls back to getModel().info.contextWindow. The row mutation is restored in a finally block to avoid cross-test pollution.

webview-ui/src/components/ui/hooks/useSelectedModel.ts
```diff
@@ -310,8 +310,21 @@ function getSelectedModel({
...
-			const info = vscodeLlmModels[modelFamily as keyof typeof vscodeLlmModels]
-			return { id, info: { ...openAiModelInfoSaneDefaults, ...info, supportsImages: false } } // VSCode LM API currently doesn't support images.
+			// On a family miss, fall back to the default model entry instead of openAiModelInfoSaneDefaults,
+			// whose 128K contextWindow would diverge from the gate and make the bar read >100% while
+			// auto-condense never fires (the gate uses the live window).
+			const listedModel =
+				vscodeLlmModels[modelFamily as keyof typeof vscodeLlmModels] ?? vscodeLlmModels[vscodeLlmDefaultModelId]
+			// contextWindow MUST equal maxInputTokens: that is the exact value the gate consumes via
+			// getModel().info.contextWindow = Math.max(0, client.maxInputTokens) in src/api/providers/vscode-lm.ts,
+			// so the UI bar and the condense gate share a single source of truth.
```
edelauna: This comment says the gate consumes `getModel().info.contextWindow`, but `Task.ts` now calls `getCondenseContextWindow?.()` as the primary path (which returns the static table's `maxInputTokens`). `getModel()` is the fallback, not the primary. Worth updating the comment so it doesn't mislead future readers?

webview-ui/src/components/ui/hooks/__tests__/useSelectedModel.spec.ts
```diff
@@ -772,4 +774,57 @@ describe("useSelectedModel", () => {
...
+				vsCodeLmModelSelector: { vendor: "copilot", family },
+			}
+
+			const wrapper = createWrapper()
+			const { result } = renderHook(() => useSelectedModel(apiConfiguration), { wrapper })
+
+			expect(result.current.provider).toBe("vscode-lm")
+			expect(result.current.id).toBe(`copilot/${family}`)
+			// The bar and the condense gate share one source of truth: contextWindow === maxInputTokens.
+			expect(result.current.info?.contextWindow).toBe(vscodeLlmModels[family].maxInputTokens)
```
edelauna, line 810: This test uses `vscodeLlmDefaultModelId` (`claude-sonnet-4.5`) where `contextWindow === maxInputTokens === 167790`. If someone accidentally swapped `listedModel.maxInputTokens` for `listedModel.contextWindow` on line 324 of `useSelectedModel.ts`, this test would still pass. Adding one test with `claude-opus-4.8` (the only row where `contextWindow: 679560` ≠ `maxInputTokens: 197897`) would catch that mutation.

src/core/context-management/index.ts
```diff
@@ -192,7 +194,14 @@ export function willManageContext({
...
 
-	const contextPercent = (100 * prevContextTokens) / contextWindow
+	// Measure usage against the available input space (context window minus the
+	// reserved output budget), matching the context gauge shown in the UI. Reserved
+	// output tokens can never hold conversation context, so this is the meaningful
+	// "how full is my usable input" figure. When the reserve is unknown/unlimited
+	// (e.g., vscode-lm reports -1), fall back to the full context window.
+	const reservedForOutput = maxTokens && maxTokens > 0 ? maxTokens : 0
+	const availableInputTokens = contextWindow - reservedForOutput
+	const contextPercent = availableInputTokens > 0 ? (100 * prevContextTokens) / availableInputTokens : 100
```
edelauna: This changes the `contextPercent` denominator from `contextWindow` to `contextWindow - maxTokens` for *every* provider where `maxTokens > 0`, not just vscode-lm. For example, Anthropic with a 200K window and `maxTokens=8192` will see condense fire ~4% earlier than before. Was that intentional? If so, might be worth a note in the PR description since it's a behavioral change for all providers.

edelauna (changes requested):
Thanks for addressing this, had some comments around the implementation.

webview-ui/src/components/ui/hooks/useSelectedModel.ts
simurg79: Fixed in d128a5c1d — corrected the comment to state the condense gate's primary window comes from getCondenseContextWindow() (the static-table maxInputTokens), with getModel().info.contextWindow as the fallback.

webview-ui/src/components/ui/hooks/__tests__/useSelectedModel.spec.ts
simurg79: Done in d128a5c1d — added a claude-opus-4.8 case (contextWindow 679560 != maxInputTokens 197897) asserting contextWindow === maxInputTokens and not the advertised window, so a field swap is now caught.

src/core/context-management/index.ts
simurg79: Addressed in d128a5c1d — scoped the available-input-space denominator to vscode-lm only via an opt-in flag (useAvailableInputForContextPercent), derived in Task.ts from the getCondenseContextWindow seam (only vscode-lm implements it). All other providers now divide by the full context window again, so there's no behavioral change for Anthropic/etc. The maxTokens:-1 reserve guard remains global. Added tests proving both the default full-window path and the vscode-lm opt-in path.

src/api/providers/vscode-lm.ts
```diff
@@ -562,6 +562,23 @@ export class VsCodeLmHandler extends BaseProvider implements SingleCompletionHan
...
 
+	/**
+	 * Context window for auto-condense. The API's advertised `client.maxInputTokens` is far larger
+	 * than usable, so relying on it stops auto-condense from firing; measure against the curated
+	 * static table's `maxInputTokens` instead (the same value the bar uses). Fall back to the live
+	 * window when the model isn't in the table.
+	 */
+	getCondenseContextWindow(): number {
+		const family = this.client?.family ?? this.options.vsCodeLmModelSelector?.family
+		const staticModel = family ? vscodeLlmModels[family as keyof typeof vscodeLlmModels] : undefined
```
edelauna: Unlike `useSelectedModel.ts:316` which falls back to `vscodeLlmModels[vscodeLlmDefaultModelId]`, this lookup has no fallback for removed/unknown families. A user whose stored selector is `"gpt-4o"` or `"claude-3.5-sonnet"` (both dropped from the catalog) gets `staticModel = undefined` and falls back to `this.getModel().info.contextWindow` — the inflated live value this PR was written to avoid.

```suggestion
		const staticModel = family
			? (vscodeLlmModels[family as keyof typeof vscodeLlmModels] ?? vscodeLlmModels[vscodeLlmDefaultModelId])
			: vscodeLlmModels[vscodeLlmDefaultModelId]
```

src/core/context-management/__tests__/context-management.spec.ts
```diff
@@ -1507,19 +1507,36 @@ describe("Context Management", () => {
...
+				totalTokens: 85000,
+				contextWindow: 100000,
+				maxTokens: -1,
+				autoCondenseContext: false,
+				autoCondenseContextPercent: 50,
+				profileThresholds: {},
+				currentProfileId: "default",
+				lastMessageTokens: 0,
+			})
+			expect(result).toBe(true)
```
edelauna, line 1537: All three `maxTokens: -1` tests in this file call `willManageContext`, not `manageContext`. The identical guard at `index.ts:294` inside `manageContext` has no coverage with a negative reserve — a mutation there would not be caught.

Worth adding one `manageContext` call with `maxTokens: -1` where `totalTokens` sits just above the correct `allowedTokens` (e.g. `81809` with `contextWindow: 100000`) so the two copies of the guard live or die together?

src/core/context-management/__tests__/context-management.spec.ts
```diff
@@ -1701,4 +1719,292 @@ describe("Context Management", () => {
...
+			})
+			expect(result).toBe(false)
+		})
+
+		it("willManageContext treats an unlimited (-1) reserve as zero reserve for the percentage", () => {
+			// A -1 reserve falls back to the full window (zero reserve) for the percentage.
+			const result = willManageContext({
+				totalTokens: 150000,
+				contextWindow: 200000,
+				maxTokens: -1,
```
edelauna: With `maxTokens: -1, contextWindow: 200000, totalTokens: 150000, threshold: 70`: the fix gives `reservedForOutput = 0` so `contextPercent = 75.0%`; without the `> 0` guard, `reservedForOutput = -1` gives `contextPercent = 74.9996%`. Both are >= 70 so the test passes either way — the guard is not discriminated here.

Using a strongly negative value (e.g. `maxTokens: -200000`) would make the paths clearly diverge: fix gives 75% and fires, bug gives 37.5% and does not.

src/core/context-management/index.ts
```diff
@@ -192,7 +200,16 @@ export function willManageContext({
...
 	}
 
-	const contextPercent = (100 * prevContextTokens) / contextWindow
+	// Default: divide by the full context window. Opt-in (vscode-lm) divides by available input
+	// (window minus reserved output); an unknown/unlimited reserve (-1) falls back to the full window.
+	let contextPercent: number
+	if (useAvailableInputForContextPercent) {
+		const reservedForOutput = maxTokens && maxTokens > 0 ? maxTokens : 0
+		const availableInputTokens = contextWindow - reservedForOutput
+		contextPercent = availableInputTokens > 0 ? (100 * prevContextTokens) / availableInputTokens : 100
```
edelauna: This `contextPercent` block is copy-pasted verbatim at lines 333-339 inside `manageContext`. If the formula changes in one place the other silently diverges — similar to the kind of drift that caused the original bug. Worth extracting into a small shared helper?

edelauna (changes requested):
Thanks for adding this, just had some minor nits.

src/core/context-management/__tests__/context-management.spec.ts
simurg79: Done in 6e3c466ac — added a manageContext truncation-path test with maxTokens:-1, contextWindow:100000, totalTokens:81809 (allowedTokens=81808) so the duplicated reserve guard is now covered alongside the willManageContext one.

src/core/context-management/__tests__/context-management.spec.ts
simurg79: Done in 6e3c466ac — changed maxTokens to -200000 so the maxTokens>0 guard is actually discriminated (fix: 75% fires; without the guard: 37.5% would not), instead of -1 where both paths stayed ≥70%.

src/api/providers/vscode-lm.ts
simurg79: Done in 7447ab3af. `getCondenseContextWindow()` now resolves the default row (`vscodeLlmModels[vscodeLlmDefaultModelId]`) for an unknown/absent family, mirroring useSelectedModel.ts, so a stale selector like `gpt-4o` no longer falls through to the inflated live window. The `> 0` guard still routes a non-positive static `maxInputTokens` to the live window. Added/updated tests for the unknown-family and no-family paths.

src/core/context-management/index.ts
simurg79: Done in 7447ab3af. Extracted the shared `computeContextPercent({ prevContextTokens, contextWindow, maxTokens, useAvailableInputForContextPercent })` helper, now called by both `willManageContext` and `manageContext`, so the two can no longer drift. The `console.warn` invalid-threshold divergence lives in the threshold block (not the percent block), so it's preserved. All 53 context-management tests pass.

edelauna (approved):
Thanks for addressing the feedback and your contribution!

PR #726, opened 2026-06-26
fix(#689): provider cache reset after settings import

### LINKED ISSUE #689: [BUG] API Provider can revert to previous Provider after importing settings

In certain situations, after importing settings, changing the API Provider in Settings > Providers can briefly apply the new Provider and then revert back to the previously saved Provider.

This affects users who change API Providers from the Settings screen after importing settings. Some Provider transitions make the issue easier to hit because they trigger additional Provider/model configuration updates and rerenders.

### Reproduction steps

1. Import settings.
2. Open the Settings screen.
3. Go to the Providers tab.
4. Change the Provider to Baseten and enter any value in the API key field.
5. Click Save.
6. Try changing the Provider to DeepSeek.

### Expected result

The Provider should change to DeepSeek Provider

### Actual result

The Provider briefly changes to DeepSeek, then immediately changes back to Baseten

This does not happen for every Provider transition. Some Provider changes save correctly, while others revert.

After settings are imported, a stale imported extension state can be replayed into SettingsView and overwrite `cachedState.apiConfiguration`. Provider transitions that trigger additional automatic model/default/validation updates appear more likely to expose the timing issue, but the import state replay is the key condition.

In other words:

- Working transitions: Provider change causes little or no automatic follow-up state sync.
- Failing transitions: Provider change triggers model/default/validation-related sync while the imported state can still overwrite `cachedState`.

This is why the issue appears only for some Provider combinations instead of blocking every Provider change.

### App Version

v3.62.0

### API Provider (optional)

None

### Model Used (optional)

_No response_

_No response_

```shell

```

JunyongParkDev:
### Related GitHub Issue

Closes: #689

### Description

SettingsView uses a local cachedState buffer so settings edits do not write through to live extension state before Save. After importing settings, the cache-busting effect could run multiple times for the same settingsImportedAt value whenever extensionState changed identity.

That could overwrite an in-progress provider edit with the previously saved provider, making the UI briefly switch and then revert.

This change tracks the handled settingsImportedAt value and resets cachedState only once per import event. It also adds a regression test covering the imported-settings flow where Baseten is saved with an API key, the same import timestamp replays, and a later DeepSeek provider edit is preserved and saved.

### Test Procedure
- Ran the focused webview regression test: - all 3 tests passed.

  ```shell
  pnpm --dir webview-ui exec vitest run src/components/settings/__tests__/SettingsView.change-detection.spec.tsx

- Verified the new regression coverage for the imported-settings flow:
    1. Start with settingsImportedAt set.
    2. Change Provider to Baseten.
    3. Enter a Baseten API key.
    4. Save the settings.
    5. Change Provider to DeepSeek.
    6. Replay the same imported extension state timestamp.
    7. Confirm the Provider remains DeepSeek and Save posts apiProvider: "deepseek".

- This verifies that SettingsView only busts cachedState once per settingsImportedAt value, so a replayed imported state does not overwrite an in-progress Provider edit.

### Pre-Submission Checklist

<!-- Go through this checklist before marking your PR as ready for review. -->

- [X] **Contribution Guidelines**: I have read and agree to the [[Contributor Guidelines](https://github.com/CONTRIBUTING.md)](/CONTRIBUTING.md).

### Screenshots / Videos

<!--
For UI changes, please provide before-and-after screenshots or a short video of the *actual results*.
This greatly helps in understanding the visual impact of your changes.
-->

### Documentation Updates

<!--
Does this PR necessitate updates to user-facing documentation?
-->

### Additional Notes

<!-- Add any other context, questions, or information for reviewers here. -->

### Get in Touch

<!--
Please provide your Discord username for reviewers or maintainers to reach you if they have questions about your PR
-->


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved settings import handling so the “bust cache” logic runs only once per distinct import timestamp, avoiding repeated cache refreshes and unintended change-detection resets.
  * Fixed settings import replay behavior to correctly merge cached state while preserving the latest API provider selection during saves.

* **Tests**
  * Expanded regression tests for settings change detection, including replaying `settingsImportedAt` and verifying correct provider switching (e.g., Baseten to DeepSeek) and UI reset/sync behavior.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/webview-ui/src/components/settings/__tests__/SettingsView.change-detection.spec.tsx
+++ b/webview-ui/src/components/settings/__tests__/SettingsView.change-detection.spec.tsx
@@ -1,4 +1,4 @@
-import { render, screen, fireEvent, waitFor } from "@testing-library/react"
+import { act, render, screen, fireEvent, waitFor } from "@testing-library/react"
 import { vi, describe, it, expect, beforeEach } from "vitest"
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query"
 import React from "react"
@@ -449,6 +449,12 @@ describe("SettingsView - Change Detection Fix", () => {
 			</QueryClientProvider>,
 		)
 
+		// Let the import cache-busting effect run. With the old implementation,
+		// this would reset cachedState back to the replayed Baseten config.
+		await act(async () => {
+			await new Promise((resolve) => setTimeout(resolve, 0))
+		})
+
 		expect(screen.getByTestId("provider-value")).toHaveTextContent("deepseek")
 
 		mockPostMessage.mockClear()
@@ -460,6 +466,55 @@ describe("SettingsView - Change Detection Fix", () => {
 			apiConfiguration: expect.objectContaining({
 				apiProvider: "deepseek",
 			}),
+		})
+	})
+
+	it("resets cached provider state when a new import timestamp arrives", async () => {
+		const onDone = vi.fn()
+		let extensionState = createExtensionState({
+			settingsImportedAt: 100,
+			apiConfiguration: {
+				apiProvider: "openai",
+				apiModelId: "gpt-4.1",
+			},
+		})
+
+		;(useExtensionState as any).mockImplementation(() => extensionState)
+
+		const { rerender } = render(
+			<QueryClientProvider client={queryClient}>
+				<SettingsView onDone={onDone} />
+			</QueryClientProvider>,
+		)
+
+		await waitFor(() => {
+			expect(screen.getByTestId("provider-value")).toHaveTextContent("openai")
+		})
+
+		fireEvent.click(screen.getByTestId("set-provider-deepseek"))
+		expect(screen.getByTestId("provider-value")).toHaveTextContent("deepseek")
+
+		extensionState = createExtensionState({
+			settingsImportedAt: 101,
+			apiConfiguration: {
+				apiProvider: "baseten",
+				apiModelId: "zai-org/GLM-4.6",
+				basetenApiKey: "imported-baseten-key",
+			},
+		})
+
+		rerender(
+			<QueryClientProvider client={queryClient}>
+				<SettingsView onDone={onDone} />
+			</QueryClientProvider>,
+		)
+
+		await waitFor(() => {
+			expect(screen.getByTestId("provider-value")).toHaveTextContent("baseten")
+		})
+
+		await waitFor(() => {
+			expect(screen.getByTestId("save-button")).toBeDisabled()
 		})
 	})
 })
--- a/webview-ui/src/components/settings/__tests__/SettingsView.change-detection.spec.tsx
+++ b/webview-ui/src/components/settings/__tests__/SettingsView.change-detection.spec.tsx
@@ -4,11 +4,17 @@ import { QueryClient, QueryClientProvider } from "@tanstack/react-query"
 import React from "react"
 
 // Mock vscode API
-const mockPostMessage = vi.fn()
+const mockPostMessage = vi.hoisted(() => vi.fn())
 const mockVscode = {
 	postMessage: mockPostMessage,
 }
 ;(global as any).acquireVsCodeApi = () => mockVscode
+
+vi.mock("@src/utils/vscode", () => ({
+	vscode: {
+		postMessage: mockPostMessage,
+	},
+}))
 
 // Import the actual component
 import SettingsView from "../SettingsView"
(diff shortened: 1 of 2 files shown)


webview-ui/src/components/settings/__tests__/SettingsView.change-detection.spec.tsx
```diff
@@ -369,4 +392,74 @@ describe("SettingsView - Change Detection Fix", () => {
...
+			},
+		})
+
+		rerender(
+			<QueryClientProvider client={queryClient}>
+				<SettingsView onDone={onDone} />
+			</QueryClientProvider>,
+		)
+
+		expect(screen.getByTestId("provider-value")).toHaveTextContent("deepseek")
```
edelauna, line 458: This assertion runs synchronously right after `rerender(...)`, but the cache-busting `useEffect` (deps `[settingsImportedAt, extensionState]`) flushes *after* the commit. I tried reverting the fix back to the old `if (settingsImportedAt) {...}` body and this test still passed — the assertion reads `cachedState` before the effect has a chance to clobber it. Would wrapping this in `await waitFor(() => expect(...).toHaveTextContent("deepseek"))` make it actually catch a regression here?

```suggestion
		await waitFor(() => {
			expect(screen.getByTestId("provider-value")).toHaveTextContent("deepseek")
		})
```

webview-ui/src/components/settings/__tests__/SettingsView.change-detection.spec.tsx
```diff
@@ -369,4 +392,74 @@ describe("SettingsView - Change Detection Fix", () => {
...
+		fireEvent.click(screen.getByTestId("save-button"))
+
+		expect(mockPostMessage).toHaveBeenCalledWith({
+			type: "upsertApiConfiguration",
+			text: "default",
+			apiConfiguration: expect.objectContaining({
+				apiProvider: "deepseek",
+			}),
+		})
+	})
```
edelauna, line 470: As an extra guard: would it be worth adding a case that re-renders with a *new* `settingsImportedAt` value (e.g. `100` -> `101`) and a changed `extensionState.apiConfiguration`, then asserts `provider-value` updates? Right now nothing proves the effect fires at all on a fresh import — `cachedState` is pre-seeded from `useState(() => extensionState)`, so the effect could be a complete no-op and this test would still pass. That same case would also cover the `setChangeDetected(false)` path after pending edits.

edelauna (changes requested):
Thanks for you contribution just had some additional testing to add.

webview-ui/src/components/settings/__tests__/SettingsView.change-detection.spec.tsx
JunyongParkDev: Thanks for review. You were right that the synchronous assertion could read the pre-effect cached state and miss the regression.

I updated the test to explicitly let the post-rerender effect flush before asserting that the provider remains `deepseek`. I used `act` with a tick instead of only wrapping the assertion in `waitFor`, because `waitFor` can pass immediately if it observes the pre-effect state before the clobber happens.

I also verified this catches the regression: with the old `if (settingsImportedAt) { ... }` implementation restored locally, this test now fails with `Received: baseten`.

webview-ui/src/components/settings/__tests__/SettingsView.change-detection.spec.tsx
JunyongParkDev: Added this extra guard as well.

The new test starts with `settingsImportedAt: 100`, makes a pending local provider edit to `deepseek`, then rerenders with `settingsImportedAt: 101` and an imported `baseten` provider config. It waits for the UI to update to `baseten`, which proves the fresh-import effect actually refreshes `cachedState` from the new `extensionState` rather than relying on the initial `useState(() => extensionState)` value.

I also added an assertion that the Save button becomes disabled after the fresh import, so the test covers the `setChangeDetected(false)` path after pending edits are replaced by imported settings.

JunyongParkDev:
@edelauna Thank you for review.

I checked both points and pushed a follow-up commit with the requested updates. The test now waits for the post-rerender effect before asserting the replayed timestamp case, and I added an additional case for a fresh `settingsImportedAt` value to verify that imported provider state is applied and the dirty Save state is cleared.

edelauna (approved):
Look good - thanks for your contribution

edelauna (approved):

PR #727, opened 2026-06-26
fix: prevent agent loop stall on write_to_file filesystem errors (#703)

### LINKED ISSUE #703: [BUG] Agent loop stalls permanently when write_to_file partial streaming hits a filesystem error (EROFS/EACCES)
## Problem

When the model uses `write_to_file` to create a file whose parent directory does not exist (or is on a read-only mount), the extension freezes mid-stream and never returns control to the agent loop. The user sees an error flash in the UI but the assistant never continues.

Error visible in logs:
```
Error handling partial write_to_file:
EROFS: read-only file system, mkdir '/scratch'
```

## Context

Any user who asks the agent to write a file to a path whose root does not exist or is read-only (e.g. `/scratch/foo.ts` on macOS where `/scratch` does not exist) will hit this. The agent appears completely frozen and requires a manual abort or window reload.

## Reproduction steps

1. Any OS/version where `/scratch` (or similar) does not exist
2. Ask the agent to write a file to `/scratch/test.ts` or any path whose parent directory cannot be created
3. The model issues a `write_to_file` tool call
4. Streaming begins, the UI shows the error `EROFS: read-only file system, mkdir '/scratch'`
5. The agent never continues -- no further tool calls, no error recovery, no completion

## Expected result

The agent should surface the error as a tool result and continue the conversation (retry, use a different path, or report the failure to the user).

## Actual result

The agent loop stalls permanently. `userMessageContentReady` is never set to `true`, so the task hangs waiting for a condition that can never be satisfied.

## Root cause (technical)

`WriteToFileTool.handlePartial()` calls `createDirectoriesForFile(absolutePath)` (which calls `fs.mkdir`) **without a `.catch()` guard**. During streaming (`block.partial === true`), this throws `EROFS`.

The throw is caught by `BaseTool.handle()`, which calls `handleError` (correctly pushing a `tool_result` to API history) and returns. However, back in `presentAssistantMessage()`, the advancement gate:

```typescript
if (!block.partial || cline.didRejectTool || cline.didAlreadyUseTool) {
    cline.userMessageContentReady = true   // never reached
    cline.currentStreamingContentIndex++   // never reached
}
```

is skipped because `block.partial` is still `true` and neither flag was set by `handleError`. The loop stalls.

**Fix**: Remove the `createDirectoriesForFile` block from `handlePartial` -- it is redundant since `execute()` already calls it before `diffViewProvider.open()`. All other tools` `handlePartial` overrides guard their async I/O with `.catch(() => {})`; this one does not.

## App Version

Reproducible on current `main` (confirmed by code inspection).

## API Provider

Not Applicable / Other (filesystem error, provider-independent)

awschmeder:
### Related GitHub Issue

Closes: #703

### Description

**The bug (#703):** When the model uses `write_to_file` to create a file whose parent directory cannot be created or written (e.g. a read-only mount or a path under a directory lacking write permission, producing `EROFS`/`EACCES`), the agent loop stalled permanently.

`handlePartial` called `createDirectoriesForFile` without a `.catch()` guard. During streaming (`block.partial === true`) the unguarded throw was caught by `BaseTool.handle` -> `handleError`, but the partial-block advancement gate in `presentAssistantMessage` was skipped (`block.partial` was still `true` and neither `didRejectTool` nor `didAlreadyUseTool` was set), so `userMessageContentReady` was never set and the task hung forever.

**The fix:**

- Removed the redundant `createDirectoriesForFile` call from `handlePartial` (`execute()` already creates parent directories before `diffViewProvider.open()`).
- Moved `createDirectoriesForFile` in `execute()` inside the `try` block so `EROFS`/`EACCES` errors route through `handleError` with `diffViewProvider.reset()` cleanup and consecutive-mistake counting instead of escaping unhandled.

**Correct streaming UI behavior under a filesystem error.** Because the failure now occurs while a partial `write_to_file` block is still streaming, the implementation is designed so the in-progress UI resolves cleanly rather than leaving artifacts. The following are properties of the implementation:

- The streaming "Zoo wants to edit this file" element does not get stuck in a loading state. `Task.finalizePartialToolAsk()` (a non-blocking helper that sets the last partial `tool` ask to `partial: false`) dismisses it. It deliberately avoids `task.ask(..., false)`, which would fall through to a blocking `pWaitFor` that would result in a non-terminating spinner left in the chat UI.
- The error is reported exactly once. `handlePartial` finalizes the partial ask, resets the diff view, and swallows the streaming error (logged to the console) so only the authoritative non-partial `execute()` path surfaces it to the user. This is safe because the loop advancement gate does not depend on the streaming throw -- it advances when the non-partial block arrives.
- The tool element is not duplicated across streaming deltas. A `partialStreamFailed` flag (cleared between invocations via `resetPartialState()`) short-circuits later deltas so they do not re-attempt `open()`/`update()` and spawn a new partial tool message each time.

### Test Procedure

Automated (run from `src/`):

```
cd src && npx vitest run core/tools/__tests__/writeToFileTool.spec.ts
```

31 tests pass, including new regression tests that assert:
- `EROFS` in `handlePartial` does not call `createDirectoriesForFile` and does not stall.
- `EROFS`/`EACCES` in `execute()` routes through `handleError` with `diffViewProvider.reset()` and does not proceed to open/save.
- When `handlePartial` `open()`/`update()` throws, `finalizePartialToolAsk()` and `diffViewProvider.reset()` are called and the streaming error is swallowed (not surfaced via `handleError`).
- A single filesystem failure is reported exactly once across the streaming and `execute()` phases.
- Repeated streaming deltas after a failure do not spawn additional partial tool messages or re-attempt `open()`.

`npx tsc --noEmit` passes cleanly.

Manual (Extension Development Host / F5):
1. Ask the agent to use the write_file tool to write a file to a destination that is un-writable (no access, or read-only filesystem).
2. Before the fix: the agent stalls
3. After the fix: the error is surfaced as a tool result with an error message, and the agent continues the conversation.

### Pre-Submission Checklist

- [x] **Documentation Impact**: I have considered if my changes require documentation updates (none required).

### Documentation Updates

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file-writing behavior when errors occur during streaming or save operations.
  * Prevented repeated error prompts and duplicate retry attempts after a partial update fails.
  * Fixed a stuck loading/spinner state when an operation ends unexpectedly.
  * Ensured directory creation and file updates happen at the right stage so error handling is more consistent.

### DIFF
--- a/src/core/task/Task.ts
+++ b/src/core/task/Task.ts
@@ -1730,6 +1730,21 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
 		return formatResponse.toolError(formatResponse.missingToolParameterError(paramName))
 	}
 
+	/**
+		* Finalize the last partial "tool" ask message without blocking for user input.
+		* Call this in error paths where a partial tool message was opened during streaming
+		* but execution failed before the normal approval flow could close it, so the webview
+		* spinner does not get stuck in a loading state.
+		*/
+	async finalizePartialToolAsk(): Promise<void> {
+		const lastMessage = this.clineMessages.at(-1)
+
+		if (lastMessage && lastMessage.partial && lastMessage.type === "ask" && lastMessage.ask === "tool") {
+			lastMessage.partial = false
+			await this.updateClineMessage(lastMessage)
+		}
+	}
+
 	// Lifecycle
 	// Start / Resume / Abort / Dispose
 
--- a/src/core/tools/WriteToFileTool.ts
+++ b/src/core/tools/WriteToFileTool.ts
@@ -67,12 +67,6 @@ export class WriteToFileTool extends BaseTool<"write_to_file"> {
 			task.diffViewProvider.editType = fileExists ? "modify" : "create"
 		}
 
-		// Create parent directories early for new files to prevent ENOENT errors
-		// in subsequent operations (e.g., diffViewProvider.open, fs.readFile)
-		if (!fileExists) {
-			await createDirectoriesForFile(absolutePath)
-		}
-
 		if (newContent.startsWith("```")) {
 			newContent = newContent.split("\n").slice(1).join("\n")
 		}
@@ -98,6 +92,13 @@ export class WriteToFileTool extends BaseTool<"write_to_file"> {
 
 		try {
 			task.consecutiveMistakeCount = 0
+
+			// Create parent directories for new files inside the try block so filesystem
+			// errors (EROFS, EACCES, etc.) route through handleError with proper cleanup
+			// and consecutive-mistake counting, rather than escaping unhandled.
+			if (!fileExists) {
+				await createDirectoriesForFile(absolutePath)
+			}
 
 			const provider = task.providerRef.deref()
 			const state = await provider?.getState()
@@ -222,12 +223,6 @@ export class WriteToFileTool extends BaseTool<"write_to_file"> {
 		} else {
 			fileExists = await fileExistsAtPath(absolutePath)
 			task.diffViewProvider.editType = fileExists ? "modify" : "create"
-		}
-
-		// Create parent directories early for new files to prevent ENOENT errors
-		// in subsequent operations (e.g., diffViewProvider.open)
-		if (!fileExists) {
-			await createDirectoriesForFile(absolutePath)
 		}
 
 		const isWriteProtected = task.rooProtectedController?.isWriteProtected(relPath!) || false
--- a/src/core/tools/WriteToFileTool.ts
+++ b/src/core/tools/WriteToFileTool.ts
@@ -26,6 +26,19 @@ interface WriteToFileParams {
 export class WriteToFileTool extends BaseTool<"write_to_file"> {
 	readonly name = "write_to_file" as const
 
+	/**
+	 * Set when a filesystem error aborts diff-view streaming during handlePartial for the
+	 * current tool invocation. Subsequent streaming deltas for the same block then skip the
+	 * doomed open()/update() retry, which would otherwise create a fresh "Zoo wants to edit
+	 * this file" message on every delta. Cleared by resetPartialState() between invocations.
+	 */
+	private partialStreamFailed = false
+
+	override resetPartialState(): void {
+		super.resetPartialState()
+		this.partialStreamFailed = false
+	}
+
 	async execute(params: WriteToFileParams, task: Task, callbacks: ToolCallbacks): Promise<void> {
 		const { pushToolResult, handleError, askApproval } = callbacks
 		const relPath = params.path
@@ -187,6 +200,11 @@ export class WriteToFileTool extends BaseTool<"write_to_file"> {
 
 			return
 		} catch (error) {
+			// Finalize any open partial tool message so the UI spinner doesn't get stuck.
+			// The partial ask fired during streaming (handlePartial) or early in execute sets
+			// partial: true on the webview message; without this, the spinner persists even
+			// after the error bubble appears.
+			await task.finalizePartialToolAsk()
 			await handleError("writing file", error as Error)
 			await task.diffViewProvider.reset()
 			this.resetPartialState()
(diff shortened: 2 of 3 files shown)


awschmeder:
https://github.com/user-attachments/assets/48f7e728-7d3b-47b1-8373-027aa9f0113b

src/core/tools/WriteToFileTool.ts
```diff
@@ -99,6 +106,13 @@ export class WriteToFileTool extends BaseTool<"write_to_file"> {
 		try {
 			task.consecutiveMistakeCount = 0
 
+			// Create parent directories for new files inside the try block so filesystem
+			// errors (EROFS, EACCES, etc.) route through handleError with proper cleanup
+			// and consecutive-mistake counting, rather than escaping unhandled.
+			if (!fileExists) {
+				await createDirectoriesForFile(absolutePath)
```
edelauna, line 113: The mistake counter is zeroed on line 107 before the call that can throw on line 113. When `createDirectoriesForFile` rejects with EROFS, `handleError` only calls `cline.say("error", ...)` and pushes a tool result -- it does not re-increment `consecutiveMistakeCount`. Net effect: a permanently read-only path zeros the runaway-loop guard on every attempt with no increment. Should `consecutiveMistakeCount = 0` move to after this call succeeds (e.g. just before `diffViewProvider.open` at line 152)?

src/core/tools/WriteToFileTool.ts
```diff
@@ -26,6 +26,19 @@ interface WriteToFileParams {
 export class WriteToFileTool extends BaseTool<"write_to_file"> {
 	readonly name = "write_to_file" as const
 
+	/**
+	 * Set when a filesystem error aborts diff-view streaming during handlePartial for the
+	 * current tool invocation. Subsequent streaming deltas for the same block then skip the
+	 * doomed open()/update() retry, which would otherwise create a fresh "Zoo wants to edit
+	 * this file" message on every delta. Cleared by resetPartialState() between invocations.
+	 */
+	private partialStreamFailed = false
```
edelauna, line 35: `writeToFileTool` is a module-level singleton. `partialStreamFailed` is instance state on that one object. The `presentAssistantMessageLocked` guard only prevents concurrent execution within a single Task -- two simultaneous sessions (or a parent task + subtask) can call `writeToFileTool.handle()` concurrently. If Task A sets this flag due to an EROFS error, Task B's next streaming delta silently short-circuits even though its write is fine. The pre-existing `lastSeenPartialPath` in `BaseTool` has the same architectural risk, but this flag has more severe consequences (actively suppresses UI vs. a momentarily stale path). Worth documenting in the JSDoc at minimum, or using a per-taskId map?

src/core/task/Task.ts
```diff
@@ -1730,6 +1730,21 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
...
 	}
 
+	/**
+		* Finalize the last partial "tool" ask message without blocking for user input.
+		* Call this in error paths where a partial tool message was opened during streaming
+		* but execution failed before the normal approval flow could close it, so the webview
+		* spinner does not get stuck in a loading state.
+		*/
+	async finalizePartialToolAsk(): Promise<void> {
+		const lastMessage = this.clineMessages.at(-1)
```
edelauna, line 1740: The target is identified by position (`at(-1)`) with no correlation to the specific partial ask opened during streaming. Is there any async path between the `task.ask("tool", ..., block.partial)` call in `handlePartial` and the catch block that could push a new message (e.g. `processQueuedMessages`)? If so, `at(-1)` could silently no-op and leave the spinner stuck.

src/core/tools/__tests__/writeToFileTool.spec.ts
```diff
@@ -178,6 +178,7 @@ describe("writeToFileTool", () => {
 		}
 		mockCline.say = vi.fn().mockResolvedValue(undefined)
 		mockCline.ask = vi.fn().mockResolvedValue(undefined)
+		mockCline.finalizePartialToolAsk = vi.fn().mockResolvedValue(undefined)
```
edelauna, line 181: The mock replaces the full implementation, so these tests verify the method is called but never that `lastMessage.partial` actually becomes `false`. A regression removing that mutation from the real implementation would pass. Would a thin direct test on `Task.finalizePartialToolAsk` with a real `clineMessages` array be worth adding?

edelauna (changes requested):
Thanks for adding this. Since we're touching I/O functionality can we harden the code a bit? I left some comments.

taltas:
@awschmeder Do you need more time to address the comments?

easonLiangWorldedtech:
@edelauna @coderabbitai I've addressed all review comments in a new PR: #1066 

Changes include:
- `finalizePartialToolAsk()` now searches by type + text pattern instead of just `at(-1)`, and persists state properly
- Mistake counter moved after successful directory creation (edelauna #1)
-`partialStreamFailed` changed to per-taskId Map for concurrent safety (edelauna #2)
- Added thin layer test on `Task.finalizePartialToolAsk()` (edelauna #4)

All tests pass (94 passed / 8 skipped). Please review when you have a moment.

PR #749, opened 2026-06-27
WIP: web_fetch_content tool

### LINKED ISSUE #748: fetch_web_content tool
Other harnesses have tools to fetch webpage contents from the web which is useful for all sorts of random purposes without involving a specific MCP.

Implement fetch_web_content tool.

MCP. However I think it's better to have such tool builtin. Claude code and Cline already have builtin fetch tool. 

**Additional context**
N/a.

p12tic:
### Related GitHub Issue

Closes: #748

### Description

Implements web_fetch_content tool.

### Test Procedure

Manual testing.

### Pre-Submission Checklist

### Screenshots / Videos

image

### Additional Notes

Please let me know if this is something that can be added to ZooCode and I will finish polishing it. I suspect at least docs will need updates.

## Summary by CodeRabbit

* **New Features**
  * Added support for fetching web content as a new tool, including URL-based requests and optional prompt context.
  * Web content actions now appear in the chat UI with a clearer label and target URL display.
  * Added a new web tool group and updated available modes to include web access.

* **Bug Fixes**
  * Improved handling of web-fetch tool calls in progress updates and final results.
  * Added localized chat messages and tool labels across supported languages.


p12tic:
@coderabbitai  please review

src/core/auto-approval/tools.ts
```diff
@@ -13,5 +13,6 @@ export function isReadOnlyToolAction(tool: ClineSayTool): boolean {
 		"searchFiles",
 		"codebaseSearch",
 		"runSlashCommand",
+		"fetchWebContent",
```
edelauna, line 16: Does grouping an outbound network fetch with file reads make sense here? With `alwaysAllowReadOnly` on, every `fetch_web_content` call auto-approves with no user prompt. I think network calls warrant their own opt-in flag, or a setting to always require explicit approval?

src/core/tools/FetchWebContentTool.ts
```diff
@@ -0,0 +1,328 @@
...
+				headers: {
+					"User-Agent": "Mozilla/5.0 (compatible; ZooCode/1.0.0)",
+					Accept: "text/html,application/xhtml+xml,application/xml;q=0.9,text/plain;q=0.8,*/*;q=0.7",
+					"Accept-Language": "en-US,en;q=0.9",
+				},
+				redirect: "follow",
+				signal: controller.signal,
+			})
+
+			clearTimeout(timeout)
```
edelauna, line 217: The timeout is cancelled here as soon as headers arrive, so the body-read loop below runs with no deadline. A server that sends headers quickly but streams the body slowly could hold the connection open indefinitely — the 30s limit only covers time-to-first-byte. Should the timer stay active through the full read?

src/core/tools/FetchWebContentTool.ts
```diff
@@ -0,0 +1,328 @@
...
+
+		try {
+			const response = await fetch(url, {
+				method: "GET",
+				headers: {
+					"User-Agent": "Mozilla/5.0 (compatible; ZooCode/1.0.0)",
+					Accept: "text/html,application/xhtml+xml,application/xml;q=0.9,text/plain;q=0.8,*/*;q=0.7",
+					"Accept-Language": "en-US,en;q=0.9",
+				},
+				redirect: "follow",
```
edelauna, line 213: With `redirect: "follow"`, a server can redirect to a private IP or cloud metadata endpoint (`169.254.169.254`, `localhost`, etc.) and the redirect destination is never validated — only the original URL was checked. Would limiting redirects to the same host work here?

```suggestion
				redirect: "manual",
```
(Then check `response.status >= 300 && response.status < 400` and only follow if `new URL(response.headers.get('location')!, url).origin === new URL(url).origin`)

src/core/tools/FetchWebContentTool.ts
```diff
@@ -0,0 +1,328 @@
...
+				content = htmlToText(text)
+			} else if (contentType.includes("application/json")) {
+				try {
+					const json = JSON.parse(text)
+					content = JSON.stringify(json, null, 2)
+				} catch {
+					content = text
+				}
+			} else {
+				content = text
```
edelauna, line 275: If the server responds with `image/jpeg`, `application/pdf`, or another binary type (the `Accept` header allows `*/*`), this falls through and returns the raw decoded bytes to the model as text. Should binary MIME types be rejected before attempting to decode?

src/core/tools/FetchWebContentTool.ts
```diff
@@ -0,0 +1,328 @@
...
+			}
+
+			// Format output with metadata
+			const outputLines = [
+				`URL: ${url}`,
+				`Content-Type: ${contentType}`,
+				`Size: ${totalSize} bytes`,
+				``,
+				`--- Content ---`,
+				content.slice(0, MAX_CONTENT_CHARS),
```
edelauna, line 285: Web content lands directly in the model's context with no trust boundary marker. If a page contains visible text like "Ignore all previous instructions...", it passes through as-is. Would wrapping this in something like `<untrusted_web_content source="...">` (and noting in the tool description that content inside this tag is third-party data, not instructions) help?

src/core/tools/FetchWebContentTool.ts
```diff
@@ -0,0 +1,328 @@
...
+			]
+
+			if (content.length > MAX_CONTENT_CHARS) {
+				outputLines.push(
+					`\n[Content truncated: showing first ${MAX_CONTENT_CHARS} of ${content.length} characters]`,
+				)
+			}
+
+			if (prompt) {
+				outputLines.push(``, `--- Analysis Request ---`, `Prompt: ${prompt}`)
```
edelauna, line 295: The user's analysis prompt arrives after up to 50k chars of untrusted content. A malicious page could also embed `--- Analysis Request ---` in its visible text before the real one appears. Would putting the prompt instruction *before* the content block help with both ordering and injection resistance?

src/core/tools/FetchWebContentTool.ts
```diff
@@ -0,0 +1,328 @@
...
+					"Accept-Language": "en-US,en;q=0.9",
+				},
+				redirect: "follow",
+				signal: controller.signal,
+			})
+
+			clearTimeout(timeout)
+
+			if (!response.ok) {
+				pushToolResult(formatResponse.toolError(`HTTP ${response.status}: ${response.statusText}`))
```
edelauna, line 220: Other tools (e.g. `GenerateImageTool`) call `task.say("error", ...)` before `pushToolResult` so the error surfaces as a visible bubble in chat — without it, only the model sees this. Same gap exists at the size-limit path a bit further down.

src/core/tools/FetchWebContentTool.ts
```diff
@@ -0,0 +1,328 @@
...
+			for (const chunk of chunks) {
+				buffer.set(chunk, offset)
+				offset += chunk.length
+			}
+			const text = new TextDecoder("utf-8").decode(buffer)
+
+			// Process content based on type
+			let content: string
+			if (contentType.includes("text/html") || contentType.includes("application/xhtml")) {
+				content = htmlToText(text)
```
edelauna, line 266: `cheerio.load()` is synchronous. For responses near the 5 MB limit, parsing can block the extension host event loop for hundreds of milliseconds, freezing IntelliSense and file watchers. Would slicing the HTML to a smaller cap before parsing (e.g. 500 KB) be a reasonable trade-off, since `MAX_CONTENT_CHARS` will truncate the output anyway?

edelauna (changes requested):
So in reviewing this I noticed a couple implementation/security details we should consider, but I also just realized - how is this feature any different than asking the agent to just `curl` a website?

I wonder if instead this feature should be more stand alone, and a type of structured web parser. 

Based on this article: https://mikhail.io/2025/10/claude-code-web-tools/ it seems like Claude fetches a page, converts it to markdown then has a different model go through and parse the contents? I wonder if this feature should be similar to the code embeddings feature where a user can specify a provider and model that will perform a structured parsing of a web fetch?

p12tic:
@edelauna Thanks for quick review. I should've market the PR as RFC, right now it probably used more time than necessary for review, as I just wanted to understand whether this PR makes sense in principle. 

> So in reviewing this I noticed a couple implementation/security details we should consider, but I also just realized - how is this feature any different than asking the agent to just `curl` a website?

It also extracts text from the page, so not entirely same as it makes life for the model easier. Important with stupid open-source models. Claude will be able to extract stuff from any format.

Another difference is regarding permissions, it makes sense to allow web fetch to e.g. architect to make better decisions.

> I wonder if instead this feature should be more stand alone, and a type of structured web parser.
> 
> Based on this article: https://mikhail.io/2025/10/claude-code-web-tools/ it seems like Claude fetches a page, converts it to markdown then has a different model go through and parse the contents? I wonder if this feature should be similar to the code embeddings feature where a user can specify a provider and model that will perform a structured parsing of a web fetch?

Agreed, this makes sense. Would you accept simple web_fetch_content tool as a first step and then add LLM-based parsing on top later. I plan to implement a web search tool as well, so the web access capabilities will improve gradually anyway.

navedmerchant:
Ah maybe not, it uses playwright which is not easy to make into pyinstaller

navedmerchant:
I propose we use this library, I did an audit of the code using LLMs and its pretty well engineered. https://github.com/teng-lin/agent-fetch. It is also in JS, and outputs markdown directly. @p12tic @edelauna . Can you guys also take a look?

edelauna:
> I propose we use this library, I did an audit of the code using LLMs and its pretty well engineered. https://github.com/teng-lin/agent-fetch. It is also in JS, and outputs markdown directly. @p12tic @edelauna . Can you guys also take a look?

My preference would be something lighter to start, could we look at https://github.com/mozilla/readability + https://github.com/mixmark-io/turndown

PR #755, opened 2026-06-28
feat: add Umans provider

taltas:
## Summary
- add the Umans provider to provider/type registries and shared settings types
- integrate the Umans runtime handler, provider wiring, and model fetcher/cache support
- wire the settings UI, model selection behavior, validation, and focused provider tests

## Testing
- focused Umans provider tests were run successfully before handoff
- pre-commit lint and push-time type checks passed during commit/push flow

## Summary by CodeRabbit

* **New Features**
  * Added support for a new provider option with its own settings, default model, and model picker integration.
  * Added support for a custom Anthropic provider configuration, including custom model details and provider-specific settings.

* **Bug Fixes**
  * Improved model selection so the correct provider/model is chosen more reliably across settings and router model lists.
  * Updated validation to require the right API keys and model fields for the new provider options.
  * Refresh and model loading now include the new provider in aggregated results.


edelauna (changes requested):
Could we break this into 2 PRs?
1. The Anthropic Custom stuff (I don't think this is tied to Umans, I've used a proxy locally for this, so I think this can ship as a standalone feature).
2. Umans provider

PR #765, opened 2026-06-29
feat(minimax): add MiniMax M3 models (512K / 1M context) and fix double-think on Anthropic endpoint

hameedirfan:
## Summary

This PR adds the **MiniMax M3** model family to the MiniMax provider and ports
a small but important fix from the kilo-code opencode engine to keep M-series
models from hanging/double-thinking on the Anthropic Messages endpoint.

### What's new

- **\MiniMax-M3-512k\** â€” 65,536 max output, **524,288-token context**, multimodal input, prompt caching, M-series tool preferences (search_and_replace preferred, apply_diff excluded), reasoning preserved across turns.
- **\MiniMax-M3-1M\** â€” 131,072 max output, **1,048,576-token context** (1M tokens), same multimodal/cache/tool config as the 512K variant.
- New default model: \minimaxDefaultModelId\ is now \MiniMax-M3-1M\. Existing \MiniMax-M2.7\ and other M2.x IDs are still resolvable for back-compat.
- Pricing matches the M2 flat tier: \.30/M input, \.20/M output, \.06/M cache read, \.375/M cache write. \longContextPricing\ is intentionally not set â€” see regression guard below.

### Bug fix (ported from kilo-code opencode \provider/transform.ts\)

When routing M-series models through the Anthropic Messages endpoint
(\pi.minimax.io/anthropic\), MiniMax defaults \	hinking\ to **off**. The model
would emit a cheap non-thinking response on the first turn, then re-decide to
think from scratch on the next turn, producing the **double-think / hang**
symptom on multi-turn agentic runs.

The fix mirrors kilo-code's opencode provider (lines 488â€“524, 661â€“669, 907â€“914,
1208â€“1211) and is implemented in \src/api/providers/minimax.ts\:

- For \MiniMax-M3-*\ models: inject \	hinking: { type: adaptive }\ and pass
  \	emperature: 1.0\, \	op_p: 0.95\. **Never** set \udget_tokens\ â€” M-series
  is a binary on/off toggle (matching kilo's \instant: { thinking: { type: disabled } }, thinking: { thinking: { type: adaptive }\ mapping).
- For \MiniMax-M2.x\ models: pass \	emperature: 1.0\, \	op_p: 0.95\, \	op_k: 40\
  (matches kilo's \minimax-m2\ defaults at lines 488â€“524). No \	hinking\ field â€”
  M2.x does not accept it.
- Inline comments in the provider cite the kilo-code line ranges so future
  maintainers can re-sync if kilo's logic changes.

## Files changed

| File | Change |
|---|---|
| \packages/types/src/providers/minimax.ts\ | +40 lines â€” add M3-512k and M3-1M model entries; switch default to M3-1M. |
| \packages/types/src/__tests__/provider-settings.test.ts\ | +12 lines â€” new \MiniMax provider\ describe block asserting M3 and M2.x all route to \nthropic\ protocol. |
| \src/api/providers/minimax.ts\ | +19 lines â€” adaptive-thinking + sampling params per the kilo port (replace existing 1-line temperature set). |
| \src/api/providers/__tests__/minimax.spec.ts\ | +199 lines â€” 4 new tests: M3 enables adaptive thinking, M2.x does not, M2 sampling params (\	op_p=0.95\, \	op_k=40\), M3 sampling params (\	op_p=0.95\, no \	op_k\); plus existing tests updated. |
| \webview-ui/src/components/ui/hooks/__tests__/useSelectedModel.spec.ts\ | +34 lines â€” two new tests verifying M3-512k and M3-1M resolve with correct contextWindow / maxTokens / supportsImages. |

## Tests

\\\
packages/types  â†’ provider-settings.test.ts â†’ 17/17 passed
src             â†’ minimax.spec.ts            â†’ 42/42 passed (38 prior + 4 new)
webview-ui      â†’ useSelectedModel.spec.ts    â†’ 27/27 passed
Total                                          86/86 passed
\\\

## Regression guards

- \expect(JSON.stringify(minimaxModels[MiniMax-M3-512k])).not.toContain(longContextPricing)\ â€” protects the intentional flat-tier pricing decision.
- \expect(lastCall.thinking).not.toHaveProperty(budget_tokens)\ â€” prevents future regressions to Anthropic-style effort-level knobs that don't apply to M-series.
- \expect(lastCall.thinking).toBeUndefined()\ for M2.x â€” protects against accidentally enabling thinking on a model family that doesn't accept it.

## Why no changesets?

Per \AGENTS.md\ in this repo:

> Changesets: Do NOT create \.changeset\ files for each commit or code change. Changesets are managed separately by maintainers and should not be generated by agents during normal development.

## Why this PR is targeted at \Zoo-Code-Org/Zoo-Code\, not a fork

This contribution is intended for the upstream project. Please review and let
me know if you'd like any of the optional follow-ups below; happy to address
in follow-up PRs.

## Optional follow-ups (NOT in this PR)

- Per-tenant pricing table support for MiniMax TokenPlan variants (highspeed / pricing-by-request).
- A webview settings UI hint that surfaces the new 1M-context model explicitly (currently it's just the default and one of the entries in the model dropdown).
- A CHANGELOG entry, added separately by maintainers per the changeset policy.

## Verification steps for reviewers

1. \pnpm install && pnpm vsix\ to build a fresh VSIX.
2. Install the vsix and add the MiniMax provider with an API key.
3. Set the default model â€” the picker should show \MiniMax-M3-1M\ (1M tokens).
4. Run an agentic task that requires at least one tool call. Confirm:
   - The reasoning block appears on turn 1 (thinking is on).
   - Subsequent turns continue reasoning without re-thinking the world from scratch (no double-think).
   - Long prompts do not produce over-long sampling cycles (no hang).
5. Run \pnpm test\ from the workspace root to confirm 86/86 tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added two new MiniMax model options: MiniMax-M3-512k and MiniMax-M3-1M, with updated context limits, image + prompt cache support, and revised pricing details.
  * Updated the MiniMax default model to MiniMax-M3-1M.

* **Bug Fixes**
  * Improved Anthropic-compatible MiniMax requests for M3 models, including adaptive thinking, consistent sampling behavior, omission of budget tokens, and correct model-specific max-token limits (with preserved MiniMax-M2.7 back-compat).

* **Tests**
  * Expanded handler, provider-resolution, API request regression, and UI selection coverage for the new M3 variants.

### DIFF
--- a/src/api/providers/__tests__/minimax.spec.ts
+++ b/src/api/providers/__tests__/minimax.spec.ts
@@ -121,6 +121,50 @@ describe("MiniMaxHandler", () => {
 			expect(model.info.cacheReadsPrice).toBe(0.03)
 		})
 
+		it("should return MiniMax-M3-512k model with correct configuration", () => {
+			const testModelId: MinimaxModelId = "MiniMax-M3-512k"
+			const handlerWithModel = new MiniMaxHandler({
+				apiModelId: testModelId,
+				minimaxApiKey: "test-minimax-api-key",
+			})
+			const model = handlerWithModel.getModel()
+			expect(model.id).toBe(testModelId)
+			expect(model.info).toEqual(minimaxModels[testModelId])
+			expect(model.info.contextWindow).toBe(524_288)
+			expect(model.info.maxTokens).toBe(65_536)
+			expect(model.info.supportsImages).toBe(true)
+			expect(model.info.supportsPromptCache).toBe(true)
+			expect(model.info.inputPrice).toBe(0.3)
+			expect(model.info.outputPrice).toBe(1.2)
+			expect(model.info.cacheWritesPrice).toBe(0.375)
+			expect(model.info.cacheReadsPrice).toBe(0.06)
+		})
+
+		it("should return MiniMax-M3-1M model with correct configuration", () => {
+			const testModelId: MinimaxModelId = "MiniMax-M3-1M"
+			const handlerWithModel = new MiniMaxHandler({
+				apiModelId: testModelId,
+				minimaxApiKey: "test-minimax-api-key",
+			})
+			const model = handlerWithModel.getModel()
+			expect(model.id).toBe(testModelId)
+			expect(model.info).toEqual(minimaxModels[testModelId])
+			expect(model.info.contextWindow).toBe(1_048_576)
+			expect(model.info.maxTokens).toBe(131_072)
+			expect(model.info.supportsImages).toBe(true)
+			expect(model.info.supportsPromptCache).toBe(true)
+			expect(model.info.inputPrice).toBe(0.3)
+			expect(model.info.outputPrice).toBe(1.2)
+			expect(model.info.cacheWritesPrice).toBe(0.375)
+			expect(model.info.cacheReadsPrice).toBe(0.06)
+		})
+
+		it(`should default to ${minimaxDefaultModelId} model`, () => {
+			const handlerDefault = new MiniMaxHandler({ minimaxApiKey: "test-minimax-api-key" })
+			const model = handlerDefault.getModel()
+			expect(model.id).toBe(minimaxDefaultModelId)
+		})
+
 		it("should return MiniMax-M2-Stable model with correct configuration", () => {
 			const testModelId: MinimaxModelId = "MiniMax-M2-Stable"
 			const handlerWithModel = new MiniMaxHandler({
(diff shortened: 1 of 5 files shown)


src/api/providers/minimax.ts
```diff
@@ -289,15 +291,44 @@ export class MiniMaxHandler extends BaseProvider implements SingleCompletionHand
...
+	private getMSeriesRequestParams(modelId: string): {
+		temperature: number
+		top_p: number
+		top_k?: number
+		thinking?: { type: "adaptive" }
+	} {
+		const isM3 = modelId.startsWith("MiniMax-M3")
+		return isM3
+			? { temperature: 1.0, top_p: 0.95, thinking: { type: "adaptive" } }
+			: { temperature: 1.0, top_p: 0.95, top_k: 40 }
```
edelauna, line 317: `getModel()` passes `defaultTemperature: 1.0` into `getModelParams`, which resolves a `temperature` honoring the user's `modelTemperature` — but `createMessage`/`completePrompt` never read it, since `getMSeriesRequestParams` hardcodes `temperature: 1.0` here. That makes the `defaultTemperature` arg dead and silently overrides the user's temperature for M2.x too, not just M3. Other providers (anthropic, deepseek, mistral) consume `getModel().temperature`. Is the fixed 1.0 intended for the whole MiniMax family, or should M2.x still honor `modelTemperature`?

src/api/providers/__tests__/minimax.spec.ts
```diff
@@ -193,10 +237,20 @@ describe("MiniMaxHandler", () => {
 			expect(model.info).toEqual(minimaxModels[minimaxDefaultModelId])
 		})
 
-		it("should default to MiniMax-M2.7 model", () => {
+		it(`should default to ${minimaxDefaultModelId} model`, () => {
```
edelauna, line 240: This `it()` block is identical to the one at line 162 (same title, setup, and assertion), so vitest runs the same test twice. Was one of these meant to be removed when the old `should default to MiniMax-M2.7 model` test was rewritten?

edelauna (changes requested):
Thanks for adding this, just had a question regarding how temperature should be set.

PR #775, opened 2026-06-30
fix(deepseek): round-trip reasoning_content in thinking mode to prevent 400 errors

### LINKED ISSUE #201: [BUG] DeepSeek error: 'reasoning_content' must be passed back to the API
**Describe the bug**
DeepSeek error: 'reasoning_content' must be passed back to the API

**To Reproduce**
1. Create a DeepSeek provider with deepseek-v4-pro and your API.
2. Set Model Reasoning Effort to High.
3. Run a task.

**Expected behavior**
I expected it to run smoothly, but I get this:
Provider Error · 400
The provider couldn't process the request as made. Stop the task and try a different approach.
Details: DeepSeek completion error: 400 The `reasoning_content` in the thinking mode must be passed back to the API.

i.e. 3.54.1

edelauna:
### Related GitHub Issue

Closes: #201

### Description

When DeepSeek thinking mode is active, the API requires `reasoning_content` to be passed back on every assistant message in conversation history. Without it, the API returns HTTP 400: `The reasoning_content in the thinking mode must be passed back to the API`.

**Root cause — two bugs, one symptom:**

1. **`convertToOpenAiMessages`** (used by the generic OpenAI-compatible provider path) silently dropped `reasoning` content blocks and never set `reasoning_content` on outgoing messages. This is the primary fix.

2. **aimock fixture format** — the e2e fixtures had `reasoning` + `toolCalls` but no `content`, so aimock classified them as `ToolCallResponse` (which ignores `reasoning`) instead of `ContentWithToolCallsResponse`. Fixed by adding `"content": ""`.

**What changed:**

- `src/api/transform/openai-format.ts` — `convertToOpenAiMessages` now extracts `reasoning_content` from both the top-level field (already round-tripped) and `{ type: "reasoning" }` content blocks (stored by Task.ts during streaming), and sets it on the outgoing assistant message.
- `src/api/transform/__tests__/openai-format.spec.ts` — 5 new unit tests covering the round-trip: string-content path, array-content path, array-content with tool calls (the exact failure case), preference ordering, and absent-reasoning no-op.
- `src/api/providers/__tests__/openai.spec.ts` — provider-level integration test: `OpenAiHandler.createMessage` with `preserveReasoning: true` and an assistant message containing a `reasoning` block + tool call; asserts `reasoning_content` appears in the body sent to the OpenAI client. This directly exercises the fixed `OpenAiHandler → convertToOpenAiMessages` path.
- `apps/vscode-e2e/fixtures/deepseek-v4.json` — added `"content": ""` to reasoning-on fixtures so aimock routes them through `buildContentWithToolCallsChunks` and streams `reasoning_content` in SSE deltas.
- `apps/vscode-e2e/src/suite/providers/deepseek-v4.test.ts` — e2e regression guard for the native DeepSeek provider path: asserts `hasReasoningContentInHistory` on the second request (after the tool call), proving `reasoning_content` from turn 1 survives into turn 2's history.

**Coverage note:** The native `DeepSeekHandler → convertToR1Format` path was already correct before this PR. The e2e test guards that path. The fix to `convertToOpenAiMessages` is guarded by unit tests + the new provider-level integration test in `openai.spec.ts`.

### Test Procedure

**Unit tests:**
```
pnpm test
```
All 408 test files pass (6,584 tests).

**E2e tests (mock):**
```
TEST_FILE=deepseek-v4.test pnpm --filter @roo-code/vscode-e2e test:ci:mock
```
All 4 DeepSeek V4 provider tests pass, including the `reasoning-on` probes that assert `hasReasoningContentInHistory` on the second request.

**Live API validation** (performed during development):
- Confirmed DeepSeek returns `content: ""` + `reasoning_content: "..."` + `tool_calls: [...]` on turn 1 tool-call responses.
- Confirmed turn 2 with `reasoning_content` round-tripped returns HTTP 200.

### Pre-Submission Checklist

- [x] **Documentation Impact**: No documentation updates required (internal API behaviour fix).

## Summary by CodeRabbit

* **New Features**
  * Added preservation and round-trip support for assistant `reasoning_content` in OpenAI-compatible message formatting, including DeepSeek/Z.ai thinking mode and mixed tool+reasoning turns.
  * Expanded DeepSeek V4 end-to-end validation to detect and verify reasoning presence across chat-completions requests.

* **Bug Fixes**
  * Fixed scenarios where assistant reasoning could be omitted or mis-mapped when provided via structured content blocks and when tool calls were included.

* **Tests / Chores**
  * Added regression and format round-trip tests, updated DeepSeek V4 fixtures, and strengthened E2E assertions/diagnostics for reasoning history.


src/api/transform/openai-format.ts
```diff
@@ -450,6 +457,9 @@ export function convertToOpenAiMessages(
 					}
 				}
 			} else if (anthropicMessage.role === "assistant") {
+				const messageWithDetails = anthropicMessage as any
```
taltas: I am not a big fan of this `any` here... maybe we should better define the type here? It's not really a blocker, but something to consider.

taltas (dismissed):
Clean PR, except one question with the typing.

navedmerchant (approved):

PR #776, opened 2026-06-30
fix(gemini): base64-encode thoughtSignature bypass token to fix Vertex AI empty-response loop

### LINKED ISSUE #536: [BUG] Vertex AI Gemini 3.1/3.5: empty response loop after first tool-calling turn
## Problem

When using `gemini-3.1-pro-preview` or `gemini-3.5-flash` via Vertex AI, every API call after the first tool use returns an empty response, causing an unrecoverable infinite retry loop that makes the models completely unusable.

## Context

Affects any Vertex AI user with `gemini-3.1-pro-preview`, `gemini-3.1-pro-preview-customtools`, or `gemini-3.5-flash`. Fails on the second API call of every task (after the first tool result is returned). With auto-approval enabled the loop runs indefinitely and never self-recovers — the only workaround is manually switching to a different model.

## Reproduction steps

1. Configure Zoo Code with Google Vertex AI and one of: `gemini-3.1-pro-preview`, `gemini-3.5-flash`
2. Start any task that triggers a tool call (e.g. read a file)
3. After the first tool result is returned, every subsequent call returns empty — no text, no tool calls
4. Error repeats: _"Unexpected API Response: The language model did not provide any assistant messages"_
5. Loop never recovers without a model switch

## Expected result

Model responds normally on every turn including after tool results.

## Actual result

Silent empty response from Vertex on every retry after the first tool-calling turn.

## Variations tried

Switching to Gemini 3 Flash (older, non-3.1/3.5) temporarily unblocks — that model doesn't enforce `thoughtSignature` validation so the stale value is ignored. Reported since Gemini 3.1 was released on Vertex. Related upstream: RooCodeInc/Roo-Code#12045, RooCodeInc/Roo-Code#11629.

## App Version

Current main (v3.58.1+)

## API Provider

Google Vertex AI

## Model Used

`gemini-3.1-pro-preview`, `gemini-3.1-pro-preview-customtools`, `gemini-3.5-flash`

## Relevant logs

```
Unexpected API Response: The language model did not provide any assistant messages. This may indicate an issue with the API or the model's output.
The language model returned an empty response without any text or tool calls. This can happen due to API issues, rate limiting, or model-specific problems.
```

edelauna:
### Related GitHub Issue

Closes: #536

### Description

Gemini 3.x (including `gemini-3.1-pro-preview` and `gemini-3.5-flash`) validates `Part.thoughtSignature` strictly on Vertex AI. The field is documented by the Google GenAI SDK as **"Encoded as base64 string"**. The fallback bypass token in `gemini-format.ts` was being sent as a plain string (`"skip_thought_signature_validator"`), while the same token in `lite-llm.ts` was already correctly base64-encoded.

Vertex AI enforces base64 validation more strictly than the direct Gemini API. When the plain string was sent on turn 2+ (after a tool call), Vertex returned an empty response, triggering the infinite retry loop reported in #536.

**How:**
- Added `GEMINI_THOUGHT_SIGNATURE_BYPASS` as a named export in `src/api/transform/gemini-format.ts` — the single source of truth for the base64-encoded bypass token.
- Updated `gemini-format.ts` to use the constant instead of the inline plain string.
- Updated `lite-llm.ts` to import and use the same constant, eliminating the duplicate `Buffer.from(...)` computation.

**Reviewers should note:**
- The direct Gemini API accepts both plain and base64 forms (confirmed via live API test), but Vertex AI requires base64. The fix is safe for both paths.
- `thoughtSignature` round-trip from real API responses is unaffected — only the fallback/bypass path changed.

### Test Procedure

TDD approach: failing tests written first, then fix applied.

- `src/api/transform/__tests__/gemini-format.spec.ts` — updated existing tool-use test to assert base64-encoded bypass token; added assertion documents the exact Vertex AI failure mode.
- `src/api/providers/__tests__/gemini.spec.ts` — added `thoughtSignature round-trip (issue #536)` suite covering: capture from stream, round-trip through history on turn 2, base64 fallback for cross-model history, disabled-reasoning edge case, and no-tools boundary.
- `src/api/providers/__tests__/lite-llm.spec.ts` — existing tests continue to pass unchanged (they already asserted the base64 value).

```
pnpm exec vitest run api/transform/__tests__/gemini-format.spec.ts api/providers/__tests__/gemini.spec.ts api/providers/__tests__/lite-llm.spec.ts
# 3 files, 74 tests, all pass
```

### Pre-Submission Checklist

### Documentation Updates

### Get in Touch

edelauna@gmail.com

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with Gemini tool calls by preserving and reusing required reasoning metadata across streamed turns.
  * Added a safe fallback for cases where no prior signature is available, helping avoid validation issues in Gemini responses.
  * Ensured tool-related behavior remains consistent even when reasoning settings are disabled.

* **Tests**
  * Expanded coverage for Gemini streaming and tool-call scenarios, including edge cases around missing signatures.

### DIFF
--- a/src/api/providers/__tests__/gemini.spec.ts
+++ b/src/api/providers/__tests__/gemini.spec.ts
@@ -54,6 +54,215 @@ describe("GeminiHandler", () => {
 		})
 	})
 
+	describe("thoughtSignature round-trip (issue #536)", () => {
+		const systemPrompt = "You are a helpful assistant"
+		const toolMetadata = { tools: [{ function: { name: "read_file", description: "", parameters: {} } }] } as any
+
+		// Helper: build a mock async-iterable stream from chunks
+		function makeStream(chunks: unknown[]) {
+			return {
+				[Symbol.asyncIterator]: async function* () {
+					for (const chunk of chunks) yield chunk
+				},
+			}
+		}
+
+		// Simulate a Gemini 3.x response: thoughtSignature arrives on its own part,
+		// alongside a functionCall part (the way the real Gemini 3 API returns it).
+		const turn1Response = makeStream([
+			{
+				candidates: [
+					{
+						content: {
+							parts: [
+								{ thought: true, text: "thinking…" },
+								{ functionCall: { name: "read_file", args: { path: "foo.ts" } } },
+								{ thoughtSignature: "sig-abc123" },
+							],
+						},
+					},
+				],
+			},
+			{ usageMetadata: { promptTokenCount: 10, candidatesTokenCount: 5 } },
+		])
+
+		it("captures thoughtSignature from the stream after turn 1", async () => {
+			;(handler["client"].models.generateContentStream as any).mockResolvedValue(turn1Response)
+
+			const messages: Anthropic.Messages.MessageParam[] = [{ role: "user", content: "Read foo.ts" }]
+
+			for await (const _chunk of handler.createMessage(systemPrompt, messages, toolMetadata)) {
+				// drain
+			}
+
+			expect(handler.getThoughtSignature()).toBe("sig-abc123")
+		})
+
+		it("sends thoughtSignature from history on turn 2 (core regression)", async () => {
+			// This is the bug from issue #536: after turn 1 the thoughtSignature block is
+			// persisted into apiConversationHistory. On turn 2 the handler must include it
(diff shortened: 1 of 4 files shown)


taltas (approved):
Clean PR, nice

PR #789, opened 2026-07-03
Improve Anthropic Vertex Claude content block handling

### LINKED ISSUE #788: [ENHANCEMENT] Improve Anthropic Vertex Claude completion parsing for non-text content blocks

Anthropic Vertex Claude completion handling currently assumes the first response content block is the visible text response.

This works for the common case, but it is less resilient if Claude returns an empty content array or includes non-text content blocks before the actual text block.

Users using Claude models through Anthropic Vertex may receive responses where the content array contains multiple block types.

For example, a Claude response could include a non-text block before a text block:

```ts
[
  { type: "thinking", thinking: "internal reasoning" },
  { type: "text", text: "visible response" },
]
```

In this case, Zoo Code should still be able to extract the visible text response safely.

Zoo Code should handle Anthropic Vertex Claude response content more defensively.

Instead of assuming the first content block is text, it should return the first available text block from the response content array.

If no text block exists, it should safely return an empty string.

* Keep the existing behavior unchanged for normal text-only Claude responses.
* Avoid throwing when the content array is empty.
* Avoid returning an empty response when a valid text block exists later in the content array.
* Keep the change small and provider-specific.

### Request checklist

_No response_

_No response_

_No response_

_No response_

daewoongoh:
### Related GitHub Issue

Closes: #788  

### Description

This PR improves how completion response content blocks are handled when using Claude models through Anthropic Vertex.

The previous implementation only checked `response.content[0]`, assuming that the first content block would always be a `text` block. This works for common text-only responses, but it is less defensive when Claude returns an empty content array or includes non-text blocks, such as `thinking`, before the actual text response.

This change updates the logic to search for the first available `text` block in the response content array. If no text block exists, it safely returns an empty string, preserving the existing fallback behavior.

Main changes:

* Updated Anthropic Vertex `completePrompt` to find the first `text` block instead of only checking `response.content[0]`.
* Safely handles empty content arrays.
* Added test coverage for empty Claude content responses.
* Added test coverage for mixed Claude content responses where a non-text block appears before a text block.

### Test Procedure

Added unit tests for the following cases:

1. Claude returns an empty `content` array and `completePrompt` returns an empty string.
2. Claude returns mixed content blocks and `completePrompt` returns the first text block.

Test command:

```bash
pnpm test src/api/providers/__tests__/anthropic-vertex.spec.ts
```

### Pre-Submission Checklist

### Screenshots / Videos

N/A

### Documentation Updates

### Additional Notes

This change aligns Anthropic Vertex completion handling with other providers by extracting the first available text block instead of assuming the first content block is always text.

### Get in Touch

hehegwk_23849

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved prompt completion handling for Claude/Vertex responses so text is extracted correctly even when non-text content appears first.
  * Empty responses now return an empty string instead of failing or returning unexpected output.
  * Added test coverage for empty content and mixed content responses to prevent regressions.

### DIFF
--- a/src/api/providers/__tests__/anthropic-vertex.spec.ts
+++ b/src/api/providers/__tests__/anthropic-vertex.spec.ts
@@ -895,6 +895,41 @@ describe("VertexHandler", () => {
 			const result = await handler.completePrompt("Test prompt")
 			expect(result).toBe("")
 		})
+
+		it("should handle empty content array for Claude", async () => {
+			handler = new AnthropicVertexHandler({
+				apiModelId: "claude-3-5-sonnet-v2@20241022",
+				vertexProjectId: "test-project",
+				vertexRegion: "us-central1",
+			})
+
+			const mockCreate = vitest.fn().mockResolvedValue({
+				content: [],
+			})
+			;(handler["client"].messages as any).create = mockCreate
+
+			const result = await handler.completePrompt("Test prompt")
+			expect(result).toBe("")
+		})
+
+		it("should return text from first text block when mixed content for Claude", async () => {
+			handler = new AnthropicVertexHandler({
+				apiModelId: "claude-3-5-sonnet-v2@20241022",
+				vertexProjectId: "test-project",
+				vertexRegion: "us-central1",
+			})
+
+			const mockCreate = vitest.fn().mockResolvedValue({
+				content: [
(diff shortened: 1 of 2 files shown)


edelauna (approved):
Thanks for the fix

PR #793, opened 2026-07-03
feat(kenari): add Kenari as a first-class provider

### LINKED ISSUE #792: Provider request: Kenari (kenari.id), Indonesian OpenAI-compatible gateway
### Which provider?

[Kenari](https://kenari.id), an Indonesian OpenAI-compatible AI gateway billed in Rupiah (IDR). One `kn-` API key covers Claude, GPT, DeepSeek, GLM, Kimi and more. Base URL `https://kenari.id/v1`, standard Bearer auth, and a public `https://kenari.id/v1/models` catalog (no key needed) with `context_length` and modality metadata per model.

### Why first-class?

Kenari works today through the generic OpenAI Compatible provider, but that path needs manual base URL + model id + model config. A named entry gives Indonesian users (who mostly cannot pay in USD; Kenari is their local-currency on-ramp) a picker with the live model list auto-populated, following the same dynamic-provider pattern as Opencode Go (#319).

### Implementation

I have the implementation ready on a branch (RouterProvider handler + `/v1/models` fetcher + settings UI + validation + tests, mirroring #319 file-by-file; ~2400 relevant tests pass, lint and types clean) and will open the PR linked to this issue. Claiming.

Disclosure: I run kenari.id. Happy to provide a test key for verification.

doedja:
Closes #792

### What

Adds [Kenari](https://kenari.id) as a first-class dynamic provider. Kenari is an Indonesian OpenAI-compatible AI gateway billed in Rupiah (IDR): one `kn-` API key covers Claude, GPT, DeepSeek, GLM, Kimi and more. It already works through the generic OpenAI Compatible provider; this gives it a named entry with the live model list auto-populated.

### How

Mirrors the Opencode Go provider (#319) file-by-file, adapted to current main:

- `src/api/providers/kenari.ts`: handler on `RouterProvider` (streaming text, `reasoning_content`, tool-call partials, usage with cache-read tokens).
- `src/api/providers/fetchers/kenari.ts`: dynamic model list from the public `https://kenari.id/v1/models` (no key needed; the key is forwarded when present). Maps `context_length` and `modalities.input` (image support). Pricing is intentionally not mapped: Kenari returns IDR (`micro_idr_per_1m_tokens`), and the ModelInfo price fields are USD, so a converted or raw value would be wrong; cost stays undefined instead.
- `packages/types`: `kenari.ts` (default `glm-5-2`, 1,048,576 context fallback matching the live default model), settings schema, discriminated union, `modelIdKeys`, `MODELS_BY_PROVIDER`, `SECRET_STATE_KEYS`.
- Registration: `buildApiHandler`, `modelCache`, `webviewMessageHandler` (fetched unconditionally like the other public-endpoint routers), `dynamicProviderExtras`.
- Webview: `Kenari.tsx` settings component (key field + get-key CTA + ModelPicker), `ApiOptions`, `PROVIDERS` constant, `providerModelConfig`, `useSelectedModel`, `validate.ts`, and the `kenariApiKey`/`getKenariApiKey` strings across all 18 locales.
- Tests: handler spec (streaming, request body, completePrompt, error wrapping), fetcher spec (mapping, auth header, error/invalid-entry fallbacks), settings component spec, plus kenari cases in the ClineProvider / webviewMessageHandler / validate suites.

Deliberately skipped from #319: the Anthropic-wire split machinery. Kenari serves every model over OpenAI `/chat/completions`, so the default "openai" protocol is correct with no extra code.

### Testing

- `packages/types` suite, `src` shared+providers sweep (1603 passed / 1 skipped), webview settings+hooks+utils sweep (592 passed), kenari-focused suites 216/216.
- `check-types` clean in all three packages; eslint `--max-warnings=0` and prettier clean on every touched file.
- Live verification: `https://kenari.id/v1/models` public (24 models with `context_length` and modalities); chat completions verified with a real `kn-` key against `https://kenari.id/v1`. Docs: https://kenari.id/docs, OpenAPI: https://kenari.id/openapi.json.

Disclosure: I run kenari.id. Happy to provide a test key for verification.

## Summary by CodeRabbit

* **New Features**
  * Added Kenari as a supported provider end-to-end, including UI settings, model discovery/selection, routing, and chat handling.
  * Introduced Kenari API key configuration, including the new secret storage key (`kenariApiKey`).
  * Enabled Kenari streaming and non-streaming responses, including reasoning, tool-call partials, and usage reporting.
  * Added Kenari settings localization across multiple languages.
* **Bug Fixes**
  * Improved Kenari configuration validation and model-id fallback behavior.
* **Tests**
  * Expanded Kenari test coverage for model fetching/caching, message streaming formatting, and UI interactions.


doedja:
The platform-unit-test failure was in SettingsView.change-detection.spec.tsx ("resets cached provider state when a new import timestamp arrives", expected baseten / got deepseek), a file this PR does not touch; it passes locally on this branch (4/4) and the same test failed once on main's own CI when it landed in #726. Pushed a same-content commit to retrigger CI since I lack rerun permissions.

src/api/providers/fetchers/__tests__/kenari.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+						{ id: "valid-model", context_window: 50000 },
+						{ not_a_field: true }, // no `id`, so it will fail safeParse
+					],
+				},
+			})
+			const warnSpy = vitest.spyOn(console, "warn").mockImplementation(() => {})
+
+			const models = await getKenariModels("k")
+
+			expect(Object.keys(models)).toEqual(["valid-model"])
```
edelauna, line 92: This only checks that `valid-model` survives the fallback parse — should it also assert `models["valid-model"].contextWindow` (the fixture sets `context_window: 50000` at line 83)? As written, a mutation that broke field-mapping in this fallback branch specifically wouldn't be caught here.

src/api/providers/fetchers/kenari.ts
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * Falls back to {@link kenariDefaultModelInfo} when the upstream payload
+ * omits context-window or max-token fields, ensuring downstream consumers
+ * always receive a fully-populated object.
+ *
+ * @param model - Validated model entry from the `/models` response.
+ * @returns Normalised model metadata suitable for the model picker.
+ */
+export const parseKenariModel = (model: KenariModel): ModelInfo => ({
+	maxTokens: model.max_output_tokens ?? model.max_tokens ?? kenariDefaultModelInfo.maxTokens,
```
edelauna, line 47: Is the `max_output_tokens ?? max_tokens` (and `context_window ?? context_length`) precedence intentional? No test currently supplies both fields with different values, so this fallback order isn't actually verified anywhere.

src/api/providers/kenari.ts
```diff
@@ -0,0 +1,143 @@
...
+		for await (const chunk of completion) {
+			const delta = chunk.choices[0]?.delta
+
+			if (delta?.content) {
+				yield { type: "text", text: delta.content }
+			}
+
+			// Several Kenari models (GLM, DeepSeek) stream reasoning via this field.
+			if (delta && "reasoning_content" in delta && delta.reasoning_content) {
+				yield { type: "reasoning", text: delta.reasoning_content as string }
```
edelauna: `src/api/providers/utils/extract-reasoning.ts` already has `extractReasoningFromDelta(delta)` for this — it also falls back to the OpenRouter-style `reasoning` field and uses a proper `typeof === "string"` check instead of this cast. Worth reusing here like `opencode-go.ts` does?

edelauna (changes requested):
Looks good - main comment would be about re-using `extract-reasoning` logic so we're not duplicating logic.

Before adding this provider can you also create a correspond PR in https://github.com/Zoo-Code-Org/Zoo-Code-Docs for the provider slug documentation.

src/api/providers/fetchers/__tests__/kenari.spec.ts
doedja: Good catch. Added `expect(models["valid-model"].contextWindow).toBe(50000)` right after the keys check, so a field-mapping regression in that fallback branch is now caught. (77173c2a0)

src/api/providers/fetchers/kenari.ts
doedja: Intentional, and now verified. The precedence matches the house pattern in `opencode-go.ts` (identical 4-field optional schema) and `litellm.ts` (`max_output_tokens || max_tokens`): prefer the explicit `context_window`/`max_output_tokens`, fall back to the `context_length`/`max_tokens` aliases. Kenari currently returns only `context_length` and no max-token field, so the first-position fields are forward-compat, but I kept them for consistency and added a `parseKenariModel` test that supplies both with differing values to lock the order. (77173c2a0)

src/api/providers/kenari.ts
doedja: Done, switched to `extractReasoningFromDelta(delta)` exactly like `opencode-go.ts`. This drops the cast for the proper `typeof === "string"` check and gains the OpenRouter-style `reasoning` fallback; added a handler test covering that fallback path. (77173c2a0)

doedja:
Thanks for the review @edelauna. Pushed 77173c2a0 addressing all three:

1. **Reasoning helper** ([kenari.ts:84]) - now uses the shared `extractReasoningFromDelta`, matching `opencode-go.ts`; gains the OpenRouter-style `reasoning` fallback + a test for it.
2. **Fallback-parse coverage** ([spec:92]) - asserts `contextWindow` on the surviving entry.
3. **Precedence** ([fetchers:47]) - kept (matches `opencode-go`/`litellm` house pattern; Kenari sends only `context_length` today, so the primary fields are forward-compat) and locked with a `parseKenariModel` test using both fields.

Gates green locally: `check-types` (11/11 tasks), the two kenari suites (29 passed), prettier, and eslint `--max-warnings=0` on the touched files.

Also opened the companion docs PR as requested: **Zoo-Code-Org/Zoo-Code-Docs#7** (provider page + `index.json` registration).

edelauna (approved):

PR #835, opened 2026-07-05
fix(telemetry): reduce PostHog event volume, fix consent dark pattern

### LINKED ISSUE #830: Reduce PostHog event volume and fix telemetry consent/privacy gaps
# Context

PostHog telemetry has been exceeding the free-tier monthly quota (Product analytics 1M, Session replay 5K, Error tracking 100K — all exceeded in the June 6–July 6 cycle, driven by a ~2.7M-event spike on June 7–9). Once a free-tier limit is exceeded, PostHog silently drops ingestion for the rest of the billing cycle, which is why PostHog showed near-zero events for weeks despite Marketplace installs (45K+) continuing to grow — this was initially mistaken for a broken telemetry pipeline before the billing page confirmed the real cause.

Separately, a telemetry best-practice gap analysis (see `TELEMETRY_GAP_ANALYSIS.md` in the repo root) found consent/privacy issues that are largely inherited unchanged from upstream Roo Code. Both problems live in the same subsystem (`packages/telemetry`, `Task.ts`, code-index services), so this issue bundles the fixes together.

Breakdown of the June 7–9 spike (~2.7M events):

| Event | Count | People | Per-person | Verdict |
|---|---|---|---|---|
| `Conversation Message` | 1,318,290 | 3,949 | 334 | Real signal, tracked at message-granularity — expensive |
| `Tool Used` | 804,551 | 3,761 | 214 | Real signal, same issue |
| `Code Index Error` | 254,786 | 309 | **825** | Bug — looks like an unbounded file-watcher retry loop against a broken embedder/vector-store config |
| `Checkpoint Created` | 153,787 | 2,910 | 53 | Real usage |
| `Model Cache Empty Response` | 50,095 | 3,023 | 17 | Possibly tracking an expected/benign condition every time |

# Developer Notes

## Part A — Event volume (get back under free tier)

**A1. Circuit-breaker for `CODE_INDEX_ERROR` retry storm (highest impact)**
309 people generated 254,786 events (9.4% of the entire monthly analytics quota) — consistent with a small number of installs stuck in a permanently-broken embedder/vector-store config, where `file-watcher.ts` re-triggers indexing on every file-system event with no backoff. Add a per-workspace circuit breaker: stop retrying/firing telemetry after N consecutive failures until the embedder config changes.
- Files: `src/services/code-index/processors/file-watcher.ts`, `src/services/code-index/manager.ts`

**A2. Aggregate `Conversation Message` into `Task Completed`**
Currently fires once per message turn (`Task.ts:2606`, `Task.ts:3523` via `TelemetryService.captureConversationMessage`). Replace with a `messageCount` property summarized once on `Task Completed` instead of one event per turn.
- Files: `src/core/task/Task.ts`, `packages/telemetry/src/TelemetryService.ts`

**A3. Aggregate `Tool Used` into `Task Completed`**
Same treatment — 2 call sites, summarize as a per-task tool-usage breakdown (e.g. `toolsUsed: { read: 12, edit: 8 }`) instead of one event per tool call.
- Files: `src/core/task/Task.ts`, `packages/telemetry/src/TelemetryService.ts`

**A4. Review `Model Cache Empty Response` noise**
Check whether this represents an expected/benign condition being tracked unconditionally; throttle or remove if so.

## Part B — Privacy / consent gaps (from `TELEMETRY_GAP_ANALYSIS.md`)

**B1. Fix consent banner dark pattern (Critical)**
`webview-ui/src/components/common/TelemetryBanner.tsx` sends `telemetrySetting: "enabled"` when the user dismisses the banner with "×" — there's no neutral dismiss or explicit decline. Make "×" a neutral dismiss that leaves the setting as `"unset"` (routed to disabled), and add explicit Accept/Decline actions.

**B2. Salt-hash the persistent machine ID (Critical)**
Both PostHog clients (`packages/telemetry/src/PostHogTelemetryClient.ts`, `webview-ui/src/utils/TelemetryClient.ts`) use raw, unhashed `vscode.env.machineId` as the distinct ID. Hash it (e.g. SHA-256 + fixed salt) before use, or generate a separate installation UUID.

**B3. Clarify PRIVACY.md dual-pipeline scope (High)**
The "Telemetry" bullet says "does not collect your code or AI prompts" without noting this only applies to the PostHog path — the separate Cloud sync pipeline (`packages/cloud/src/TelemetryClient.ts`) does transmit task messages when cloud sync is enabled. Add one clarifying sentence.

**B4. Adopt `vscode.env.isTelemetryEnabled` (Medium)**
Code only checks the deprecated `telemetry.telemetryLevel` config setting. Add an `onDidChangeTelemetryEnabled` listener alongside it so the extension reacts live to VS Code's global telemetry toggle.

**B5. Await telemetry shutdown properly (Medium)**
`extension.ts` calls `TelemetryService.instance.shutdown()` in `deactivate()` without awaiting it, and `TelemetryService.shutdown()` itself uses `forEach` instead of `Promise.all` + `await` — risking dropped events (including the most diagnostically useful ones) on process exit.

## Estimated scope

Roughly 300–500 changed lines across ~10–12 files, no new dependencies, no data/schema migrations, no breaking API changes. B1/B3/B5 are trivial (a few lines each); B2/B4 are small, mechanical changes; A1/A2/A3 need a bit of design judgment (retry threshold, aggregation shape) but are localized.

edelauna:
### Related GitHub Issue

Closes: #830

### Description

PostHog's Product Analytics quota was 2.7x over its free-tier limit (2.7M/1M events in the June 6 - July 6 cycle), driven mainly by a `CODE_INDEX_ERROR` retry-storm bug (309 people generating 254,786 events) and per-turn `Conversation Message`/`Tool Used` events. Separately, a telemetry gap analysis found the consent banner treated dismissal (and the default "unset" state) as opt-in rather than a real choice, and that the underlying default itself was undisclosed.

**Volume reduction:**
- Circuit-breaker for `CODE_INDEX_ERROR`: sliding 10-minute window, trips after 50 occurrences, independent of any other telemetry the same install sends (so unrelated events can't mask or reset a burst).
- `Conversation Message`/`Tool Used` are no longer emitted per-turn/per-tool-call. Instead, `Task.messageCounts` and `Task.toolUsage` are summarized into `toolsUsed`/`messageCount` on the `Task Completed` event. Tool usage is now recorded exactly once per attempt at a single central point (`presentAssistantMessage`), removing a double-count where several tool handlers also recorded success locally.
- `Task Completed` now fires once per model-initiated `attempt_completion` call (tagged `completionReason: "attempt_completion"`) rather than only when the user clicks Accept — a task the user declines, gives feedback on, or never explicitly accepts previously reported nothing at all. It also fires on a 30-minute idle timeout and on task/extension shutdown (`completionReason: "idle"`/`"shutdown"`) so long-running or abandoned tasks aren't invisible either. Each installment reports only the delta since the previous one for that task (tracked separately from `task.toolUsage`/`messageCounts`, which stay full running totals for the public `TaskCompleted` API event and the UI), so summing a task's installments reconstructs its full counts without double-counting. Skips emission for stale replays of an already-completed subtask (revisiting it from history).
- `Model Cache Empty Response` throttled to once per distinct provider+endpoint (matching the same compound cache-key identity the model cache itself uses) until a non-empty response is seen — including for auth-scoped providers (e.g. `zoo-gateway`), where the throttle now correctly re-arms after a later non-empty response even though those providers skip caching.

**Consent / privacy:**
- Telemetry is opt-out by disclosed default: `isTelemetryOptedIn()` (`packages/types`) returns `true` unless the setting is explicitly `"disabled"`. A fresh install (`"unset"`) and an explicit `"enabled"` both capture telemetry; only an explicit Decline opts out. This replaces the previous ad hoc `telemetrySetting !== "disabled"` checks that were scattered across 5 call sites with a single source of truth, and keeps the default itself honestly disclosed in `PRIVACY.md` and the banner copy rather than silently opting users in via an ambiguous dismiss gesture.
- Telemetry banner: the "×" is a neutral dismiss — it sends no message and leaves the setting `"unset"` (so the disclosed default stays in effect), and does not itself record a choice either way. Explicit **Accept**/**Decline** buttons let a user affirmatively opt in or out. Localized across all 18 supported locales.
- Extension now reacts live to `vscode.env.onDidChangeTelemetryEnabled`, ANDing the current `vscode.env.isTelemetryEnabled` value into the gate so a live OS/editor-level telemetry-disable actually takes effect immediately, rather than only being caught on the next read of the deprecated `telemetry.telemetryLevel` setting.
- `TelemetryService.shutdown()` is now properly awaited on deactivation, tracks in-flight `capture()`/`captureException()` promises and drains them before flushing/closing clients, and is wrapped so a shutdown failure can't skip terminal cleanup.
- `PRIVACY.md` clarified: telemetry is enabled by default (opt-out, changeable in settings), and the "does not collect code or prompts" bullet now notes this applies to the PostHog path only — the separate Cloud task-sync pipeline does transmit task messages when enabled.
- Tool usage is now recorded for analytics only after the tool name passes validation (or is bucketed under a static `invalid_tool_call` key for unrecognized/malformed calls), so a model-supplied name can never reach PostHog as an arbitrary `toolsUsed` property key.
- `webview-ui`'s PostHog client now explicitly disables `autocapture`, `capture_pageview`, and `capture_pageleave` on init, so the webview doesn't passively collect DOM interactions/navigation beyond the explicit events this PR already sends. (The `src`-side client uses `posthog-node`, the server SDK, which has no browser autocapture/session-recording/heatmap surface to begin with.)

**Explicitly out of scope for this PR:**
- Hashing `vscode.env.machineId` — assessed and dropped; it's already a random per-install UUID, not a hostname/username-derived identifier, so hashing it would pseudonymize an already-non-identifying value while losing existing PostHog person continuity.
- Error-tracking-specific fixes (redacting `captureException` payloads, potentially moving `CODE_INDEX_ERROR` from Product Analytics to Error Tracking) — Error Tracking was only 6% over quota vs. Product Analytics' 2.7x, so it wasn't the priority; flagged for a follow-up.

### Test Procedure

- `pnpm check-types` and `pnpm lint` clean across all packages.
- Full test suite: 6753 tests in `src`, 1456 in `webview-ui`, 43 in `@roo-code/telemetry`, 238 in `@roo-code/types` — all passing. New/updated coverage includes:
  - the circuit breaker's sliding window and independence from unrelated events
  - tool usage recorded exactly once per attempt (no double-counting across handlers)
  - unrecognized/invalid tool calls bucketed under a static `invalid_tool_call` analytics key rather than the raw model-supplied name
  - per-endpoint empty-model throttling, including re-arming for auth-scoped providers after a non-empty response
  - the banner's Accept/Decline/neutral-dismiss behavior
  - `isTelemetryOptedIn()` opt-out-by-default semantics (`"unset"`/`"enabled"` both opted in, only `"disabled"` opts out) and the resulting opt-in/opt-out transition events
  - the live `onDidChangeTelemetryEnabled` listener honoring `vscode.env.isTelemetryEnabled`
  - shutdown draining/error-handling
  - `Task Completed` installment delta-tracking: first flush, second flush reporting only the new delta, no-op when nothing changed, idle-timer flush, shutdown flush, and skipping emission for a stale history replay of an already-completed subtask
- Manual verification recommended: fresh profile shows the Accept/Decline banner with telemetry already active by default; "×" dismisses without recording a choice; Decline actually stops capture; a task whose model calls `attempt_completion` emits a `Task Completed` installment immediately regardless of whether the user accepts/declines/gives feedback, with correct (non-duplicated) `toolsUsed`/`messageCount` deltas and zero `Conversation Message`/`Tool Used` events.

### Pre-Submission Checklist

### Documentation Updates

- [x] Yes, documentation updates are required. `PRIVACY.md` updated to state telemetry is enabled by default (opt-out) and to clarify the Cloud task-sync pipeline's scope relative to PostHog telemetry.

## Summary by CodeRabbit

* **New Features**
  * Added explicit Accept/Decline actions to the telemetry banner (closing it no longer submits a choice).
  * Telemetry consent now reacts live to VS Code’s telemetry enable/disable toggle.
  * Task completion analytics now includes completion reason plus summarized tool usage and user/assistant message counts.
* **Bug Fixes**
  * Reduced telemetry noise during error bursts with a per-event circuit breaker and safer shutdown draining.
  * Improved model-cache empty-response throttling and concurrent fetch coordination.
  * Hardened telemetry so invalid or model-supplied tool names are safely bucketed.
* **Documentation**
  * Updated privacy and in-product telemetry text to clarify what is collected and what is not (and added a cloud task sync disclosure).


src/core/tools/__tests__/editFileTool.spec.ts
```diff
@@ -560,7 +560,6 @@ describe("editFileTool", () => {
 
 			expect(mockTask.diffViewProvider.saveChanges).toHaveBeenCalled()
 			expect(mockTask.didEditFile).toBe(true)
-			expect(mockTask.recordToolUsage).toHaveBeenCalledWith("edit_file")
```
edelauna, line 563: removed since we're no longer emitting per message, but rather as a summary once the task completes

edelauna: No change needed: telemetry is intentionally emitted in task-completion summaries rather than per edit operation, so the removed assertion no longer represents the behavior.

webview-ui/src/i18n/locales/hi/settings.json
```diff
@@ -963,8 +963,8 @@
 	},
 	"footer": {
 		"telemetry": {
-			"label": "गुमनाम त्रुटि और उपयोग रिपोर्टिंग की अनुमति दें",
-			"description": "गुमनाम उपयोग डेटा और त्रुटि रिपोर्ट भेजकर Zoo Code को बेहतर बनाने में मदद करें। यह टेलीमेट्री कोड, प्रॉम्प्ट या व्यक्तिगत जानकारी एकत्र नहीं करती। अधिक विवरण के लिए हमारी <privacyLink>गोपनीयता नीति</privacyLink> देखें। आप इसे कभी भी बंद कर सकते हैं।"
+			"label": "त्रुटि और उपयोग रिपोर्टिंग की अनुमति दें",
+			"description": "आपके नाम या खाते के बजाय एक इंस्टॉल-लिंक्ड पहचानकर्ता से जुड़े उपयोग डेटा और त्रुटि रिपोर्ट भेजकर Zoo Code को बेहतर बनाने में मदद करें। यह टेलीमेट्री आपका कोड या प्रॉम्प्ट एकत्र नहीं करती। अधिक विवरण के लिए हमारी <privacyLink>गोपनीयता नीति</privacyLink> देखें। आप इसे कभी भी बंद कर सकते हैं।"
```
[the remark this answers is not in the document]
edelauna: Addressed in the current branch: the Hindi telemetry description now states that telemetry can be disabled at any time.

edelauna:
broken into a smaller PR

PR #842, opened 2026-07-06
fix(anthropic): honor custom apiModelId instead of silently defaulting to claude-sonnet-4-5

### LINKED ISSUE #418: [BUG] Azure AI Foundry (Anthropic endpoint): model fallback to claude-sonnet-4-5 ignores configured apiModelId
**Describe the bug**
When using the Anthropic provider with a custom Azure AI Foundry base URL, Zoo Code makes a preflight call to /v1/models on startup. Azure AI Foundry does not support this endpoint — it returns 404 api_not_supported. Zoo Code then falls back to its hardcoded default claude-sonnet-4-5 for all subsequent requests, completely ignoring the apiModelId set in settings. Every request fails with DeploymentNotFound for claude-sonnet-4-5.

**To Reproduce**
1. Set API Provider to Anthropic
2. Enable "Use custom base URL" → https://<your-resource>.services.ai.azure.com/anthropic
3. Enable "Pass Anthropic API Key as Authorization header instead of X-Api-Key"
4. Set Model to claude-sonnet-4-6-1 (any Azure deployment name)
5. Send any message

**Expected behavior**
Request is sent using the configured apiModelId (claude-sonnet-4-6-1).

**Actual behavior**

Zoo Code calls GET /v1/models first. Azure returns:

{"error":{"code":"api_not_supported","message":"Requested API is currently not supported"}}

Zoo Code then falls back to claude-sonnet-4-5 and the request fails:

404 {"error":{"code":"DeploymentNotFound","message":"The API deployment claude-sonnet-4-5 does not exist."}}
Evidence that the endpoint and model work correctly

Direct curl with the configured model succeeds:

curl -X POST "https://<resource>.services.ai.azure.com/anthropic/v1/messages" \
  -H "Authorization: Bearer <key>" \
  -H "Content-Type: application/json" \
  -H "anthropic-version: 2023-06-01" \
  -d '{"model":"claude-sonnet-4-6-1","max_tokens":10,"messages":[{"role":"user","content":"hi"}]}'
# Returns 200 with a valid response

**Proposed fix**

When /v1/models returns a non-200 response, fall back to the user-configured apiModelId rather than a hardcoded default. Alternatively, skip the /v1/models preflight entirely when a custom base URL is set.

**Environment**

Zoo Code version: 3.56.0
VS Code on Windows 11
API Provider: Anthropic with Azure AI Foundry custom base URL

 3.56.0

### LINKED ISSUE #843: [BUG] Custom/unrecognized Anthropic model ID gets wrong capability info, sends thinking.type=enabled and gets rejected with 400
**Describe the bug**

When using the Anthropic provider with a custom model ID that isn't a key in the built-in `anthropicModels` table (e.g. a custom deployment/proxy name for a Sonnet 5-class model), `AnthropicHandler.getModel()` looks up capabilities using `anthropicModels[anthropicDefaultModelId]` ("claude-sonnet-4-5"'s info) instead of the real underlying model's capabilities.

Since `claude-sonnet-4-5` doesn't have `supportsReasoningBinary`, the request builds `thinking: {type: "enabled", budget_tokens: N}` (the legacy shape) instead of `thinking: {type: "adaptive"}`. Real Sonnet 5 / Opus 4.7+ class models reject the legacy shape:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"\"thinking.type.enabled\" is not supported for this model. Use \"thinking.type.adaptive\" and \"output_config.effort\" to control thinking behavior."}}
```

**To Reproduce**
1. Set API Provider to Anthropic with a custom base URL
2. Set Model to a custom deployment name that maps to a Sonnet 5 / Opus 4.7+ class model (e.g. `claude-sonnet-5-bf`)
3. Enable reasoning/extended thinking
4. Send any message

**Expected behavior**
Capabilities for an unrecognized model ID should be best-effort guessed from known model-family substrings (the same pattern already used in `BedrockHandler.guessModelInfoFromId`), so the correct `thinking` shape is used.

**Related**
This is the second half of the bug reported in #418 (same `getModel()` function). A single PR fixes both: #842

grizmin:
## Summary
- `AnthropicHandler.getModel()` coerced any `apiModelId` not present in the static `anthropicModels` table down to `anthropicDefaultModelId` ("claude-sonnet-4-5") — and that coerced id is what actually got sent as `model` in the API request, silently ignoring a user-configured custom model name.
- The same fallback broke capability lookups used to build the `thinking` request param: an unrecognized id inherited the default model's (older) capability profile, so custom deployments of newer models (e.g. Sonnet 5 class) got the legacy `thinking: {type: "enabled", budget_tokens}` shape instead of `{type: "adaptive"}`, which those models reject with a 400.
- Fix: the model id sent to the API always honors a user-configured `apiModelId`. For unrecognized values, capabilities are best-effort guessed via known model-family substring matching (same pattern as the existing `BedrockHandler.guessModelInfoFromId`).

Fixes #418
Fixes #843

## Test plan
- [x] Added regression tests in `anthropic.spec.ts` covering: custom model id preserved in `getModel().id`, capability guessing via substring match, fallback to default info when no family matches, and `createMessage` sending the raw custom id with `thinking: {type: "adaptive"}`.
- [x] `pnpm --filter zoo-code test` - 50/50 passing
- [x] `pnpm --filter zoo-code check-types` - clean
- [x] `pnpm --filter zoo-code lint` - clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Custom Anthropic model IDs are now honored as provided, instead of being silently replaced with a default model.
  * When an ID isn’t recognized, model capability details used for reasoning/thinking are now determined via best-effort family matching for more reliable behavior.
  * If the ID matches no known family, the request still uses a sensible fallback model contract while preserving the chosen model ID.
* **Tests**
  * Added coverage for custom/unrecognized Anthropic model IDs and reasoning-related request parameters.


src/api/providers/anthropic.ts
```diff
@@ -353,10 +353,25 @@ export class AnthropicHandler extends BaseProvider implements SingleCompletionHa
 		}
 	}
 
+	// Guesses capabilities for an unrecognized model ID via known-family substring match.
+	private guessModelInfoFromId(modelId: string): ModelInfo {
+		const matchedId = (Object.keys(anthropicModels) as AnthropicModelId[])
+			.sort((a, b) => b.length - a.length)
```
navedmerchant: lets hoist a sorted list out of this method so we dont have to needlessly sort every time

src/api/providers/anthropic.ts
```diff
@@ -353,10 +353,25 @@ export class AnthropicHandler extends BaseProvider implements SingleCompletionHa
...
+			.find((knownId) => modelId.includes(knownId))
```
navedmerchant: we should lowercase the modelId before matching so the matching is case insensitve

navedmerchant (commented):
Thanks for this fix! Couple of small comments and should be good to merge

navedmerchant (approved):

PR #848, opened 2026-07-07
fix: Ollama provider tool result handling and premature context condensing

### LINKED ISSUE #847: [BUG] Ollama provider: premature context condensing and incorrect tool result handling
**Describe the bug**

The native Ollama provider has three issues that degrade tool-calling behavior and context management:

1. **Premature context condensing**: `parseOllamaModel` sets `maxTokens` to the full `contextWindow`. Since `getModelMaxOutputTokens` reserves 20% of the window for output, this causes the context to be condensed far earlier than necessary.
2. **Tool results sent as user messages**: `convertToOllamaMessages` always sends tool results with the `"user"` role, so the model cannot distinguish tool results from actual user messages.
3. **`additionalProperties` in tool schema**: The tool parameter schema includes `additionalProperties`, which is not part of Ollama's tool schema definition and breaks tool-calling templates on some models.

**To Reproduce**
1. Configure a native Ollama model with a large context window.
2. Use a tool-calling workflow with the model.
3. Observe premature condensing and/or malformed tool-call behavior.

**Expected behavior**
- `maxTokens` should reflect max output tokens (not the full context window), inheriting the sane default (4096).
- Tool results should use Ollama's native `"tool"` role with `tool_name` when the corresponding `tool_use` block is known.
- `additionalProperties` should be stripped from tool parameter schemas.

latest

**Additional context**
The fix tracks `tool_use` IDs to tool names so tool results can be sent with the correct role, inherits the default `maxTokens`, and strips `additionalProperties` from tool schemas.

navedmerchant:
### Related GitHub Issue

Closes: #847
Also addresses: #851

### Description

This PR fixes three issues in the native Ollama provider, plus a follow-up batch addressing review feedback on this PR and the related issue #851:

1. **Premature context condensing** — `parseOllamaModel` previously set `maxTokens` to the full `contextWindow`. Since `getModelMaxOutputTokens` reserves 20% of the window for output, this caused condensing to trigger far earlier than necessary. The fix inherits the sane default (4096) from `ollamaDefaultModelInfo` instead, so `maxTokens` correctly represents max output tokens rather than the full context window.

2. **Tool results sent as user messages** — `convertToOllamaMessages` always sent tool results with the `"user"` role, making it impossible for the model to distinguish tool results from actual user messages. The fix tracks `tool_use` IDs to tool names via a `Map`, and when a matching `tool_use` block is found, sends the result with Ollama's native `"tool"` role and `tool_name` field.

3. **`additionalProperties` in tool schema** — The tool parameter schema included `additionalProperties`, which is not part of Ollama's tool schema definition and broke tool-calling templates on some models. The fix strips `additionalProperties` from the parameters before sending tools to Ollama.

#### Review feedback (#848) and issue #851 follow-ups

4. **Concurrent model fetch single-point-of-failure (#851)** — `getOllamaModels` retrieved model details concurrently with `Promise.all` over individual `/api/show` POST requests. A single failing request (corrupt model manifest, timeout, or server overload) rejected the whole `Promise.all`, wiping out all otherwise healthy models. Each per-model promise now has a `.catch()` that logs the individual failure and resolves cleanly, so `Promise.all` completes and the remaining healthy models still load.

5. **Tool results must be text-only** — Tool results were attaching `images` to native `tool`-role messages, which Ollama's `tool` role does not support and can invalidate the request. The `toolResultImages` accumulator was also reused across iterations, so one image could leak into later tool results. Tool results are now text-only (`content` + `tool_name`); images extracted from tool results are moved onto the adjacent `user` message (the only role that supports images), and the per-result image accumulator is reset for each tool result to prevent leakage.

6. **Recursive `additionalProperties` stripping** — `delete parameters.additionalProperties` only stripped the field at the top level, leaving nested `properties`/`items` schemas leaking `additionalProperties` into Ollama payloads. A recursive `stripAdditionalProperties` helper now sanitizes the schema at every nesting level while preserving the rest of the schema shape.

### Test Procedure

- Unit tests added/updated in `src/api/providers/__tests__/native-ollama.spec.ts` covering:
  - Tool results using the `"tool"` role with `tool_name` when the `tool_use` ID is known
  - Tool results falling back to the `"user"` role when the `tool_use` ID is unknown
  - `additionalProperties` being stripped from tool parameter schemas (top-level and nested)
  - `maxTokens` inheriting the default rather than the full context window
  - Tool results staying text-only with images moved onto the adjacent user message
  - No image leakage between sibling tool results
- Unit tests added/updated in `src/api/providers/fetchers/__tests__/ollama.test.ts` covering:
  - `maxTokens` expectations aligned with the inherited default (4096)
  - Healthy models still listed when a single `/api/show` request fails (issue #851 regression test)
- Run tests with:
  - `cd src && npx vitest run api/providers/__tests__/native-ollama.spec.ts`
  - `cd src && npx vitest run api/providers/fetchers/__tests__/ollama.test.ts`

### Pre-Submission Checklist

- [x] **Issue Linked**: This PR is linked to an approved GitHub Issue (#847) and also addresses #851.
- [x] **Scope**: My changes are focused on the linked issues (one major feature/fix per PR).

### Screenshots / Videos

N/A — no UI changes.

### Documentation Updates

### Additional Notes

Also resolves the CodeRabbit review comments on this PR (maxTokens test expectations, text-only tool results / image relocation, and recursive `additionalProperties` stripping).

### Get in Touch

@navedmerchant

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tool-result handling to use the correct tool role and tool name when a matching tool call exists; unmatched tool results now follow a safe fallback.
  * Prevented image leakage across multiple tool results; images now route to the appropriate adjacent messages.
  * Removed unsupported fields from tool schemas before sending, including nested schema structures.
  * Improved model metadata so token limits behave correctly, and improved model loading so one failed model fetch won’t block others.
* **Tests**
  * Expanded coverage for tool calling, schema sanitization, and partial model-fetch failures.


src/api/providers/native-ollama.ts
```diff
@@ -193,14 +205,24 @@ export class NativeOllamaHandler extends BaseProvider implements SingleCompletio
...
-					name: tool.function.name,
-					description: tool.function.description,
-					parameters: tool.function.parameters as OllamaTool["function"]["parameters"],
-				},
-			}))
+			.map((tool) => {
+				// Strip additionalProperties from the parameters schema.
+				// This field is not part of Ollama's tool schema definition and can
+				// cause issues with some models' tool-calling templates.
+				const rawParams = tool.function.parameters as Record<string, unknown> | undefined
```
edelauna, line 299: The `delete parameters.additionalProperties` only strips the property at the top level of the parameters object. If your tools define nested properties (for example, object properties or items of type object), those nested schemas will still leak `additionalProperties: false`.

Since Ollama model templates are highly sensitive to these fields, consider recursing through the schema to sanitize it completely. You can introduce a recursive helper like:

```typescript
function stripAdditionalProperties(schema: unknown): unknown {
	if (!schema || typeof schema !== "object") return schema;
	if (Array.isArray(schema)) return schema.map(stripAdditionalProperties);
	const result: Record<string, unknown> = {};
	for (const [key, value] of Object.entries(schema)) {
		if (key === "additionalProperties") continue;
		result[key] = stripAdditionalProperties(value);
	}
	return result;
}
```

edelauna (commented):
Also noticed this pre-existing issue: https://github.com/Zoo-Code-Org/Zoo-Code/issues/851 -  could maybe include a fix in this PR since it's small enough.

src/api/providers/__tests__/native-ollama.spec.ts
```diff
@@ -1238,5 +1250,355 @@ describe("NativeOllamaHandler", () => {
...
+
+			const stream = handler.createMessage("System", messages)
+			for await (const _ of stream) {
+				// consume stream
+			}
+
+			const callArgs = mockChat.mock.calls[0][0] as any
+
+			// The tool result uses the native "tool" role and is text-only.
+			const toolMessage = callArgs.messages.find((m: any) => m.role === "tool")
```
edelauna, line 1536: The tests "should keep tool results text-only and move images on..." and "should not leak images..." do not assert on the text content of the generated tool messages. A mutation that corrupts the tool content or leaks the `(see following user message for image)` placeholder to sibling results could survive undetected.

We recommend explicitly asserting the expected `content` string inside each tool result block, for example:
```typescript
expect(toolMessage.content).toBe("screenshot\n(see following user message for image)")
```

edelauna (approved):
Looks like LM Studio suffers from the same.

PR #872, opened 2026-07-10
fix(settings): buffer Save-managed settings in cachedState until Save

### LINKED ISSUE #862: [BUG] Settings controls persist changes before Save (Save/Discard contract broken)
**Describe the bug**

Several controls inside `SettingsView` write their changes to the extension host **immediately**, instead of buffering them in the local `cachedState` and waiting for the user to click **Save**.

Because **Discard** reverts edits by resetting `cachedState` back to the currently persisted state (`setCachedState(extensionState)`), any control that already sent `updateSettings` has poisoned that source of truth — so clicking Discard can no longer undo the change. The "Save / Discard" promise is silently broken.

This affects five Save-managed values:

- `allowedCommands` (Auto Approve → Execute)
- `deniedCommands` (Auto Approve → Execute)
- `profileThresholds` (Context Management → Condensing Threshold, non-Default profile)
- `includeTaskHistoryInEnhance` (Prompts → Enhance)
- `mcpEnabled` (MCP → Enable MCP)

All five are part of the `SettingsView` Save payload (`handleSubmit`), so per the repository's Settings View Pattern they must bind to local `cachedState`, not live state.

**To Reproduce**

Steps to reproduce (using Allowed Commands as the example):

1. Open Settings and go to **Auto Approve**.
2. Enable **Execute**.
3. In the allowed-commands input, type `npm test` and click **Add**.
4. Do **not** click Save. Click **Done** and choose **Discard changes**.
5. Reopen Settings → Auto Approve → Execute.
6. See that `npm test` is still there — the discarded edit was persisted anyway.

The same pattern reproduces for each affected value:

- **Denied commands:** add/remove a denied command → Discard → entry persists.
- **Profile thresholds:** enable *Automatically condense context*, pick a **non-Default** profile, move the threshold slider → Discard → value persists. (Selecting Default moves a different value and does not reproduce this.)
- **Include task history:** toggle the checkbox in Prompts → Enhance → Discard → toggle persists.
- **MCP enabled:** toggle the top-level **Enable MCP** switch → Discard → toggle persists. (Note: per-server toggles/delete/restart are intentionally immediate and are **not** part of this bug.)

**Expected behavior**

Discarding unsaved changes leaves `allowedCommands`, `deniedCommands`, `profileThresholds`, `includeTaskHistoryInEnhance`, and `mcpEnabled` unchanged unless **Save** was clicked. These controls should follow the same Save/Discard contract as the rest of `SettingsView`:

- No `updateSettings` is posted before Save for these Save-managed settings.
- Save posts the new values once.
- Discard fully reverts the edits.

**Screenshots**

N/A (state/behavior bug; see reproduction steps).

**Video**

N/A.

3.66.0

**Additional context**

**Root cause (code locations):**

- `webview-ui/src/components/settings/SettingsView.tsx` — these values are included in the Save payload (`handleSubmit`) and Discard reverts via `setCachedState(extensionState)`.
- `webview-ui/src/components/settings/AutoApproveSettings.tsx` — command add/remove handlers posted `updateSettings` immediately.
- `webview-ui/src/components/settings/ContextManagementSettings.tsx` — the non-default profile branch posted `profileThresholds` immediately.
- `webview-ui/src/components/settings/PromptsSettings.tsx` — the checkbox posted `includeTaskHistoryInEnhance` immediately.
- `webview-ui/src/components/mcp/McpEnabledToggle.tsx` — bound to live extension state and persisted immediately.

The extension host side (`src/core/webview/webviewMessageHandler.ts`) is correct:
persisting on receipt of `updateSettings` is expected. The fix belongs on the webview side — stop sending it before Save.

**Correct precedent (not a bug):** `autoApprovalEnabled` and the per-server MCP actions (`toggleMcpServer`, `deleteMcpServer`, `restartMcpServer`, `updateMcpTimeout`) use dedicated messages, are **not** in the Save payload, and are genuinely immediate — these should stay as they are.

**Suggested fix:** route all five Save-managed values exclusively through `cachedState` / `setCachedStateField` while `SettingsView` is open (remove the immediate `updateSettings` calls). For `mcpEnabled`, which is also read by `McpView` to gate the server list and was bound to live state, adopt the existing "props first, fall back to context" pattern (as used by `PromptsSettings`) so `SettingsView` can pass the buffered value and a `cachedState`-backed setter.

**Suggested tests** (place at the lowest layer that would have failed — the `webview-ui` component/unit layer):

- `AutoApproveSettings` — adding/removing allowed & denied commands buffers via `setCachedStateField` and posts no `updateSettings`.
- `ContextManagementSettings` — changing a non-default profile threshold buffers and posts no `updateSettings`.
- `PromptsSettings` — toggling include-task-history calls the setter and posts no `updateSettings`.
- `McpEnabledToggle` — controlled usage buffers via the setter prop and posts nothing; uncontrolled usage keeps the original immediate behavior.
- `SettingsView` — update existing allowed-command tests to assert no pre-Save persistence, and add a Discard regression test.

JunyongParkDev:
### Related GitHub Issue

Closes: #862 

### Description

Several controls inside `SettingsView` wrote their changes to the extension host **immediately** via `updateSettings`, instead of buffering them in the local `cachedState` and waiting for the user to click **Save**. Because **Discard** reverts edits by resetting `cachedState` back to the persisted state (`setCachedState(extensionState)`), any control that had already posted `updateSettings` poisoned that source of truth — so Discard could no longer undo the change, silently breaking the Save/Discard contract.

This PR routes all five Save-managed values exclusively through `cachedState` / `setCachedStateField` while `SettingsView` is open:

- **`allowedCommands` / `deniedCommands`** (`AutoApproveSettings`) — drop the immediate `updateSettings` on command add/remove.
- **`profileThresholds`** (`ContextManagementSettings`) — drop the immediate `updateSettings` on the non-default profile branch; remove the now-unused `vscode` import.
- **`includeTaskHistoryInEnhance`** (`PromptsSettings`) — drop the immediate `updateSettings` on toggle.
- **`mcpEnabled`** (`McpEnabledToggle` / `McpView`) — adopt the existing "props first, fall back to context" pattern so `SettingsView` can pass the buffered value and a `cachedState`-backed setter. Because `mcpEnabled` is also read by `McpView` to gate the server list, the prop is threaded through `McpView` down to the toggle.

**Design choices / trade-offs:**

- The extension host side (`webviewMessageHandler.ts`) is correct and left untouched — persisting on receipt of `updateSettings` is expected. The fix belongs on the webview side: stop sending it before Save.
- `McpEnabledToggle` and `McpView` stay backward compatible: when used **uncontrolled** (no props), they fall back to live extension state and keep the original immediate-persist behavior. Only the `SettingsView` (controlled) path buffers.
- Genuinely-immediate actions are intentionally left unchanged because they are **not** part of the Save payload: `autoApprovalEnabled` and the per-server MCP actions (`toggleMcpServer`, `deleteMcpServer`, `restartMcpServer`, `updateMcpTimeout`).

### Test Procedure

**Automated (webview-ui component/unit layer — the lowest layer that would have failed):**

```bash
cd webview-ui
npx vitest run \
  src/components/settings/__tests__/AutoApproveSettings.spec.tsx \
  src/components/settings/__tests__/ContextManagementSettings.spec.tsx \
  src/components/settings/__tests__/PromptsSettings.spec.tsx \
  src/components/mcp/__tests__/McpEnabledToggle.spec.tsx \
  src/components/settings/__tests__/SettingsView.spec.tsx
```

Added / updated tests:

- `AutoApproveSettings` — adding/removing allowed & denied commands buffers via `setCachedStateField` and posts **no** `updateSettings`.
- `ContextManagementSettings` — changing a non-default profile threshold buffers and posts **no** `updateSettings`.
- `PromptsSettings` — toggling include-task-history calls the setter and posts **no** `updateSettings`.
- `McpEnabledToggle` — controlled usage buffers via the setter prop and posts nothing; uncontrolled usage keeps the original immediate behavior.
- `SettingsView` — existing allowed-command tests updated to assert no pre-Save persistence, plus a new Discard regression test.

**Manual (reproduces the original bug; using Allowed Commands as the example):**

1. Open Settings → **Auto Approve** → enable **Execute**.
2. Type `npm test` in the allowed-commands input and click **Add**.
3. Do **not** click Save. Click **Done** and choose **Discard changes**.
4. Reopen Settings → Auto Approve → Execute.
5. **Before:** `npm test` is still there. **After this PR:** it is gone.

The same flow reproduces for denied commands, non-default profile thresholds,
Prompts → Enhance → include task history, and MCP → Enable MCP.

### Pre-Submission Checklist

### Additional Notes

The behavior now matches the repository's Settings View Pattern: Save-managed values bind to local `cachedState`, `Save` posts them once via `handleSubmit`, and `Discard` fully reverts. No changes to the extension host or message protocol.

## Summary by CodeRabbit

* **New Features**
  * Improved MCP enablement in Settings to support buffered, controlled edits with Save/Discard behavior.
* **Bug Fixes**
  * Prevented immediate persistence for command allow/deny edits, non-default context threshold changes, and the task-history prompt toggle; these now update locally until Save.
* **Tests**
  * Added/updated Vitest coverage to verify Save/Discard buffering for MCP, Auto-Approve commands, context thresholds, prompt preferences, and discard flows.

### DIFF
--- a/webview-ui/src/components/settings/PromptsSettings.tsx
+++ b/webview-ui/src/components/settings/PromptsSettings.tsx
@@ -208,6 +208,17 @@ const PromptsSettings = ({
 										}
 
 										setIncludeTaskHistoryInEnhance(target.checked)
+
+										// Controlled: setter buffers into cachedState (Save persists).
+										// Uncontrolled: context setter is local-only, so persist here.
+										if (!propsSetIncludeTaskHistoryInEnhance) {
+											vscode.postMessage({
+												type: "updateSettings",
+												updatedSettings: {
+													includeTaskHistoryInEnhance: target.checked,
+												},
+											})
+										}
 									}}>
 									<span className="font-medium">
 										{t("prompts:supportPrompts.enhance.includeTaskHistory")}
--- a/webview-ui/src/components/settings/PromptsSettings.tsx
+++ b/webview-ui/src/components/settings/PromptsSettings.tsx
@@ -208,11 +208,6 @@ const PromptsSettings = ({
 										}
 
 										setIncludeTaskHistoryInEnhance(target.checked)
-
-										vscode.postMessage({
-											type: "updateSettings",
-											updatedSettings: { includeTaskHistoryInEnhance: target.checked },
-										})
 									}}>
 									<span className="font-medium">
 										{t("prompts:supportPrompts.enhance.includeTaskHistory")}
--- a/webview-ui/src/components/settings/__tests__/SettingsView.spec.tsx
+++ b/webview-ui/src/components/settings/__tests__/SettingsView.spec.tsx
@@ -777,9 +777,9 @@ describe("SettingsView - Duplicate Commands", () => {
 		)
 	})
 
-	it("does not persist allowed commands when discarding unsaved changes", () => {
+	it("reverts and does not persist allowed commands when discarding unsaved changes", () => {
 		// Render once and get the activateTab helper
-		const { activateTab, getSettingsContent } = renderSettingsView()
+		const { onDone, activateTab, getSettingsContent } = renderSettingsView()
 
 		// Activate the autoApprove tab
 		activateTab("autoApprove")
@@ -794,6 +794,9 @@ describe("SettingsView - Duplicate Commands", () => {
 		fireEvent.change(input, { target: { value: "npm test" } })
 		const addButton = within(content).getByTestId("add-command-button")
 		fireEvent.click(addButton)
+
+		// Verify the command was buffered and rendered before discarding it
+		expect(within(content).getByText("npm test")).toBeInTheDocument()
 
 		// Click Done, which opens the unsaved-changes dialog since a change is detected
 		const doneButton = screen.getByText("settings:common.done")
@@ -810,5 +813,9 @@ describe("SettingsView - Duplicate Commands", () => {
 				updatedSettings: expect.objectContaining({ allowedCommands: ["npm test"] }),
 			}),
 		)
+
+		// The buffered edit must be reverted before leaving Settings
+		expect(within(getSettingsContent()).queryByText("npm test")).not.toBeInTheDocument()
+		expect(onDone).toHaveBeenCalledTimes(1)
 	})
 })
--- a/webview-ui/src/components/settings/__tests__/SettingsView.spec.tsx
+++ b/webview-ui/src/components/settings/__tests__/SettingsView.spec.tsx
@@ -629,13 +629,13 @@ describe("SettingsView - Allowed Commands", () => {
 		// Verify command was added
 		expect(within(content).getByText("npm test")).toBeInTheDocument()
 
-		// Verify VSCode message was sent
-		expect(vscode.postMessage).toHaveBeenCalledWith({
-			type: "updateSettings",
-			updatedSettings: {
-				allowedCommands: ["npm test"],
-			},
-		})
+		// Adding a command must NOT persist before Save; it only buffers in cachedState.
+		expect(vscode.postMessage).not.toHaveBeenCalledWith(
+			expect.objectContaining({
+				type: "updateSettings",
+				updatedSettings: expect.objectContaining({ allowedCommands: ["npm test"] }),
+			}),
+		)
 	})
 
 	it("removes command from the list", () => {
@@ -663,13 +663,13 @@ describe("SettingsView - Allowed Commands", () => {
 		// Verify command was removed
 		expect(within(content).queryByText("npm test")).not.toBeInTheDocument()
 
-		// Verify VSCode message was sent
-		expect(vscode.postMessage).toHaveBeenLastCalledWith({
-			type: "updateSettings",
-			updatedSettings: {
-				allowedCommands: [],
-			},
-		})
+		// Removing a command must NOT persist before Save; it only buffers in cachedState.
+		expect(vscode.postMessage).not.toHaveBeenCalledWith(
+			expect.objectContaining({
+				type: "updateSettings",
+				updatedSettings: expect.objectContaining({ allowedCommands: expect.anything() }),
+			}),
+		)
 	})
 
 	describe("SettingsView - Tab Navigation", () => {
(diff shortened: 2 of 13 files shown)


webview-ui/src/components/settings/__tests__/SettingsView.spec.tsx
```diff
@@ -776,4 +776,39 @@ describe("SettingsView - Duplicate Commands", () => {
...
+		fireEvent.click(discardButton)
+
+		// Discarding must never persist the buffered command edit
+		expect(vscode.postMessage).not.toHaveBeenCalledWith(
+			expect.objectContaining({
+				type: "updateSettings",
+				updatedSettings: expect.objectContaining({ allowedCommands: ["npm test"] }),
+			}),
+		)
+	})
```
edelauna, line 820: This only asserts that `postMessage` was not called with the buffered value — it never checks that the UI actually reverted. If the Discard handler stopped resetting `cachedState` (e.g. the `setCachedState(extensionState)` call got dropped), this test would still pass, because the buffered command is never posted during the discard click itself. Worth also asserting the command is gone from the UI after discard?

```suggestion
		// Discarding must never persist the buffered command edit
		expect(vscode.postMessage).not.toHaveBeenCalledWith(
			expect.objectContaining({
				type: "updateSettings",
				updatedSettings: expect.objectContaining({ allowedCommands: ["npm test"] }),
			}),
		)

		// ...and the buffered edit must actually revert in the UI
		expect(within(getSettingsContent()).queryByText("npm test")).not.toBeInTheDocument()
	})
```

edelauna (changes requested):
looks good - thanks for this commit had a minor test coverage comment

webview-ui/src/components/settings/__tests__/SettingsView.spec.tsx
JunyongParkDev: Thanks for the thorough review and for catching this gap! I’ve strengthened the test to verify that the command is present before Discard, removed from the UI afterward, and that `onDone` is called. This now covers both non-persistence and the actual `cachedState` rollback behavior.

webview-ui/src/components/settings/PromptsSettings.tsx
```diff
@@ -208,11 +208,6 @@ const PromptsSettings = ({
 										}
 
 										setIncludeTaskHistoryInEnhance(target.checked)
-
-										vscode.postMessage({
```
edelauna, line 212: When rendered uncontrolled (when `propsSetIncludeTaskHistoryInEnhance` is omitted), removing `vscode.postMessage` here causes toggling `includeTaskHistoryInEnhance` to update local React context state only, without sending `updateSettings` to the VS Code extension host.

edelauna (changes requested):
Noticed something related to the design, even though there are no uncontrolled references in the code base for `PromptsSettings`, can we handle the uncontrolled case since it was included in the design.

JunyongParkDev:
> Noticed something related to the design, even though there are no uncontrolled references in the code base for `PromptsSettings`, can we handle the uncontrolled case since it was included in the design.

Good point. I appreciate the review. The component is designed for uncontrolled use (props-first, context fallback), so I've handled that path instead of letting it silently break.

webview-ui/src/components/settings/PromptsSettings.tsx
JunyongParkDev: Thanks, fixed. I branch on the setter prop (`!propsSetIncludeTaskHistoryInEnhance`, since `false` is a valid value): the controlled path buffers into cachedState so Save/Discard still works, while the uncontrolled path restores the immediate `postMessage` — consistent with `McpEnabledToggle`. Added tests for both paths.

JunyongParkDev: Thanks, fixed. I branch on the setter prop (`!propsSetIncludeTaskHistoryInEnhance`, since `false` is a valid value): the controlled path buffers into cachedState so Save/Discard still works, while the uncontrolled path restores the immediate `postMessage` — consistent with `McpEnabledToggle`. Added tests for both paths.

edelauna (approved):
Thanks so much for working through the feedback!

PR #878, opened 2026-07-11
fix(ollama): always respond to model refresh, surface errors, and use form-edited base URL

### LINKED ISSUE #877: [BUG] Ollama model refresh stuck in loading, silently fails on errors, and uses stale base URL
**Describe the bug**

When using the Ollama provider in the settings panel, the "Refresh Models" flow has several problems:

1. **Stuck loading state**: The `requestOllamaModels` handler in the extension only posts an `ollamaModels` message back to the webview when models are *found* (`Object.keys(ollamaModels).length > 0`). When no models are returned — or when the fetch throws an error — no response is ever sent, so the webview stays stuck in a perpetual "loading" state with no feedback to the user.

2. **Silent failures on fetch errors**: When the Ollama fetch fails (e.g. Ollama isn't running, wrong base URL, connection refused), the handler catches the error and silently logs it with `console.debug`. The user sees no error message and has no idea why the refresh produced no models.

3. **Stale base URL used for refresh**: The refresh always uses the *saved* `ollamaBaseUrl` / `ollamaApiKey` from extension state, ignoring any unsaved edits the user has made in the settings form. If a user types a new base URL and clicks refresh *before* saving, the refresh queries the old URL — producing confusing results.

4. **No dedicated refresh button / status feedback**: The Ollama settings panel has no visible refresh button with loading, success, or error status indicators, so users cannot tell whether a refresh is in progress, succeeded, or failed.

**To Reproduce**

1. Open the Zoo Code settings panel and select the Ollama provider.
2. Enter a base URL for an Ollama instance that is **not running** (or enter an incorrect URL).
3. Trigger a model refresh (the panel refreshes models on mount / provider switch).
4. Observe that no models appear and there is no error message or loading indicator — the panel gives no feedback at all.
5. Edit the base URL to a correct value but click refresh **before** saving the setting.
6. Observe that the refresh still queries the old (saved) URL, not the one currently typed in the form.

**Expected behavior**

- The refresh should **always** post a response back to the webview, even when no models are found or the fetch errors, so the UI can transition out of the loading state.
- Fetch errors should be **logged to the output channel** and the error message should be **surfaced in the settings panel** so the user understands why no models were returned.
- The refresh should use the **base URL and API key currently entered in the form** (unsaved edits), not the stale saved state, so the user can validate a URL before committing it.
- The Ollama panel should have a **dedicated refresh button** with visible loading / success / error status feedback.

**Screenshots**

N/A — the bug is about missing feedback and stale state, not a visual rendering issue.

**Video**

N/A

3.66.0

**Additional context**

The root cause is in the `requestOllamaModels` case of [`webviewMessageHandler.ts`](src/core/webview/webviewMessageHandler.ts) and the [`Ollama.tsx`](webview-ui/src/components/settings/providers/Ollama.tsx) settings component. The handler silently swallows errors and conditionally suppresses empty responses, while the component relies on `useEvent` from `react-use` with no refresh status state machine.

navedmerchant:
### Related GitHub Issue

Closes: #877

### Description

This PR fixes the Ollama model refresh flow so the settings panel always gives the user feedback and uses the URL they're actually looking at. The key implementation details:

**Extension handler ([`webviewMessageHandler.ts`](src/core/webview/webviewMessageHandler.ts))**
- The `requestOllamaModels` case now prefers `baseUrl` / `apiKey` from the incoming message `values` (which reflect the user's unsaved form edits) over the saved extension state, falling back to state only when values are absent. This lets a user validate a URL *before* saving it.
- The handler **always** posts an `ollamaModels` response back to the webview — even when no models are found — so the UI can transition out of the loading state instead of hanging forever.
- On fetch failure, the error is now logged to the output channel via `provider.log(...)` and an empty `ollamaModels` response is posted **with the error message included**, so the webview can display it directly. The previous `console.debug` silent swallow is removed.

**Webview ([`ApiOptions.tsx`](webview-ui/src/components/settings/ApiOptions.tsx))**
- The `requestOllamaModels` message dispatched on provider switch now includes `values: { baseUrl, apiKey }` from the current `apiConfiguration`, matching the new handler contract.

**Ollama settings component ([`Ollama.tsx`](webview-ui/src/components/settings/providers/Ollama.tsx))**
- Replaced the `useEvent` hook from `react-use` with a native `window.addEventListener("message", ...)` inside a `useEffect` (with proper cleanup), removing the `react-use` dependency from this component.
- Added a `refreshStatus` state machine (`idle` → `loading` → `success` | `error`) and a `refreshError` state to hold the backend error message.
- Added a dedicated **Refresh Models** button (using the shared `Button` component) that shows a spinner while loading, disables itself during refresh, and displays success / error status text below the button. The error text prefers the backend-provided message and falls back to a localized generic error string.
- The message listener only transitions status when the component is in the `loading` state, so unsolicited `ollamaModels` messages (e.g. from mount refresh) don't clobber the status display.

### Test Procedure

**Unit tests — extension handler** (`src/core/webview/__tests__/webviewMessageHandler.spec.ts`)
- `cd src && npx vitest run core/webview/__tests__/webviewMessageHandler.spec.ts`
- New cases verify: empty models response is posted when no models found; error response with message is posted and logged on fetch failure; `baseUrl`/`apiKey` from message `values` take precedence over saved state.

**Unit tests — Ollama component** (`webview-ui/src/components/settings/providers/__tests__/Ollama.spec.tsx`)
- `cd webview-ui && npx vitest run src/components/settings/providers/__tests__/Ollama.spec.tsx`
- New `describe("Ollama Component - refresh models")` block verifies: idle button render, correct `requestOllamaModels` payload with `values` on click, loading/disabled state, success state on models arrival, error state on empty models, backend error message display, and no-op for unsolicited messages when not loading.

**Manual testing**
1. Select the Ollama provider in settings with Ollama **not running**.
2. Click "Refresh Models" — the button shows a spinner, then displays the error message (e.g. "Connection refused") in red.
3. Start Ollama, click refresh again — the button shows a success status and the model picker populates.
4. Edit the base URL field but click refresh **before** saving — the refresh now queries the newly typed URL, not the stale saved one.

### Pre-Submission Checklist

- [x] **Issue Linked**: This PR is linked to approved GitHub Issue #877.
- [x] **Scope**: My changes are focused on the linked issue (Ollama refresh UX).
- [x] **Testing**: New and updated tests have been added covering the handler and component changes.
- [x] **Documentation Impact**: I have considered documentation updates (see below).

### Screenshots / Videos

N/A — the refresh button and status text use existing shared UI primitives (`Button`, codicon icons, VS Code CSS variables) with no new visual design.

### Documentation Updates

- [x] No documentation updates are required. The refresh button is self-explanatory and uses existing localized strings (`settings:providers.refreshModels.*`).

### Additional Notes

The `react-use` `useEvent` import was removed from [`Ollama.tsx`](webview-ui/src/components/settings/providers/Ollama.tsx) in favor of a native `useEffect`-based listener. This is a minor cleanup that reduces the component's dependency surface and makes the listener lifecycle explicit. The corresponding test mock for `react-use` was also removed.

### Get in Touch

navedmerchant (GitHub) / navedmerchant on Discord

## Summary by CodeRabbit

* **New Features**
  * Added an Ollama “Refresh models” button in provider settings with loading, success, and error feedback.
  * Refresh requests now send the current Ollama base URL and API key.
* **Bug Fixes**
  * Model refresh now always returns an `ollamaModels` payload to the UI, including empty results and explicit error details.
  * Refresh fails during cache refresh or model retrieval now surface clear error messages.
* **Tests**
  * Expanded test coverage for empty results, request failures, credential precedence, and ignoring unsolicited updates.


taltas:
@roomote Please review this PR

webview-ui/src/components/settings/providers/Ollama.tsx
```diff
@@ -35,20 +37,42 @@ export const Ollama = ({ apiConfiguration, setApiConfigurationField }: OllamaPro
...
-		switch (message.type) {
-			case "ollamaModels":
-				{
-					const newModels = message.ollamaModels ?? {}
-					setOllamaModels(newModels)
+				if (refreshStatus === "loading") {
+					if (Object.keys(newModels).length > 0) {
+						setRefreshStatus("success")
+					} else {
+						setRefreshStatus("error")
```
edelauna: When Ollama is running but no models are installed, the backend posts `{ ollamaModels: {} }` with no `error` field — so `message.error` is `undefined` here and the component shows the generic error string. Is a healthy-but-empty Ollama server really the same as a fetch failure?

```suggestion
				} else if (message.error) {
					setRefreshStatus("error")
					setRefreshError(message.error)
				} else {
					// Ollama reachable but no models installed yet
					setRefreshStatus("success")
```

webview-ui/src/components/settings/providers/Ollama.tsx
```diff
@@ -35,20 +37,42 @@ export const Ollama = ({ apiConfiguration, setApiConfigurationField }: OllamaPro
...
+			}
 		}
-	}, [])
 
-	useEvent("message", onMessage)
+		window.addEventListener("message", handleMessage)
+		return () => {
+			window.removeEventListener("message", handleMessage)
+		}
+	}, [refreshStatus])
```
edelauna: The `[refreshStatus]` dep tears down and re-adds the listener on the idle→loading transition — which fires synchronously right before `vscode.postMessage(...)` at line 66. If Ollama responds very quickly, could the response arrive in the brief gap and be silently dropped, leaving the UI stuck in loading?

A `useRef` for `refreshStatus` with a stable `[]` dep would avoid the re-subscription.

webview-ui/src/components/settings/ApiOptions.tsx
```diff
@@ -218,7 +218,13 @@ const ApiOptions = ({
 					},
 				})
 			} else if (selectedProvider === "ollama") {
-				vscode.postMessage({ type: "requestOllamaModels" })
+				vscode.postMessage({
+					type: "requestOllamaModels",
+					values: {
+						baseUrl: apiConfiguration?.ollamaBaseUrl,
+						apiKey: apiConfiguration?.ollamaApiKey,
```
edelauna, line 226: `ollamaApiKey` is now forwarded in `values`, but it’s not in the `useDebounce` dep array (further down around the dep list). Should it be, so a key-only change also triggers a re-fetch?

src/core/webview/webviewMessageHandler.ts
```diff
@@ -1203,23 +1203,38 @@ export const webviewMessageHandler = async (
...
+				provider.postMessageToWebview({ type: "ollamaModels", ollamaModels: ollamaModels })
 			} catch (error) {
-				// Silently fail - user hasn't configured Ollama yet
-				console.debug("Ollama models fetch failed:", error)
+				// Log the error to the output channel for debugging, but still
+				// post an empty response so the webview doesn't stay stuck in
+				// the loading state. Include the error message so the webview
+				// can display it directly in the settings panel.
+				const errorMsg = error instanceof Error ? error.message : String(error)
+				provider.log(`[requestOllamaModels] Failed to fetch models: ${errorMsg}`)
```
edelauna: When the user has typed a new URL but not saved, this log won’t show which host was actually attempted. Worth including `baseUrl`?

```suggestion
				provider.log(`[requestOllamaModels] Failed to fetch models from ${baseUrl}: ${errorMsg}`)
```

src/core/webview/webviewMessageHandler.ts
```diff
@@ -1203,23 +1203,38 @@ export const webviewMessageHandler = async (
...
 
 				const ollamaModels = await getModels(ollamaOptions)
 
-				if (Object.keys(ollamaModels).length > 0) {
-					provider.postMessageToWebview({ type: "ollamaModels", ollamaModels: ollamaModels })
-				}
+				// Always post a response so the webview refresh status can
+				// transition out of "loading" — even when no models are found.
+				provider.postMessageToWebview({ type: "ollamaModels", ollamaModels: ollamaModels })
 			} catch (error) {
```
edelauna, line 1258: Both `flushModels` and `getModels` can throw into this catch, but the log says "Failed to fetch models" in both cases. A cache-layer failure and a network failure would look identical in the output channel — worth distinguishing?

src/core/webview/__tests__/webviewMessageHandler.spec.ts
```diff
@@ -331,6 +331,70 @@ describe("webviewMessageHandler - requestOllamaModels", () => {
...
+		})
+
+		// Should use the URL from message values, not the saved state
+		expect(mockGetModels).toHaveBeenCalledWith({
+			provider: "ollama",
+			baseUrl: "https://ollama.example.com",
+			apiKey: "secret-key",
+		})
+
+		expect(mockClineProvider.postMessageToWebview).toHaveBeenCalledWith({
```
edelauna, line 423: `mockGetModels` is asserted with the right options, but `mockFlushModels` isn’t checked here. If `flushModels` used the stale saved-state URL while `getModels` used the form-edited one, this test would still pass. Worth adding:
```
expect(mockFlushModels).toHaveBeenCalledWith(
  { provider: "ollama", baseUrl: "https://ollama.example.com", apiKey: "secret-key" },
  true
)
```

edelauna (changes requested):

edelauna (approved):

PR #880, opened 2026-07-11
fix(terminal): tell model when command was aborted by the user

### LINKED ISSUE #833: [BUG] Zoo Code does not tell model when command Abort button was pressed
When the user clicks the `Abort` button to abort a running command, the model only sees a `SIGKILL` exit code and thus suspects e.g. the Linux OOM-killer, which leads it to wrong decisions.

Image

**Expected behavior**

Zoo Code should include in the message that communicates the command exit code to the model whether the user clicked `Abort.`

**Screenshots**

Image

3.66

umi008:
### Related GitHub Issue

Closes: #833

### Description

When the user clicked the Abort button on a running shell command, `ExecaTerminalProcess` emitted `shell_execution_complete` with `{ exitCode: 0 }` regardless of whether the process completed normally or was killed by the user. The model received no indication that the user intervened — it would see either a successful exit or SIGKILL and could suspect Linux OOM-killer or other system events, leading to incorrect diagnosis and wrong follow-up actions.

**Fix**: 
1. Added optional `aborted` field to `ExitCodeDetails` interface
2. `ExecaTerminalProcess.run()` emits `{ exitCode: 0, aborted: this.aborted }` so the flag reflects whether `abort()` was called
3. `formatExitStatus()` checks `aborted` first and returns `"Command was aborted by the user."` — clear, unambiguous, and takes priority over exit code or signal info

### Test Procedure

1. Start Zoo Code in any mode that can run terminal commands
2. Send a long-running command (e.g., `sleep 60`)
3. Click the Abort button in the UI while the command is running
4. Verify the model receives: `Command executed in terminal... Command was aborted by the user.`
5. Verify the model does NOT speculate about OOM-killer, SIGKILL, or other system events

### Pre-Submission Checklist

- [x] **Issue Linked**: Closes #833
- [x] **Scope**: One focused fix — communicates user abort to the model
- [x] **Self-Review**: Backward-compatible; `aborted` is optional so all existing callers work unchanged
- [x] **Testing**: TypeScript compilation verified; abort path in ExecaTerminalProcess tested end-to-end

### Screenshots / Videos

N/A — backend message formatting, no UI changes.

### Documentation Updates

### Additional Notes

The VSCode terminal path (`TerminalProcess`) sends Ctrl+C instead of SIGKILL when aborting, which correctly produces SIGINT in the exit details — that path was not affected by this bug. The fix targets the execa-based fallback path specifically.

## Summary by CodeRabbit

* **Bug Fixes**
  * Show clearer messaging when a command is intentionally aborted by the user.
  * Improve shell command completion reporting by consistently including an `aborted` flag in completion event payloads.
  * Refine exit-status text formatting for aborted vs non-aborted outcomes.
* **Tests**
  * Expanded terminal process tests to verify `aborted` on success and added abort/error scenarios.
  * Added unit coverage for exit-status formatting across aborted, signal, and undefined inputs.


src/integrations/terminal/types.ts
```diff
@@ -72,4 +72,10 @@ export interface ExitCodeDetails {
 	signal?: number | undefined
 	signalName?: string
 	coreDumpPossible?: boolean
+	/**
+	 * Set to true when the user explicitly pressed the Abort button
```
edelauna: Should this wording be a little less user-specific? `abort()` is also used by timeout and task teardown paths, so this flag seems to mean the process was intentionally aborted rather than necessarily that the user clicked Abort.

src/integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts
```diff
@@ -129,7 +129,7 @@ describe("ExecaTerminalProcess", () => {
 			const spy = vitest.fn()
 			terminalProcess.on("shell_execution_complete", spy)
 			await terminalProcess.run("echo test")
-			expect(spy).toHaveBeenCalledWith({ exitCode: 0 })
+			expect(spy).toHaveBeenCalledWith({ exitCode: 0, aborted: false })
```
edelauna, line 132: Could we add one regression case for the actual abort path? This assertion covers normal completion with `aborted: false`, but it would not catch dropping the `aborted: true` propagation or the model-facing `Command was aborted by the user.` formatting branch.

edelauna (changes requested):
Nice! Thanks for taking this on, this has bugged me for a while.

Since we're touching these files, can we also increase coverage so that the codecov plugin is happy as well.

src/integrations/terminal/types.ts
umi008: updated the docstring to say "intentionally aborted (via abort(), timeout, or task teardown)" instead of "user explicitly pressed the Abort button" to reflect all abort paths

src/integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts
umi008: added: new test "should emit aborted: true when abort() is called during execution" covers the actual abort path — calls abort() mid-execution and verifies shell_execution_complete includes aborted: true. the error-path tests (ExecaError and generic) assert aborted: false. together these cover all emission paths.

umi008:
addressed all review feedback from @edelauna:

1. **coverage increased** — added 7 new tests:
   - abort path test: calls abort() mid-execution, verifies aborted: true emission
   - 2 error-path tests: ExecaError + generic error both assert aborted: false
   - 5 formatExitStatus unit tests: aborted=true msg, normal exit, signal, undefined
   - codecov patch coverage is now 100%

2. **wording fix** — docstring updated to reflect abort() is also called by timeout/teardown, not just user button press

both threads replied inline

PR #881, opened 2026-07-11
fix(webview): anchor thinking timer to message timestamp to survive Virtuoso remounts

### LINKED ISSUE #656: [BUG] Timer of Thinking will reset to 0 after expand and collapse it

The timer for the thinking section will reset to 0 after expanding and collapsing it.

N/A

### Reproduction steps

1. Have a conversation with long thinking and wait for several seconds
2. Expand the thinking section 
3. Collapse the thinking section
4. You will notice the timer for the thinking is reset to 0, rather than its original timer.

### Expected result

Timer goes as usual

### Actual result

Timer is reset to 0

_No response_

### App Version

3.60.0

### API Provider (optional)

None

### Model Used (optional)

_No response_

_No response_

```shell

```

umi008:
### Related GitHub Issue

Closes: #656

### Description

The thinking/reasoning timer (e.g., "Thought for 30s") was resetting to 0 or disappearing when the user expanded and then collapsed the thinking section. This happened because `ReasoningBlock` used `Date.now()` as the timer anchor at component mount time — but inside a `Virtuoso` virtual list, expand/collapse can trigger DOM recycling that unmounts and remounts the component. On remount, `startTimeRef` gets a fresh `Date.now()`, and since `isStreaming` is already `false` (the message finished streaming before the user interacted), the timer effect never runs — leaving `elapsed` at 0.

This was more visible with zh_CN locale because CJK character widths in the header text produced different DOM geometry, pushing the layout past Virtuoso recycle thresholds in viewports where Latin text would not trigger a remount.

**Fix**: The component now uses the message creation timestamp (`ts` prop, already passed from `ChatRow.tsx` but previously discarded in the destructuring) as the timer anchor. On remount after streaming has finished, `elapsed` is initialized from `Date.now() - ts`, giving a reasonable approximation of the thinking duration instead of 0.

### Test Procedure

1. Start a chat with any provider that supports thinking/reasoning (Claude with extended thinking, DeepSeek, etc.)
2. Wait for the thinking section to appear with a timer (e.g., "30s")
3. Expand the thinking section, then collapse it
4. Verify the timer still shows the original duration — not 0s or hidden
5. Repeat with different locales (en, zh_CN) to confirm the fix is locale-independent

### Pre-Submission Checklist

- [x] **Issue Linked**: Closes #656
- [x] **Scope**: One focused fix — anchors timer to stable message data
- [x] **Self-Review**: Props interface already had `ts`; component just wasn't using it — this is a pure hook into existing data
- [x] **Testing**: TypeScript compilation verified; tested the remount scenario with the init function fallback

### Screenshots / Videos

N/A — state management fix, no visual change beyond the timer not resetting.

### Documentation Updates

### Additional Notes

Edge case: for messages scrolled away for minutes and then scrolled back, `Date.now() - ts` will show longer than the actual thinking time. This is cosmetic and far preferable to showing 0 or hiding the timer entirely — and only affects the unlikely case of a late remount after significant wall-clock time.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved elapsed-time display for streaming and non-streaming reasoning by basing the timer on the message’s creation time rather than when the UI mounts.
  * Elapsed-time now starts at 0 while streaming, stays consistent if the view is reopened for the same message, and stops updating when streaming ends.
* **Tests**
  * Added automated coverage for elapsed-time rendering, streaming updates, caching across remounts, and collapsed/zero-duration display behavior.

### DIFF
--- a/webview-ui/src/components/chat/ReasoningBlock.tsx
+++ b/webview-ui/src/components/chat/ReasoningBlock.tsx
@@ -14,14 +14,19 @@ interface ReasoningBlockProps {
 	metadata?: any
 }
 
-export const ReasoningBlock = ({ content, isStreaming, isLast }: ReasoningBlockProps) => {
+export const ReasoningBlock = ({ content, isStreaming, isLast, ts }: ReasoningBlockProps) => {
 	const { t } = useTranslation()
 	const { reasoningBlockCollapsed } = useExtensionState()
 
 	const [isCollapsed, setIsCollapsed] = useState(reasoningBlockCollapsed)
 
-	const startTimeRef = useRef<number>(Date.now())
-	const [elapsed, setElapsed] = useState<number>(0)
+	// Anchor the elapsed timer to the message creation timestamp (ts)
+	// rather than component mount time. When Virtuoso recycles or
+	// remounts this component (e.g. during expand/collapse in a
+	// virtualized list), the timer survives because ts is a stable
+	// prop from the message data rather than a fresh Date.now().
+	const startTimeRef = useRef<number>(ts)
+	const [elapsed, setElapsed] = useState<number>(() => (isStreaming ? 0 : Math.max(0, Date.now() - ts)))
 	const contentRef = useRef<HTMLDivElement>(null)
 
 	useEffect(() => {
--- a/webview-ui/src/components/chat/ReasoningBlock.tsx
+++ b/webview-ui/src/components/chat/ReasoningBlock.tsx
@@ -14,6 +14,16 @@ interface ReasoningBlockProps {
 	metadata?: any
 }
 
+/**
+ * Module-level cache that persists elapsed reasoning durations across
+ * React component remounts. When Virtuoso recycles a ReasoningBlock
+ * (e.g., during expand/collapse or scroll), the cache provides the
+ * final elapsed value from the previous mount cycle instead of
+ * recomputing from Date.now() - ts, which would be wrong if minutes
+ * have passed since the message was created.
+ */
+const elapsedCache = new Map<number, number>()
+
 export const ReasoningBlock = ({ content, isStreaming, isLast, ts }: ReasoningBlockProps) => {
 	const { t } = useTranslation()
 	const { reasoningBlockCollapsed } = useExtensionState()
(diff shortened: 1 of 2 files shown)


webview-ui/src/components/chat/ReasoningBlock.tsx
```diff
@@ -14,14 +14,19 @@ interface ReasoningBlockProps {
...
 
-	const startTimeRef = useRef<number>(Date.now())
-	const [elapsed, setElapsed] = useState<number>(0)
+	// Anchor the elapsed timer to the message creation timestamp (ts)
+	// rather than component mount time. When Virtuoso recycles or
+	// remounts this component (e.g. during expand/collapse in a
+	// virtualized list), the timer survives because ts is a stable
+	// prop from the message data rather than a fresh Date.now().
+	const startTimeRef = useRef<number>(ts)
+	const [elapsed, setElapsed] = useState<number>(() => (isStreaming ? 0 : Math.max(0, Date.now() - ts)))
```
edelauna: If a completed reasoning block remounts minutes later or history is reopened, won’t this recompute elapsed from message creation to now even though streaming already stopped? Is there a persisted finish time or duration we can anchor to instead?

edelauna (changes requested):
Thanks for your contribution, had a question of the state mechanics around keeping track of the timer value.

webview-ui/src/components/chat/ReasoningBlock.tsx
umi008: good catch. the fix uses a module-level Map<number, number> keyed by message timestamp that caches the final elapsed duration when streaming stops. since module state survives react component unmount/remount cycles, late remounts (minutes later, history reopened) now get the cached correct duration instead of recomputing Date.now() - ts. the Map is bounded by the number of reasoning blocks in a session — no leak in practice.

umi008:
addressed @edelauna review feedback:

added a module-level elapsedCache (Map<number, number>) that persists the final thinking duration across React remounts. on late remounts (scrolled away for minutes, history reopened), the cached value is returned instead of recomputing Date.now() - ts, which would give an inflated elapsed time.

commit f957b59b0 on fix/656-reasoning-timer-reset

PR #886, opened 2026-07-13
feat(friendli): add reasoning params support

Lee-Si-Yoon:
## Summary

Add Friendli-specific reasoning parameter support for GLM-5.x and MiniMax-M2.5 models served via Friendli Model APIs.

Friendli extends the OpenAI Chat Completions API with non-standard reasoning fields (`reasoning_effort`, `chat_template_kwargs`, `parse_reasoning`, `include_reasoning`) that the base `BaseOpenAiCompatibleProvider` does not handle. This PR overrides `createStream` and `completePrompt` in `FriendliHandler` to inject these parameters.

**Ref**: https://friendli.ai/docs/guides/reasoning

### Model catalog (`packages/types/src/providers/friendli.ts`)
- GLM-5.2 / GLM-5.1: add `supportsReasoningEffort: ["minimal","low","medium","high","xhigh","max"]` + `reasoningEffort: "high"` (controllable reasoning models)
- MiniMax-M2.5: add `supportsReasoningBinary: true` (always-reasoning model)

### Handler (`src/api/providers/friendli.ts`)
- Override `createStream` and `completePrompt` to inject Friendli reasoning params via `buildFriendliReasoningParams()`
- **Controllable reasoning models (GLM-5.x)**: send `reasoning_effort` + `chat_template_kwargs: { enable_thinking: true }` + `parse_reasoning`/`include_reasoning` when user selects a valid effort
- **Always-reasoning models (MiniMax-M2.5)**: send `parse_reasoning`/`include_reasoning` only
- **Non-reasoning models (DeepSeek-V3.2)**: no extra params — reasoning_effort silently ignored
- When user sets reasoning_effort to `"none"` or `"disable"`, or `enableReasoningEffort=false`, no Friendli reasoning params are sent — the model runs at its default behavior
- `delta.reasoning_content` streaming already handled by `extractReasoningFromDelta` in base class

### Tests (`src/api/providers/__tests__/friendli.spec.ts`)
- 8 reasoning-specific tests added (29 total pass):
  - GLM-5.2 with reasoning enabled → full params sent
  - enableReasoningEffort=false → no params (model default)
  - reasoningEffort="none" → no params (model default)
  - reasoningEffort="disable" → no params (model default)
  - MiniMax-M2.5 always-reasoning → parse_reasoning/include_reasoning only
  - DeepSeek-V3.2 non-reasoning → no params
  - delta.reasoning_content stream parsing
  - completePrompt reasoning params

## Design decisions
- **No `enable_thinking: false`**: When reasoning is disabled (none/disable), we send nothing rather than explicitly disabling thinking. This lets the model use its default behavior.
- **No `ultracode`**: Friendli API supports `ultracode` in the enum but we do not surface it — only the standard efforts up to `max`.
- **No `preserveReasoning`**: Friendli only supports stateless responses API — multi-turn reasoning re-submission has no effect.
- **Non-reasoning models**: If reasoning_effort is set for a model that does not support it, params are silently omitted (no error).

## Test plan
- [x] `pnpm vitest run api/providers/__tests__/friendli.spec.ts` — 29/29 pass
- [x] `pnpm run build` — success
- [x] `pnpm run lint` (src + packages/types) — clean

## Summary by CodeRabbit

- **New Features**
  - Added reasoning-effort controls for supported Friendli GLM models, including per-model allowed effort levels and a default “high” setting for GLM-5.1/5.2.
  - Friendli requests now apply reasoning settings consistently for both streaming and non-streaming, including streamed reasoning output when enabled.
- **Bug Fixes**
  - Stopped sending reasoning-related parameters to non-reasoning models.
  - Properly disables/omits reasoning fields when reasoning is turned off.
- **Tests**
  - Expanded coverage for reasoning parameter inclusion/exclusion and streamed reasoning chunk emission.

### DIFF
--- a/packages/types/src/providers/friendli.ts
+++ b/packages/types/src/providers/friendli.ts
@@ -20,6 +20,8 @@ export const friendliModels = {
 		outputPrice: 4.4,
 		cacheWritesPrice: 0,
 		cacheReadsPrice: 0.26,
+		supportsReasoningEffort: ["minimal", "low", "medium", "high", "xhigh", "max"],
+		reasoningEffort: "high",
 		description:
 			"GLM-5.2 is Zhipu's flagship model with a 1M context window and 128k max output, served via Friendli Model APIs. It delivers top-tier long-context reasoning, coding, and agentic performance for extended engineering sessions.",
 	},
@@ -33,6 +35,8 @@ export const friendliModels = {
 		outputPrice: 4.4,
 		cacheWritesPrice: 0,
 		cacheReadsPrice: 0.26,
+		supportsReasoningEffort: ["minimal", "low", "medium", "high", "xhigh", "max"],
+		reasoningEffort: "high",
 		description:
 			"GLM-5.1 is Zhipu's most capable model with a 200k context window and 128k max output, served via Friendli Model APIs. It delivers top-tier reasoning, coding, and agentic performance.",
 	},
--- a/src/api/providers/__tests__/friendli.spec.ts
+++ b/src/api/providers/__tests__/friendli.spec.ts
@@ -142,7 +142,16 @@ describe("FriendliHandler", () => {
 		},
 	])(
 		"should expose newly added model $modelId",
-		({ modelId, contextWindow, maxTokens, supportsMaxTokens, inputPrice, outputPrice, cacheWritesPrice, cacheReadsPrice }) => {
+		({
+			modelId,
+			contextWindow,
+			maxTokens,
+			supportsMaxTokens,
+			inputPrice,
+			outputPrice,
+			cacheWritesPrice,
+			cacheReadsPrice,
+		}) => {
 			expect(friendliModels[modelId]).toBeDefined()
 			const info = friendliModels[modelId] as import("@roo-code/types").ModelInfo
 			expect(info.maxTokens).toBe(maxTokens)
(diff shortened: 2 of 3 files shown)


packages/types/src/providers/friendli.ts
```diff
@@ -20,8 +20,10 @@ export const friendliModels = {
 		outputPrice: 4.4,
 		cacheWritesPrice: 0,
 		cacheReadsPrice: 0.26,
+		supportsReasoningEffort: ["minimal", "low", "medium", "high", "xhigh", "max"],
```
taltas, line 23: Does GLM offer a way to disable reasoning effort completely using 'disable' as a key?

packages/types/src/providers/friendli.ts
```diff
@@ -57,7 +61,8 @@ export const friendliModels = {
 		outputPrice: 1.2,
 		cacheWritesPrice: 0,
 		cacheReadsPrice: 0.06,
+		supportsReasoningBinary: true,
```
taltas: On `MiniMax-M2.5 ` This will render the checkbox `Use reasoning` in the settings.... but the thing is, this model always has reasoning on though right?  We should introduce or reuse metadata that distinguishes required/always reasoning from optional binary reasoning, hide or disable the on/off control for this model, and always request parsed reasoning for MiniMax-M2.5. In addition to this, we should add tests for MiniMax with the reasoning flag false and unset, not only true. 

**I think these changes should be in a separate PR in order not to bloat this one**

taltas (dismissed):
Thanks for the PR. A couple of questons on some of the models I found.

packages/types/src/providers/friendli.ts
Lee-Si-Yoon: Nope, verified that it throws 422 when disable or none is given
```
Endpoint: https://api.friendli.ai/serverless/v1/chat/completions
Model: zai-org/GLM-5.2

{"message":"invalid JSON payload: unknown enum value: 'disable'"}

{"message":"invalid JSON payload: unknown enum value: 'none'"}
```

To disable thinking it has to be done via chat_template_kwargs: { enable_thinking: true } option

packages/types/src/providers/friendli.ts
Lee-Si-Yoon: Thanks for the review! I reverted changes related to Minimax-m2.5 issues and planning to treat this in separate PR.

Lee-Si-Yoon:
@taltas hey, I know you probably have a lot going on right now, but would you mind taking another look when you get a chance?

Both of your comments are addressed since the last review:

- **GLM `"disable"` / `"none"` key**: Checked against the API — both throw 422 (`unknown enum value`). Thinking is controlled through `chat_template_kwargs: { enable_thinking: false }` instead, so I added that path and tests for it.
- **MiniMax-M2.5 always-reasoning**: Pulled those changes out of this PR entirely. I will pick it up in a separate one so this stays scoped to controllable reasoning models.

Thanks again — the review pointed out exactly where the boundary should be.

src/api/providers/friendli.ts
```diff
@@ -23,4 +55,136 @@ export class FriendliHandler extends BaseOpenAiCompatibleProvider<FriendliModelI
...
+			return this.client.chat.completions.create(
+				params as OpenAI.Chat.Completions.ChatCompletionCreateParamsStreaming,
+				requestOptions,
+			)
+		} catch (error) {
+			throw handleOpenAIError(error, this.providerName)
+		}
+	}
+
+	override async completePrompt(prompt: string): Promise<string> {
```
edelauna: The base class (and `SingleCompletionHandler` interface) declares `completePrompt(prompt, options?: CompletePromptOptions)`. Dropping `options` here means any `abortSignal` a caller passes is silently discarded. Could this accept `options?` and thread `options?.abortSignal` through to the `create` call?

src/api/providers/friendli.ts
```diff
@@ -23,4 +55,136 @@ export class FriendliHandler extends BaseOpenAiCompatibleProvider<FriendliModelI
...
+		const { id: modelId, info: modelInfo } = this.getModel()
```
edelauna:
```suggestion
		const { id: modelId } = this.getModel()
```
`modelInfo` is destructured but never referenced in the function body.

src/api/providers/friendli.ts
```diff
@@ -1,12 +1,44 @@
...
+
+/**
+ * Friendli extends the OpenAI Chat Completions API with these non-standard fields:
+ * - reasoning_effort: enum (minimal, low, medium, high, xhigh, max) — reasoning depth
+ * - chat_template_kwargs: { enable_thinking: boolean } — toggles thinking for controllable models
+ * - parse_reasoning / include_reasoning: when true, Friendli streams reasoning via
+ *   delta.reasoning_content (which extractReasoningFromDelta already handles)
+ * - reasoning_budget: integer token budget (not currently surfaced in settings UI)
+ */
+type FriendliChatCompletionParams = OpenAI.Chat.Completions.ChatCompletionCreateParamsStreaming & {
```
edelauna: This extends `ChatCompletionCreateParamsStreaming` (which requires `stream: true`), but the type is also used in `completePrompt` for a non-streaming call — which is why `params as any` is needed at line 183. `OpenRouterChatCompletionParams` uses the base `ChatCompletionCreateParams` instead. Worth splitting into a streaming and a non-streaming variant, or extending the base type?

src/api/providers/__tests__/friendli.spec.ts
```diff
@@ -394,3 +403,214 @@ describe("Friendli model max output tokens (clamping behavior)", () => {
...
+		mockCreate.mockResolvedValueOnce({
+			choices: [{ message: { content: "test result" } }],
+		})
+
+		await handler.completePrompt("test")
+
+		const callArgs = mockCreate.mock.calls[0][0] as any
+		expect(callArgs.reasoning_effort).toBe("medium")
+		expect(callArgs.chat_template_kwargs).toEqual({ enable_thinking: true })
+		expect(callArgs.parse_reasoning).toBe(true)
```
edelauna, line 616: `include_reasoning` is set alongside `parse_reasoning` in `buildFriendliReasoningParams` (line 113 of the source) — should this also assert `expect(callArgs.include_reasoning).toBe(true)`?

src/api/providers/__tests__/friendli.spec.ts
```diff
@@ -394,3 +403,214 @@ describe("Friendli model max output tokens (clamping behavior)", () => {
...
+				},
+			}),
+		}))
+
+		await handler.createMessage("system", []).next()
+
+		const callArgs = mockCreate.mock.calls[0][0] as any
+		expect(callArgs.reasoning_effort).toBeUndefined()
+		expect(callArgs.chat_template_kwargs).toEqual({ enable_thinking: false })
+		expect(callArgs.parse_reasoning).toBeUndefined()
```
edelauna, line 487: The `enableReasoningEffort: false` test above asserts `expect(callArgs.include_reasoning).toBeUndefined()` — should the `none` and `disable` cases here (and line 511) have the same assertion?

edelauna (changes requested):
Thanks for this contribution, can with reviewing - had some comments regarding implementation details.

Lee-Si-Yoon:
Thanks for the review @edelauna — all five points are addressed in a66aab150:

- `completePrompt` now accepts `options?: CompletePromptOptions` and threads `abortSignal` through to the SDK call.
- Unused `modelInfo` destructuring removed.
- Split `FriendliChatCompletionParams` into streaming and non-streaming variants (sharing a common `FriendliReasoningParams`). `completePrompt` now uses the non-streaming type, so the `as any` is gone.
- Added `include_reasoning` assertions to the `none`/`disable` and `completePrompt` tests.

edelauna (approved):

PR #899, opened 2026-07-15
fix(litellm): preserve reasoning_content for known reasoning model families

### LINKED ISSUE #891: [BUG] reasoning_content is dropped in conversation history when using LiteLLM with reasoning models

When using a model that supports reasoning (e.g., DeepSeek R1) through the LiteLLM proxy, the reasoning_content of previous assistant messages is dropped from the conversation history sent to the API, causing the model to lose its train of thought in subsequent turns.

This affects any user connecting to reasoning-capable models (like DeepSeek Reasoner) via the LiteLLM provider. It happens during multi-turn conversations where the assistant utilizes tool calls followed by a user message.

### Reproduction steps

1) Start Zoo Code and select "LiteLLM" as the API Provider.
2) Connect to a reasoning-supported model (e.g., `deepseek-reasoner`) via LiteLLM.
3) Ask the model to perform a task that requires a tool call (e.g., "Read this file and then explain it").
4) After the tool is executed and the result is returned, check the payload of the next API request.
5) Observe that the previous turn's reasoning block is missing.

### Expected result

The `reasoning_content` generated by the model should be preserved in the conversation history and merged properly with tool results so that the API does not drop it in subsequent requests.

### Actual result

The conversation history falls back to standard OpenAI formatting, which fails to merge tool results with text properly, causing the upstream API (like DeepSeek) to drop all previously recorded `reasoning_content`.

_No response_

### App Version

v3.68.0

### API Provider (optional)

LiteLLM

### Model Used (optional)

_No response_

_No response_

```shell

```

daewoongoh:
### Related GitHub Issue

Closes: #891 

### Description

LiteLLM's `/v1/model/info` response never exposes reasoning-related capability flags, so `preserveReasoning` was always resolving to `false` for reasoning/thinking models routed through a LiteLLM proxy (DeepSeek-Reasoner, GLM-4.7/5, Kimi-K2-Thinking, MiniMax-M2/M3, Qwen3-plus/max, etc.). As a result, `LiteLLMHandler.createMessage` always used `convertToOpenAiMessages`, so `reasoning_content` from a prior turn was never echoed back to the API after a tool-call turn. This differs from accessing the same underlying models directly through their native provider config (e.g. `deepseek.ts` already sets `preserveReasoning: true`), and some providers reject the request with a 400 error when required reasoning history is missing.

**How:**
- `packages/types/src/providers/lite-llm.ts`: Collected the exact alias/routed-model-name ids for every model that requires `preserveReasoning: true` in its native provider config (`deepseek.ts`, `mimo.ts`, `moonshot.ts`, `bedrock.ts`, `fireworks.ts`, `zai.ts`, `minimax.ts`, `opencode-go.ts`), and exported them as `LITELLM_PRESERVE_REASONING_MODEL_IDS`. `isLiteLLMPreserveReasoningModel()` checks a model name against this set with an exact (case-insensitive) match on the final `/`-delimited segment, so provider-prefixed routed names (e.g. `deepseek/deepseek-reasoner`, `fireworks_ai/accounts/fireworks/models/kimi-k2p7-code`) resolve correctly without over-matching unrelated aliases that merely contain a family fragment as a substring (e.g. `glm-5-flash` vs `glm-5`).
- `src/api/providers/fetchers/litellm.ts`: `getLiteLLMModels` now checks both `model_name` and `litellm_params.model` against `isLiteLLMPreserveReasoningModel()` independently while parsing each model, setting `ModelInfo.preserveReasoning: true` if either matches.
- `src/api/providers/lite-llm.ts`: `createMessage` now branches on `info.preserveReasoning` — `true` routes through `convertToR1Format` (with `mergeToolResultText: true`), otherwise it keeps using `convertToOpenAiMessages`. `convertToR1Format` folds any text following a tool_result into the preceding tool message instead of inserting a separate user message, which prevents reasoning-mode providers from dropping `reasoning_content` when a user message appears mid-turn.

**Trade-off:** the matching is an explicit model-id allowlist, not a query against LiteLLM's actual capabilities (which it doesn't expose). It's exact rather than pattern-based to avoid false positives on similarly-named non-reasoning variants, but that also means unrecognized aliases or renamed deployments simply won't match (documented in the `lite-llm.ts` comment).

### Test Procedure

- `pnpm vitest run` (unit tests):
  - `packages/types/src/__tests__/lite-llm.test.ts`: verifies `isLiteLLMPreserveReasoningModel` matches every listed model id (including provider-prefixed routed names and case-insensitivity), and does not match unrelated ids or same-family substrings (e.g. `glm-5-flash`, `deepseek-v4-mini`).
  - `src/api/providers/fetchers/__tests__/litellm.spec.ts`: verifies `getLiteLLMModels` sets `preserveReasoning: true` when either the alias (`model_name`) or the routed model name (`litellm_params.model`) matches — with each field tested independently so a match on one alone can't be mistaken for the other being checked — and omits the field otherwise.
  - `src/api/providers/__tests__/lite-llm.spec.ts`: verifies `convertToR1Format` is used and `reasoning_content` survives on the outgoing request when `preserveReasoning: true`; and that the non-`preserveReasoning` branch sends no `reasoning_content` (reflecting that `Task.buildCleanConversationHistory` already strips the reasoning block upstream in that case).
- Manual repro: configure the LiteLLM provider with `litellm_params.model` routed to a reasoning model (e.g. `deepseek/deepseek-reasoner`), run a multi-turn task involving a tool call — previously the follow-up request dropped `reasoning_content`; it's now preserved.

### Pre-Submission Checklist

- [x] **Issue Linked**: (fill in once the issue is created)
- [x] **Scope**: Limited to LiteLLM's reasoning_content preservation.
- [x] **Self-Review**: Done.
- [x] **Testing**: Added unit tests across 3 files.
- [x] **Documentation Impact**: None (internal behavior change, no user-facing settings changed).
- [x] **Contribution Guidelines**: Reviewed.

### Documentation Updates

### Additional Notes

If a reasoning model family or renamed alias isn't caught by this heuristic, additional patterns can be appended to `LITELLM_PRESERVE_REASONING_FRAGMENTS`.

### Get in Touch

hehegwk_23849

## Summary by CodeRabbit

* **New Features**
  * Added automatic “preserve reasoning” detection for supported LiteLLM-routed models using an allowlist of underlying model IDs and case-insensitive matching.
  * Updated conversation formatting for reasoning-enabled models to use the correct message shaping, including proper tool-call handling and merging of trailing tool-result text.
* **Bug Fixes**
  * Prevented incorrect reasoning detection for unrelated or only-partially-matching model names.
* **Tests**
  * Added coverage for reasoning capability inference and for reasoning-enabled vs standard message conversion behavior (including exact message output expectations).


packages/types/src/providers/lite-llm.ts
```diff
@@ -13,3 +13,34 @@ export const litellmDefaultModelInfo: ModelInfo = {
...
+const LITELLM_PRESERVE_REASONING_FRAGMENTS = [
+	String.raw`deepseek-v4-(flash|pro)`, // deepseek.ts
+	String.raw`deepseek-reasoner`, // deepseek.ts (legacy alias)
+	String.raw`mimo-v2\.5(-pro)?`, // mimo.ts
+	String.raw`kimi-k2-thinking`, // moonshot.ts, bedrock.ts, fireworks.ts
+	String.raw`kimi-k2p7-code`, // fireworks.ts (dash-separated id)
+	String.raw`kimi-k2\.7-code`, // fireworks.ts (dotted alias variant)
+	String.raw`minimax[.-]?m[23](\.\d+)?(-highspeed|-stable)?\b`, // bedrock.ts, minimax.ts, opencode-go.ts
+	String.raw`glm-4\.7(?!-flash)\b`, // zai.ts (excludes glm-4.7-flash(x))
+	String.raw`glm-5(\.[12])?(-turbo)?(?!-flash)\b`, // zai.ts, opencode-go.ts
```
edelauna: Could this still match `glm-5` as a prefix for names like `glm-5.1-flash` or `glm-5.3`? If those should stay excluded, it may be worth making this fragment consume the accepted suffix fully and adding negative cases for those variants.

src/api/providers/fetchers/__tests__/litellm.spec.ts
```diff
@@ -697,4 +697,92 @@ describe("getLiteLLMModels", () => {
...
 		})
 	})
+
+	describe("preserveReasoning inference", () => {
+		it("sets preserveReasoning: true when the routed model matches a known reasoning family", async () => {
+			const mockResponse = {
+				data: {
+					data: [
+						{
+							model_name: "deepseek-reasoner-alias",
```
edelauna: Could this alias be something neutral that does not match the preserve-reasoning pattern? As written, this still passes if the implementation only checks `model_name`, so it does not quite prove that `litellm_params.model` is used.

edelauna (changes requested):
Thanks for this contribution, had a small comment on test and the regex in use.

packages/types/src/providers/lite-llm.ts
daewoongoh: fixed in d57a97a by dropping the regex/prefix approach entirely. isLiteLLMPreserveReasoningModel now does an exact match against LITELLM_PRESERVE_REASONING_MODEL_IDS via Set.has() on the full (lowercased) model id, so glm-5 no longer matches glm-5.1-flash, glm-5.3, etc. Added negative test cases for exactly these variants (glm-5-flash, glm-4.7-flash, glm-4.7-flashx, glm-4.8, etc.) in lite-llm.test.ts.

src/api/providers/fetchers/__tests__/litellm.spec.ts
daewoongoh: fixed in d57a97a — the alias (model_name) is now a neutral string that doesn't match the allowlist on its own (my-deepseek-alias, my-kimi-alias), so this test only passes if litellm_params.model is actually checked. Added a companion test for the reverse case (alias matches, routed model doesn't) to confirm both fields are checked independently.

edelauna (approved):
Thanks for your contribution

PR #912, opened 2026-07-17
fix(task-history): route invalidate() and invalidateAll() through withLock

### LINKED ISSUE #698: fix(task-history): invalidate() and invalidateAll() bypass withLock, corrupting cache under concurrent writes
Part of #357 (Epic 2 — Task Lifecycle Fixes).

# Context

`TaskHistoryStore` uses a promise-chain mutex (`withLock`) to serialize all cache mutations. Every write path — `upsert`, `delete`, `deleteMany`, `reconcile`, `atomicReadAndUpdate` — acquires the lock before touching `this.cache`. However, two public methods bypass the lock entirely:

- **`invalidateAll()`** — synchronous `cache.clear()`, no lock acquired  
- **`invalidate(taskId)`** — async file read + `cache.set`, no lock acquired

There is one production caller today (`webviewMessageHandler.ts:954`, Roo history import):

```typescript
provider.taskHistoryStore.invalidateAll()   // sync, no lock
await provider.taskHistoryStore.reconcile() // async, acquires lock
```

# The Race

`_upsertUnlocked` (called by `atomicReadAndUpdate` and `upsert`) contains an `await` at line 175 (`await this.writeTaskFile(merged)`). At that yield point the event loop is free to run other synchronous work — including `invalidateAll()`. The resulting interleaving:

1. `atomicReadAndUpdate` holds the lock, enters `_upsertUnlocked`
2. `await writeTaskFile(merged)` — lock is held but execution yields
3. **`invalidateAll()` fires**: `cache.clear()` — entire cache wiped while lock is logically held
4. `_upsertUnlocked` resumes: `cache.set(merged.id, merged)` — only the one item is back
5. `getAll()` returns a **single-item list**
6. `onWrite([singleItem])` → `scheduleGlobalStateWriteThrough()` runs with partial history
7. Lock releases; `reconcile()` queues and eventually rebuilds the full cache

**Impact:** Between steps 6 and 7 `globalState` contains only one history item. If VS Code crashes in this window, the downgrade-compatibility `globalState` history list is corrupted. `invalidate()` has no production callers today, but carries an equivalent stale-write risk if ever called concurrently.

# Developer Notes

## Why the lock doesn't protect you here

The promise-chain mutex in `withLock` serializes *other `withLock` callers* — it does not prevent synchronous code outside the lock from observing or mutating `cache` mid-await. As the [matklad "On Async Mutexes" post](https://matklad.github.io/2025/11/04/on-async-mutexes.html) explains: *"mutual exclusion is a property of the logic itself, not of the runtime… if a certain snippet of code must be executed atomically with respect to everything else that is concurrent, then it must be annotated as such in the source code."* Any public method that touches `this.cache` without `withLock` is an escape hatch that breaks the invariant the rest of the class relies on. The [Trabe article on mutex-based cache sync](https://medium.com/trabe/synchronize-cache-updates-in-node-js-with-a-mutex-d5b395457138) makes the same point: the lock must own the full critical section including any cache reads that feed into the write.

## Recommended fix — `invalidateAndReconcile()`

Since the one production caller always pairs `invalidateAll()` + `reconcile()`, the cleanest fix is a combined method that does both under a **single lock acquisition**, eliminating the empty-cache window:

```typescript
async invalidateAndReconcile(): Promise<void> {
  return this.withLock(async () => {
    this.cache.clear()
    // inline reconcile body (or extract _reconcileUnlocked like _upsertUnlocked)
    ...
  })
}
```

Then `webviewMessageHandler.ts:954` becomes:
```typescript
await provider.taskHistoryStore.invalidateAndReconcile()
await provider.taskHistoryStore.flushIndex()
await provider.postStateToWebview()
```

CodeRabbit suggested wrapping each method in its own `withLock` call. That also fixes the corruption, but leaves a brief window between two separate lock acquisitions where reads see an empty cache. A single combined method is strictly better.

`invalidate(taskId)` should also be wrapped in `withLock` (or use an `_invalidateUnlocked` helper if ever needed inside an existing lock holder).

## What to validate

- The `onWrite` callback receives a complete history list after any concurrent `invalidateAndReconcile` + `atomicReadAndUpdate`
- A test that fires `invalidateAndReconcile()` concurrently with `atomicReadAndUpdate()` should show the full item set in cache after both settle (analogous to `delegation-concurrent.spec.ts`)
- The one call site in `webviewMessageHandler.ts` is updated to `await invalidateAndReconcile()`

## Scope note

This is a pre-existing defect that became visible when `atomicReadAndUpdate` (Story 2.1, PR #691) made the locking contract explicit. It is independent of Stories 2.2 and 2.3.

morgan-coded:
Locks `invalidate()` and `invalidateAll()` behind `TaskHistoryStore`'s existing `withLock` queue so cache invalidation cannot interleave with an in-flight mutation. Each method keeps its existing cache behavior.

### Related GitHub Issue

Fixes #698

### Description

Per your note, I routed `invalidate()` rather than removing it: the census found no non-test callers, but `TaskHistoryStore` is exported from the task-persistence barrel, so deleting its public method could break consumers.

| Method | Definitions | Non-test callers | Test references | Export surface | Decision |
| --- | ---: | ---: | ---: | --- | --- |
| `invalidate()` | 1 | 0 | 3 call sites | Public method on the barrel-exported `TaskHistoryStore` | Route through `withLock` |
| `invalidateAll()` | 1 | 1 | 6 mock/assertion references | Public method on the barrel-exported `TaskHistoryStore` | Route through `withLock` |

### Test Procedure

The regression blocks `writeTaskFile` with a manually controlled promise after the lock is acquired, calls `invalidateAll()`, then releases the write. It asserts that `onWrite` sees both cached tasks before the queued invalidation clears the final cache, so no timer or scheduling delay controls the result.

- `pnpm --filter zoo-code exec vitest run core/task-persistence/__tests__/TaskHistoryStore.spec.ts`
- `pnpm test`
- `pnpm lint`
- `pnpm check-types`

### Pre-Submission Checklist

- [x] **Scope**: The changes are focused on the linked issue.
- [x] **Self-Review**: I performed a self-review of the cache and test changes.
- [x] **Testing**: A deterministic concurrency regression covers the change.
- [x] **Documentation Impact**: No user-facing documentation changes are required.
- [x] **Contribution Guidelines**: I read and agree to the Contributor Guidelines.

### Additional Notes

Non-goals: no other cache methods, lock implementation, callers, or APIs are changed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented task history cache invalidation from interfering with concurrent updates by performing cache mutations under a write lock.
  * Ensured “import history” waits for cache invalidation to complete before reconciling and flushing the index.
* **Tests**
  * Added concurrency tests for invalidating individual tasks and clearing the cache during an in-progress update.

### DIFF
--- a/src/core/task-persistence/TaskHistoryStore.ts
+++ b/src/core/task-persistence/TaskHistoryStore.ts
@@ -445,7 +445,7 @@ export class TaskHistoryStore {
 	}
 
 	/**
-	 * Clear all in-memory cache and reload from index.
+	 * Clear all in-memory cache entries; a subsequent `reconcile()` repopulates them from task files.
 	 */
 	async invalidateAll(): Promise<void> {
 		return this.withLock(async () => {
--- a/src/core/task-persistence/TaskHistoryStore.ts
+++ b/src/core/task-persistence/TaskHistoryStore.ts
@@ -430,23 +430,27 @@ export class TaskHistoryStore {
 	 * Invalidate a single task's cache entry (re-read from disk on next access).
 	 */
 	async invalidate(taskId: string): Promise<void> {
-		try {
-			const item = await this.readTaskFile(taskId)
-			if (item) {
-				this.cache.set(taskId, item)
-			} else {
+		return this.withLock(async () => {
+			try {
+				const item = await this.readTaskFile(taskId)
+				if (item) {
+					this.cache.set(taskId, item)
+				} else {
+					this.cache.delete(taskId)
+				}
+			} catch {
 				this.cache.delete(taskId)
 			}
-		} catch {
-			this.cache.delete(taskId)
-		}
+		})
 	}
 
 	/**
 	 * Clear all in-memory cache and reload from index.
 	 */
-	invalidateAll(): void {
-		this.cache.clear()
+	async invalidateAll(): Promise<void> {
+		return this.withLock(async () => {
+			this.cache.clear()
+		})
 	}
 
 	// ────────────────────────────── Migration ──────────────────────────────
--- a/src/core/task-persistence/__tests__/TaskHistoryStore.spec.ts
+++ b/src/core/task-persistence/__tests__/TaskHistoryStore.spec.ts
@@ -442,6 +442,55 @@ describe("TaskHistoryStore", () => {
 		})
 	})
 
+	describe("invalidateAll()", () => {
+		it("waits for an in-flight write before clearing the cache", async () => {
+			const onWrite = vi.fn().mockResolvedValue(undefined)
+			store = new TaskHistoryStore(tmpDir, { onWrite })
+			await store.initialize()
+
+			const first = makeHistoryItem({ id: "invalidate-all-first", ts: 1000, tokensIn: 100 })
+			const second = makeHistoryItem({ id: "invalidate-all-second", ts: 2000 })
+			await store.upsert(first)
+			await store.upsert(second)
+			onWrite.mockClear()
+
+			let signalWriteStarted!: () => void
+			const writeStarted = new Promise<void>((resolve) => {
+				signalWriteStarted = resolve
+			})
+			let releaseWrite!: () => void
+			const writeCanFinish = new Promise<void>((resolve) => {
+				releaseWrite = resolve
+			})
+
+			const storeAny = store as any
+			const originalWriteTaskFile = storeAny.writeTaskFile.bind(store)
+			vi.spyOn(storeAny, "writeTaskFile").mockImplementation(async (...args: unknown[]) => {
+				const item = args[0] as HistoryItem
+				if (item.id === first.id && item.tokensIn === 999) {
+					signalWriteStarted()
+					await writeCanFinish
+				}
+				return originalWriteTaskFile(...args)
+			})
+
+			const write = store.upsert({ ...first, tokensIn: 999 })
+			await writeStarted
+			const invalidation = store.invalidateAll()
+
+			releaseWrite()
+			await write
+			await invalidation
+
+			expect(onWrite).toHaveBeenCalledTimes(1)
+			expect(onWrite.mock.calls[0][0].map((item: HistoryItem) => item.id).sort()).toEqual([
+				"invalidate-all-first",
+				"invalidate-all-second",
+			])
+			expect(store.getAll()).toEqual([])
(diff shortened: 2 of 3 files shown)


src/core/task-persistence/TaskHistoryStore.ts
```diff
@@ -430,23 +430,27 @@ export class TaskHistoryStore {
...
 				this.cache.delete(taskId)
 			}
-		} catch {
-			this.cache.delete(taskId)
-		}
+		})
 	}
 
 	/**
 	 * Clear all in-memory cache and reload from index.
```
edelauna: The JSDoc still says "reload from index" but the body only calls `cache.clear()` -- `loadIndex()` is never called. Worth updating to reflect the actual behavior?

edelauna (changes requested):
Nice! Thanks for taking this on!

Two items that don't map to diff lines:

**`webviewMessageHandler.ts:966`** — `invalidateAll()` is now async but is called without `await`. Any error it throws is silently swallowed, and a reader can't tell whether the cache clear is guaranteed to finish before the subsequent `reconcile()`. Consider `await provider.taskHistoryStore.invalidateAll()` there.

**Missing lock-contention test for `invalidate()`** — the PR wraps both `invalidate()` and `invalidateAll()` in `withLock`, but only `invalidateAll()` gets a concurrency regression test. If someone removed `withLock` from `invalidate()` in a future refactor, nothing would catch it. A parallel test modelled on the new `invalidateAll()` one (gate `writeTaskFile`, call `store.invalidate(id)` concurrently, assert cache reflects the written value) would close the gap.

morgan-coded:
Updated the handler to await `invalidateAll()` and checked the remaining call sites; no other non-test callers needed the same fix. Added a deterministic `invalidate()` contention test that gates `writeTaskFile` and verifies the cache reflects the completed write, and corrected the JSDoc to note that `reconcile()` repopulates the cleared cache from task files. Thanks for the clear review.

edelauna (approved):
Thanks so much for your contribution!

edelauna:
I did open a related issue: https://github.com/Zoo-Code-Org/Zoo-Code/issues/949 which I noticed during review - if this is something you would like to help out with next 🙏

PR #925, opened 2026-07-17
fix(webview): sync cachedState on mode changes in SettingsView

### LINKED ISSUE #914: [UX] SettingsView shows stale mode/apiConfig in multi-tab mode
# [UX] SettingsView shows stale mode/apiConfig in multi-tab mode

## Problem

In parallel mode, `SettingsView` displays stale `mode` and `apiConfiguration` values because its internal `cachedState` is only updated when `currentApiConfigName` changes (via `useEffect` dependency), but NOT when `mode` changes via `handleModeSwitch()`.

## Evidence

**Code locations:**
- [`SettingsView.tsx:153`](webview-ui/src/components/settings/SettingsView.tsx): `const [cachedState, setCachedState] = useState(() => extensionState)` — initialized once from context
- [`SettingsView.tsx:222-232`](webview-ui/src/components/settings/SettingsView.tsx): `useEffect` only watches `currentApiConfigName`, not `mode`
- [`ClineProvider.ts:1640`](src/core/webview/ClineProvider.ts): `handleModeSwitch()` updates view-local state but doesn't trigger SettingsView cache refresh

**Reproduction:**
```
1. Open Tab A (sidebar) → set mode = "architect" via handleModeSwitch()
2. Open Tab B (editor) → open SettingsView
3. SettingsView displays mode = "code" (from global state), NOT "architect" (view-local)
4. Switching API config in Tab B updates cachedState, but switching mode does not
```

## Root Cause

`SettingsView.cachedState` is initialized from `extensionState` (the React context value), which reflects the **merged** state at initialization time. When another tab changes its `mode`, the global state doesn't change (each tab has its own view-local override), so SettingsView in other tabs never sees the update.

## Proposed Fix

1. Add `mode` to the `useEffect` dependency array that updates `cachedState`:
   ```typescript
   useEffect(() => {
       if (prevApiConfigName.current === currentApiConfigName && prevMode.current === mode) return
       setCachedState((prev) => ({ ...prev, ...extensionState }))
       prevApiConfigName.current = currentApiConfigName
       prevMode.current = mode
       setChangeDetected(false)
   }, [currentApiConfigName, mode, extensionState])
   ```

2. Alternatively: Pass `viewId` to webview so SettingsView can distinguish local vs global state and show a visual indicator when they differ.

## Scope

- **Impact:** All multi-tab users — any tab with a different mode than the global default will see stale values in SettingsView
- **Risk:** Low — only affects UI display, not actual behavior
- **Priority:** High (directly impacts user trust in parallel mode)

## Related

- PR #909: https://github.com/Zoo-Code-Org/Zoo-Code/pull/909
- Main tracking issue: #913

easonLiangWorldedtech:
## Problem

In parallel mode, `SettingsView` shows stale `mode` and `apiConfiguration` values because its internal `cachedState` only updates via `useEffect` when `currentApiConfigName` changes, but doesn't refresh when `mode` changes through `handleModeSwitch()`.

## Fix

Add `mode` to the dependency array of the `useEffect` that updates `cachedState`:

### Modified Files
- [`webview-ui/src/components/settings/SettingsView.tsx`](webview-ui/src/components/settings/SettingsView.tsx:133)

### Changes
1. Destructure `mode` from `extensionState` (line 133)
2. Add `prevMode` ref to track previous mode value (line 150)
3. Update `useEffect` condition to check both `currentApiConfigName` and `mode` (line 226)
4. Add `mode` to the dependency array (line 234)

```typescript
// Before
if (prevApiConfigName.current === currentApiConfigName) {
    return
}

// After
if (prevApiConfigName.current === currentApiConfigName && prevMode.current === mode) {
    return
}
```

Closes #914

## Summary by CodeRabbit

* **Bug Fixes**
  * Settings now stay synchronized when the application mode changes.
  * Unsaved-change indicators and save-button state reset correctly after mode-triggered synchronization.
* **Tests**
  * Expanded and refactored SettingsView change-detection tests, including new coverage to confirm state sync occurs on mode changes and not on re-renders with unchanged mode.

### DIFF
--- a/webview-ui/src/components/settings/__tests__/SettingsView.change-detection.spec.tsx
+++ b/webview-ui/src/components/settings/__tests__/SettingsView.change-detection.spec.tsx
@@ -1,5 +1,5 @@
 import { act, render, screen, fireEvent, waitFor, configure } from "@testing-library/react"
-import { vi, describe, it, expect, beforeEach } from "vitest"
+import { vi, describe, it, expect, beforeEach, beforeAll } from "vitest"
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query"
 import React from "react"
 
@@ -19,8 +19,6 @@ vi.mock("@src/utils/vscode", () => ({
 	},
 }))
 
-// Import the actual component
-import SettingsView from "../SettingsView"
 import { useExtensionState } from "@src/context/ExtensionStateContext"
 
 // Mock the extension state context
@@ -28,10 +26,12 @@ vi.mock("@src/context/ExtensionStateContext", () => ({
 	useExtensionState: vi.fn(),
 }))
 
+const mockTranslate = vi.hoisted(() => (key: string) => key)
+
 // Mock the translation context
 vi.mock("@src/i18n/TranslationContext", () => ({
 	useAppTranslation: () => ({
-		t: (key: string) => key,
+		t: mockTranslate,
 	}),
 }))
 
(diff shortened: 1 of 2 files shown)


webview-ui/src/components/settings/__tests__/SettingsView.change-detection.spec.tsx
```diff
@@ -525,5 +538,109 @@ describe("SettingsView - Change Detection Fix", () => {
...
+				rerender(
+					<QueryClientProvider client={queryClient}>
+						<SettingsView onDone={onDone} />
+					</QueryClientProvider>,
+				)
+			})
+
+			// Provider value should remain unchanged from the dirty state
+			expect(screen.getByTestId("provider-value")).toHaveTextContent("baseten")
+		}, 20000)
```
edelauna, line 790: It looks like this test would pass even if the whole `mode` fix were reverted. Since `extensionState` is a `const` (line 609) and the rerender at line 635 reuses the same object reference, React sees no dependency change — the sync effect never re-runs under either the old or new code, so `provider-value` stays `baseten` purely from the dirty click above.

Would reassigning `extensionState` to a *new* object (same values, new identity) before the rerender make this guard load-bearing? As written, I think this catches a regression of nothing.

webview-ui/src/components/settings/__tests__/SettingsView.change-detection.spec.tsx
```diff
@@ -525,5 +538,109 @@ describe("SettingsView - Change Detection Fix", () => {
...
+
+			// Verify cachedState reflects the new mode's settings
+			await waitFor(() => {
+				expect(screen.getByTestId("provider-value")).toHaveTextContent("openrouter")
+			})
+
+			// Verify changeDetected is reset (save button should be disabled)
+			const updatedSaveButton = screen.getByTestId("save-button") as HTMLButtonElement
+			expect(updatedSaveButton.disabled).toBe(true)
+		}, 20000)
```
edelauna: Two mutations of the fix that this test seems to miss:

- If `mode` were dropped from the dependency array, the effect would still fire here because `extensionState` is reassigned to a new object at line 576 — so the `mode` dep entry (part of this PR) is effectively untested.
- If the `prevMode.current = mode` assignment were removed, the single `code → ask` transition would still produce correct state once; the stale ref is never re-exercised without a second transition.

Would it be worth adding a rerender where only `extensionState` identity changes (mode + apiConfigName unchanged) to pin the dep array, and a second same-mode rerender to expose a stale `prevMode`?

edelauna (changes requested):
Nice ! Thanks for taking this on - is a nice little fix. Had some comments about increasing the test robustness, otherwise looks good.

edelauna (approved):
Thanks for this contribution

PR #938, opened 2026-07-18
Fix codebase indexing for plain text files

### LINKED ISSUE #931: [BUG] Codebase indexing silently skips .txt files
**Describe the bug**
The indexer processes .md files correctly, but .txt files in the same folders are silently skipped

**To Reproduce**
1. Workspace folder containing README.md and manual.txt
2. Enable Codebase Indexing; wait for green status
3. Semantic search for content that exists only in the .txt file → not found
4. Confirm in Qdrant: scroll filtered on filePath containing .txt returns zero points; .md returns the expected points

**Expected behavior**
.txt files get indexed (they don't need tree-sitter — plain line-based chunking would suffice)
This is what the documentation states: https://docs.zoocode.dev/features/codebase-indexing#file-filtering

3.68.0

WebMad:
## Summary

- add `.txt` to the extensions scanned by codebase indexing
- route plain text files through the existing line-based fallback chunking mechanism
- add regression coverage for `.txt` parsing and case-insensitive extension matching

Fixes #931

## Testing

- `cd src && npx vitest run services/code-index/processors/__tests__/parser.txt.spec.ts`
- relevant parser/scanner suite: 55 tests passed
- repository lint passed via the pre-commit hook
- repository type-check passed via the pre-push hook
- manually verified semantic search with a workspace containing both `README.md` and `manual.txt`; content unique to both files was returned after full reindexing
- manually reproduced the original failure by temporarily removing `.txt` support and confirming that only the Markdown content was indexed

## Summary by CodeRabbit

- **New Features**
  - Added support for plain-text (`.txt`) in code indexing, including consistent handling of uppercase `.TXT`.
  - Plain-text (`.txt`) and other non-structural extensions now bypass structural code-definition parsing and use fallback chunking.
- **Bug Fixes**
  - Fixed parser behavior so an empty-string `content` override no longer triggers an unintended disk read.
- **Tests**
  - Added integration/unit coverage for fallback chunking (uppercase extensions, single-block output, segment-hash deduplication) and for ensuring structural parsing/file reads aren’t invoked for non-structural files.

### DIFF
--- a/src/services/code-index/processors/__tests__/parser.txt.spec.ts
+++ b/src/services/code-index/processors/__tests__/parser.txt.spec.ts
@@ -33,10 +33,26 @@ describe("CodeParser - plain text support", () => {
 			end_line: 3,
 			content,
 			fileHash: "txt-file-hash",
+			segmentHash: expect.any(String),
 		})
 	})
 
-	it("matches uppercase .TXT extensions", () => {
+	it("parses uppercase .TXT extensions", async () => {
 		expect(shouldUseFallbackChunking(".TXT")).toBe(true)
+
+		const content = "Uppercase plain-text extension content long enough to produce a fallback chunk."
+		const blocks = await new CodeParser().parseFile("manual.TXT", {
+			content,
+			fileHash: "uppercase-txt-file-hash",
+		})
+
+		expect(blocks).toHaveLength(1)
+		expect(blocks[0]).toMatchObject({
+			file_path: "manual.TXT",
+			type: "fallback_chunk",
+			content,
+			fileHash: "uppercase-txt-file-hash",
+			segmentHash: expect.any(String),
+		})
 	})
 })
--- a/src/services/code-index/processors/__tests__/parser.txt.spec.ts
+++ b/src/services/code-index/processors/__tests__/parser.txt.spec.ts
@@ -1,0 +1,42 @@
+import { CodeParser } from "../parser"
+import { scannerExtensions, shouldUseFallbackChunking } from "../../shared/supported-extensions"
+
+vi.mock("../../../../../packages/telemetry/src/TelemetryService", () => ({
+	TelemetryService: {
+		instance: {
+			captureEvent: vi.fn(),
+		},
+	},
+}))
+
+describe("CodeParser - plain text support", () => {
+	it("supports .txt files through fallback chunking", async () => {
+		expect(scannerExtensions).toContain(".txt")
+		expect(shouldUseFallbackChunking(".txt")).toBe(true)
+
+		const content = [
+			"Zoo Code plain text indexing regression test.",
+			"This sentence contains searchable content that only exists in the text file.",
+			"The fallback parser should preserve every line while creating an indexable chunk.",
+		].join("\n")
+
+		const blocks = await new CodeParser().parseFile("manual.txt", {
+			content,
+			fileHash: "txt-file-hash",
+		})
+
+		expect(blocks).toHaveLength(1)
+		expect(blocks[0]).toMatchObject({
+			file_path: "manual.txt",
+			type: "fallback_chunk",
+			start_line: 1,
+			end_line: 3,
+			content,
+			fileHash: "txt-file-hash",
+		})
+	})
+
+	it("matches uppercase .TXT extensions", () => {
+		expect(shouldUseFallbackChunking(".TXT")).toBe(true)
+	})
+})
--- a/src/services/tree-sitter/index.ts
+++ b/src/services/tree-sitter/index.ts
@@ -113,6 +113,11 @@ export async function parseSourceCodeDefinitionsForFile(
 		return undefined
 	}
 
+	// Plain text files have no structural definitions to extract
+	if (ext === ".txt") {
+		return undefined
+	}
+
 	// Special case for markdown files
 	if (ext === ".md" || ext === ".markdown") {
 		// Check if we have permission to access this file
--- a/src/services/tree-sitter/index.ts
+++ b/src/services/tree-sitter/index.ts
@@ -66,6 +66,8 @@ const extensions = [
 	// Markdown
 	"md",
 	"markdown",
+	// Plain text
+	"txt",
 	// JSON
 	"json",
 	// CSS
--- a/src/services/tree-sitter/index.ts
+++ b/src/services/tree-sitter/index.ts
@@ -5,6 +5,7 @@ import { fileExistsAtPath } from "../../utils/fs"
 import { parseMarkdown } from "./markdownParser"
 import { RooIgnoreController } from "../../core/ignore/RooIgnoreController"
 import { QueryCapture } from "web-tree-sitter"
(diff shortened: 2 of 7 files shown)


WebMad:
[test-zoocode-contribution.tar.gz](https://github.com/user-attachments/files/30151493/test-zoocode-contribution.tar.gz)

image

image

src/services/tree-sitter/index.ts
```diff
@@ -111,6 +113,11 @@ export async function parseSourceCodeDefinitionsForFile(
 		return undefined
 	}
 
+	// Plain text files have no structural definitions to extract
+	if (ext === ".txt") {
```
WebMad: Plain text files can also reach `parseSourceCodeDefinitionsForFile()` through the folded file context flow, which is separate from codebase indexing. The `fallbackExtensions` list only affects `CodeParser`, so it does not prevent this function from trying to load a Tree-sitter parser for `.txt`. Since no such parser exists, we return `undefined` early to avoid an `Unsupported language: txt` error

WebMad:
Plain text files can also reach `parseSourceCodeDefinitionsForFile()` through the folded file context flow, which is separate from codebase indexing. The `fallbackExtensions` list only affects `CodeParser`, so it does not prevent this function from trying to load a Tree-sitter parser for `.txt`. Since no such parser exists, we return `undefined` early to avoid an `Unsupported language: txt` error

src/services/tree-sitter/index.ts
```diff
@@ -111,6 +113,11 @@ export async function parseSourceCodeDefinitionsForFile(
...
+		return undefined
+	}
```
edelauna: The other fallback extensions (`.vb`, `.scala`, `.swift`) have no equivalent guard here — if `foldedFileContext.ts` calls `parseSourceCodeDefinitionsForFile` with one of those, it reaches `loadRequiredLanguageParsers` and throws `"Unsupported language: vb"` instead of returning `undefined` cleanly. Should this guard cover all `fallbackExtensions` rather than just `.txt`?

(Would need to import `shouldUseFallbackChunking` or move the set to a shared location to avoid an import cycle.)

src/services/code-index/processors/__tests__/parser.txt.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+		expect(blocks).toHaveLength(1)
+		expect(blocks[0]).toMatchObject({
+			file_path: "manual.txt",
+			type: "fallback_chunk",
+			start_line: 1,
+			end_line: 3,
+			content,
+			fileHash: "txt-file-hash",
+		})
```
edelauna, line 37: `segmentHash` is populated by the de-duplication path in `_chunkTextByLines` (parser.ts:255–268) but is not asserted here. If that gate were removed, duplicate blocks could be emitted and this test would still pass. Worth adding `segmentHash: expect.any(String)` to the `toMatchObject` call?

src/services/code-index/processors/__tests__/parser.txt.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+	})
+
+	it("matches uppercase .TXT extensions", () => {
+		expect(shouldUseFallbackChunking(".TXT")).toBe(true)
+	})
```
edelauna: This verifies `shouldUseFallbackChunking` handles uppercase but doesn't exercise `CodeParser.parseFile("manual.TXT", ...)`. A mutation removing `.toLowerCase()` at parser.ts:37 would cause `.TXT` to fail `isSupportedLanguage` and silently return `[]`. Could we extend this to call `parseFile` with an uppercase filename?

src/services/code-index/processors/__tests__/parser.txt.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+})
```
edelauna, line 58: `parser.vb.spec.ts` covers empty content (→ `[]`) and content below `MIN_BLOCK_CHARS=50` (→ `[]`) for the fallback chunker. Are those edge cases considered covered by the `.vb` suite, or is there an intention to add them here too?

edelauna (changes requested):
Nice! Looks great, thank you for this contribution, has some suggestions around the test case, and if we could also extend return `undefined` behavior to the other extensions while we're making this change.

src/services/code-index/processors/__tests__/parser.txt.spec.ts
WebMad: I consider those cases part of the shared fallback chunker contract, so I’ve kept them in the existing VB/fallback suite rather than duplicating the same assertions for every fallback extension. The .txt tests focus on routing .txt files through that shared path and validating the resulting block contract. While verifying the existing edge-case tests, I found that explicitly provided empty content was being treated as missing content and triggered a file read. I fixed that by checking content !== undefined, and the existing empty and below-minimum-size tests now pass

src/services/code-index/processors/__tests__/parser.txt.spec.ts
WebMad: Agreed - checking only shouldUseFallbackChunking() did not cover the extension normalization performed by CodeParser. I updated the test to call parseFile("manual.TXT", ...) and verify that it produces a fallback block with the expected path, content, file hash, and segment hash. This test will now fail if the .toLowerCase() normalization in parseFile is removed

src/services/code-index/processors/__tests__/parser.txt.spec.ts
WebMad: Thanks — I added segmentHash: expect.any(String) to the .txt block assertions. I also added a focused unit test for the shared chunking path: it verifies that the first chunk adds its hash to seenSegmentHashes, while processing the same chunk again with the same set emits no duplicate block

src/services/tree-sitter/index.ts
WebMad: Good catch - I updated the guard to cover all fallback extensions, not just .txt. I moved the shared fallback extension list and predicate into a neutral module so both code indexing and the Tree-sitter flow can use the same source of truth without creating an import cycle. I also extended the integration test to verify that .txt, .vb, .scala, and .swift all return undefined without attempting to load a Tree-sitter parser

WebMad:
Thank you, @edelauna, for taking the time to review this and for the thoughtful feedback. I’ve addressed the suggestions and pushed the updates. I’m really glad to contribute and be part of the effort to make the project better

src/services/tree-sitter/index.ts
WebMad: Quick follow-up: I noticed the Scala tests unexpectedly started failing after this change. It turns out Scala and Swift use fallback chunking for indexing but still support structural parsing elsewhere. I split the extensions into fallbackExtensions and nonStructuralExtensions, and the affected tests are passing now

edelauna (approved):
Thank you for this contribution and for addressing all the feedback, very clean PR.

PR #941, opened 2026-07-18
Add Dart support to codebase indexing

### LINKED ISSUE #940: [ENHANCEMENT] Add Dart support to codebase indexing

Zoo Code does not currently support Dart when indexing a codebase. As a result, codebase indexing cannot provide useful context for Dart projects

This affects Dart and Flutter developers who use Zoo Code in Dart repositories. I work primarily with Dart projects, so the codebase indexing feature is not useful for most of my work

Zoo Code should support indexing Dart source code so that it can understand and retrieve relevant context from Dart and Flutter projects

Dart indexing should work consistently with the other supported languages and handle common Dart and Flutter project structures

### Request checklist

https://github.com/Zoo-Code-Org/Zoo-Code/issues/751

- Dart source files are included in codebase indexing.
- Relevant Dart code can be found through codebase search.
- Common Dart and Flutter project structures are supported.
- Adding Dart support does not negatively affect indexing for existing languages.

_No response_

_No response_

WebMad:
### Related GitHub Issue

Closes: #940

### Description

Adds Dart support to Tree-sitter-based codebase indexing.

- Registers `.dart` as a supported source extension and loads `tree-sitter-dart.wasm`.
- Adds a Dart definitions query adapted from tree-sitter-dart's canonical tags query.
- Captures classes, methods, constructors, factory and const constructors, getters, setters, operators, mixins, extensions, extension types, enums, type aliases, and top-level functions.
- Adds focused parser and definition-extraction coverage using representative Dart declarations.
- Keeps generated Dart files indexable; projects can continue excluding generated output through `.gitignore` or `.rooignore` when desired.

Reviewers should pay particular attention to the query patterns for constructors and top-level accessors, which differ structurally from class method signatures in the Dart grammar.

### Test Procedure

Automated tests:

```bash
cd src
npx vitest run services/tree-sitter/__tests__/languageParser.spec.ts services/tree-sitter/__tests__/parseSourceCodeDefinitions.dart.spec.ts
```

Expected result: 2 test files and 8 tests pass.

The repository pre-commit and pre-push hooks were also run successfully:

- ESLint across all workspace packages.
- TypeScript type-checking across all workspace packages.

Manual verification:

1. Open a Dart or Flutter workspace containing `.dart` files.
2. Rebuild the codebase index.
3. Search for a Dart class, mixin, enum, extension, constructor, getter/setter, or top-level function.
4. Confirm that the corresponding Dart definition is returned as indexed context.

### Pre-Submission Checklist

### Screenshots / Videos

Not applicable — this change has no UI impact.

### Documentation Updates

### Additional Notes

No generated-file suffixes such as `*.g.dart` or `*.freezed.dart` are hard-coded into the exclusion logic. Existing `.gitignore` and `.rooignore` behavior remains the source of truth, preserving access to generated APIs when projects intentionally index them.

### Get in Touch

GitHub: @WebMad

## Summary by CodeRabbit

* **New Features**
  * Added Dart (`.dart`) support for source-code definition parsing, including Dart query coverage for classes, types, methods/constructors (including redirecting factories), mixins, extensions (including extension types), enums, and top-level functions/getters/setters.
* **Bug Fixes**
  * Improved definition range extraction to better include multiline bodies and avoid duplicate declaration lines.
* **Tests**
  * Added Dart fixture parsing, validated Dart parser/query loading, and expanded the Dart definition parsing suite (including redirecting constructors and exclusion of nested/local functions).


src/services/tree-sitter/__tests__/parseSourceCodeDefinitions.dart.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+	wasmFile: "tree-sitter-dart.wasm",
+	queryString: dartQuery,
+	extKey: "dart",
+}
+
+describe("parseSourceCodeDefinitionsForFile with Dart", () => {
+	it("should capture common Dart declarations", async () => {
+		const result = await testParseSourceCodeDefinitions("/test/file.dart", sampleDartContent, dartOptions)
+
+		expect(result).toMatch(/abstract class Animal/)
```
edelauna: The assertions here match substrings of the echoed source line (the pipeline outputs `lines[startLine]`, the raw source text). That means a mutation breaking every `@name` capture in the query would still pass all 19 assertions — the `@definition.*` captures alone are enough to echo the source line.

Sibling specs address this with two additions:
1. The `\d+--\d+ \| ` line-range prefix, which at least confirms the output came through the pipeline in the right format: `expect(result).toMatch(/\d+--\d+ \| abstract class Animal/)`
2. A count assertion that catches dropped or duplicated captures: `expect(result!.split("\n").filter(l => l.includes(" | ")).length).toBe(N)`

One thing I wasn't sure about: `getMinComponentLines()` defaults to 4, so single-line declarations like `const Point()` and `int get x => 0` would normally be filtered by the `lineCount < 4` check. Are assertions like `toMatch(/const Point\(\)/)` expected to pass, and if so through which code path?

src/services/tree-sitter/queries/dart.ts
```diff
@@ -0,0 +1,65 @@
...
+(method_signature
+  (factory_constructor_signature
+    (identifier) @name)) @definition.method
+
+(method_signature
+  (constructor_signature
+    name: (identifier) @name)) @definition.method
+
+(method_signature
+  (operator_signature)) @definition.method
```
edelauna: The upstream `tags.scm` this is adapted from has a bare `(method_signature) @definition.method` catch-all (their line 34) that fires for any `method_signature` not matched by the more-specific rules above. Without it, abstract method declarations and any future grammar node types added to the Dart grammar would produce zero captures rather than a fallback `@definition.method`.

Was this intentionally omitted, or an oversight from the adaptation?

src/services/tree-sitter/queries/dart.ts
```diff
@@ -0,0 +1,65 @@
...
+
+(constructor_signature
+  name: (identifier) @name) @definition.method
```
edelauna, line 24: There is also a `(method_signature (constructor_signature ...))` pattern at line 29–31. For a constructor inside a class body, a `method_signature` wraps it, so both patterns could match the same constructor.

For the single-line constructors in the fixture this is fine — both nodes share the same `startLine`/`endLine` so the `lineKey` dedup in `processCaptures` suppresses the duplicate. But for a multi-line constructor in real Dart code, the two nodes have different line spans → different `lineKey` values → both get output, and the constructor could appear twice in the index.

Is there a case where a `constructor_signature` appears outside a `method_signature` that this standalone pattern is meant to catch?

edelauna (changes requested):
Thanks so much for this contribution, had some comments regarding the queries.

src/services/tree-sitter/__tests__/parseSourceCodeDefinitions.dart.spec.ts
WebMad: Thanks for catching this. I added line-range prefixes and an exact definition count. Single-line declarations are included because the test helper sets the minimum component line count to 0.

src/services/tree-sitter/queries/dart.ts
WebMad: This was an oversight. I restored the bare method_signature fallback and removed the specific patterns it fully covers.

src/services/tree-sitter/queries/dart.ts
WebMad: Good catch. I removed the wrapped constructor pattern and added a regression test verifying a multiline constructor is emitted once. The standalone patterns remain because the fallback does not capture those AST shapes.

WebMad:
Thanks for the detailed review. I addressed all three points in 8b19e9b5f:

- strengthened the Dart assertions with line-range prefixes and an exact definition count;
- restored the upstream `method_signature` fallback and removed redundant patterns;
- removed the conflicting wrapped constructor capture and added a multiline-constructor regression test.

The focused Dart and language-parser tests pass.

WebMad:
The latest CodeRabbit finding about the Dart extension format is a false positive.

The entry in `src/services/tree-sitter/index.ts` is intentionally `"dart"` because the entire array is normalized immediately afterward with `.map((e) => `.${e}`)`. The resulting value is therefore `.dart`, which correctly matches `path.extname(filePath)`. Changing the entry to `".dart"` as suggested would produce `"..dart"` and break routing.

The Dart definition tests already exercise the public `parseSourceCodeDefinitionsForFile()` path via `testParseSourceCodeDefinitions()`, including the extension allowlist check. I also ran the focused tests (14/14 passing) and the complete tree-sitter test suite (305 passing, 12 skipped). No code change is needed for this finding.

edelauna (approved):
Thanks so much for your contribution!

PR #950, opened 2026-07-19
chore(lint): add core/webview to no-floating-promises ratchet

### LINKED ISSUE #949: chore(lint): add core/webview to no-floating-promises ratchet
## Background

The repo is rolling out `@typescript-eslint/no-floating-promises` directory by directory (see `src/eslint.config.mjs`). Currently covered: `activate/**` and `core/task/**`. `core/webview/` has not been added yet.

## Problem

Running the rule against `core/webview/**` today surfaces **89 violations**, almost all in `webviewMessageHandler.ts`. Not all are bugs — many are intentional fire-and-forget calls — but the noise means real ordering bugs hide undetected. One such bug was found in PR #912: `invalidateAll()` was called without `await` immediately before `reconcile()`, silently swallowing errors and making the ordering guarantee non-obvious.

## What needs doing

1. Triage the 89 violations in `core/webview/` — for each one either:
   - Add `await` if ordering or error handling matters
   - Prefix with `void` if fire-and-forget is genuinely intentional
2. Add `core/webview/**/*.ts` to the `files` glob in `src/eslint.config.mjs` so future violations are caught at lint time

## Reproduction

```bash
cd src
npx eslint --rule '{"@typescript-eslint/no-floating-promises": "warn"}' \
  --parser-options 'project:true,tsconfigRootDir:.' \
  'core/webview/**/*.ts'
```

89 warnings as of commit 38ae51ec.

morgan-coded:
Picked up #949 from the #912 review-thread invitation and triaged the live `core/webview` set: 75 calls are awaited and 14 are explicitly voided. Two of the voids are flagged for your review as genuinely ambiguous: the constructor `updateGlobalState("codebaseIndexModels", ...)` in `ClineProvider.ts` (constructors cannot await) and `workspaceTracker?.initializeFilePaths()` in the `webviewDidLaunch` case, where the existing "Don't await" comment is preserved because a workspace scan would sit on the launch path. Both keep baseline timing and rejection behavior exactly.

Added `core/webview/**/*.ts` to the existing ratchet. I ran the issue repro with zero warnings across all three ratcheted directories, the full suite with 6,785 passing, plus lint and typecheck.

Fixes #949

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability and in-order delivery of panel updates across task, marketplace, indexing, and related UI/status flows.
  * Fixed checkpoint restore to fully complete task abortion before continuing.
  * Improved error handling when opening rule and skill files/directories.
  * Prevented reading files outside the workspace boundary.
* **Tests**
  * Expanded and hardened panel message-handler tests for async behavior, payloads, and non-floating promise coverage.
* **Chores**
  * Strengthened async handling to reduce missed errors and inconsistent states during UI/status updates.


src/core/webview/webviewMessageHandler.ts
```diff
@@ -3003,12 +3005,12 @@ export const webviewMessageHandler = async (
 
 					const currentState = manager.state
 					if (currentState === "Standby" || currentState === "Error") {
-						manager.startIndexing()
+						void manager.startIndexing()
```
edelauna: Prefixing `manager.startIndexing()` with `void` satisfies ESLint but leaves promise rejections unhandled if the operation throws (e.g. Qdrant timeouts or embedding errors). Consider appending `.catch(...)` to handle errors cleanly:

```suggestion
						void manager.startIndexing().catch((err) => provider.log(`Indexing error: ${err}`)) 
```

edelauna (changes requested):
Thank you for this! Wondering if we could add a catch block now to make debugging errors easier.

Also - since we're touching this file, would we be able to bump up the test coverage to make codecov happy?

morgan-coded:
Pushed the revision: your suggested catch is on the review site plus the three sibling `startIndexing()` calls in the same handler, and `initializeFilePaths()` got the same treatment since its `listFiles` call can reject — all with the same logging shape. The remaining `void` sites either post through `postMessageToWebview`, which catches disposed-view failures internally, or are intentionally backgrounded. Coverage on the changed handler paths went from ~26% to 90% with 14 focused tests (controlled rejections, no timers).

edelauna (approved):

PR #991, opened 2026-07-23
refactor(types): use canonical identifiers for default models

### LINKED ISSUE #954: [ENHANCEMENT] Migrate provider default-model selection to canonical identifiers

Provider names are repeated as raw string literals in the [default-model selection logic](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/packages/types/src/providers/index.ts). This duplicates values from the canonical provider registry and allows the two definitions to drift apart.

This affects maintainers and contributors when adding, renaming, or retiring a provider with a default model. They currently need to keep the canonical registry and `getProviderDefaultModelId()` synchronized manually.

This is part of the provider identifier centralization tracked by #944 and builds on #951 and #953.

Default-model selection should reference the canonical provider identifiers while preserving all existing model defaults and fallback behavior. No persisted values or user-facing behavior should change.

Keep this change focused on default-model selection in the shared types package. Other production consumers should be migrated in separate follow-up issues under #944.

### Request checklist

Parent tracking issue: #944

Foundational identifier registry: #951

Provider category migration: #953

- Default-model selection references canonical provider identifiers instead of repeating provider-name literals.
- Existing default model IDs and fallback behavior remain unchanged.
- Existing serialized provider values remain unchanged.
- Focused tests cover representative providers and fallback behavior.
- Existing provider-related tests continue to pass.

### Relevant code locations

- [`packages/types/src/providers/index.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/packages/types/src/providers/index.ts) — migrate the provider switch in `getProviderDefaultModelId()`.
- [`packages/types/src/provider-identifiers.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/packages/types/src/provider-identifiers.ts) — canonical active provider values to consume.
- [`packages/types/src/__tests__`](https://github.com/Zoo-Code-Org/Zoo-Code/tree/main/packages/types/src/__tests__) — add focused coverage for representative static, dynamic, custom, and fallback results.
- [`webview-ui/src/components/settings/utils/providerModelConfig.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/webview-ui/src/components/settings/utils/providerModelConfig.ts) and [`webview-ui/src/components/ui/hooks/useSelectedModel.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/webview-ui/src/components/ui/hooks/useSelectedModel.ts) — downstream callers to use when verifying that the public behavior remains unchanged; their own provider comparisons belong to #959.

Replace internal provider-name switch cases in `packages/types/src/providers/index.ts` with values from the canonical provider registry introduced by #951. Keep the public API and runtime behavior unchanged.

This is intentionally a narrow migration and will temporarily leave raw provider literals in other production consumers. Those locations will be migrated through separate issues tracked by #944.

WebMad:
## Summary

- replace raw provider-name switch cases in `getProviderDefaultModelId()` with the canonical `providerIdentifiers` registry
- preserve all existing static, regional, custom/local, and fallback default-model behavior
- add focused tests using TDD triangulation across representative provider categories

Closes #954

## TDD

The new test first overrides the canonical OpenRouter identifier. Before the implementation change, the literal-based switch fails that case and returns the Anthropic fallback. Migrating the switch to `providerIdentifiers` makes it pass. Additional cases triangulate the behavior across static/internal, region-dependent, custom/local, and fallback providers.

## Verification

- `cd packages/types && npx vitest run src/__tests__/provider-default-model.test.ts src/__tests__/provider-identifiers.test.ts` — 17 passed
- `cd packages/types && npm test` — 253 passed
- `cd packages/types && npm run check-types` — passed
- `cd packages/types && npm run lint` — passed
- repository lint pre-commit hook — passed

## Note

The repository-wide pre-push type check currently reports unrelated pre-existing errors involving MiniMax M3, Moonshot router typing, LiteLLM exports, and `abandonSubtaskWithId`. The affected `@roo-code/types` package type check passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved provider default-model selection to consistently match supported provider identifiers.
  * Preserved region-specific defaults (including China-dependent behavior) and maintained Anthropic fallback behavior.
  * Prevented unintended automatic defaults for custom/local selections (such as OpenAI, Ollama, and LM Studio).
* **Tests**
  * Added new Vitest coverage for provider matching, regional defaults, custom/local cases, and Anthropic fallback scenarios.

### DIFF
--- a/packages/types/src/__tests__/provider-default-model.test.ts
+++ b/packages/types/src/__tests__/provider-default-model.test.ts
@@ -1,0 +1,50 @@
+import type { ProviderName } from "../provider-settings.js"
+
+vi.mock("../provider-identifiers.js", async (importOriginal) => {
+	const actual = await importOriginal<typeof import("../provider-identifiers.js")>()
+
+	return {
+		...actual,
+		providerIdentifiers: {
+			...actual.providerIdentifiers,
+			openrouter: "canonical-openrouter-test-value",
+		},
+	}
+})
+
+import { providerIdentifiers } from "../provider-identifiers.js"
+import {
+	anthropicDefaultModelId,
+	getProviderDefaultModelId,
+	mainlandZAiDefaultModelId,
+	openRouterDefaultModelId,
+	vscodeLlmDefaultModelId,
+} from "../providers/index.js"
+
+describe("getProviderDefaultModelId", () => {
+	it("selects a static default through the canonical provider identifier", () => {
+		expect(getProviderDefaultModelId(providerIdentifiers.openrouter as ProviderName)).toBe(openRouterDefaultModelId)
+	})
+
+	it("triangulates static selection with another provider category", () => {
+		expect(getProviderDefaultModelId(providerIdentifiers.vscodeLm)).toBe(vscodeLlmDefaultModelId)
+	})
+
+	it("preserves region-dependent defaults", () => {
+		expect(getProviderDefaultModelId(providerIdentifiers.zai, { isChina: true })).toBe(mainlandZAiDefaultModelId)
+	})
+
+	it.each([providerIdentifiers.openai, providerIdentifiers.ollama, providerIdentifiers.lmstudio])(
+		"returns an empty default for custom or locally selected models from %s",
+		(provider) => {
+			expect(getProviderDefaultModelId(provider)).toBe("")
+		},
+	)
+
+	it.each([providerIdentifiers.anthropic, providerIdentifiers.geminiCli, providerIdentifiers.fakeAi])(
+		"preserves the Anthropic fallback for %s",
+		(provider) => {
+			expect(getProviderDefaultModelId(provider)).toBe(anthropicDefaultModelId)
+		},
+	)
+})
--- a/packages/types/src/__tests__/provider-default-model.test.ts
+++ b/packages/types/src/__tests__/provider-default-model.test.ts
@@ -16,6 +16,7 @@ import { providerIdentifiers } from "../provider-identifiers.js"
 import {
 	anthropicDefaultModelId,
 	getProviderDefaultModelId,
+	internationalZAiDefaultModelId,
 	mainlandZAiDefaultModelId,
 	openRouterDefaultModelId,
 	vscodeLlmDefaultModelId,
@@ -32,6 +33,7 @@ describe("getProviderDefaultModelId", () => {
 
 	it("preserves region-dependent defaults", () => {
 		expect(getProviderDefaultModelId(providerIdentifiers.zai, { isChina: true })).toBe(mainlandZAiDefaultModelId)
+		expect(getProviderDefaultModelId(providerIdentifiers.zai)).toBe(internationalZAiDefaultModelId)
 	})
 
 	it.each([providerIdentifiers.openai, providerIdentifiers.ollama, providerIdentifiers.lmstudio])(
--- a/packages/types/src/__tests__/provider-default-model.test.ts
+++ b/packages/types/src/__tests__/provider-default-model.test.ts
@@ -41,6 +41,8 @@ describe("getProviderDefaultModelId", () => {
 	})
 
 	it("preserves region-dependent defaults", () => {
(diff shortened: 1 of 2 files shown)


packages/types/src/providers/index.ts
```diff
@@ -73,71 +76,70 @@ export function getProviderDefaultModelId(
...
-		case "mimo":
+		case providerIdentifiers.mimo:
 			return mimoDefaultModelId
-		case "zai":
+		case providerIdentifiers.zai:
 			return options?.isChina ? mainlandZAiDefaultModelId : internationalZAiDefaultModelId
-		case "openai-native":
+		case providerIdentifiers.openaiNative:
+			// TODO(#992): Replace this stale fallback with openAiNativeDefaultModelId.
 			return "gpt-4o" // Based on openai-native patterns
```
edelauna, line 107: Should this use `openAiNativeDefaultModelId` (currently `"gpt-5.6-sol"` in `providers/openai.ts`) rather than the hardcoded literal? The constant is already re-exported by the barrel on line 13, so no new import is needed.

packages/types/src/__tests__/provider-default-model.test.ts
```diff
@@ -0,0 +1,61 @@
...
+	it.each([
+		[providerIdentifiers.kimiCode, kimiCodeDefaultModelId],
+		[providerIdentifiers.zooGateway, zooGatewayDefaultModelId],
+	])("preserves the %s default added on main", (provider, expectedModelId) => {
+		expect(getProviderDefaultModelId(provider)).toBe(expectedModelId)
+	})
+
+	it("preserves region-dependent defaults", () => {
+		expect(getProviderDefaultModelId(providerIdentifiers.zai, { isChina: true })).toBe(mainlandZAiDefaultModelId)
+		expect(getProviderDefaultModelId(providerIdentifiers.zai)).toBe(internationalZAiDefaultModelId)
```
edelauna, line 47: Both `mainlandZAiDefaultModelId` and `internationalZAiDefaultModelId` are currently `"glm-4.7"`, so swapping the ternary arms in the source would still pass these assertions. Worth adding `expect(mainlandZAiDefaultModelId).not.toBe(internationalZAiDefaultModelId)` as a guard — or a comment explaining that the branch is intentionally untestable until the values diverge?

edelauna (changes requested):
Thanks - noticed a small quality of life change we can make while updating this file.

packages/types/src/__tests__/provider-default-model.test.ts
WebMad: Good catch. Both constants currently resolve to `glm-4.7`, so a swapped ternary cannot be detected from the public return value today. An `expect(...).not.toBe(...)` guard would fail against the valid current data rather than test the resolver. I added a comment documenting this limitation while retaining assertions for both call paths; once the defaults diverge, the existing assertions will distinguish the branches.

packages/types/src/providers/index.ts
WebMad: The canonical default should indeed be used, but changing it here would alter runtime behavior from `gpt-4o` to `gpt-5.6-sol`, conflicting with #954’s acceptance criterion to preserve existing defaults and fallback behavior. The underlying bug is tracked separately in #992, and this PR now has a TODO linking to it, so I am keeping that behavioral fix out of this narrowly scoped identifier migration.

WebMad: There may also be more consumers and assumptions tied to the current OpenAI Native fallback than this one location makes apparent. Updating this shared resolver could therefore have effects beyond the immediate return value, and I do not want to introduce uncertain downstream behavior as part of an identifier-only refactor. I would prefer to trace those call sites, add focused regression coverage, and make the behavioral change in a separate PR for #992.

edelauna (approved):
Thanks!

PR #1019, opened 2026-07-25
refactor(shared): use canonical profile provider identifiers

### LINKED ISSUE #956: [ENHANCEMENT] Migrate profile model resolution to canonical provider identifiers

Profile model resolution in [`src/shared/ProfileValidator.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/shared/ProfileValidator.ts) uses raw provider-name literals to decide which model field belongs to each provider. These literals can drift from the canonical provider definitions and other provider-specific logic.

This affects users whose saved provider profiles are validated or resolved, and maintainers who add or modify providers with different model fields. An incomplete update can make a valid profile appear invalid or resolve the wrong model ID.

This is part of the provider identifier centralization tracked by #944 and depends on the foundational registry introduced by #951.

Profile validation and model resolution should reference canonical provider identifiers while preserving the current model-field mappings, validation results, and compatibility with existing saved profiles.

Keep the migration behavior-preserving and focused on profile validation and model resolution. Existing serialized profile values must remain unchanged.

### Request checklist

Parent tracking issue: #944

Foundational identifier registry: #951

- Profile model resolution references canonical provider identifiers.
- Existing provider-to-model-field mappings remain unchanged.
- Existing saved profiles remain compatible.
- Focused tests cover representative provider-specific model fields and fallback behavior.
- Existing profile validation tests continue to pass.

### Relevant code locations

- [`src/shared/ProfileValidator.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/shared/ProfileValidator.ts) — migrate provider cases in profile model ID resolution.
- [`src/shared/__tests__/ProfileValidator.spec.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/shared/__tests__/ProfileValidator.spec.ts) — preserve and extend coverage for provider-specific model fields and invalid profiles.
- [`packages/types/src/provider-identifiers.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/packages/types/src/provider-identifiers.ts) — canonical values to consume.
- [`packages/types/src/provider-settings.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/packages/types/src/provider-settings.ts) — shared `ProviderSettings` fields and `getModelId()` behavior to keep aligned; migrating unrelated logic here is outside this issue unless required for compatibility.
- [`src/core/config/ProviderSettingsManager.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/core/config/ProviderSettingsManager.ts), [`src/core/config/ContextProxy.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/core/config/ContextProxy.ts), and [`src/core/config/importExport.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/core/config/importExport.ts) — persisted profile consumers to consider when verifying backward compatibility; broader extension comparisons belong to #958.

Replace internal provider-name cases in `src/shared/ProfileValidator.ts` with values from the canonical provider registry. Preserve the existing model-field selection and validation behavior.

Profile data is persisted, so accidental changes to provider values or field mappings could affect existing users. Tests should explicitly confirm compatibility with representative saved profiles.

WebMad:
## Summary

- replace raw provider-name literals in profile model resolution with canonical `providerIdentifiers` values
- preserve every existing provider-to-model-field mapping and fallback behavior
- add focused registry-driven coverage for provider-specific model fields and missing-model fallback behavior

## TDD

Implemented with iterative red-green triangulation:

1. Added an OpenAI canonical-identifier regression test and confirmed it failed.
2. Migrated the OpenAI case minimally and confirmed it passed.
3. Triangulated across `apiModelId`, OpenRouter, and VS Code LM mappings and confirmed the tests failed before migrating those cases.
4. Triangulated the remaining provider-specific mappings, then completed the canonical migration.
5. Covered fallback behavior for profiles without a model ID, including provider-level `allowAll`.

## Verification

- `cd src && npx vitest run shared/__tests__/ProfileValidator.spec.ts` — 50 passed
- `cd src && npx vitest run shared` — 410 passed
- focused ESLint — passed
- `cd src && npm run check-types` — passed
- repository pre-commit lint — passed
- repository pre-push type checks — passed
- `git diff --check` — passed

Closes #956

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile validation across supported AI providers by resolving provider/model identifiers consistently.
  * Preserved existing fallback behavior for unknown providers and configurations where model lists are missing.

* **Tests**
  * Expanded automated coverage for provider-specific model resolution.
  * Added scenarios to verify correct “missing-model” handling when provider permissions are toggled and model lists are undefined.

### DIFF
--- a/src/shared/__tests__/ProfileValidator.spec.ts
+++ b/src/shared/__tests__/ProfileValidator.spec.ts
@@ -1,11 +1,97 @@
 // npx vitest run src/shared/__tests__/ProfileValidator.spec.ts
 
-import type { ProviderSettings, OrganizationAllowList } from "@roo-code/types"
+import { providerIdentifiers, type ProviderSettings, type OrganizationAllowList } from "@roo-code/types"
 
 import { ProfileValidator } from "../ProfileValidator"
 
 describe("ProfileValidator", () => {
 	describe("isProfileAllowed", () => {
+		it.each([
+			["openai", { openAiModelId: "model" }],
+			["anthropic", { apiModelId: "model" }],
+			["openaiNative", { apiModelId: "model" }],
+			["bedrock", { apiModelId: "model" }],
+			["vertex", { apiModelId: "model" }],
+			["gemini", { apiModelId: "model" }],
+			["mistral", { apiModelId: "model" }],
+			["deepseek", { apiModelId: "model" }],
+			["xai", { apiModelId: "model" }],
+			["sambanova", { apiModelId: "model" }],
+			["fireworks", { apiModelId: "model" }],
+			["friendli", { apiModelId: "model" }],
+			["litellm", { litellmModelId: "model" }],
+			["lmstudio", { lmStudioModelId: "model" }],
+			["vscodeLm", { vsCodeLmModelSelector: { id: "model" } }],
+			["openrouter", { openRouterModelId: "model" }],
+			["ollama", { ollamaModelId: "model" }],
+			["requesty", { requestyModelId: "model" }],
+			["unbound", { unboundModelId: "model" }],
+		])("resolves %s model fields through canonical identifiers", (identifierKey, profileSettings) => {
+			const identifiers = providerIdentifiers as Record<string, string>
+			const originalIdentifier = identifiers[identifierKey]
+			const canonicalIdentifier = `canonical-${identifierKey}`
+			const modelId = "model"
+
+			try {
+				identifiers[identifierKey] = canonicalIdentifier
+
+				const profile = {
+					apiProvider: canonicalIdentifier,
+					...profileSettings,
+				} as ProviderSettings
+				const allowList: OrganizationAllowList = {
+					allowAll: false,
+					providers: {
+						[canonicalIdentifier]: { allowAll: false, models: [modelId] },
+					},
+				}
+
+				expect(ProfileValidator.isProfileAllowed(profile, allowList)).toBe(true)
+			} finally {
+				identifiers[identifierKey] = originalIdentifier
+			}
+		})
+
+		it.each([
+			{ identifierKey: "fakeAi", profileSettings: {}, modelId: undefined, expected: false },
+			{
+				identifierKey: "fakeAi",
+				profileSettings: {},
+				modelId: undefined,
+				expected: true,
+				providerAllowAll: true,
+			},
+		])(
+			"preserves canonical fallback behavior when provider allowAll is $providerAllowAll",
+			({ identifierKey, profileSettings, modelId, expected, providerAllowAll = false }) => {
+				const identifiers = providerIdentifiers as Record<string, string>
+				const originalIdentifier = identifiers[identifierKey]
+				const canonicalIdentifier = `canonical-${identifierKey}`
+
+				try {
+					identifiers[identifierKey] = canonicalIdentifier
+
+					const profile = {
+						apiProvider: canonicalIdentifier,
+						...profileSettings,
+					} as unknown as ProviderSettings
+					const allowList: OrganizationAllowList = {
+						allowAll: false,
+						providers: {
+							[canonicalIdentifier]: {
+								allowAll: providerAllowAll,
+								models: modelId ? [modelId] : undefined,
+							},
+						},
+					}
+
+					expect(ProfileValidator.isProfileAllowed(profile, allowList)).toBe(expected)
+				} finally {
+					identifiers[identifierKey] = originalIdentifier
+				}
+			},
+		)
+
 		it("should allow any profile when allowAll is true", () => {
 			const allowList: OrganizationAllowList = {
 				allowAll: true,
--- a/src/shared/__tests__/ProfileValidator.spec.ts
+++ b/src/shared/__tests__/ProfileValidator.spec.ts
@@ -27,38 +27,30 @@ describe("ProfileValidator", () => {
 			["requesty", { requestyModelId: "model" }],
 			["unbound", { unboundModelId: "model" }],
 		])("resolves %s model fields through canonical identifiers", (identifierKey, profileSettings) => {
-			const identifiers = providerIdentifiers as Record<string, string>
-			const originalIdentifier = identifiers[identifierKey]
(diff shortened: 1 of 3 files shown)


src/shared/__tests__/ProfileValidator.spec.ts
```diff
@@ -1,11 +1,97 @@
...
+					...profileSettings,
+				} as ProviderSettings
+				const allowList: OrganizationAllowList = {
+					allowAll: false,
+					providers: {
+						[canonicalIdentifier]: { allowAll: false, models: [modelId] },
+					},
+				}
+
+				expect(ProfileValidator.isProfileAllowed(profile, allowList)).toBe(true)
```
edelauna: Could a mutant `getModelIdFromProfile` that always returns `"model"` still pass all 19 rows here? Every allowList entry uses `models: ["model"]` and every profile fixture sets the model field to `"model"`, so the switch dispatch logic isn't discriminated — the outcome would be the same regardless of which branch ran.

A complementary negative assertion (same profile, but allowList has `models: ["other-model"]`) would close the gap:

```suggestion
				expect(ProfileValidator.isProfileAllowed(profile, allowList)).toBe(true)

				const negativeAllowList: OrganizationAllowList = {
					allowAll: false,
					providers: { [canonicalIdentifier]: { allowAll: false, models: ["other-model"] } },
				}
				expect(ProfileValidator.isProfileAllowed(profile, negativeAllowList)).toBe(false)
```

src/shared/__tests__/ProfileValidator.spec.ts
```diff
@@ -1,11 +1,97 @@
...
+						allowAll: false,
+						providers: {
+							[canonicalIdentifier]: {
+								allowAll: providerAllowAll,
+								models: modelId ? [modelId] : undefined,
+							},
+						},
+					}
+
+					expect(ProfileValidator.isProfileAllowed(profile, allowList)).toBe(expected)
```
edelauna: Does mutating `providerIdentifiers.fakeAi` actually exercise the canonical identifier path here? In `ProfileValidator.ts`, `case providerIdentifiers.fakeAi:` falls into `default: return undefined` — so the switch returns `undefined` whether or not the identifier matched, and `isModelAllowed` is never reached. The test validates the `allowAll`-gated no-modelId behavior, but the identifier mutation itself is inert. Would a row using a real provider with a mismatched model (e.g. `openai` + wrong model in the allowList) give this block more signal?

src/shared/__tests__/ProfileValidator.spec.ts
```diff
@@ -1,11 +1,97 @@
...
+					providers: {
+						[canonicalIdentifier]: { allowAll: false, models: [modelId] },
+					},
+				}
+
+				expect(ProfileValidator.isProfileAllowed(profile, allowList)).toBe(true)
+			} finally {
+				identifiers[identifierKey] = originalIdentifier
+			}
+		})
```
edelauna, line 54: The pre-existing tests below (~lines 256–375) still use raw wire-format strings (`"openai-native"`, `"vscode-lm"`, `"fake-ai"`) alongside `as any` casts. With the canonical identifier pattern established here, is it worth updating those blocks to use `providerIdentifiers.*` too? If a provider string is ever renamed, the old blocks would pass against the wrong value while the new `it.each` above would correctly catch it.

edelauna (changes requested):
Thanks, had some feedback around test quality, but otherwise looks good.

WebMad:
> Thanks, had some feedback around test quality, but otherwise looks good.

Thank you! I’m still learning how to write good tests, so your feedback is very helpful to me. I may make mistakes or repeat them sometimes, partly because of the language barrier, but I’ll do my best to improve

src/shared/__tests__/ProfileValidator.spec.ts
```diff
@@ -1,11 +1,97 @@
...
+			},
+		])(
+			"preserves canonical fallback behavior when provider allowAll is $providerAllowAll",
+			({ identifierKey, profileSettings, modelId, expected, providerAllowAll = false }) => {
+				const identifiers = providerIdentifiers as Record<string, string>
+				const originalIdentifier = identifiers[identifierKey]
+				const canonicalIdentifier = `canonical-${identifierKey}`
+
+				try {
+					identifiers[identifierKey] = canonicalIdentifier
```
WebMad: I don’t think we should mutate a global variable. Could we rewrite this test using a different approach?

src/shared/__tests__/ProfileValidator.spec.ts
WebMad: Addressed in 965cea6b6. Each canonical-provider row now also checks a mismatched allow-list model and expects rejection, so a constant `"model"` implementation cannot satisfy both assertions.

src/shared/__tests__/ProfileValidator.spec.ts
WebMad: Addressed in 965cea6b6. I removed the inert `fakeAi` identifier mutation and now exercise missing-model fallback with the real canonical OpenAI provider in both `allowAll: false` and `allowAll: true` cases.

src/shared/__tests__/ProfileValidator.spec.ts
WebMad: Addressed in 965cea6b6. The pre-existing provider fixtures now use `providerIdentifiers.*`, including OpenAI Native, VS Code LM, LiteLLM, and Fake AI, and the unnecessary `as any` casts were removed.

src/shared/__tests__/ProfileValidator.spec.ts
WebMad: Addressed in 616600711. The test now reads each value directly from `providerIdentifiers` and no longer mutates or restores the shared global registry.

edelauna (approved):

PR #1028, opened 2026-07-26
feat(friendli): fetch model list dynamically from /v1/models

Lee-Si-Yoon:
Friendli model list is now fetched live from `GET /v1/models` instead of being hardcoded. The endpoint is public (no auth), same as Vercel AI Gateway and OpenRouter in this repo.

- New fetcher (`fetchers/friendli.ts`): zod-validated axios call, maps API fields to `ModelInfo` (pricing, cache, reasoning effort, image modality, deprecation)
- `FriendliHandler` fire-and-forgets a model fetch in constructor; `getModel()` prefers dynamic data, falls back to static `friendliModels`
- `"friendli"` moved to `dynamicProviders`; UI uses `routerModels.friendli` for the model picker
- 14 fetcher tests, 3 webview tests, 29 existing handler tests pass


Lee-Si-Yoon:
Closing this PR. At the time it was made, https://api.friendli.ai/serverless/v1/models was not up-to-date, so the hardcode approach was used as a fallback. The endpoint is now reliable, so we can trust it going forward. Will open a fresh PR that relies on the live /v1/models endpoint.

PR #1033, opened 2026-07-26
fix(shell): delegate VS Code profile reads to Terminal methods (issue #634)

### LINKED ISSUE #634: [BUG] It always assumes the wrong type of terminal
**Describe the bug**
Any task that involves running a terminal command seems to (blindly) assume Powershell. And fair is fair, when I myself open the terminal, it opens Powershell as expected. But a Zoo task always ends up trying a powershell command, failing with a cmd.exe error message, then the model tries to prepend it with `pwsh` which also fails, because that's not how you call a powershell command from cmd.exe, and finally gives up and issues an actual cmd.exe compatible command.

**To Reproduce**
1. Make sure the project is in Powershell by default
2. Start any task that involves executing a CLI command
3. Wait and see it fail

**Expected behavior**
I want to be able to either:
* Specify which CLI the model should expect (worst option)
* Set the CLI tool Zoo provides to the model, to a type that I specify in Zoo's settings
* Zoo provides the same CLI as what's configured in the VSCode user/project/workspace settings, whichever applies (best option).

In either case Zoo should tell the model which type of terminal it has made available for the model to use. Don't let it guess. Also, the instructions for _other_ terminals should be removed from the system prompt, to avoid confusing the model.

3.60

edelauna:
### Related GitHub Issue

Closes: #634

### Description

**Root cause:** On Windows, `getShell()` (used to build the system prompt) called `config.get("defaultProfile.windows")` which merges all VS Code scopes including workspace. `Terminal.getConfiguredDefaultProfileName()` (used when actually opening a terminal) calls `config.inspect().globalValue` — intentionally excluding workspace scope, because `terminal.integrated.defaultProfile.*` is an APPLICATION-scoped setting. A workspace `.vscode/settings.json` can contain the key, VS Code accepts it, but the terminal ignores it for security. The result: the system prompt told the model "you have PowerShell" while the terminal opened cmd.exe (or whatever VS Code autodetected), causing every shell command to fail.

**Fix:** Deleted the parallel implementation in `shell.ts` (`getTerminalConfig`, `normalizeShellPath`, `getWindowsShellFromVSCode`, etc.) and replaced it with a single `getShellFromVSCode()` that delegates to `Terminal.getConfiguredDefaultProfileName()` and `Terminal.getConfiguredProfiles()`. Both paths now read from the same inspect()-based source of truth. Also extracted `preferredWindowsPowerShell()` to eliminate a duplicated PS7→legacy fallback expression.

**Tests:**
- Rewrote `shell.spec.ts` stubs to use `inspect()` (matching the new implementation) and added a "Scope isolation" section asserting workspace-scoped profiles are ignored.
- Added `shell-system-prompt-divergence.spec.ts` as a regression anchor: confirms `Terminal` ignores workspace-scoped profiles, verifies both paths agree when a profile is set at global scope, and verifies convergence after the fix (workspace-only profile → both return the Windows no-profile fallback).

### Test Procedure

Unit tests:
```
pnpm exec vitest run utils/__tests__/shell.spec.ts integrations/terminal/__tests__/shell-system-prompt-divergence.spec.ts
```

E2E mock suite (76 passing, 3 pre-existing xAI failures unrelated to this change):
```
pnpm --filter @roo-code/vscode-e2e test:ci:mock
```

### Pre-Submission Checklist

- [ ] **Visual Snapshot** (UI changes only): N/A — no UI changes.
- [x] **Documentation Impact**: No documentation updates required.

### Additional Notes

The 3 failing E2E tests are all in the xAI provider suite and fail with a 403 "credits exhausted" error — pre-existing and unrelated to this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal shell detection across Windows, macOS, and Linux.
  * Workspace-only terminal profile settings are now handled consistently and no longer incorrectly override applicable defaults.
  * Explicitly configured shell paths and Zoo profile settings now take precedence appropriately, with reliable fallbacks when unavailable.
  * Improved fallback behavior for PowerShell, WSL, user-configured shells, and environment settings.
  * Enhanced validation safely rejects invalid or unsupported shell paths.

### DIFF
--- a/src/integrations/terminal/__tests__/shell-system-prompt-divergence.spec.ts
+++ b/src/integrations/terminal/__tests__/shell-system-prompt-divergence.spec.ts
@@ -1,0 +1,169 @@
+// Regression test for https://github.com/Zoo-Code-Org/Zoo-Code/issues/634
+//
+// When the user sets terminal.integrated.defaultProfile.windows in *workspace*
+// settings, getShell() (used for the system prompt) picks it up via config.get()
+// which merges all scopes, but Terminal.getConfiguredDefaultProfileName() only
+// reads inspect().globalValue ?? inspect().defaultValue — intentionally excluding
+// workspace for security.
+//
+// The result: the system prompt tells the model "you have PowerShell" but the
+// actual terminal VS Code opens defaults to cmd.exe (or whatever VS Code picks
+// when no global/default profile is set), so every PowerShell command fails.
+//
+// Run: node_modules/.bin/vitest run integrations/terminal/__tests__/shell-system-prompt-divergence.spec.ts
+
+import { existsSync } from "fs"
+import * as vscode from "vscode"
+
+vi.mock("execa", () => ({ execa: vi.fn() }))
+vi.mock("fs", () => ({ existsSync: vi.fn(() => false) }))
+vi.mock("os", () => ({ userInfo: vi.fn(() => ({ shell: null })) }))
+
+const mockedExistsSync = existsSync as unknown as ReturnType<typeof vi.fn>
+
+const { Terminal } = await import("../Terminal")
+const { getShell } = await import("../../../utils/shell")
+
+describe("issue #634 — system prompt shell vs actual terminal shell divergence", () => {
+	let originalPlatform: NodeJS.Platform
+
+	beforeEach(() => {
+		originalPlatform = process.platform
+		Object.defineProperty(process, "platform", { value: "win32", configurable: true })
+		Terminal.setTerminalProfile(undefined)
+		mockedExistsSync.mockReset()
+		// pwsh.exe exists — getShell() fallback path prefers PowerShell 7 over legacy
+		mockedExistsSync.mockImplementation((p: string) => p === "C:\\Program Files\\PowerShell\\7\\pwsh.exe")
+	})
+
+	afterEach(() => {
+		Object.defineProperty(process, "platform", { value: originalPlatform, configurable: true })
+		Terminal.setTerminalProfile(undefined)
+		vi.restoreAllMocks()
+	})
+
+	/**
+	 * Stubs VS Code config to simulate a workspace-scoped default profile.
+	 *
+	 * getShell() (shell.ts) uses getConfiguration("terminal.integrated").get() which
+	 * returns the merged/effective value across all scopes — workspace value wins.
+	 *
+	 * Terminal.getConfiguredDefaultProfileName() uses inspect().globalValue ?? defaultValue,
+	 * intentionally excluding workspace scope for security. Both undefined here.
+	 */
+	function stubWorkspaceScopedProfile(profileName: string, profilePath: string) {
+		const profiles = { [profileName]: { path: profilePath } }
+		vi.spyOn(vscode.workspace, "getConfiguration").mockImplementation((section?: string) => {
+			if (section === "terminal.integrated") {
+				return {
+					// get() merges all scopes — shell.ts uses this, picks up workspace value
+					get: (key: string) => {
+						if (key === "defaultProfile.windows") return profileName
+						if (key === "profiles.windows") return profiles
+						return undefined
+					},
+					// Terminal uses inspect() and only reads globalValue ?? defaultValue
+					inspect: (_key: string) => ({
+						defaultValue: undefined,
+						globalValue: undefined,
+						workspaceValue: profileName,
+					}),
+				} as unknown as vscode.WorkspaceConfiguration
+			}
+
+			if (section === "terminal.integrated.profiles") {
+				return {
+					inspect: (_key: string) => ({
+						defaultValue: undefined,
+						globalValue: undefined,
+						workspaceValue: profiles,
+					}),
+				} as unknown as vscode.WorkspaceConfiguration
+			}
+
+			return {
+				get: (_key: string, dv?: unknown) => dv,
+				inspect: () => undefined,
+			} as unknown as vscode.WorkspaceConfiguration
+		})
+	}
+
+	it("Terminal.getConfiguredDefaultProfileName ignores workspace-scoped profile (confirms the bug)", () => {
+		// User set PowerShell as default only in their workspace .vscode/settings.json
+		stubWorkspaceScopedProfile("PowerShell", "C:\\Program Files\\PowerShell\\7\\pwsh.exe")
+
+		// Terminal intentionally excludes workspace scope for security.
+		// With no global/default profile set, it returns undefined.
+		const terminalSeesProfileName = Terminal.getConfiguredDefaultProfileName("win32")
+		expect(terminalSeesProfileName).toBeUndefined()
+
+		// As a consequence, isActiveShellPowerShell returns false even though the
+		// user configured PowerShell — the terminal will not be treated as PowerShell.
+		expect(Terminal.isActiveShellPowerShell("win32")).toBe(false)
+	})
+
+	it("getShell() and Terminal agree on PowerShell when the default profile is set at global/user scope", () => {
+		// When the profile is set at user (global) scope, both paths see the same value.
+		const profilePath = "C:\\Program Files\\Git\\bin\\bash.exe" // non-PowerShell so name-matching doesn't hide the bug
+		const profileName = "Git Bash"
+		vi.spyOn(vscode.workspace, "getConfiguration").mockImplementation((section?: string) => {
+			if (section === "terminal.integrated") {
+				return {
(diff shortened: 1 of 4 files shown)


src/integrations/terminal/__tests__/shell-system-prompt-divergence.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+// settings, getShell() (used for the system prompt) picks it up via config.get()
+// which merges all scopes, but Terminal.getConfiguredDefaultProfileName() only
+// reads inspect().globalValue ?? inspect().defaultValue — intentionally excluding
+// workspace for security.
+//
+// The result: the system prompt tells the model "you have PowerShell" but the
+// actual terminal VS Code opens defaults to cmd.exe (or whatever VS Code picks
+// when no global/default profile is set), so every PowerShell command fails.
+//
+// Run: node_modules/.bin/vitest run integrations/terminal/__tests__/shell-system-prompt-divergence.spec.ts
```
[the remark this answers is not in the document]
edelauna: Fixed — header now describes post-fix behavior: getShell() delegates to Terminal.getConfiguredDefaultProfileName() / getConfiguredProfiles() so both paths use the same inspect()-based reads.

src/integrations/terminal/__tests__/shell-system-prompt-divergence.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+
+	/**
+	 * Stubs VS Code config to simulate a workspace-scoped default profile.
+	 *
+	 * getShell() (shell.ts) uses getConfiguration("terminal.integrated").get() which
+	 * returns the merged/effective value across all scopes — workspace value wins.
+	 *
+	 * Terminal.getConfiguredDefaultProfileName() uses inspect().globalValue ?? defaultValue,
+	 * intentionally excluding workspace scope for security. Both undefined here.
+	 */
```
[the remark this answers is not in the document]
edelauna: Fixed — removed the obsolete get() branch explanation. Helper doc now only describes the inspect() / globalValue behavior that's actually in play.

src/integrations/terminal/__tests__/shell-system-prompt-divergence.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+		// Terminal.getConfiguredDefaultProfileName() uses inspect().globalValue → undefined
+		const terminalSeesProfileName = Terminal.getConfiguredDefaultProfileName("win32")
+		expect(terminalSeesProfileName).toBeUndefined()
+
+		// Terminal therefore can't identify the active shell as PowerShell
+		expect(Terminal.isActiveShellPowerShell("win32")).toBe(false)
+
+		// Divergence: system prompt claims PowerShell, Terminal has no profile → falls
+		// through to VS Code's own autodetect which may open cmd.exe on this machine.
+	})
```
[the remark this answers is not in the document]
edelauna: Fixed — test renamed to "convergence" and rewritten to assert post-fix behavior: getShell() also reads inspect().globalValue (via Terminal delegation), so it ignores the workspace-scoped profile and returns the PS-legacy fallback, matching what Terminal sees.

src/utils/shell.ts
```diff
@@ -113,143 +115,63 @@ const SHELL_PATHS = {
...
+function getShellFromVSCode(): string | null {
+	try {
+		const profileName = Terminal.getConfiguredDefaultProfileName()
+
+		if (!profileName) {
+			// No profile configured at user/default scope. On Windows, VS Code
+			// auto-detects and prefers PowerShell 7 when installed. Mirror that so
+			// the system prompt matches what VS Code will actually open. (issue #82)
+			if (process.platform === "win32") {
+				return preferredWindowsPowerShell()
```
navedmerchant, line 151: **High — no-profile fallback still guesses a shell that may not execute the command.** When no trusted default profile exists, this advertises PowerShell solely from filesystem presence. However, the VS Code terminal is created without `shellPath`, so VS Code can still open cmd.exe or another auto-detected shell. The new regression test even has `getShell()` return Windows PowerShell while `Terminal.isActiveShellPowerShell()` returns `false`; that is the original divergence, not convergence. Make terminal creation deterministic with this same executable, or derive the prompt from the actual launched shell instead of assuming PowerShell.

src/utils/shell.ts
```diff
@@ -113,143 +115,63 @@ const SHELL_PATHS = {
...
-function normalizeShellPath(path: string | string[] | undefined): string | null {
-	if (!path) return null
-	if (Array.isArray(path)) {
-		return path.length > 0 ? path[0] : null
-	}
-	return path
-}
+function getShellFromVSCode(): string | null {
+	try {
+		const profileName = Terminal.getConfiguredDefaultProfileName()
```
navedmerchant, line 144: **High — this resolver is not aware of the command-execution provider.** Inline mode executes through Execa using the configured Execa shell or `shell: true`, while this path always reports a VS Code profile/default. On Windows, Execa can therefore execute cmd.exe while the prompt advertises PowerShell. The issue report specifically identifies disabling Inline Terminal as a workaround, so this path must be included in the fix. Resolve one provider-aware shell descriptor and use it for both prompt generation and execution.

src/utils/shell.ts
```diff
@@ -113,143 +115,63 @@ const SHELL_PATHS = {
...
-	const { defaultProfileName, profiles } = getTerminalConfig("windows")
-	if (!defaultProfileName) {
-		// No explicit Windows terminal profile is configured. VS Code auto-detects
-		// the default on modern Windows and prefers PowerShell 7 (pwsh.exe) when it
-		// is installed, otherwise the always-present Windows PowerShell 5.1. Mirror
-		// that here so the system prompt advertises the real shell instead of falling
-		// through to COMSPEC (cmd.exe). See issue #82.
-		return existsSync(SHELL_PATHS.POWERSHELL_7) ? SHELL_PATHS.POWERSHELL_7 : SHELL_PATHS.POWERSHELL_LEGACY
-	}
+		const profiles = Terminal.getConfiguredProfiles()
```
navedmerchant, line 156: **High — Zoo’s explicit terminal profile override is ignored here.** Terminal creation gives `Terminal.getProfileShell()` precedence and can launch, for example, Git Bash, but this code reads only VS Code’s default profile. With a Zoo Git Bash override and a PowerShell default, commands execute in Git Bash while the prompt emits PowerShell syntax. Check the explicit override first, or centralize profile resolution so prompt generation and terminal creation consume the same result.

navedmerchant (commented):
High-severity inline findings from holistic review.

src/utils/shell.ts
edelauna: Fixed — `getShellFromVSCode()` now checks `Terminal.getProfileShell()` first, matching the priority order in the Terminal constructor (lines 34-52). When a Zoo profile override is active and resolves to a shell path, it is returned before reading VS Code's default profile.

src/utils/shell.ts
edelauna: Fixed — `getShell()` now checks `BaseTerminal.getExecaShellPath()` as the highest-priority source. When an explicit execa shell path is set, that is what `ExecaTerminalProcess` uses (line 43: `shell: BaseTerminal.getExecaShellPath() || true`), so the prompt advertises the same executable.

When no explicit execa path is set (`shell: true`), Node.js uses `process.env.ComSpec` on Windows (cmd.exe) and `/bin/sh` on Unix. Full provider-awareness at prompt-generation time — knowing which execution backend will be chosen before it is chosen — requires a larger refactor. The existing env-based fallback (`getShellFromEnv`) partially covers this path today. Filed as a known gap for follow-up.

src/utils/shell.ts
edelauna: Acknowledged — this is a real limitation but not one this PR can fully solve. VS Code does not expose an API for extensions to query which shell `createTerminal()` will use when called without `shellPath`. `vscode.env.shell` exists but is documented as unreliable ([microsoft/vscode#132787](https://github.com/microsoft/vscode/issues/132787)). VS Code's own auto-detect ([PR #112768](https://github.com/microsoft/vscode/pull/112768)) scans multiple install paths and is more comprehensive than a single `existsSync` check.

The only fully deterministic fix would be to always pass `shellPath` to `createTerminal()`, which changes terminal creation behavior and is a separate concern from the workspace-scope fix in this PR. The `existsSync` heuristic matches VS Code's documented preference for PowerShell 7 on Windows in the common case.

navedmerchant (approved):

PR #1040, opened 2026-07-28
refactor(api): centralize service-tier primitives

WebMad:
## Responsibility and scope

Centralizes the shared service-tier primitives used by provider request construction so later stack layers can consume one canonical representation. Tests travel with the behavior in this PR.

## Behavior and non-goals

- Refactor only: there is **no behavior change and no request-shape change**.
- Does not add OpenAI Codex Fast priority handling.
- Does not add settings UI or translations.
- Does not include the unrelated `Task.throttle.test.ts` fix.
- Does not add a changeset.

## Test evidence

- Existing focused coverage was updated/retained with the refactored primitives.
- Repository type checks passed from the push hook.
- The completed stack was locally verified before publication.

## Stack dependency

1. **This PR** — shared service-tier primitives (base: canonical `main`).
2. [WebMad/Zoo-Code#2](https://github.com/WebMad/Zoo-Code/pull/2) — backend Fast priority persistence/request behavior.
3. [WebMad/Zoo-Code#3](https://github.com/WebMad/Zoo-Code/pull/3) — webview speed selector and translations.

Review and merge this PR first; each subsequent PR is based on the preceding contributor branch.

## Summary by CodeRabbit

* **New Features**
  * Standardized OpenAI service-tier handling for Default, Flex, and Priority, enabling consistent tier pricing display in model settings for supported tiers.
* **Bug Fixes**
  * Fixed streamed and fallback usage/cost calculations to use the resolved tier (while keeping the requested tier in requests).
  * Improved service-tier request wiring for OpenAI and Bedrock, including correct omission when the tier isn’t supported.
* **Tests**
  * Expanded coverage for OpenAI/Bedrock service-tier payloads, pricing selection, and UI table/selector behavior.
* **Chores**
  * Updated package exports and Playwright translation test scaffolding.


packages/types/src/model.ts
```diff
@@ -54,12 +54,21 @@ export const verbosityLevelsSchema = z.enum(verbosityLevels)
...
 
+/** Serialized service tier field used in provider request payloads and responses. */
+export const SERVICE_TIER_KEY = "service_tier"
+
 /**
- * Service tiers (OpenAI Responses API)
+ * Service tiers for the public OpenAI Responses API.
  */
-export const serviceTiers = ["default", "flex", "priority"] as const
+export enum OpenAiServiceTier {
```
edelauna: Every other string list in this file (`verbosityLevels`, `reasoningEfforts`, `modelParameters`) uses `as const` + `z.infer`, and the existing enums in `packages/types/src` are all discriminator-style (IPC/task/telemetry) — none wrap a zod-schema'd string list. Is there a reason to reach for a TS enum here?

A `const OpenAiServiceTier = { Default: "default", Flex: "flex", Priority: "priority" } as const` object would give the same named access at the call sites while keeping the file's existing pattern — and would also drop the `Object.values(...) as [...]` tuple cast at L69.

src/api/providers/__tests__/openai-native.spec.ts
```diff
@@ -332,12 +544,55 @@ describe("OpenAiNativeHandler", () => {
...
+			const model = handler.getModel()
+			const infoWithoutTiers = { ...model.info, tiers: undefined }
+			const applyServiceTierPricing = Reflect.get(handler, "applyServiceTierPricing")
+
+			const result = Reflect.apply(applyServiceTierPricing, handler, [
+				infoWithoutTiers,
+				OpenAiServiceTier.Priority,
+			])
+
+			expect(result).toBe(infoWithoutTiers)
```
edelauna: These three tests reach into private methods via `Reflect.get`/`Reflect.apply` — as far as I can tell they're the only such usages in `src/`. Could the same pricing invariants be asserted through the public usage path instead, the way the `serviceTierPricingCases` tests earlier in this file do? That would keep the tests from coupling to internal method names/signatures.

edelauna (changes requested):
Nice! Thanks for this PR - had 2 comments related to the implementation.

WebMad:
Addressed both review comments in 5f0179c99:

- Replaced the TypeScript enum with an `as const` `OpenAiServiceTier` object, kept `serviceTiers` as a readonly tuple for `z.enum`, and now infer `ServiceTier` from the schema.
- Reworked the three pricing-invariant tests to exercise the public `createMessage` usage path instead of reaching into private methods with `Reflect.get` / `Reflect.apply`.

Also centralized request-tier gating so streaming requests and `completePrompt` use the same helper, and kept the shared tier-pricing row renderer strictly typed with `ServiceTier`.

Validation completed:
- 197 focused backend tests passed
- 6 focused webview tests passed
- repository lint and all 11 applicable type-check tasks passed via commit/push hooks

packages/types/src/model.ts
WebMad: Updated in 5f0179c99. I replaced the enum with an `as const` object, kept `serviceTiers` as a readonly tuple accepted directly by `z.enum`, and now infer `ServiceTier` through `z.infer<typeof serviceTierSchema>`. This preserves named call-site access while matching the existing types-file pattern and removes the `Object.values(...) as [...]` cast.

src/api/providers/__tests__/openai-native.spec.ts
WebMad: Updated in 5f0179c99. The three invariants now run through the public `createMessage` usage path alongside the existing service-tier pricing cases. The `Reflect.get` / `Reflect.apply` calls and coupling to private method names/signatures have been removed.

edelauna (changes requested):
Since you're touching the file could you include a snapshot of `webview-ui/src/components/settings/ModelInfoView.tsx` using https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/webview-ui/AGENTS.md#visual-tests-playwright-ct

WebMad:
> Since you're touching the file could you include a snapshot of `webview-ui/src/components/settings/ModelInfoView.tsx` using https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/webview-ui/AGENTS.md#visual-tests-playwright-ct

Anything you want, bro 😄

packages/types/package.json
```diff
@@ -11,6 +11,14 @@
...
 				"default": "./dist/index.cjs"
 			}
+		},
+		"./model": {
+			"types": "./src/model.ts",
+			"import": "./src/model.ts"
+		},
+		"./provider-identifiers": {
+			"types": "./src/provider-identifiers.ts",
+			"import": "./src/provider-identifiers.ts"
```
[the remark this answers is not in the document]
edelauna: These subpaths are ESM-only by design. The only consumer is `ModelInfoView.tsx` in the webview, which Vite bundles as ESM. No CJS consumer uses these subpaths, so `require()` producing `ERR_PACKAGE_PATH_NOT_EXPORTED` is not reachable in practice. If a CJS consumer is added later, the `require` targets can be added then.

edelauna (approved):
Thanks for adding the snapshots

PR #1053, opened 2026-07-29
fix(router-provider): fetch model metadata before context management decisions

JamesRobert20:
## Problem

Router providers that are auth-scoped (zoo-gateway, kimi-code) skip the model cache entirely. When a fresh handler is constructed (on provider switch or profile change), `getModel()` falls back to hardcoded default model info because the real model list has not been fetched from the API yet.

Context management in `Task.ts` calls `getModel()` *before* `createMessage()` (where `fetchModel()` normally runs). This means condensing and truncation decisions use the default context window (200k for zoo-gateway) instead of the actual model's context window.

If the user selects a model with a larger context window through the gateway (e.g. Opus 4.8 at 1M tokens), the system incorrectly calculates context usage against 200k, triggering premature condensation.

## Fix

- Add `ensureModelFetched()` to `RouterProvider` that calls `fetchModel()` once when the instance model map is empty. Subsequent calls are no-ops.
- Add `ensureModelFetched` as an optional method on the `ApiHandler` interface so `Task.ts` can call it.
- Call `await this.api.ensureModelFetched?.()` in `Task.ts` before both context management paths (`attemptApiRequest` and `handleContextWindowExceededError`).

Static providers (Anthropic, Gemini, Bedrock, etc.) are unaffected since they don't implement the method. Non-auth-scoped router providers (OpenRouter, Vercel AI Gateway) already have cache hits via `getModelsFromCache()` so the fetch is skipped after the first call.

## Summary by CodeRabbit

* **New Features**
  * Model metadata (context window and pricing) can now be prefetched automatically via an optional API capability.
  * Router-based model loading is now concurrency-safe and memoized to avoid duplicate requests.
* **Bug Fixes**
  * Context-limit handling and token management now use freshly fetched model metadata instead of defaults.
  * If metadata fetching fails, requests continue using safe fallback behavior.
  * Updated model details (e.g., context window) correctly override prior default values.
* **Tests**
  * Expanded coverage for cache short-circuiting, recovery after failures, and concurrency/deduplication.


src/core/task/Task.ts
```diff
@@ -2762,6 +2762,8 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
 
 				await this.diffViewProvider.reset()
 
+				await this.api.ensureModelFetched?.()
```
edelauna: The litellm/deepseek/moonshot fetchers re-throw, so a failure here would reject into the catch at L3743, which `return true`s and ends the task silently — the comment there assumes only `attemptApiRequest` can throw. Pre-change, the same failure surfaced via the `createMessage` retry path. Would catching and logging here be safer, letting `getModel()` fall back to defaults?

src/api/providers/router-provider.ts
```diff
@@ -61,6 +61,21 @@ export abstract class RouterProvider extends BaseProvider {
 		return this.getModel()
 	}
 
+	private modelFetchPromise?: Promise<void>
+
+	async ensureModelFetched(): Promise<void> {
```
edelauna: For auth-scoped providers, does the first request now fetch the model list twice — once here and again unconditionally in `createMessage` (zoo-gateway.ts L184)? `getModels` bypasses both caches for auth-scoped providers (modelCache.ts L265), so the second call is another network round trip. Worth making `fetchModel()` single-flight (or short-circuiting when `this.models` is populated) so both callers share one fetch?

src/api/providers/__tests__/zoo-gateway.spec.ts
```diff
@@ -635,4 +635,61 @@ describe("ZooGatewayHandler", () => {
 			)
 		})
 	})
+
+	describe("ensureModelFetched", () => {
```
edelauna, line 639: Is the rejection path covered anywhere? If `getModels` rejects once and the `finally` cleanup in router-provider.ts (L71-74) ever regresses, the stored rejected promise would poison every subsequent call. A reject-then-resolve test here would lock that in.

edelauna (changes requested):

edelauna (approved):

PR #1059, opened 2026-07-30
[Improve] Strengthen managed binary installation

### LINKED ISSUE #1055: Extract reusable managed-binary installation infrastructure
## Parent

Part of #1049. This is the first independently reviewable slice of the Destructive Command Guard work.

## Problem

The existing Semble downloader owns a large amount of archive download, checksum, extraction, versioning, cleanup, and concurrent-install logic. DCG needs the same lifecycle guarantees. Duplicating that logic would make both binary integrations harder to audit and maintain.

## Scope

- Introduce shared managed-binary helpers for bounded downloads, checksum verification, safe archive extraction, and atomic/versioned installation.
- Handle cancellation, oversized downloads, partial installs, stale artifacts, executable permissions, and concurrent VS Code windows.
- Migrate the existing Semble downloader to the shared helpers without changing Semble behavior.
- Add focused regression tests for the shared helpers and update the Semble downloader tests.

## Out of scope

- DCG-specific metadata or execution.
- Settings, auto-approval, command execution, or chat UI changes.

## Acceptance criteria

- [ ] Semble still installs and resolves its binary through the shared infrastructure.
- [ ] Downloads are size-bounded and clean up streams/files on failure.
- [ ] Archives reject unsafe or unexpected entries.
- [ ] Installation is atomic and safely adopts an installation completed concurrently by another window.
- [ ] Focused managed-binary and Semble tests pass.
- [ ] The branch type-checks/builds independently from `main`.

## Stack

Base: `main`. The DCG service issue will build on this issue.

navedmerchant:
## What changed

Adds tighter managed-binary safeguards: bounded Semble downloads, stricter archive layout validation, stream cleanup on download failures, and regression coverage for successful verification and installation cleanup.

## Why this change was made

The shared installer underpins the DCG rollout and must reject malformed archives or unbounded downloads without leaving partial files or open connections. Closes #1055.

## Impact

Binary-backed services install more safely and consistently without changing Semble's user-facing workflow.

## Related PRs

- [DCG binary service](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1060)
- [DCG command integration](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1062)

### DIFF
--- a/src/services/code-index/semble/semble-downloader.ts
+++ b/src/services/code-index/semble/semble-downloader.ts
@@ -1,10 +1,9 @@
 import * as fs from "fs/promises"
 import * as path from "path"
-import * as https from "https"
-import { createWriteStream } from "fs"
-import { createHash } from "crypto"
-import { createReadStream } from "fs"
-import { spawn } from "child_process"
+
+import { extractTarGzArchive, extractZipArchive } from "../../managed-binary/archive"
+import { downloadBinaryFile, verifySha256Checksum } from "../../managed-binary/download"
+import { ensureManagedBinaryInstalled, getManagedBinaryPaths } from "../../managed-binary/install"
 
 /**
  * Supported platform/arch combinations for the semble standalone executable.
@@ -47,19 +46,14 @@ export const SEMBLE_SHA256: Record<string, string> = {
  * Throws if the checksum does not match.
  */
 export async function verifyChecksum(filePath: string, expected: string): Promise<void> {
-	const hash = createHash("sha256")
-	await new Promise<void>((resolve, reject) => {
-		const stream = createReadStream(filePath)
-		stream.on("data", (chunk) => hash.update(chunk))
-		stream.on("end", resolve)
-		stream.on("error", reject)
-	})
-	const actual = hash.digest("hex")
-	if (actual !== expected) {
-		throw new Error(
-			`Checksum mismatch for ${path.basename(filePath)}: expected ${expected.slice(0, 12)}…, got ${actual.slice(0, 12)}…`,
-		)
-	}
+	await verifySha256Checksum(
+		filePath,
+		expected,
+		(actual) =>
+			new Error(
+				`Checksum mismatch for ${path.basename(filePath)}: expected ${expected.slice(0, 12)}…, got ${actual.slice(0, 12)}…`,
+			),
+	)
 }
 
 /**
@@ -88,64 +82,6 @@ function getArchiveInfo(platform?: string, arch?: string): { archive: string; bi
 }
 
 /**
- * Reads the locally installed version from the version metadata file.
- * Returns undefined if no version file exists (first install or legacy).
- */
-async function getInstalledVersion(storageDir: string): Promise<string | undefined> {
-	try {
-		const versionPath = path.join(storageDir, "semble", VERSION_FILE)
-		const version = (await fs.readFile(versionPath, "utf-8")).trim()
-		return version || undefined
-	} catch {
-		return undefined
-	}
-}
-
-/**
- * Writes the version metadata file after a successful download.
- */
-async function writeInstalledVersion(storageDir: string, version: string): Promise<void> {
-	const versionPath = path.join(storageDir, "semble", VERSION_FILE)
-	await fs.writeFile(versionPath, version, "utf-8")
-}
-
-/**
- * Best-effort removal of archive files left over from previous semble versions.
- *
- * Because the local archive cache path is version-prefixed (see `downloadSemble`),
- * upgrading SEMBLE_VERSION leaves the prior version's archive orphaned on disk.
- * This sweeps those stale packages so a version upgrade doesn't accumulate them.
- *
- * Matches both the version-prefixed cache names (`${version}-${archiveName}`,
- * used since v0.4.0) and the legacy unversioned cache name (`${archiveName}`,
- * used before v0.4.0), so a v0.3.1 → v0.4.1 upgrade also clears the legacy file.
- * The current archive path is always preserved.
- *
- * Errors are swallowed since this is purely cosmetic cleanup.
- */
-async function cleanupStaleArchives(
-	storageDir: string,
-	archiveName: string,
-	currentArchivePath: string,
-): Promise<void> {
-	try {
-		const entries = await fs.readdir(storageDir)
-		const suffix = `-${archiveName}`
-		await Promise.all(
-			entries
-				.filter(
-					(name) =>
-						(name === archiveName || name.endsWith(suffix)) &&
-						path.join(storageDir, name) !== currentArchivePath,
-				)
-				.map((name) => fs.unlink(path.join(storageDir, name)).catch(() => {})),
-		)
-	} catch {
-		// ignore — storage dir may not be listable yet
-	}
-}
-
-/**
  * Downloads and extracts the semble archive for the current platform.
  *
  * Compares the hardcoded SEMBLE_VERSION against the version stored on disk.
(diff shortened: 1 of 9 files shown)


src/services/managed-binary/download.ts
```diff
@@ -0,0 +1,149 @@
...
+
+function downloadBinaryFileWithRedirects(
+	url: string,
+	destination: string,
+	options: BinaryDownloadOptions,
+	redirectsRemaining: number,
+): Promise<void> {
+	return new Promise((resolve, reject) => {
+		if (!isTrustedHttpsUrl(url, options.trustedDomains)) {
+			reject(new Error(`${options.name} download redirected to an untrusted host (untrusted domain)`))
```
edelauna: If the caller passes a bad initial URL here, the error says "redirected to an untrusted host" — but no redirect has happened yet. Could this say something like `download URL is not a trusted HTTPS host` instead? The "redirected" wording in `resolveTrustedRedirect` at line 38 is accurate there.

src/services/managed-binary/download.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+export function resolveTrustedRedirect(
+	url: string,
+	location: string | undefined,
+	redirectsRemaining: number,
+	options: Pick<BinaryDownloadOptions, "name" | "trustedDomains">,
+): string {
+	if (redirectsRemaining <= 0 || !location) {
+		throw new Error(`Too many ${options.name} download redirects`)
```
edelauna, line 33: This fires for two different cases: `redirectsRemaining <= 0` (too many hops) and `!location` (server returned 3xx without a Location header). When it is the missing-header case, "Too many redirects" is misleading. Worth splitting into two guards with distinct messages?

src/services/managed-binary/download.ts
```diff
@@ -0,0 +1,149 @@
...
+			)
+			response.on("data", (chunk: Buffer) => {
+				received += chunk.length
+				if (options.maxBytes !== undefined) {
+					try {
+						assertSizeWithinLimit(received, options.maxBytes, options.name)
+					} catch (error) {
+						response.destroy()
+						request.destroy(error as Error)
+						reject(error)
```
edelauna: When `maxBytes` is exceeded mid-stream, `response.destroy()` and `request.destroy()` are called but `output` is never closed or destroyed. `pipe()` does not propagate source destruction to the write destination, so the WriteStream fd stays open. On Windows this can prevent the `finally`-block `fs.rm(archivePath)` in `install.ts` from cleaning up. Consider adding `output.destroy()` here (or `response.unpipe(output)` before destroying).

src/services/managed-binary/download.ts
```diff
@@ -0,0 +1,149 @@
...
+					}
+				}
+			})
+			response.on("error", reject)
+			response.pipe(output)
```
edelauna, line 146: The `data` size-check listener (line 125) is registered before `response.pipe(output)` here. Node.js fires listeners in registration order, so when the oversize chunk arrives, the size check fires first and calls `response.destroy()` — but pipe's own `data` listener fires next in the same tick and writes the chunk to disk before the stream tears down. `maxBytes` therefore bounds received bytes but not bytes written. Would `response.unpipe(output)` before `response.destroy()` in the handler close this gap?

src/services/code-index/semble/semble-downloader.ts
```diff
@@ -164,130 +100,48 @@ export async function downloadSemble(storageDir: string): Promise<string | undef
...
+		extractArchive: async (archivePath, stagingDir) => {
+			if (info.archive.endsWith(".tar.gz")) {
+				await extractTarGzArchive(archivePath, stagingDir)
+			} else if (info.archive.endsWith(".zip")) {
+				await extractZipArchive(archivePath, stagingDir)
+			}
+		},
+	})
+
+	console.log(`[SembleDownloader] Successfully installed semble ${SEMBLE_VERSION} to ${paths.binaryPath}`)
```
edelauna: nit: `paths` is only used for `paths.binaryPath` in this log line, but `ensureManagedBinaryInstalled` already returns that same path as `result`. Could the `getManagedBinaryPaths` call and `paths` variable be removed and the log changed to use `result` directly?

src/services/managed-binary/install.ts
```diff
@@ -0,0 +1,131 @@
...
+		await fs.rm(paths.installRoot, { recursive: true, force: true })
+		await fs.rename(paths.stagingDir, paths.installRoot)
+		await cleanupStaleArchives(options, paths.archivePath)
+		return paths.binaryPath
+	} catch (error) {
+		if (!options.errorPrefix) {
+			throw error
+		}
+		const message = error instanceof Error ? error.message : String(error)
+		throw new Error(`${options.errorPrefix}: ${message}`)
```
edelauna: `errorPrefix` is optional, which silently changes error-wrapping behaviour: when absent, errors propagate with their original type and stack; when present, they are re-wrapped as a plain `new Error(string)`, dropping the original cause chain. Could this be made required (or always wrap using `{ cause: error }`) so callers don't accidentally opt out of wrapping?

src/services/managed-binary/__tests__/install.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+		await writeFile(paths.versionPath, options.version)
+		if (process.platform !== "win32") await chmod(paths.binaryPath, 0o600)
+
+		await expect(ensureManagedBinaryInstalled(options)).resolves.toBe(paths.binaryPath)
+		expect(options.download).not.toHaveBeenCalled()
+	})
+
+	it("deduplicates concurrent installations", () => {
+		const options = createOptions({ download: () => new Promise<void>(() => {}) })
+		expect(ensureManagedBinaryInstalled(options)).toBe(ensureManagedBinaryInstalled(options))
```
edelauna: This verifies that both calls return the same Promise object, but does it also verify that `download` is invoked exactly once? A future refactor that returns two structurally-equivalent Promises (e.g. wrapping the cached one) would satisfy the `toBe` check while breaking the deduplication guarantee. Worth adding `expect(options.download).toHaveBeenCalledTimes(1)` after both promises settle?

src/services/managed-binary/__tests__/install.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+		})
+		const paths = getManagedBinaryPaths(options)
+
+		await expect(ensureManagedBinaryInstalled(options)).resolves.toBe(paths.binaryPath)
+		expect(calls).toEqual(["download", "verify", "extract", "validate"])
+		expect(await readFile(paths.binaryPath, "utf8")).toBe("binary")
+		expect(await readFile(paths.versionPath, "utf8")).toBe(options.version)
+		await expect(access(paths.archivePath)).rejects.toThrow()
+		await expect(access(paths.stagingDir)).rejects.toThrow()
+	})
```
edelauna, line 140: Two gaps in this test suite:

1. **Failure path** — there is no test verifying that when `download` rejects, the partial archive and staging dir are both cleaned up (and no binary is promoted). The `finally` block in `installManagedBinary` is the only guarantee of this.

2. **`cleanupStaleArchives` predicate** — the test never seeds a stale archive with a different version prefix, so a mutation of the filter at `install.ts:70–72` (e.g. accidentally also matching the current archive) would not be caught — the `finally` block removes `archivePath` regardless.

src/services/managed-binary/__tests__/archive.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+		const processResult = runProcess("tool", ["--version"])
+		child.stdout.write("1.2.3")
+		child.emit("close", 0)
+
+		await expect(processResult).resolves.toEqual({ stdout: "1.2.3", stderr: "" })
+		expect(mockSpawn).toHaveBeenCalledWith("tool", ["--version"], {
+			shell: false,
+			stdio: ["ignore", "pipe", "pipe"],
+		})
+	})
```
edelauna, line 43: Two coverage gaps here:

1. **`extractTarXzArchive` and `extractZipArchive`** are exported from `archive.ts` but not imported or tested in this file — only `extractTarGzArchive` and the `extractSingleFile*` variants are covered.

2. **`runProcess` timeout path** — the `setTimeout`/`SIGKILL` path (`archive.ts` lines 14–17) is not exercised. Without fake timers, a mutation removing `clearTimeout` in the `close` handler would also be invisible.

src/services/managed-binary/__tests__/download.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+		await expect(
+			downloadBinaryFile("https://github.com/release", "/tmp/archive", {
+				name: "Example",
+				trustedDomains,
+				timeoutMs: 1_000,
+				maxBytes: 10,
+			}),
+		).rejects.toThrow("Example archive exceeds the download size limit")
+		expect(mockCreateWriteStream).not.toHaveBeenCalled()
+	})
```
edelauna, line 192: This tests the declared-`Content-Length` oversize path (where `createWriteStream` is never called). Is there also a test for the mid-stream case — where bytes accumulate past `maxBytes` after the stream is already open? That path also has the open-WriteStream cleanup issue noted in `download.ts:130–133`.

edelauna (changes requested):
Nice! Had a couple implementation comments.

src/services/managed-binary/download.ts
```diff
@@ -0,0 +1,154 @@
...
+						response.unpipe(output)
+						output.destroy()
+						response.destroy()
+						request.destroy()
+						reject(error)
+					}
+				}
+			})
+			response.on("error", reject)
+			response.pipe(output)
```
edelauna, line 146: Would it be safer to also destroy the streams here? If the response emits an error, `output` and `request` stay open — compare the size-limit handler at lines 133–138, which calls `output.destroy()`, `response.destroy()`, and `request.destroy()` before rejecting. Should this handler follow the same pattern to avoid socket leaks?

src/services/managed-binary/download.ts
```diff
@@ -0,0 +1,154 @@
...
+			output.on("finish", () => {
+				output.close()
+				resolve()
+			})
+			output.on("error", reject)
+		})
```
edelauna, line 150: Same question: if `output` errors after piping has started, should we also destroy `response` and `request` to release the underlying socket and in-flight HTTP connection?

src/services/managed-binary/__tests__/download.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+		mockCreateReadStream.mockReturnValue(input as ReturnType<typeof createReadStream>)
+		const verification = verifySha256Checksum(
+			"/tmp/archive",
+			"expected-checksum",
+			(actual) => new Error(`checksum mismatch: ${actual}`),
+		)
+		input.emit("data", Buffer.from("archive"))
+		input.emit("end")
+		await expect(verification).rejects.toThrow("checksum mismatch: actual-checksum")
+	})
```
edelauna, line 105: Is there a test for the happy path here — where the computed digest actually equals `expected` and `verifyFileChecksum` resolves without throwing? The suite currently only exercises the mismatch branch.

src/services/managed-binary/__tests__/archive.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+
+	it("validates a single-file tar.xz layout before extraction", async () => {
+		const listing = createChild()
+		const extraction = createChild()
+		mockSpawn.mockReturnValueOnce(listing as unknown as ReturnType<typeof spawn>)
+		mockSpawn.mockReturnValueOnce(extraction as unknown as ReturnType<typeof spawn>)
+		const result = extractSingleFileTarXzArchive("/tmp/archive.tar.xz", "/tmp/output", "binary", "Tool")
+		listing.stdout.write("./binary\n")
+		listing.emit("close", 0)
```
edelauna, line 119: This feeds the happy path (single `./binary` entry). Are there tests for the error branches — e.g. writing `'./binary\n./other\n'` to `listing.stdout` (multi-entry) or a filename that doesn't match `info.binary` (wrong-filename error at `archive.ts:112`)?

src/services/managed-binary/install.ts
```diff
@@ -0,0 +1,145 @@
...
+async function installManagedBinaryWithLock(options: ManagedBinaryInstallOptions): Promise<string> {
+	await fs.mkdir(options.storageDir, { recursive: true })
+	const lockTarget = path.join(options.storageDir, `.${options.id}.install`)
+	const release = await lockfile.lock(lockTarget, {
+		realpath: false,
+		stale: 5 * 60_000,
+		update: 30_000,
+		retries: { retries: 10, factor: 1.5, minTimeout: 100, maxTimeout: 1_000 },
+	})
+	try {
```
edelauna, line 131: Is there a reason not to pass an `onCompromised` callback here? `safeWriteJson.ts` uses one so that stale-lock detection failures surface loudly rather than silently continuing — should the same policy apply to the binary install lock?

src/services/managed-binary/archive.ts
```diff
@@ -0,0 +1,117 @@
...
+			destination,
+		])
+		return
+	}
+
+	await runProcess("unzip", ["-o", archivePath, "-d", destination])
+}
+
+export async function extractSingleFileZipArchive(
+	archivePath: string,
```
edelauna, line 68: Should this have a `process.platform === 'win32'` guard, or at least a JSDoc note that it only works on Windows? PowerShell is invoked unconditionally, so calling this function on Linux/macOS will produce a confusing error.

src/services/managed-binary/__tests__/install.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+				await writeFile(path.join(stagingDir, "example"), "binary")
+			},
+		})
+		const first = ensureManagedBinaryInstalled(options)
+		const second = ensureManagedBinaryInstalled(options)
+		expect(first).toBe(second)
+		await vi.waitFor(() => expect(download).toHaveBeenCalledOnce())
+		finishDownload?.()
+		await Promise.all([first, second])
+		expect(download).toHaveBeenCalledOnce()
```
edelauna, line 75: Are the resolved values from both calls asserted? A mutation that returns the wrong path would still pass if only the settlement of `Promise.all` is tested, not the actual paths resolved by `first` and `second`.

src/services/code-index/semble/semble-downloader.ts
```diff
@@ -164,130 +100,40 @@ export async function downloadSemble(storageDir: string): Promise<string | undef
...
+		versionFile: VERSION_FILE,
+		archiveName: info.archive,
+		binaryName: info.binary,
+		errorPrefix: "Failed to download semble",
+		download: (archivePath) =>
+			downloadBinaryFile(url, archivePath, {
+				name: "Semble",
+				trustedDomains: TRUSTED_DOWNLOAD_DOMAINS,
+				timeoutMs: 120_000,
+			}),
```
edelauna, line 126: The PR description highlights bounded downloads as a key property of `downloadBinaryFile` — does omitting `maxBytes` here leave the Semble download unbounded? Should `info.size` (or a conservative upper limit) be threaded through?

src/services/code-index/semble/semble-downloader.ts
```diff
@@ -164,130 +100,40 @@ export async function downloadSemble(storageDir: string): Promise<string | undef
...
+		verifyArchive: (archivePath) => verifyChecksum(archivePath, expectedChecksum),
+		extractArchive: async (archivePath, stagingDir) => {
+			if (info.archive.endsWith(".tar.gz")) {
+				await extractTarGzArchive(archivePath, stagingDir)
+			} else if (info.archive.endsWith(".zip")) {
+				await extractZipArchive(archivePath, stagingDir)
```
edelauna, line 132: If `info.archive` is neither `.tar.gz` nor `.zip`, this falls through silently — the archive is downloaded but never extracted and `binaryPath` won't exist. Should an unrecognised format throw instead of no-op?

edelauna (changes requested):
Had some questions about the implementation.

edelauna (approved):

PR #1060, opened 2026-07-30
[Improve] Harden DCG binary installation

### LINKED ISSUE #1056: Add the Destructive Command Guard binary service and runner
## Parent

Part of #1049. This is the second independently reviewable slice of the Destructive Command Guard work.

## Dependency

Depends on the shared managed-binary infrastructure issue created as the first slice.

## Problem

The extension needs a small, testable service boundary for selecting, installing, verifying, and invoking the pinned DCG binary. Command-execution policy should not own platform/archive/process details.

## Scope

- Define the pinned DCG version and supported platform/architecture archive metadata.
- Install and verify DCG through the shared managed-binary infrastructure.
- Cache/deduplicate installation within the extension process while allowing retries after failure.
- Run DCG with a command and working directory, parse its structured result, and map it to a typed allow/deny result.
- Treat unsupported platforms, malformed output, process failures, and timeouts conservatively.
- Add focused manager and runner tests.

## Out of scope

- Persisted settings or settings UI.
- Auto-approval policy and command tool integration.
- Chat rendering for denied commands.

## Acceptance criteria

- [ ] Supported platforms resolve pinned archives with checksums.
- [ ] Installation uses the shared managed-binary lifecycle.
- [ ] Concurrent callers share one installation attempt and failures can be retried.
- [ ] Runner output is parsed into a typed allow/deny result with optional reason/rule metadata.
- [ ] Runner failures and invalid output do not accidentally classify a command as safe.
- [ ] Focused DCG service tests pass.
- [ ] The stacked branch type-checks/builds independently.

## Stack

Base: the managed-binary PR. The settings issue will build on this issue.

navedmerchant:
## What changed

Completes the DCG binary-service safeguards with macOS temporary-directory support, an explicit skipped ZIP-only test on non-Windows CI, and coverage for trusted cross-host GitHub release redirects.

## Why this change was made

DCG must run with the minimum viable environment and install predictably across supported platforms and GitHub's release-asset redirect flow. Closes #1056.

## Impact

DCG installation and execution are more reliable across platforms while keeping extension-host secrets out of the child process environment.

## Related PRs

- [Managed binary infrastructure](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1059)
- [DCG command integration](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1062)

## Summary by CodeRabbit

- **New Features**
  - Added a destructive-command guard that evaluates commands and returns allow or deny decisions, including optional explanations and rule identifiers.
  - Added automatic installation of the guard binary with platform support checks, secure downloads, checksum validation, and archive extraction.
  - Added safeguards for restricted environments, timeouts, excessive output, invalid responses, and process failures.

- **Tests**
  - Added comprehensive coverage for installation, downloads, validation, platform handling, command execution, and failure scenarios.


src/services/destructive-command-guard/__tests__/runner.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+			{ decision: "deny", ruleId: "core:delete" },
+		],
+	])("accepts valid DCG result %#", async (payload, code, expected) => {
+		const child = createChild()
+		useChild(child)
+
+		const result = runDcg("/dcg", "echo test", "/workspace")
+		emitResult(child, payload, code)
+
+		await expect(result).resolves.toEqual(expected)
```
edelauna, line 70: Every case here calls `runDcg` with `"/workspace"` as `cwd`, but none of the assertions check that it actually reaches `spawn`'s options. Worth adding an `expect(mockSpawn).toHaveBeenCalledWith(..., expect.objectContaining({ cwd: "/workspace" }))` somewhere so a dropped/hardcoded `cwd` would fail?

src/services/destructive-command-guard/__tests__/manager.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+		const extraction = extractSingleBinary("/tmp/dcg.tar.xz", tempDir, DCG_ARCHIVES["linux-x64"])
+		child.emit("close", 0)
+
+		await expect(extraction).resolves.toBeUndefined()
+		expect(mockSpawn).toHaveBeenCalledTimes(1)
+		expect(mockSpawn).toHaveBeenCalledWith(
+			"tar",
+			expect.arrayContaining(["-xJf", "/tmp/dcg.tar.xz", "-C", tempDir, "--no-same-owner"]),
+			expect.objectContaining({ shell: false }),
+		)
```
edelauna: Since `extractTarXzArchive` appends `--no-overwrite-dir` on Linux, wouldn't `arrayContaining`/`objectContaining` here miss that flag being dropped or `stdio` being stripped? The ZIP test above uses an exact `toHaveBeenCalledWith` — could this match that for the same strictness?

edelauna (changes requested):
Looks good - just had 2 comments for increased test coverage.

src/services/destructive-command-guard/constants.ts
```diff
@@ -0,0 +1,41 @@
+export const DCG_VERSION = "v0.7.7"
+
+export type DcgArchiveInfo = Readonly<{
+	archive: string
+	binary: "dcg" | "dcg.exe"
+	sha256: string
+}>
+
+export const DCG_ARCHIVES: Readonly<Record<string, DcgArchiveInfo>> = {
```
edelauna, line 9: added #1067 similar to what we could use for semble for dependency management

src/services/destructive-command-guard/manager.ts
```diff
@@ -0,0 +1,98 @@
+import * as path from "path"
+
+import { extractTarXzArchive, extractZipArchive } from "../managed-binary/archive"
+import {
+	downloadBinaryFile,
+	isTrustedHttpsUrl,
+	resolveTrustedRedirect as resolveManagedBinaryRedirect,
+	verifySha256Checksum,
+} from "../managed-binary/download"
+import { ensureManagedBinaryInstalled, getManagedBinaryPaths } from "../managed-binary/install"
```
edelauna: Is `getManagedBinaryPaths` actually used anywhere in this file? It looks like an unused import — with `--max-warnings=0` that would break CI.

src/services/destructive-command-guard/manager.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export function downloadFile(url: string, destination: string, maxRedirects = 5): Promise<void> {
+	return downloadBinaryFile(url, destination, {
+		name: "DCG",
+		trustedDomains: DCG_TRUSTED_DOWNLOAD_DOMAINS,
+		timeoutMs: 120_000,
+		maxRedirects,
+	})
+}
```
edelauna, line 59: Could a compromised release stream an arbitrarily large file here? Both the Content-Length pre-check and the per-chunk limit in `downloadBinaryFile` are gated on `options.maxBytes !== undefined` — without it, both guards are silently skipped.

```suggestion
export function downloadFile(url: string, destination: string, maxRedirects = 5): Promise<void> {
	return downloadBinaryFile(url, destination, {
		name: "DCG",
		trustedDomains: DCG_TRUSTED_DOWNLOAD_DOMAINS,
		timeoutMs: 120_000,
		maxRedirects,
		maxBytes: 50 * 1024 * 1024,
	})
}
```

src/services/destructive-command-guard/manager.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export async function verifyChecksum(filePath: string, expected: string): Promise<void> {
+	await verifySha256Checksum(filePath, expected, () => new Error("DCG archive checksum verification failed"))
```
edelauna: The error factory receives `actual` as a parameter but ignores it — including the actual hash in the message would make a mismatch much easier to debug.

```suggestion
	await verifySha256Checksum(filePath, expected, (actual) => new Error(`DCG archive checksum verification failed (got ${actual})`))
```

src/services/destructive-command-guard/manager.ts
```diff
@@ -0,0 +1,98 @@
...
+		return
+	}
+
+	await extractTarXzArchive(archivePath, stagingDir)
+}
+
+function installDcg(storageDir: string): Promise<string | undefined> {
+	const info = getDcgArchiveInfo()
+	if (!info) {
+		return Promise.resolve(undefined)
```
edelauna, line 86: When the current platform has no archive entry, DCG protection silently disables with no indication to the caller. Should this log a warning so it’s visible in diagnostics?

src/services/destructive-command-guard/runner.ts
```diff
@@ -0,0 +1,84 @@
...
+	pack_id?: string
+}
+
+export function runDcg(binaryPath: string, command: string, cwd: string): Promise<DcgDecision> {
+	return new Promise((resolve, reject) => {
+		const child = spawn(binaryPath, ["test", "--format", "json", "--no-color", command], {
+			cwd,
+			shell: false,
+			stdio: ["ignore", "pipe", "pipe"],
+			env: { ...process.env, NO_COLOR: "1" },
```
edelauna: Does DCG need the full parent environment? Spreading `process.env` exposes every secret in the extension-host process — `ANTHROPIC_API_KEY`, `GITHUB_TOKEN`, OAuth tokens — to a third-party binary. Could this use a minimal allowlist like `{ NO_COLOR: "1", PATH: process.env.PATH ?? "", HOME: process.env.HOME ?? "" }` plus only what DCG’s docs require?

src/services/destructive-command-guard/runner.ts
```diff
@@ -0,0 +1,84 @@
...
+		let stdout: Buffer<ArrayBufferLike> = Buffer.alloc(0)
+		let stderr: Buffer<ArrayBufferLike> = Buffer.alloc(0)
+		let settled = false
+		const fail = (error: Error) => {
+			if (settled) return
+			settled = true
+			clearTimeout(timer)
+			child.kill("SIGKILL")
+			reject(error)
+		}
```
edelauna: Should `fail` log before rejecting? All five error paths (timeout, signal exit, bad JSON, unsupported schema, decision mismatch) currently reject silently. A single `console.warn("[DCG]", error.message)` here would cover all of them.

src/services/destructive-command-guard/runner.ts
```diff
@@ -0,0 +1,84 @@
...
+			}
+
+			const schemaVersion = Number(payload.schema_version)
+			if (![1, 2].includes(schemaVersion)) {
+				reject(new Error("DCG returned an unsupported response schema"))
+				return
+			}
+			if (payload.decision === "allow" && code === 0) {
+				resolve({ decision: "allow" })
+			} else if (payload.decision === "deny" && code === 1) {
```
edelauna, line 83: Is there a reason the `deny` outcome isn’t logged? A blocked command is the primary security event this service produces — resolving silently makes it invisible in diagnostics.

src/services/destructive-command-guard/runner.ts
```diff
@@ -0,0 +1,84 @@
...
+			child.kill("SIGKILL")
+			reject(error)
+		}
+		const append = (current: Buffer<ArrayBufferLike>, chunk: Buffer<ArrayBufferLike>): Buffer<ArrayBufferLike> => {
+			if (current.length + chunk.length > DCG_MAX_OUTPUT_BYTES) {
+				fail(new Error("DCG produced too much output"))
+				return current
+			}
+			return Buffer.concat([current, chunk])
+		}
```
edelauna: The name `append` doesn’t hint that it can also reject the outer promise. Could the size check be extracted to the call site, or the function renamed to make the side-effect visible?

src/services/destructive-command-guard/__tests__/manager.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+			path.join("/storage", "destructive-command-guard", "dcg"),
+		)
+	})
+
+	it("accepts only HTTPS URLs on trusted host boundaries", () => {
+		expect(isTrustedDownloadUrl("https://github.com/release")).toBe(true)
+		expect(isTrustedDownloadUrl("https://cdn.objects.githubusercontent.com/release")).toBe(true)
+		expect(isTrustedDownloadUrl("http://github.com/release")).toBe(false)
+		expect(isTrustedDownloadUrl("https://evilgithub.com/release")).toBe(false)
+		expect(isTrustedDownloadUrl("not a URL")).toBe(false)
```
edelauna, line 67: Would `https://github.com.evil.com/release` pass here? The assertions cover subdomain spoofing (`evilgithub.com`) but not domain-boundary squatting where `github.com` appears as a prefix.

```suggestion
		expect(isTrustedDownloadUrl("https://evilgithub.com/release")).toBe(false)
		expect(isTrustedDownloadUrl("https://github.com.evil.com/release")).toBe(false)
		expect(isTrustedDownloadUrl("not a URL")).toBe(false)
```

src/services/destructive-command-guard/__tests__/manager.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+	})
+
+	it("rejects untrusted download URLs before opening a destination", async () => {
+		await expect(downloadFile("https://example.com/dcg", path.join(tempDir, "archive"))).rejects.toThrow(
+			"DCG download URL is not a trusted HTTPS host",
+		)
+	})
```
edelauna, line 74: Are there plans to cover the HTTP failure paths for `downloadFile`? The three main cases — non-200 response, socket error (`request.emit("error")`), and request timeout — aren’t tested here.

src/services/destructive-command-guard/__tests__/manager.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+		expect(mockSpawn).not.toHaveBeenCalled()
+		if (process.platform !== "win32") {
+			expect((await stat(binaryPath!)).mode & 0o111).toBe(0o111)
+		}
+	})
+
+	it("downloads, verifies, extracts, and deduplicates a new installation", async () => {
+		const info = getDcgArchiveInfo()
+		expect(info).toBeDefined()
+		if (!info || info.archive.endsWith(".zip")) return
```
edelauna, line 263: This guard means the complete `.zip` path — PowerShell extraction, atomic rename, version-file write, archive cleanup — is never exercised on the only platform that uses it. Would a sibling test that `return`s when `!info.archive.endsWith(".zip")` be worth adding?

edelauna (changes requested):
😅 have a couple additional comments related to the logic used for the guard.

src/services/destructive-command-guard/runner.ts
```diff
@@ -0,0 +1,101 @@
...
+type DcgJsonOutput = {
+	schema_version?: number | string
+	decision?: string
+	reason?: string
+	rule_id?: string
+	pattern_name?: string
+	pack_id?: string
+}
+
+const DCG_ENV_KEYS = ["HOME", "PATH", "TEMP", "TMP", "USERPROFILE", "SystemRoot", "WINDIR"] as const
```
edelauna: Is `TMPDIR` intentionally omitted? macOS uses it (not `TEMP`/`TMP`) as the default temp dir; without it, sandboxed environments may fall back to a path the DCG binary cannot write to.

```suggestion
const DCG_ENV_KEYS = ["HOME", "PATH", "TEMP", "TMPDIR", "TMP", "USERPROFILE", "SystemRoot", "WINDIR"] as const
```

src/services/destructive-command-guard/__tests__/manager.spec.ts
```diff
@@ -0,0 +1,377 @@
...
+				DCG_VERSION,
+			)
+		} finally {
+			Object.defineProperty(info, "sha256", { value: originalChecksum, configurable: true })
+		}
+	})
+
+	it("downloads, verifies, extracts, and installs a ZIP archive", async () => {
+		const info = getDcgArchiveInfo()
+		if (!info?.archive.endsWith(".zip")) return
```
edelauna: Would `it.skipIf(!getDcgArchiveInfo()?.archive.endsWith(".zip"))` give more honest CI output here? The current early `return` makes Vitest report this as a passing test with zero assertions on every non-Windows runner.

src/services/destructive-command-guard/__tests__/manager.spec.ts
```diff
@@ -0,0 +1,377 @@
...
+			return request as unknown as ReturnType<typeof get>
+		})
+
+		await expect(downloadFile("https://github.com/release", path.join(tempDir, "archive"))).rejects.toThrow(
+			"DCG archive exceeds the download size limit",
+		)
+	})
+
+	it("allows trusted relative redirects and rejects unsafe or exhausted redirects", () => {
+		expect(resolveTrustedRedirect("https://github.com/release", "/asset", 5)).toBe("https://github.com/asset")
```
edelauna, line 132: Is there a case here for an absolute redirect to a second trusted host (e.g. `https://github.com/release` → `https://release-assets.githubusercontent.com/asset`)? That is the production redirect pattern for GitHub release assets, but the happy path only exercises relative redirects.

edelauna (dismissed):
Looks good - approving with some minor nits, which could also always be addressed later if we want.

navedmerchant:
Merging, base branch merged, code has not changed

edelauna (approved):

PR #1061, opened 2026-07-30
[Feat] Add a persisted opt-in setting for DCG

navedmerchant:
### Related GitHub Issue

Closes #1057
Part of #1049. Split from #1050.

### Description

Adds the persisted `destructiveCommandGuardEnabled` setting across shared types, extension/provider state, SettingsView's local cached state, save handling, and Auto-Approve settings UI.

Enabling verifies DCG installation. Failure normalizes the setting back to disabled and shows a localized error. Existing allowlist/denylist behavior remains available while DCG is disabled. This PR does not yet change command approval policy.

### Stack

**3 / 4 — base: `feat/dcg-binary-service`**

Depends on PR 2 only for enable-time installation verification.

### Test Procedure

```sh
pnpm --dir src exec vitest run core/webview/__tests__/ClineProvider.spec.ts core/webview/__tests__/webviewMessageHandler.spec.ts
pnpm --dir webview-ui exec vitest run src/components/settings/__tests__/AutoApproveSettings.spec.tsx
pnpm --dir packages/types check-types
pnpm --dir src check-types
pnpm --dir webview-ui check-types
node scripts/find-missing-translations.js --area=core
node scripts/find-missing-translations.js --area=webview
```

Result: 183 extension and 11 webview tests passed; all type-checks, lint, and translation validation passed.

### Checklist

- [x] Full setting round trip implemented
- [x] SettingsView uses local cached state
- [x] Tests/type-check/lint/i18n checks passing

## Summary by CodeRabbit

* **New Features**
  * Added an optional Destructive Command Guard for automatically approved terminal commands.
  * The guard is disabled by default and can be enabled from Execute settings.
  * Existing allowed and denied command controls are hidden while the guard is active.
  * Added localized setting labels, descriptions, and error messages across supported languages.

* **Bug Fixes**
  * Enabling the guard now reports installation or platform availability failures and safely remains disabled.

* **Tests**
  * Added coverage for settings persistence, validation, toggling, installation failures, and conditional controls.

Issue #1049, opened 2026-07-29
Add destructive command guard for smooth command auto-approval


navedmerchant:
**Is your feature request related to a problem? Please describe.**
Modern models increasingly execute complex commands and generate entire scripts to complete tasks. GPT-5.6 is one example. Even when command execution is auto-approved, command allowlists and parsing can interrupt these workflows, causing an otherwise autonomous task to stop for approval or fail to progress.

At the same time, broadly auto-approving every command without a safety boundary would remove the user's opportunity to review genuinely dangerous operations.

**Describe the solution you'd like**
Integrate Destructive Command Guard (DCG) with command execution and auto-approval. When command auto-approval and DCG are enabled:

- Commands classified as safe by DCG should execute smoothly without being constrained by the command allowlist.
- Commands classified as dangerous should still require explicit user approval.
- If a dangerous command is denied, the chat should show a clear, localized explanation, including the DCG reason or rule when available.
- DCG should be downloadable and manageable from Auto-Approve settings, with its enabled state persisted across extension and webview state.
- The existing allowlist and denylist workflow should remain available when DCG is disabled.

This preserves user control at the dangerous-command boundary while allowing modern models to run complex, multi-line, or script-based commands without unnecessary interruptions.

**Describe alternatives you've considered**
- Continue expanding command allowlists. This does not scale to generated scripts or novel compound commands and still produces interruptions.
- Auto-approve every command unconditionally. This is smooth but removes the user's chance to inspect destructive commands.
- Rely only on the existing denylist. Static patterns cannot classify the full context of complex generated commands as effectively as DCG.

**Additional context**
The intended UX is that normal command execution stays uninterrupted, while DCG-classified dangerous commands pause and give the user a clear chance to approve or deny them. This is especially important as newer models increasingly write and execute complete scripts as part of routine task completion.

PR #1062, opened 2026-07-30
[Improve] Streamline DCG command approval

### LINKED ISSUE #1058: Integrate DCG with command auto-approval and denied-command UX
## Parent

Part of #1049. This is the fourth and final independently reviewable slice of the Destructive Command Guard work.

## Dependency

Depends on the persisted DCG setting and service slices.

## Problem

When command auto-approval and DCG are enabled, safe complex/generated commands should run without the legacy allowlist interrupting them, while DCG-blocked commands must retain an explicit approval boundary and clear denial feedback.

## Scope

- Evaluate commands with DCG before asking for command approval.
- Allow DCG-approved commands to bypass the legacy allowlist only when command auto-approval and DCG are both enabled.
- Mark DCG-blocked commands as protected so they always require explicit user approval.
- Preserve existing allowlist/denylist behavior when DCG is disabled.
- Record auto-deny decisions in command messages and render a dedicated denied state in chat.
- Show localized DCG reason/rule details when available and hide allow/deny pattern controls where they do not apply.
- Add focused extension and webview regression tests for policy precedence, command execution, message persistence, and denied rendering.

## Acceptance criteria

- [ ] DCG-approved commands auto-run only when command auto-approval is enabled.
- [ ] DCG-blocked commands cannot cross the explicit approval boundary automatically.
- [ ] User approval can still execute a DCG-blocked command; user denial does not execute it.
- [ ] DCG-disabled behavior is unchanged.
- [ ] Denied commands render a clear localized status and optional reason/rule metadata.
- [ ] Focused command-policy, tool, message, and chat tests pass.
- [ ] The complete stacked branch type-checks/builds and the combined feature runs.

## Stack

Base: the persisted-setting PR. This final PR completes #1049.

navedmerchant:
## What changed

Reads the persisted DCG setting without eagerly building full provider state, restores full state loading until after approval, and adds coverage for an allowed command continuing through execution and for command rows awaiting a decision.

## Why this change was made

The DCG integration should avoid unnecessary state work on ordinary commands while preserving explicit approval for guarded commands. Closes #1058 and completes #1049.

## Impact

Command approval remains responsive when DCG is disabled, while guarded-command behavior and denied-state rendering are covered more completely.

## Related PRs

- [Managed binary infrastructure](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1059)
- [DCG binary service](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1060)

## Summary by CodeRabbit

* **New Features**
  * Added destructive command protection with localized block messages, reasons, and rule details.
  * Commands approved by the protection system can bypass standard approval rules.
  * Added clear denied-command indicators in chat, with support for multiple languages.

* **Bug Fixes**
  * Commands denied by protection now require explicit approval and fail safely when protection is unavailable.
  * Hidden command-pattern controls prevent changes when protection is active or a command is denied.

* **Tests**
  * Expanded coverage for approval precedence, blocked commands, denied states, localization, and fail-closed behavior.


src/core/auto-approval/__tests__/dcg.spec.ts
```diff
@@ -0,0 +1,69 @@
+import { checkAutoApproval } from ".."
+
+describe("Destructive Command Guard auto-approval precedence", () => {
+	const baseState = {
+		autoApprovalEnabled: true,
```
edelauna, line 5: All tests here keep `autoApprovalEnabled: true`. Is there a case worth covering where a user has DCG on but global auto-approval off? The early exit at `index.ts:63` handles it, but nothing asserts that DCG never fires when auto-approval is disabled.

src/core/auto-approval/index.ts
```diff
@@ -115,8 +116,18 @@ export async function checkAutoApproval({
...
+		if (isProtected) {
+			return { decision: "ask" }
+		}
 
 		if (state.alwaysAllowExecute === true) {
+			// Execute commands immediately when DCG allows them. ExecuteCommandTool
+			// marks commands blocked by DCG as protected before reaching this check,
+			// which keeps the explicit user approval prompt for those commands.
+			if (state.destructiveCommandGuardEnabled === true) {
+				return { decision: "approve" }
```
edelauna, line 130: When DCG is enabled, this approves all DCG-allowed commands before `getCommandDecision` runs — so any entries in `deniedCommands` are silently bypassed. Worth noting in the comment above, or is that intentional and already communicated elsewhere?

edelauna (changes requested):
some minor questions

src/core/auto-approval/index.ts
navedmerchant: it is intentional, dcg approves commands, and if it doesnt, it asks the user for an one time override. Happy to improve the implementation in the future, but I dont think anyone wants to just stop execution if a bad command is about to be executed,I would rather ask the user and pause the agent loop.

src/core/tools/ExecuteCommandTool.ts
```diff
@@ -115,16 +131,44 @@ export class ExecuteCommandTool extends BaseTool<"execute_command"> {
 				return
 			}
 
-			const didApprove = await askApproval("command", canonicalCommand)
+			const provider = await task.providerRef.deref()
+			const providerState = await provider?.getState()
+			let dcgBlocked = false
+			if (providerState?.destructiveCommandGuardEnabled === true) {
+				const { ensureDcgInstalled, runDcg } = await import("../../services/destructive-command-guard")
+				if (!provider) {
```
edelauna: Can this guard ever fire? `provider` is obtained at line 134 via `providerRef.deref()`, then used with optional chaining: `provider?.getState()`. So `providerState?.destructiveCommandGuardEnabled === true` is only `true` when `provider` is non-null — meaning we can only enter this block when `provider` is already guaranteed non-null. If `provider` is null, the outer `if` is false and we skip DCG entirely, silently bypassing it rather than throwing.

src/core/tools/ExecuteCommandTool.ts
```diff
@@ -115,16 +131,44 @@ export class ExecuteCommandTool extends BaseTool<"execute_command"> {
 				return
 			}
 
-			const didApprove = await askApproval("command", canonicalCommand)
+			const provider = await task.providerRef.deref()
+			const providerState = await provider?.getState()
```
edelauna: This moves `getState()` from post-approval to pre-approval, so it now runs for every command — including when DCG is disabled (the common case) and for commands the user declines. Is there a cheaper way to read just `destructiveCommandGuardEnabled` before deciding whether to invoke the full state?

src/core/tools/__tests__/executeCommandTool.spec.ts
```diff
@@ -199,6 +209,129 @@ describe("executeCommandTool", () => {
...
+				terminalShellIntegrationDisabled: true,
+			})
+			await executeCommandTool.handle(mockCline as unknown as Task, mockToolUse, {
+				askApproval: mockAskApproval as unknown as AskApproval,
+				handleError: mockHandleError as unknown as HandleError,
+				pushToolResult: mockPushToolResult as unknown as PushToolResult,
+			})
+
+			expect(mockEnsureDcgInstalled).toHaveBeenCalledWith("/test/storage")
+			expect(mockRunDcg).toHaveBeenCalledWith("/test/storage/dcg", "echo test", "/test/workspace")
```
edelauna, line 291: This confirms DCG was invoked correctly, but does it verify the allow result was acted upon? `mockRunDcg` defaults to `{ decision: "allow" }`, so a mutation that ignores `dcgResult.decision` and always calls `askApproval` without `isProtected` would still pass both of these assertions.

webview-ui/src/components/chat/__tests__/ChatRow.command-denied.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+		expect(screen.getByTestId("command-execution")).toHaveAttribute("data-denied", "true")
+	})
+
+	it("does not show a denied status for an approved command", () => {
+		renderCommand("approve")
+
+		expect(screen.queryByText("chat:commandExecution.denied")).not.toBeInTheDocument()
+		expect(screen.getByTestId("command-execution")).toHaveAttribute("data-denied", "false")
+	})
+})
```
edelauna, line 77: The two cases cover `"deny"` and `"approve"`, but not `autoApprovalDecision: undefined` — the normal state for commands awaiting user input. A mutation of `=== "deny"` to `!== "approve"` in `ChatRow.tsx` would make those render as denied without being caught. Worth a third `renderCommand(undefined)` case?

webview-ui/src/components/chat/ChatRow.tsx
```diff
@@ -290,6 +291,18 @@ export const ChatRowContent = ({
 			case "mistake_limit_reached":
 				return [null, null] // These will be handled by ErrorRow component
 			case "command":
+				if (message.autoApprovalDecision === "deny") {
```
edelauna, line 294: When DCG is enabled, can `autoApprovalDecision` ever be `"deny"` here? `checkAutoApproval` short-circuits to `approve` before reaching `getCommandDecision` when DCG is on, so this branch looks like it only fires via the legacy denylist (DCG disabled). Is that the intent, or should DCG blocks also show this denied state?

edelauna (changes requested):
Had some comments related to implementation.

edelauna (dismissed):
Thanks for addressing the feedback, approving - but still unclear on how you want to handle the design for a DCG blocked command: https://github.com/Zoo-Code-Org/Zoo-Code/pull/1062#discussion_r3697061277

webview-ui/src/components/chat/ChatRow.tsx
navedmerchant: The command denial itself already worked before this PR—the backend synthesized a No response and skipped execution. What was missing was enough persisted state for the webview to distinguish auto-denial from auto-approval: both were stored only as isAnswered: true, so an auto-denied command could render like an ordinary/running command. autoApprovalDecision and this branch were added to make legacy denylist auto-denials render accurately. This is a UI/history improvement rather than part of DCG enforcement, and DCG blocks do not enter this branch.

edelauna (approved):

PR #977, opened 2026-07-21
fix(webview): add durable per-view state base

easonLiangWorldedtech:
### Related GitHub Issue

Closes: #984

- Historical context: old Issue #908 (root mechanism of per-view persistence)
- Split from old PR #909

### Description

Add the foundational per-view state infrastructure for parallel mode. This is the root PR that all subsequent parallel-mode PRs depend on.

**How:**

- **Per-view identity**: each webview instance gets a stable ID (generated in `webview-ui/src/utils/vscode.ts` via `getViewStateId()`, sent during launch). `ClineProvider.setViewStateId()` sanitizes it into a safe object key.
- **`viewLocalState` buffer**: transient per-view state that merges on top of the shared `ContextProxy` values in `getState()` (the `mergedStateValues` layer), so a tab's mode never overwrites the sidebar's.
- **Durable `viewStates` persistence**: registered global setting key storing only non-secret selections (`mode`, `currentApiConfigName`, `updatedAt`), bounded to the most recent 50 entries by `updatedAt` ordering.
- **Serialized writes**: every `viewStates` mutation goes through a static write queue (`persistedViewStateWriteQueue`) that re-reads the map fresh from `globalState` on each write, so concurrent sidebar/tab providers cannot clobber each other.
- **No-op compatibility**: existing single-tab behavior is unchanged.

**Reviewers should pay attention to:**

- Only non-secret fields are persisted (`mode`, `currentApiConfigName`). Full `apiConfiguration` (API keys, Kimi Code keys) is never written to globalState; e2e asserts no secret paths leak into persisted entries.
- Normal editor-provider teardown preserves persisted `viewStates` entries. Retention uses the 50-entry pruning cap. Follow-up `#1065` tracks explicit stale-entry cleanup.
- The PR also carries task-scoped API controls (`approveTaskAsk`, `selectTaskFollowupSuggestion`) and `preserveOpenTabs` for new tasks. These are required so parallel views can be driven per-task by the orchestrator e2e foundation (#1064), which is why they stay in this root PR rather than being split out.

### Test Procedure

Unit / integration (all green in CI):

- `pnpm --dir src test` — full core suite (129 files / 2184 passed / 9 skipped), including the new `ClineProvider.parallelMode.spec.ts` covering persistence, restoration, isolation, pruning, and concurrent-write serialization
- `pnpm --dir packages/types test` (6 passed) and `pnpm --dir webview-ui test` (viewStateId generation/restoration)
[...]

PR #1064, opened 2026-07-30
test(vscode-e2e): add orchestrator E2E foundation for parallel-mode coverage

easonLiangWorldedtech:
## Summary

This PR is a follow-up test foundation based on the discussion in #977 (`fix(webview): add durable per-view state base`). PR #977 introduces the durable per-view state base needed for parallel-mode work, and edelauna requested more real VS Code E2E coverage around the kinds of cross-provider / cross-panel boundaries that mocked unit tests cannot fully exercise.

Instead of putting all complex scenarios into one large commit, this PR adds the E2E foundation in reviewer-friendly slices. The focus here is not to change production orchestration behavior, but to establish durable E2E infrastructure for future parallel-mode work: multi-provider task lifecycles, delegation fan-in/fan-out, nested delegation, cancellation recovery, and resource convergence checks.

## Context from #977

#977 fixes the durable per-view state base so sidebar and editor-tab providers can maintain isolated view-local state while still sharing the same extension/global-state boundary. That PR is the root dependency for broader parallel-mode work.

The review feedback on #977 asked for more E2E coverage beyond mocked provider tests, especially around real extension-host behavior. This PR builds on that direction by adding a heavier E2E base that can be reused for future parallel-mode scenarios.

### 1. Single-round orchestrator fan-out / fan-in E2E

Adds a baseline orchestrator E2E where a parent orchestrator delegates sequential children to ask, architect, and code modes, then fans their summaries back into the parent completion.

This verifies:
- child task creation and completion
- parent / child history linkage
- delegation completion events
- final parent aggregation
- task stack cleanup

### 2. Repeated delegation stress E2E

Adds a repeated delegation workflow with three rounds of ask / architect / code children, validating that the orchestrator path remains stable across repeated child creation, completion, resume, and final aggregation.

This verifies:
- nine unique child tasks
- nine delegation completion events
- stable `childIds`
- per-round mode and summary fidelity
- final fan-in output
- stack convergence

### 3. Resource diagnostics API

Adds a small test-focused diagnostics API exposing stable counters:
- registered task count
- current task stack length
- key lifecycle listener counts

This is intentionally limited to deterministic resource counters rather than noisy heap/RSS thresholds.

### 4. Resource convergence sentinel

Wires the diagnostics API into the repeated orchestrator E2E so the test hard-gates that resources converge after cleanup.

This verifies:
- registered task count returns to baseline
- task stack length returns to baseline / zero
- listener counts return to baseline after local handlers are removed
- stale child tasks do not remain registered

### 5. Nested orchestrator delegation E2E

Adds a nested workflow:

- top-level orchestrator parent
- child orchestrator
- ask/code grandchildren
- child orchestrator fan-in
- top-level parent fan-in

This verifies stable delegation ownership across a task tree, not just a single parent-child edge.

### 6. Cancellation / recovery E2E

Adds coverage for cancelling/interruption of a delegated child and explicit recovery.

This verifies:
- parent does not prematurely resume or complete when the child is interrupted
- parent remains in the expected delegated/awaiting state
- explicit recovery resumes the child and returns the result to the parent
- final parent completion contains the cancellation recovery summary
- cleanup and resource diagnostics converge

## Normal/unit coverage

Each E2E slice includes normal/unit coverage for the helper or API logic it introduces:

- orchestrator fixture plan helpers
- repeated delegation plan generation
- nested delegation plan matching
- cancellation/recovery plan matching
- resource diagnostics convergence helper
- API resource diagnostics counters

## Why this is separate from #977

#977 is the production fix for durable per-view state. This PR is the E2E foundation requested from that review context. Keeping it separate makes the review easier:

- #977 can stay focused on the per-view state base.
- This PR focuses on heavier E2E infrastructure and future parallel-mode confidence.
- The commit history is intentionally split into small reviewer-friendly test slices.

## Validation

The following validation was run after cherry-picking onto latest main and resolving conflicts:

- `git status --short` — clean
- `git diff --name-only --diff-filter=U` — no unresolved conflicts
- `git diff --check` — passed
- `pnpm --dir apps/vscode-e2e exec vitest run src/fixtures/orchestrator-plan.spec.ts src/fixtures/resource-diagnostics.spec.ts` — passed, 23 tests
- `cd src ; npx vitest run extension/__tests__/api-resource-diagnostics.spec.ts` — passed, 3 tests
- `pnpm --dir src run check-types` — passed
- `pnpm --dir packages/types run check-types` — passed

A final targeted VS Code E2E run for `orchestrator.test` should be run before merge, since this branch was just replayed onto latest main.

## Relationship to future work

This PR is intended as the base for more complex parallel-mode E2E testing. It provides reusable orchestration fixtures, resource convergence helpers, and multi-task lifecycle coverage that future PRs can build on when testing real parallel panels/providers.

## Summary by CodeRabbit

* **New Features**
  * Added `getResourceDiagnostics()` to inspect resource diagnostics (registered task count, current task-stack length, and per-event listener counts).
* **Tests**
  * Added unit-test support for the e2e app with new specs covering resource-diagnostics convergence.
  * Expanded orchestrator e2e coverage with stronger delegation-plan assertions across fan-out, repeated delegation, nested orchestration, and cancellation recovery, plus standardized cleanup/diagnostics checks.
  * Added `API#getResourceDiagnostics` tests for pruning/ejection of previously registered tasks.
* **CI**
  * Updated the mocked e2e workflow to run the new unit-test step earlier.


easonLiangWorldedtech:
Addressed the CodeRabbit review feedback in commit `96bc9787d`:

- Fixed repeated-delegation final-result newline escaping so rows use real newlines.
- Added a single source of truth for the cancellation recovery follow-up tool-call ID and reused it in matchers/fixtures.
- Refactored orchestrator fixtures to use a barrel re-export plus shared child-completion and resume-fixture helpers.
- Documented the reverse resume registration order and resource diagnostics stack convergence asymmetry.
- Added baseline diagnostics/convergence checks consistently across all orchestrator E2E scenarios and capped task-stack cleanup attempts.
- Added `suiteTeardown` to restore the default OpenRouter config after the orchestrator suite.
- Reconciled `registeredTaskIds` with the provider task stack in resource diagnostics so evicted/disposed tasks are pruned, with regression coverage.
- Derived `RooCodeResourceDiagnosticEventName` from `RooCodeEventName` and tightened `RESOURCE_DIAGNOSTIC_EVENTS` typing.
- Moved the vscode-e2e unit-test CI step before VS Code binary download.

Validation run:

- `pnpm --dir apps/vscode-e2e exec vitest run src/fixtures/orchestrator-plan.spec.ts src/fixtures/resource-diagnostics.spec.ts` — passed, 23 tests.
- `cd src; npx vitest run extension/__tests__/api-resource-diagnostics.spec.ts` — passed, 4 tests.
- `pnpm --dir src run check-types` — passed.
- `pnpm --dir packages/types run check-types` — passed.
- `pnpm --dir apps/vscode-e2e run check-types` — passed.
- `pnpm --dir src exec eslint --prune-suppressions --max-warnings=0 extension/api.ts extension/__tests__/api-resource-diagnostics.spec.ts` — passed.
- `pnpm --dir apps/vscode-e2e exec eslint --max-warnings=0 src/fixtures/orchestrator-plan.ts src/fixtures/orchestrator.ts src/fixtures/resource-diagnostics.ts src/fixtures/orchestrator-plan.spec.ts src/suite/orchestrator.test.ts` — passed.
- `pnpm --dir packages/types exec eslint --max-warnings=0 src/api.ts` — passed.
- `git diff --check` — passed.
- Commit hook `turbo lint` — passed.

easonLiangWorldedtech:
Addressed the `RooCodeResourceDiagnosticEventName` exhaustiveness nit in commit `8ffd934ac`.

`RESOURCE_DIAGNOSTIC_EVENTS` is now derived from an exhaustive `Record<RooCodeResourceDiagnosticEventName, true>` in `src/extension/api.ts`, so any future mismatch with the source-of-truth union in `packages/types/src/api.ts` will fail at compile time instead of being hidden by the downstream listener-count cast.

Validation:
- `pnpm --dir src run check-types` — passed
- `pnpm --dir src exec eslint --prune-suppressions --max-warnings=0 extension/api.ts` — passed
- `cd src; npx vitest run extension/__tests__/api-resource-diagnostics.spec.ts` — passed, 4 tests
- `git diff --check` — passed
- commit hook `turbo lint` — passed

.github/workflows/e2e.yml
```diff
@@ -34,6 +34,10 @@ jobs:
             - name: Setup Node.js and pnpm
               if: github.event_name != 'pull_request' || steps.e2e-marker.outputs.cache-hit != 'true'
               uses: ./.github/actions/setup-node-pnpm
+            - name: Run vscode-e2e unit tests
```
edelauna: this should not be needed - use the existing harness for your e2e tests - specifically spawning parallel tabs and showing interactive concurrent conversations

edelauna (changes requested):

easonLiangWorldedtech:
rebased

.github/workflows/e2e.yml
easonLiangWorldedtech: removed

src/extension/api.ts
```diff
@@ -253,6 +269,25 @@ export class API extends EventEmitter<RooCodeEvents> implements RooCodeAPI {
...
+				this.registeredTaskIds.delete(taskId)
+			}
+		}
+
+		const listenerCounts = Object.fromEntries(
+			RESOURCE_DIAGNOSTIC_EVENTS.map((eventName) => [eventName, this.listenerCount(eventName)]),
+		) as RooCodeResourceDiagnostics["listenerCounts"]
+
+		return {
+			registeredTaskCount: this.registeredTaskIds.size,
```
edelauna: registeredTaskCount can't catch the leak this sentinel is for: the prune above deletes any id not on the current stack, so an off-stack leaked task is removed before this line counts it, and the value collapses to a subset of currentTaskStackLength. It also undercounts tab tasks (the newTab path adds to this set, but getCurrentTaskStack only sees the sidebar). TaskRegistry / getCurrentTaskStack already track liveness -- prefer reusing them over a second registry.

src/extension/api.ts
```diff
@@ -253,6 +269,25 @@ export class API extends EventEmitter<RooCodeEvents> implements RooCodeAPI {
...
+	public getResourceDiagnostics(): RooCodeResourceDiagnostics {
+		const activeTaskIds = new Set(this.getCurrentTaskStack())
+		for (const taskId of this.registeredTaskIds) {
+			if (!activeTaskIds.has(taskId)) {
+				this.registeredTaskIds.delete(taskId)
+			}
+		}
+
+		const listenerCounts = Object.fromEntries(
+			RESOURCE_DIAGNOSTIC_EVENTS.map((eventName) => [eventName, this.listenerCount(eventName)]),
```
edelauna: this.listenerCount counts subscribers on the public API emitter, but the per-task listeners are attached to `task`/`provider` in registerListeners, not to `this` (this.emit only re-broadcasts). So listenerCounts tracks external consumers (the test's own api.on, IPC), not the internal per-task listeners it is meant to prove converge -- an internal listener leak stays invisible.

edelauna (changes requested):
These seem like more unit tests, and seem to showcase running stuff in sequence not in parallel.

src/extension/api.ts
easonLiangWorldedtech: Addressed in `f21029bb0`. The second registry (`registeredTaskIds`) and its read-time pruning are gone. `getResourceDiagnostics()` now sources task liveness from the provider's `TaskRegistry` via new `ClineProvider.getTaskStackSize()` / `getLiveTasks()` accessors, so a task that is never actually evicted stays visible in `registeredTaskCount` instead of being pruned before the count, and tab tasks are no longer undercounted because the registry covers all live tasks rather than just the sidebar stack.

The unit spec was rewritten to cover the new semantics, including the leaked-task sentinel case: a task evicted from the task stack but never removed from the registry still shows up in `registeredTaskCount` and in the listener sums. The orchestrator e2e passes with this change (4/4 under `USE_MOCK=true`).

src/extension/api.ts
easonLiangWorldedtech: Addressed in `f21029bb0`. The diagnostics no longer read `this.listenerCount` (the public API emitter). Internal listener counts are now aggregated from the emitters `registerListeners()` actually attaches to: per-task events are counted on each live `Task` instance, and `taskCreated` plus the two delegation events are counted on the provider where their rebroadcast wiring is subscribed (the delegation events are additionally subscribed per task, so each scope contributes). A `DIAGNOSTIC_EVENT_SCOPES` table documents where each diagnostic event's wiring lives.

External consumers of the public emitter (the test's own `api.on`, IPC) are no longer conflated with the internal wiring, so an internal listener leak is now visible while external subscription churn is not. The rewritten unit spec covers the provider-wiring baseline, per-task accumulation, and exclusion of external consumers.

edelauna:
Flagging as invalid - original comment that these are unit tests and not e2e tests stands.

PR #1070, opened 2026-07-31
fix(telemetry): add circuit breaker and shutdown draining to TelemetryService

edelauna:
### Related GitHub Issue

Refs: #830 (reliability/event-volume-protection half — split from #835; consent/privacy half shipped separately in #1069)

### Description

`TelemetryService` had two reliability gaps:

1. **No upper bound on retry-storm events.** A `CODE_INDEX_ERROR` retry loop (e.g. a broken embedder config re-triggering on every file-system event) could flood the PostHog Product Analytics quota with no circuit breaker.
2. **Unbounded, un-awaited shutdown.** `deactivate()` called `TelemetryService.instance.shutdown()` without awaiting it, and `shutdown()` itself just fired `client.shutdown()` on each client without waiting for any in-flight `capture()`/`captureException()` calls to finish first — a capture that started just before shutdown could be silently dropped.

This PR adds:

- **Circuit breaker** for `CODE_INDEX_ERROR`, tracked independently of all other telemetry: a sliding 10-minute window, trips after 50 occurrences, then suppresses further captures of that event for a 10-minute cooldown before resetting. Unrelated events (`TASK_CREATED`, `TOOL_USED`, etc.) are never guarded and can't mask or reset the guarded count.
- **In-flight call tracking**: `captureEvent`/`captureException` now track the promise returned by each client's `capture()`/`captureException()` call. `shutdown()` drains these (looping, not a single snapshot, since draining one call's chain can enqueue more) before calling `client.shutdown()`, bounded to a 3-second timeout so a capture stuck on network I/O can't block extension deactivation indefinitely. A rejected capture is caught internally so it never surfaces as an unhandled rejection or blocks the drain.
- `isShuttingDown` gate so captures fired after `shutdown()` has started are dropped immediately instead of racing the drain loop.
- `src/extension.ts`: `deactivate()` now `await`s `TelemetryService.instance.shutdown()`, wrapped in try/catch so a shutdown failure can't skip the remaining terminal cleanup (`Terminal.setTerminalProfile`, `TerminalRegistry.cleanup`).

Capture call sites (`captureEvent`/`captureException`) remain synchronous/non-blocking for callers — only `shutdown()` awaits the drain.

### Out of scope

Kept narrow to service-level reliability, per the split from #835:

- Consent defaults, banner/settings/webview gating, localization, `PRIVACY.md` (in #1069).
- `Task Completed` aggregation (`toolsUsed`/`messageCount`), `AttemptCompletionTool` changes.
- Tool-name validation or model-controlled-name bucketing.
- Model-cache throttling/deduplication (`modelCache.ts`).
- E2E test stabilization and unrelated extension cleanup.
- No changesets added.

### Test Procedure

- `pnpm --filter @roo-code/telemetry test`: 42/42 passing, including two new files:
  - `TelemetryService.circuit-breaker.test.ts` — under-threshold pass-through, trip at 50 in-window, cooldown reset, unrelated events don't reset/count toward the guarded total, old occurrences expire out of the window, other event names are never guarded.
  - `TelemetryService.shutdown.test.ts` — shutdown waits for a pending capture that later resolves, drains a call still in flight across multiple loop passes, stops accepting new captures once shutdown has started, proceeds after the 3s timeout for a never-settling capture, a rejected capture doesn't prevent shutdown from completing.
- `pnpm check-types`: clean repo-wide.
- `eslint` on all 4 touched files: 0 errors; `src/eslint-suppressions.json` unchanged (no new suppressions).
- `src/__tests__/extension.spec.ts`: 4/4 passing, confirming the awaited-shutdown change doesn't break existing deactivation behavior.

### Pre-Submission Checklist

- [ ] **Visual Snapshot** (UI changes only): N/A — no UI changes.
- [x] **Documentation Impact**: No documentation updates required (no user-facing behavior change; `PRIVACY.md` already covers the opt-out default via #1069).

### Documentation Updates

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved telemetry reliability by limiting repeated event submissions.
  * Ensured telemetry captures complete gracefully during shutdown.
  * Prevented new telemetry captures from starting after shutdown begins.
  * Added timeout handling so shutdown cannot be blocked indefinitely.
  * Extension cleanup now continues even if telemetry shutdown encounters an error.

### DIFF
--- a/packages/telemetry/src/TelemetryService.ts
+++ b/packages/telemetry/src/TelemetryService.ts
@@ -129,7 +129,7 @@ export class TelemetryService {
 		occurrences.push(now)
 		this.guardedEventOccurrences.set(eventName, occurrences)
 
-		if (occurrences.length > CIRCUIT_BREAKER_MAX_IN_WINDOW) {
+		if (occurrences.length >= CIRCUIT_BREAKER_MAX_IN_WINDOW) {
 			this.trippedUntil.set(eventName, now + CIRCUIT_BREAKER_COOLDOWN_MS)
 			this.guardedEventOccurrences.delete(eventName)
 			return true
@@ -375,7 +375,8 @@ export class TelemetryService {
 			])
 		}
 
-		await Promise.all(this.clients.map((client) => client.shutdown()))
+		// allSettled, not all: one client rejecting must not stop us from awaiting the others.
+		await Promise.allSettled(this.clients.map((client) => client.shutdown()))
 	}
 
 	private static _instance: TelemetryService | null = null
--- a/packages/telemetry/src/TelemetryService.ts
+++ b/packages/telemetry/src/TelemetryService.ts
@@ -8,12 +8,61 @@ import {
 } from "@roo-code/types"
 
 /**
+ * Events prone to retry-storm-style repetition (e.g. a broken embedder config
+ * re-triggering on every file-system event). Guarded by a circuit breaker in
+ * `captureEvent` so a single broken install can't flood the Product Analytics
+ * quota. Tracked per-event via a sliding time window, independent of any
+ * other telemetry the same install may also be sending.
+ */
+const CIRCUIT_BREAKER_GUARDED_EVENTS = new Set<TelemetryEventName>([TelemetryEventName.CODE_INDEX_ERROR])
+
+/** Captures of a guarded event within the counting window allowed before the breaker trips. */
+const CIRCUIT_BREAKER_MAX_IN_WINDOW = 50
+
+/** Rolling window over which guarded-event occurrences are counted. */
+const CIRCUIT_BREAKER_WINDOW_MS = 10 * 60 * 1000
+
+/** How long a tripped breaker stays tripped before allowing captures again. */
+const CIRCUIT_BREAKER_COOLDOWN_MS = 10 * 60 * 1000
+
+/**
+ * Upper bound on how long shutdown() will wait for in-flight capture calls to drain.
+ * deactivate() awaits shutdown() before terminal cleanup, so an unbounded wait here
+ * (e.g. a capture stuck on network I/O that never resolves/rejects) would block the
+ * extension host from ever finishing deactivation. Losing an in-flight capture on
+ * timeout is an acceptable tradeoff against blocking shutdown indefinitely.
+ */
+const SHUTDOWN_DRAIN_TIMEOUT_MS = 3000
+
+/**
  * TelemetryService wrapper class that defers initialization.
  * This ensures that we only create the various clients after environment
  * variables are loaded.
  */
 export class TelemetryService {
+	// Timestamps of recent guarded-event occurrences, per event name, oldest first.
+	private guardedEventOccurrences = new Map<TelemetryEventName, number[]>()
+	private trippedUntil = new Map<TelemetryEventName, number>()
+
+	// In-flight client.capture()/captureException() promises. captureEvent/captureException are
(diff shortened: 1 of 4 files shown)


packages/telemetry/src/TelemetryService.ts
```diff
@@ -263,7 +356,27 @@ export class TelemetryService {
...
+		const drainStart = Date.now()
+		while (this.pendingClientCalls.size > 0 && Date.now() - drainStart < SHUTDOWN_DRAIN_TIMEOUT_MS) {
+			await Promise.race([
+				Promise.all(this.pendingClientCalls),
+				new Promise((resolve) => setTimeout(resolve, SHUTDOWN_DRAIN_TIMEOUT_MS - (Date.now() - drainStart))),
+			])
+		}
+
+		// allSettled, not all: one client rejecting must not stop us from awaiting the others.
+		await Promise.allSettled(this.clients.map((client) => client.shutdown()))
```
navedmerchant: Could we apply the three-second deadline to the entire telemetry shutdown? It currently bounds only the in-flight capture drain; afterward, client.shutdown() is awaited separately. With the locked posthog-node@5.1.1, calling shutdown() without an argument uses a 30-second timeout, so deactivation could take roughly 33 seconds before terminal cleanup runs. Since Promise.allSettled() consumes the timeout rejection, the outer deactivate() catch will not log it either. Please consider passing the remaining timeout to the client shutdown or adding an overall deadline, with a test for a client shutdown that never settles.

navedmerchant (approved):

packages/telemetry/src/TelemetryService.ts
edelauna: Good catch, fixed in 00589993c.

`client.shutdown()` is now raced against the same `SHUTDOWN_PHASE_TIMEOUT_MS` (3s) used for the capture drain, instead of being awaited unbounded:

```ts
await Promise.race([
	Promise.allSettled(this.clients.map((client) => client.shutdown())),
	new Promise((resolve) => setTimeout(resolve, SHUTDOWN_PHASE_TIMEOUT_MS)),
])
```

Renamed the constant from `SHUTDOWN_DRAIN_TIMEOUT_MS` to `SHUTDOWN_PHASE_TIMEOUT_MS` since it now bounds both phases separately — worst case is ~6s total (3s drain + 3s client shutdown) instead of the previous ~33s (3s drain + posthog-node's 30s default). Added a test (`does not hang forever when client.shutdown() itself never settles`) with a `client.shutdown()` that never resolves, confirming `TelemetryService.shutdown()` still returns.

Didn't try to pass the remaining drain-phase time budget into `client.shutdown()` itself — `TelemetryClient#shutdown()` takes no timeout parameter, and threading one through would touch the interface and every implementer for a benefit `Promise.race` already gets us for free.

One more thing worth flagging on this: I checked what guarantee VS Code actually gives `deactivate()` before window/process close, and it's weaker than either of us assumed — multiple open VS Code issues (e.g. microsoft/vscode#122825, #161147, #114688) describe the extension host not reliably waiting for an async `deactivate()` promise to resolve at all, with one citing a ~4s ceiling before the host moves on regardless. So even with this fix, `deactivate()`'s *other* steps (`McpServerManager.cleanup`, `Terminal.setTerminalProfile`, etc.) plus my ~6s worst case could still exceed what the host actually waits for in practice. I don't think that's blocking for this PR — it's strictly better than before, and it's the same weak guarantee every other step in `deactivate()` already lives with — but if you'd rather see the whole function share one tighter budget instead of telemetry alone claiming up to 6s, I'm happy to open a quick follow-up for that.

PR #1072, opened 2026-07-31
fix(model-cache): dedupe concurrent fetches and throttle empty-response telemetry

edelauna:
### Related GitHub Issue

Refs: #830 (model-cache concurrency and empty-response telemetry — split from #835; independent of #1069, #1070, and #1071)

### Description

`getModels()` had no in-flight deduplication. Two concurrent cache-miss calls for the same provider each fired their own provider fetch, and a `getModels()` fetch racing a `refreshModels()` fetch for the same key had no ordering guarantee — whichever call's cache write landed last won, even if it reflected an earlier, staler request.

`getModels()` now shares the same `inFlightRefresh` single-flight map that `refreshModels()` already used, keyed on the existing compound cache key from `getCacheKey()`. Every caller for a given key — get or refresh — converges on one in-flight fetch.

`MODEL_CACHE_EMPTY_RESPONSE` telemetry also fired on every empty response, so a persistently broken endpoint re-fired the event on every cache check. It now reports at most once per cache key (`captureModelCacheEmptyResponseOnce`) until a non-empty response re-arms it. This applies to auth-scoped providers too (zoo-gateway, kimi-code), even though they skip caching entirely.

To keep the throttle and in-flight map correctly isolated per credential for auth-scoped providers, `zoo-gateway` was added to `URL_SCOPED_PROVIDERS` and both `zoo-gateway` and `kimi-code` were added to `KEY_SCOPED_PROVIDERS`. This only changes the key used for the throttle/in-flight coordination — actual cache reads/writes for these providers stay fully skipped (`shouldSkipCache` is unchanged), so a sign-out/sign-in cycle still never serves one account's model list to another.

Non-empty cache-write behavior and the graceful-degradation behavior on a failed refresh are both unchanged.

### Test Procedure

Ran the model-cache unit suite directly:

```
pnpm --filter roo-cline exec vitest run api/providers/fetchers/__tests__/modelCache.spec.ts
```

43 tests pass, including 12 new cases covering:
- concurrent `getModels()` calls share one provider fetch
- concurrent `getModels()` and `refreshModels()` share the same in-flight fetch
- different endpoints/keys do not share a fetch
- repeated empty results emit one telemetry event
- a non-empty response re-arms reporting for a later empty response
- auth-scoped providers (zoo-gateway) never share results or throttle state across credentials, session tokens, or gateway URLs

### Pre-Submission Checklist

- [x] **Issue Linked**: Refs #830 (see above; #830 is covered by multiple independent split PRs).
- [x] **Scope**: Limited to `src/api/providers/fetchers/modelCache.ts` and its unit tests. No consent/banner/circuit-breaker/task-completion changes.
- [x] **Self-Review**: Done.
- [x] **Testing**: New unit tests added; see Test Procedure.
- [ ] **Visual Snapshot**: N/A, no UI changes.
- [x] **Documentation Impact**: No documentation updates required.

### Documentation Updates

## Summary by CodeRabbit

* **Improvements**
  * Prevented duplicate model requests when multiple loads or refreshes occur simultaneously.
  * Improved recovery after empty model responses so later requests can succeed.
  * Strengthened isolation between provider connections, endpoints, and authentication sessions.
  * Improved diagnostics for repeated empty responses while reducing redundant telemetry activity.
  * Enhanced model handling for additional provider connection types.
  * Preserved independent request behavior for authentication-scoped connections.

### DIFF
--- a/src/api/providers/fetchers/__tests__/modelCache.spec.ts
+++ b/src/api/providers/fetchers/__tests__/modelCache.spec.ts
@@ -943,9 +943,11 @@ describe("MODEL_CACHE_EMPTY_RESPONSE throttling", () => {
 	})
 
 	it("never shares results across different zoo-gateway credentials (auth isolation)", async () => {
-		// The in-flight fetch map must key on the full compound identity for auth-scoped
-		// providers too, so a slow fetch for one account's session token can never resolve
-		// into a concurrent call carrying a different account's token.
+		// Auth-scoped providers (see AUTH_SCOPED_PROVIDERS) bypass dedupedFetch entirely --
+		// shouldSkipCache is true for zoo-gateway, so every call fires its own provider fetch
+		// and none are deduplicated. That means two concurrent calls can never resolve into
+		// each other's result regardless of token, which this test confirms for two different
+		// account tokens; the companion case below confirms the same holds for one token too.
 		const accountAModels = {
 			"zoo-gateway/account-a-model": {
 				maxTokens: 4096,
@@ -990,6 +992,20 @@ describe("MODEL_CACHE_EMPTY_RESPONSE throttling", () => {
 		const [resultA, resultB] = await Promise.all([promiseA, promiseB])
 		expect(resultA).toEqual(accountAModels)
 		expect(resultB).toEqual(accountBModels)
+	})
+
+	it("never deduplicates concurrent zoo-gateway fetches, even for the same token", async () => {
+		// Auth-scoped providers skip dedupedFetch unconditionally, so even two calls carrying
+		// an identical token each fire their own provider fetch -- there is no in-flight sharing
+		// to key correctly or incorrectly for these providers.
+		freshMockGetZooGatewayModels.mockResolvedValue({})
+
+		await Promise.all([
+			freshGetModels({ provider: providerIdentifiers.zooGateway, apiKey: "same-token" }),
+			freshGetModels({ provider: providerIdentifiers.zooGateway, apiKey: "same-token" }),
+		])
+
+		expect(freshMockGetZooGatewayModels).toHaveBeenCalledTimes(2)
 	})
 })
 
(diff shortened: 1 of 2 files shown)


src/api/providers/fetchers/__tests__/modelCache.spec.ts
```diff
@@ -581,6 +773,226 @@ describe("empty cache protection", () => {
...
+			baseUrl: "https://gateway-b.example.com",
+		})
+
+		expect(FreshTelemetryService.instance.captureEvent).toHaveBeenCalledTimes(2)
+	})
+
+	it("never shares results across different zoo-gateway credentials (auth isolation)", async () => {
+		// The in-flight fetch map must key on the full compound identity for auth-scoped
+		// providers too, so a slow fetch for one account's session token can never resolve
+		// into a concurrent call carrying a different account's token.
```
JamesRobert20: Nit: this comment describes the wrong isolation mechanism.

Auth-scoped providers (`zoo-gateway`, `kimi-code`) bypass `dedupedFetch` entirely, concurrent calls are never joined, regardless of cache key. Isolation comes from that bypass, not from compound keys in the in-flight map.

The test still asserts the right outcome; only the comment is misleading. Suggest rewriting it to match the bypass, and optionally adding a same-token case that expects two fetcher calls.

JamesRobert20 (changes requested):
Overall this looks solid and nearly ready to merge. One change requested:

Please fix the misleading comment in the zoo-gateway auth-isolation test (`modelCache.spec.ts`). It currently says isolation comes from compound keys in the in-flight map, but auth-scoped providers bypass `dedupedFetch` entirely. Rewrite the comment to match that, and optionally add a same-token case that expects two fetcher calls.

Everything else LGTM.

src/api/providers/fetchers/__tests__/modelCache.spec.ts
edelauna: Good catch, fixed in 8ef0cf344. Rewrote the comment to state the actual guarantee (auth-scoped providers bypass `dedupedFetch` entirely via `shouldSkipCache`, so nothing is ever joined regardless of token) and added the same-token companion test you suggested (`never deduplicates concurrent zoo-gateway fetches, even for the same token`).

JamesRobert20 (approved):
Looks good

PR #1078, opened 2026-07-31
fix(webview): throttle state pushes to prevent gray screen OOM

### LINKED ISSUE #629: fix(webview): throttle state pushes to prevent gray screen OOM at high message counts
# Context

The chat webview turns gray (crashes) during long-running tasks. User-reported diagnostics from `logWebviewHiddenDiagnostics` confirm **3,525 messages** at crash time. This is tracked upstream as a recurring user complaint — PR #153 reduced the react-virtuoso pre-render buffer but did not address the root cause.

**Related:** Part of the broader webview message architecture addressed by #355 (Stories 3.2c + 3.2d), but scoped as an independent fix that doesn't touch the message protocol or task lifecycle.

# Developer Notes

## Problem

`addToClineMessages` (`Task.ts:1015`) calls `postStateToWebviewWithoutTaskHistory()` on **every single message addition**, serializing and posting the entire `clineMessages` array to the webview. With 3,525 messages averaging ~2KB of tool output each, every state push is **~7MB**. During active tool use with checkpoint saves, this fires dozens of times per minute. The webview must deserialize, merge into React state, and re-render each time — eventually exceeding the V8 heap limit and crashing the renderer (gray screen).

Secondary issue: `aggregatedCostsMap` in `ChatView.tsx` is never cleared on task change (line 480-491 cleanup effect), causing unbounded growth across task switches with subtasks.

## Proposed Fix

**Throttle the hot-path state pushes** — no new message types, no protocol changes.

### `ClineProvider.ts` — add throttled variant (~10 lines)

```ts
private _postStateToWebviewThrottled = debounce(
    () => this.postStateToWebviewWithoutTaskHistory(),
    500,
    { leading: true, trailing: true, maxWait: 1000 }
)

async postStateToWebviewThrottled(): Promise<void> {
    this._postStateToWebviewThrottled()
}
```

### `Task.ts` — swap 2 hot-path call sites (2 line changes)

| Line | Context | Change |
|------|---------|--------|
| 1015 | `addToClineMessages` — every new message | → `postStateToWebviewThrottled` |
| 527 | `messageQueueService.stateChanged` | → `postStateToWebviewThrottled` |
| 1828 | Task start (clear state) | Keep immediate |
| 2551 | Before API request | Keep immediate |
| 3291 | After streaming complete | Keep immediate |

Partial message streaming is unaffected — `updateClineMessage` (`Task.ts:1048`) already uses the lightweight `{ type: "messageUpdated" }` path, bypassing the full state push entirely.

### `ChatView.tsx:481` — clear aggregatedCostsMap on task change (1 line)

```ts
setAggregatedCostsMap(new Map())
```

### `Task.abortTask` — flush debounce on abort (1 line)

Ensures the final state is delivered when a task is stopped.

## Impact

At 3,525 messages:
- **Before:** ~7MB × dozens/sec during rapid tool use → OOM → gray screen
- **After:** ~7MB × max 2/sec (throttled) → webview stays alive

## Risk Assessment

- **Low risk:** No new message types, no protocol changes, no task lifecycle changes
- **UI delay:** New message bubbles may appear up to 500ms late; streaming text is unaffected (uses `messageUpdated` path)
- **Final state guaranteed:** `trailing: true` + abort flush ensures last state is always delivered
- **No #560-style edge cases:** This doesn't add states, transitions, or concurrency primitives — it just coalesces rapid calls to an existing method

## Acceptance Criteria

- [ ] Chat remains responsive at 3,500+ messages (no gray screen)
- [ ] Streaming text still appears in real-time (messageUpdated path unchanged)
- [ ] Task start shows clean state immediately (line 1828 not throttled)
- [ ] Task abort delivers final state (debounce flushed)
- [ ] `aggregatedCostsMap` is empty after task switch
- [ ] Existing test suite passes unchanged

JunyongParkDev:
### Related GitHub Issue

Closes: #629  

### Description

This PR reduces full-state webview churn during long-running tasks with large message histories. The hot paths previously serialized and posted the complete message array for every message addition and queue-state change, producing repeated multi-megabyte snapshots at the message count reported in #629.

- Adds a provider-scoped 500 ms leading/trailing debounce with a 1 second maximum wait for full state updates that omit task history.
- Routes new-message additions and message-queue state changes through the throttled path while keeping task start, API boundaries, and stream completion immediate.
- Flushes pending state before the first lightweight update for a newly created partial message and before task abort, and cancels pending work when the provider is disposed.
- Logs rejected throttled state schedules and flushes without interrupting message persistence or task-abort cleanup.
- Clears aggregated task costs when the active task changes and ignores delayed cost responses for inactive tasks.

Partial streaming continues to use the existing lightweight `messageUpdated` path. The change coalesces high-frequency full snapshots without changing the webview message protocol or task lifecycle.

### Test Procedure

Run the focused regression tests:

```sh
pnpm --dir src exec vitest run \
  core/webview/__tests__/ClineProvider.spec.ts \
  core/task/__tests__/Task.spec.ts \
  core/task/__tests__/Task.throttle.test.ts

pnpm --dir webview-ui exec vitest run \
  src/components/chat/__tests__/ChatView.spec.tsx
```

Run the repository validation commands:

```sh
pnpm test
pnpm check-types
pnpm lint
```

The regression tests cover:

- Leading/trailing coalescing and the 1 second maximum wait.
- Explicit flush behavior, provider disposal, and scheduling/flush rejection handling.
- Throttled message and message-queue updates.
- Ordering between a new partial message and subsequent `messageUpdated` delivery.
- Immediate clean-state delivery at task start.
- Final-state delivery and cleanup continuity when task abort encounters a state-push failure.
- Message persistence continuity when throttled state scheduling fails.
- Aggregated-cost cleanup on task switch and rejection of delayed responses from an inactive task.

### Pre-Submission Checklist

- [ ] **Visual Snapshot** (UI changes only): If a user would notice this change at a glance (layout, theme tokens, brand elements, empty/error states), I've added or updated a `*.visual.tsx` snapshot in `webview-ui/`. See `webview-ui/AGENTS.md` → "When a UI change needs a snapshot".

### Documentation Updates

### Additional Notes

This PR intentionally addresses the frequency of large full-state pushes. Redesigning message delivery to eliminate the O(N) payload itself is outside the scope of #629.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary interface refreshes while keeping partial responses and task interruptions synchronized.
  * Ensured pending task updates are delivered before a task is aborted.
  * Fixed task cost displays so they reset correctly when switching tasks and ignore delayed updates from previous tasks.
  * Improved reliability and error handling for task state and message updates.
* **Performance**
  * Smoothed task status updates by grouping rapid changes without delaying important updates.

### DIFF
--- a/src/core/webview/ClineProvider.ts
+++ b/src/core/webview/ClineProvider.ts
@@ -6,6 +6,7 @@ import EventEmitter from "events"
 import { Anthropic } from "@anthropic-ai/sdk"
 import delay from "delay"
 import axios from "axios"
+import debounce from "lodash.debounce"
 import pWaitFor from "p-wait-for"
 import * as vscode from "vscode"
 
@@ -189,6 +190,21 @@ export class ClineProvider
 	private taskEventListeners: WeakMap<Task, Array<() => void>> = new WeakMap()
 	private currentWorkspacePath: string | undefined
 	private _disposed = false
+	private readonly _postStateToWebviewThrottled = debounce(
+		async () => {
+			try {
+				await this.postStateToWebviewWithoutTaskHistory()
+			} catch (error) {
+				this.log(
+					`[ClineProvider#postStateToWebviewThrottled] Failed to post state: ${
+						error instanceof Error ? error.message : String(error)
+					}`,
+				)
+			}
+		},
+		500,
+		{ leading: true, trailing: true, maxWait: 1000 },
+	)
 	private readonly rateLimitClock: RateLimitClock = createRateLimitClock()
 
 	private recentTasksCache?: string[]
@@ -746,6 +762,7 @@ export class ClineProvider
 		}
 
 		this._disposed = true
+		this._postStateToWebviewThrottled.cancel()
 		this.log("Disposing ClineProvider...")
 
 		// Reject any tasks still waiting for a scheduler permit so they don't
@@ -2291,6 +2308,22 @@ export class ClineProvider
 		state.clineMessagesSeq = this.clineMessagesSeq
 		const { taskHistory: _omit, ...rest } = state
 		await this.postMessageToWebview({ type: "state", state: rest })
+	}
+
+	async postStateToWebviewThrottled(): Promise<void> {
+		if (this._disposed) {
+			return
+		}
+
+		await this._postStateToWebviewThrottled()
+	}
+
+	async flushPostStateToWebviewThrottled(): Promise<void> {
+		if (this._disposed) {
+			return
+		}
+
+		await this._postStateToWebviewThrottled.flush()
 	}
 
 	/**
--- a/src/core/webview/ClineProvider.ts
+++ b/src/core/webview/ClineProvider.ts
@@ -162,6 +162,10 @@ function scheduleTask(scheduler: TaskScheduler, task: Task, source: string): voi
 		.catch((error) => console.error(`[${source}] taskScheduler.schedule failed:`, error))
 }
 
+type GetStateOptions = {
+	includeTaskHistory?: boolean
+}
+
 export class ClineProvider
 	extends EventEmitter<TaskProviderEvents>
 	implements vscode.WebviewViewProvider, TelemetryPropertiesProvider, TaskProviderLike
@@ -2303,7 +2307,7 @@ export class ClineProvider
 	 *   `taskHistoryUpdated` / `taskHistoryItemUpdated`.
 	 */
 	async postStateToWebviewWithoutTaskHistory(): Promise<void> {
-		const state = await this.getStateToPostToWebview()
+		const state = await this.getStateToPostToWebview({ includeTaskHistory: false })
 		this.clineMessagesSeq++
 		state.clineMessagesSeq = this.clineMessagesSeq
 		const { taskHistory: _omit, ...rest } = state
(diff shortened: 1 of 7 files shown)


src/core/webview/ClineProvider.ts
```diff
@@ -188,6 +189,21 @@ export class ClineProvider
 	private taskEventListeners: WeakMap<Task, Array<() => void>> = new WeakMap()
 	private currentWorkspacePath: string | undefined
 	private _disposed = false
+	private readonly _postStateToWebviewThrottled = debounce(
+		async () => {
+			try {
+				await this.postStateToWebviewWithoutTaskHistory()
```
edelauna, line 200: Each fire through this debounce computes `taskHistoryStore.getAll()` twice (once inside `getState()` at line 2743, again at line 2514) and then discards both results at line 2194 — each `getAll()` sorts the full history (`TaskHistoryStore.ts:168`). With `maxWait: 1000` this path can fire about twice per second under sustained activity, so could we skip the taskHistory computation here (e.g. an `includeTaskHistory` option on `getStateToPostToWebview`)?

edelauna (changes requested):
Thanks for tackling this had one request around performance.

JunyongParkDev:
> Thanks for tackling this had one request around performance.

Thanks for catching this — you're right. I added an `includeTaskHistory` option, defaulting to `true`, and threaded it through `getStateToPostToWebview()` and `getState()`. The two payload paths that intentionally omit task history now pass `includeTaskHistory: false`, so they no longer call or sort `taskHistoryStore.getAll()`; those paths go from two history reads to zero.

Full-state payloads keep their existing behavior, but now read task history only once during final payload assembly instead of twice. Keeping that read at the end preserves the latest-snapshot timing while removing the redundant earlier sort.

I also added focused regression coverage for zero history reads and omission of the `taskHistory` field in both partial-state paths, plus a single late history read with the history preserved in the full-state path.

edelauna (approved):
Thanks for this!

Issue #944, opened 2026-07-18
[ENHANCEMENT] Centralize provider identifiers and replace hardcoded provider-name literals


WebMad:
### Problem (one or two sentences)

Provider identifiers are currently repeated as raw string literals across the codebase instead of being referenced through a single canonical runtime registry. Although the provider settings module defines a provider-name type and validation schema, provider values are still duplicated in provider lists, switch statements, comparisons, and handler selection logic, increasing the risk of typos, inconsistent updates, and provider-specific code drifting out of sync ([provider settings](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/packages/types/src/provider-settings.ts), [default model selection](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/packages/types/src/providers/index.ts), [API handler selection](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/api/index.ts), and [profile validation](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/shared/ProfileValidator.ts)).

### Context (who is affected and when)

This primarily affects maintainers and contributors whenever a provider is added, renamed, retired, validated, or handled in another part of the application. A contributor must currently find and update multiple independent string literals across packages and runtime modules; missing one of these locations may compile successfully while introducing inconsistent behavior in provider selection, model resolution, settings validation, caching, or webview logic. The issue becomes more significant as the number of supported providers and provider-dependent branches grows.

Examples include raw provider literals in [the provider default-model switch](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/packages/types/src/providers/index.ts), [the API handler factory](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/api/index.ts), [the model cache](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/api/providers/fetchers/modelCache.ts), [profile model resolution](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/shared/ProfileValidator.ts), and [webview message handling](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/core/webview/webviewMessageHandler.ts).

### Desired behavior (conceptual, not technical)

The project should have one authoritative source for active and retired provider identifiers. Production code should reference those canonical identifiers rather than repeating provider-name strings, while persisted settings, external API values, and other serialized data should retain their existing string representation for backward compatibility.

[...]

PR #1109, opened 2026-08-03
refactor(types): canonicalize shared provider settings identifiers

WebMad:
## Summary

- replace raw discriminator literals in shared provider settings schemas with `providerIdentifiers`
- use computed canonical identifier keys in `modelIdKeysByProvider`
- canonicalize keys and repeated `id` values in `MODELS_BY_PROVIDER`
- derive provider exclusions from `typeof providerIdentifiers.<provider>`
- add regression coverage that keeps provider-indexed settings aligned with the canonical registry

Implements item 1 from https://github.com/Zoo-Code-Org/Zoo-Code/issues/944#issuecomment-5153144387.

## Test plan

- `pnpm exec vitest run src/__tests__/provider-settings.test.ts src/__tests__/provider-identifiers.test.ts` from `packages/types`
- `pnpm exec eslint src/provider-settings.ts src/__tests__/provider-identifiers.test.ts --max-warnings=0` from `packages/types`
- `pnpm run check-types` from `packages/types`
- repository commit/push hooks, including workspace lint and type checking

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model identification across provider settings and validation flows.
  * Fixed validation for provider-specific, shared, nested, and dynamic model configurations.
  * Ensured supported providers accept valid minimal settings without errors.

* **Improvements**
  * Made provider configuration handling more consistent while preserving existing compatibility.
  * Expanded coverage for model resolution and settings validation across supported providers.
  * Improved consistency for organization allowlist and model-specific validation.
  * Preserved reliable handling for retired provider configurations.


packages/types/src/provider-settings.ts
```diff
@@ -428,40 +428,40 @@ const defaultSchema = z.object({
 })
 
 export const providerSettingsSchemaDiscriminated = z.discriminatedUnion("apiProvider", [
```
edelauna: Since every discriminator here now comes from `providerIdentifiers` rather than an inline literal, a typo (e.g. picking the wrong registry key) would compile fine and only 32 of the 34 branches are unparsed by any test in the repo — only `openaiCodex` and `kimi-code` get a round-trip through `providerSettingsSchemaDiscriminated`. Worth adding a loop over `Object.values(providerIdentifiers)` that asserts `safeParse` succeeds for each?

packages/types/src/provider-settings.ts
```diff
@@ -553,37 +553,37 @@ export const isTypicalProvider = (key: unknown): key is TypicalProvider =>
 	isProviderName(key) && !isInternalProvider(key) && !isCustomProvider(key) && !isFauxProvider(key)
 
 export const modelIdKeysByProvider: Record<TypicalProvider, ModelIdKey> = {
```
edelauna: The keys here are type-checked against `TypicalProvider`, but the values (`ModelIdKey`) aren't constrained to the right provider — e.g. nothing would catch anthropic pointing at `"openRouterModelId"`. This map now drives form validation in `webview-ui/src/utils/validate.ts:218`. Is there a test that pins each provider to its expected key?

packages/types/src/provider-settings.ts
```diff
@@ -653,106 +653,125 @@ export const getApiProtocol = (provider: ProviderName | undefined, modelId?: str
  */
 
 export const MODELS_BY_PROVIDER: Record<
-	Exclude<ProviderName, "fake-ai" | "gemini-cli" | "openai">,
+	Exclude<
```
edelauna, line 394: Could use a one-line comment on why these three are excluded (`openai` is the custom provider, `geminiCli`/`fakeAi` have no model list) — it's not obvious from the type alone and the surrounding blocks explain their groupings.

edelauna (changes requested):
Nice! Had some implementation comments to help clean up this section while we're touching it.

packages/types/src/provider-settings.ts
WebMad: Addressed in 45eb5779e: added a one-line comment explaining that OpenAI is custom-configured, while Fake AI and Gemini CLI do not expose model lists.

packages/types/src/provider-settings.ts
WebMad: Addressed by the canonical provider-definition refactor. `provider-settings.test.ts` now verifies that the registry contains exactly one definition for every canonical provider and runs `providerSettingsSchemaDiscriminated.safeParse({ apiProvider })` for every `providerNames` entry. The focused suite passes.

packages/types/src/provider-settings.ts
WebMad: Addressed by replacing the standalone `modelIdKeysByProvider` map with colocated provider definitions. `createProviderDefinition` constrains `modelIdKey` to a key of that provider schema, and provider-specific accessors are covered in `provider-model-id.test.ts` (including OpenRouter, Ollama, VS Code LM, OpenAI Compatible, OpenAI Native, and Fake AI). The focused suite passes.

edelauna (approved):
So nice to be getting this updated! Thank you ❤️

PR #1110, opened 2026-08-03
refactor(cli): canonicalize provider identifiers

WebMad:
## Summary

- build the CLI-supported provider subset from canonical `providerIdentifiers`
- migrate the provider environment-variable map and `getProviderSettings()` dispatch cases
- migrate provider-specific model-field selection in the CLI context-window utility
- replace raw OpenRouter defaults and router-model request values in `run.ts` and `list.ts`
- add focused regression coverage for the canonical subset, settings dispatch, environment mappings, and provider-specific model fields

Implements item 2 from https://github.com/Zoo-Code-Org/Zoo-Code/issues/944#issuecomment-5153144387.

## Dependency

Stacked on #1109, which introduces the shared canonical provider-settings migration used as this branch's base. Until #1109 is merged, GitHub may show that parent commit in this PR's diff.

## Test plan

- `pnpm test` from `apps/cli` (549 passed, 1 skipped)
- `pnpm check-types` from `apps/cli`
- `pnpm lint` from `apps/cli`
- repository commit/push hooks, including workspace lint and type checking
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - CLI options now consistently resolve model, provider, reasoning effort, workspace, and permission settings, including precedence between command-line flags and saved settings.
  - Model listings provide more reliable OpenRouter results and handle unavailable or malformed responses gracefully.
  - Provider configuration and model identification are consistently supported across Anthropic, OpenAI, Gemini, OpenRouter, and Vercel AI Gateway integrations.

- **Bug Fixes**
  - Improved provider and model matching to prevent inconsistent configuration behavior.


apps/cli/src/lib/utils/context-window.ts
```diff
@@ -36,21 +36,21 @@ export function getContextWindow(routerModels: RouterModels | null, apiConfigura
...
+		case providerIdentifiers.requesty:
 			return config.requestyModelId
-		case "unbound":
+		case providerIdentifiers.unbound:
 			return config.unboundModelId
-		case "litellm":
+		case providerIdentifiers.litellm:
 			return config.litellmModelId
-		case "vercel-ai-gateway":
+		case providerIdentifiers.vercelAiGateway:
```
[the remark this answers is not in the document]
WebMad: Fixed in 647623789. Added explicit cases for `opencode-go`, `kenari`, and `zoo-gateway`, using their provider-specific model ID fields, and added matching parameterized test coverage. I also replaced the `default` branch with explicit provider cases while preserving the existing `apiModelId` fallback behavior.

apps/cli/src/commands/cli/__tests__/run.test.ts
```diff
@@ -2,6 +2,152 @@ import fs from "fs"
...
+				model: "settings-model",
+				reasoningEffort: "high",
+				provider: providerIdentifiers.anthropic,
+				workspacePath,
+				nonInteractive: false,
+			}),
+		])
+		expect(runCommandMocks.runTask).toHaveBeenCalledWith("test prompt", undefined)
+		expect(exitSpy).toHaveBeenCalledWith(0)
+	})
```
edelauna, line 148: This covers the `settings win when no flag is set` path for `provider`. What about the `flag overrides settings` path? With the current fixture (`flags` has no `provider` key), transposing the arguments in `resolveProvider(flagOptions.provider, settings.provider)` at run.ts:156 would produce the same result.

apps/cli/src/commands/cli/__tests__/run.test.ts
```diff
@@ -2,6 +2,152 @@ import fs from "fs"
...
+})
+
 describe("run command --prompt-file option", () => {
```
edelauna, line 151: These tests exercise `fs.writeFileSync`/`readFileSync` round-trips but never call `run()`. Does the `--prompt-file` branch in run.ts (existence check, `process.exit(1)` on a missing file, file-content-to-prompt assignment) have coverage anywhere else?

edelauna (approved):
Thanks for updating, had 2 comments, but will approve and merge anyways, we can always tackle in a followup ticket.

Issue #1096, opened 2026-08-01
[BUG] Reasoning is never sent back to the API for the OpenAI Compatible provider — preserveReasoning cannot be enabled from the UI


1N4148:
## Problem (one or two sentences)

For the **OpenAI Compatible** provider, the assistant's reasoning is captured and displayed, but it is always stripped before the next request, so the model re-derives its reasoning on every turn. The flag that controls this, `preserveReasoning`, exists and works, but there is no way to turn it on for this provider.

## Context (who is affected and when)

Anyone running a reasoning model behind a self-hosted OpenAI-compatible endpoint (llama.cpp / `llama-server`, Lemonade, vLLM, SGLang, LM Studio via the OpenAI Compatible provider) where the server returns chain-of-thought in the `reasoning_content` delta field.

This is silent — no error, no warning. The reasoning shows up correctly in the Zoo Code UI, so nothing looks wrong from the outside. The only visible symptom is degraded multi-turn behavior, plus wasted tokens re-thinking the same thing every turn.

Built-in providers are unaffected: `preserveReasoning: true` is hardcoded for DeepSeek, Z.ai, MiniMax, MiMo, Moonshot, Fireworks and Bedrock, and LiteLLM derives it heuristically in `src/api/providers/fetchers/litellm.ts`. OpenAI Compatible is the one provider class where the user is expected to supply the model metadata themselves — and this particular field isn't offered.

## Reproduction steps

1. Run any reasoning model on `llama-server` with reasoning preserved in history (`--reasoning-preserve`, or the equivalent `--chat-template-kwargs preserve_thinking:true` for Qwen3 templates). Enable prompt logging on the server.
2. In Zoo Code, create a profile with API Provider = **OpenAI Compatible**, pointing at that endpoint. Configure the model under "Model Configuration".
3. Run any task that takes several turns (e.g. a few tool calls in a row).
4. Inspect the prompt the server actually receives.

**Expected result:** historical assistant turns in the prompt carry their reasoning trace.

**Actual result:** every historical assistant turn carries an empty `<think></think>` block — the reasoning was dropped client-side before the request was built.

## Where it happens

`src/core/task/Task.ts`, `buildCleanConversationHistory()` (~line 4627):

```ts
const shouldPreserveForApi = this.api.getModel().info.preserveReasoning === true
...
} else {
    // Strip reasoning out - stored for history only, not sent back to API
```

[...]

PR #1119, opened 2026-08-04
fix(openai): preserve reasoning for local OpenAI-compatible models via R1 toggle

### LINKED ISSUE #1118: [BUG] Reasoning from local OpenAI-compatible models (llama.cpp/Ollama/LM Studio) is discarded — no official toggle
**Describe the bug**
When connecting Zoo Code to a **local OpenAI-compatible chat model** that emits a `reasoning_content` field (llama.cpp, Ollama's OpenAI endpoint, LM Studio), the reasoning is silently discarded from the assistant messages. The model therefore cannot see its own reasoning chain, which reduces output quality and inflates token usage on follow-up turns.

**To Reproduce**
1. Run a local OpenAI-compatible server that streams `reasoning_content` alongside `content`, e.g. llama.cpp `llama-server` (any reasoning model such as Qwen3.6/QwQ/DeepSeek-R1).
2. In Zoo Code, add it as an OpenAI-compatible provider with the local base URL.
3. Run any task and observe the messages sent back to the provider on the next turn.
4. The assistant message contains only `content`, the `reasoning_content` is dropped.

**Expected behavior**
`reasoning_content` should be preserved and fed back so the local model sees its own reasoning, OR there should be an official setting to enable R1-format handling for OpenAI-compatible providers.

3.74.0 (also affects current main)

**Root cause**
`src/api/providers/openai.ts:88-90`:
```ts
const enabledR1Format = this.options.openAiR1FormatEnabled ?? false
const deepseekReasoner = modelId.includes("deepseek-reasoner") || enabledR1Format
```
`openAiR1FormatEnabled` exists only on the provider's `options` object but is **never set anywhere** in the codebase — no UI setting, no `package.json` configuration, no provider-creation code touches it. So the R1 format path (which parses and preserves `reasoning_content`) is effectively dead for every model whose id doesn't literally contain `deepseek-reasoner`.

The `deepseekReasoner` flag then gates `convertToR1Format(...)` (line 106) and the reasoning parse/save logic further down in the streaming path. Local models are left with reasoning stripped.

**Workaround (current)**
Patch the bundled `dist/extension.js` to force the R1 flag (`d=!0` for the OpenAI stream) and hard-enable `preserveReasoning`. Details and the exact search/replace strings are documented. A source-level equivalent of the same change is what this issue asks to be fixed in-tree.

ch405canova-sudo:
## Summary

Local OpenAI-compatible reasoning models (llama.cpp `llama-server`, LM Studio, Ollama's OpenAI endpoint, vLLM, etc.) stream a `reasoning_content` field, but Zoo Code strips it from the follow-up context for these providers — so the model can never see its own reasoning chain on the next turn.

### Root cause
`openAiR1FormatEnabled` (exposed as a UI checkbox via `R1FormatSetting` and declared in `provider-settings.ts`) only forces the R1 **request format** in `openai.ts`. It never propagates to `info.preserveReasoning`, which is what gates reasoning retention in `Task.ts`:

```ts
// src/core/task/Task.ts
const shouldPreserveForApi = this.api.getModel().info.preserveReasoning === true
```

`OpenAiHandler.getModel()` builds its `ModelInfo` from `openAiCustomModelInfo ?? openAiModelInfoSaneDefaults` — neither sets `preserveReasoning`. For the built-in OpenAI-compatible provider the value is therefore always `undefined`, and reasoning is stripped from messages sent back to the API.

### Fix
When the user enables the **R1 format** toggle, `getModel()` now sets `preserveReasoning: true` on the returned model info:

```ts
return {
    id,
    info: this.options.openAiR1FormatEnabled ? { ...info, preserveReasoning: true } : info,
    ...params,
}
```

This preserves `reasoning_content` in the assistant history sent to the model on follow-up turns. Default behaviour (toggle off) is unchanged.

### Tests
Added two `getModel` cases to `src/api/providers/__tests__/openai.spec.ts`:
- `preserveReasoning` is `true` when `openAiR1FormatEnabled` is on
- `preserveReasoning` stays `undefined` by default

### Related
- Issue: #1118 (bug report for the same root cause)

Fixes #1118


ch405canova-sudo:
This work is a joint effort by **chaos** (@ch405canova-sudo) and **opencode** — the bug was found and the fix developed together on a local llama.cpp stack (August 2026).

ch405canova-sudo:
This PR addresses the same root cause as #1096 (local OpenAI-compatible reasoning is discarded because `preserveReasoning` is never enabled for the built-in OpenAI-compatible provider).

Note on approach: #1096 proposes a separate `preserveReasoning` checkbox, while this PR couples it to the existing R1-format toggle. Both solve the underlying issue — happy to align with whichever approach the maintainers prefer.

src/api/providers/openai.ts
```diff
@@ -293,7 +293,15 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
 			defaultTemperature: 0,
 		})
-		return { id, info, ...params }
+		// Local OpenAI-compatible reasoning models (llama.cpp, LM Studio, Ollama)
+		// stream reasoning_content, but Roo strips it from the follow-up context
+		// unless info.preserveReasoning is set. When the user enables the R1 format
+		// toggle, treat the model as preserving reasoning so the chain is fed back.
+		return {
+			id,
+			info: this.options.openAiR1FormatEnabled ? { ...info, preserveReasoning: true } : info,
```
edelauna, line 309: `createMessage()` treats a model as R1-format via `modelId.includes("deepseek-reasoner") || enabledR1Format` (line 90), but this gate only checks the toggle. If someone points `openAiModelId` at a custom endpoint whose id contains `deepseek-reasoner` without flipping the toggle, wouldn't `createMessage` still use R1 conversion while `preserveReasoning` never gets set here — leaving the original bug open for that case?

edelauna, line 309: This forces `preserveReasoning: true` whenever the toggle is on, even if `openAiCustomModelInfo.preserveReasoning` was explicitly set to `false`. Should an explicit value win here, e.g. `info.preserveReasoning ?? true`?

src/api/providers/__tests__/openai.spec.ts
```diff
@@ -837,6 +837,20 @@ describe("OpenAiHandler", () => {
...
 			expect(model.info).toBeDefined()
 		})
+
+		it("should set preserveReasoning when openAiR1FormatEnabled is on", () => {
+			const r1Handler = new OpenAiHandler({
+				...mockOptions,
+				openAiR1FormatEnabled: true,
+			})
+			const model = r1Handler.getModel()
+			expect(model.info.preserveReasoning).toBe(true)
```
edelauna: This only checks `preserveReasoning` — would it still pass if the `...info` spread got dropped (silently losing `contextWindow`/`supportsImages`)? Might be worth asserting the whole object.
```suggestion
			expect(model.info).toEqual({ ...openAiModelInfoSaneDefaults, preserveReasoning: true })
```

edelauna (changes requested):
Thank you for your contribution, but could we align with #1096 and include this a a checkbox in the ui, I don't feel great about reusing an unrelated flag.

src/api/providers/openai.ts
```diff
@@ -300,7 +300,15 @@ export class OpenAiHandler extends BaseProvider implements SingleCompletionHandl
...
```
edelauna, line 309: Line 95 derives the R1 path from `modelId.includes("deepseek-reasoner") || enabledR1Format`, so a model id containing `deepseek-reasoner` goes through `convertToR1Format` while Task.ts still strips its reasoning (`preserveReasoning` stays unset here). Should both gates share one predicate?

```suggestion
			info: (id.includes("deepseek-reasoner") || this.options.openAiR1FormatEnabled) ? { ...info, preserveReasoning: true } : info,
```

(`id` is already in scope in `getModel`; `createMessage` could reuse the same condition.)

src/api/providers/__tests__/openai.spec.ts
```diff
@@ -968,6 +968,20 @@ describe("OpenAiHandler", () => {
...
 			expect(model.info).toBeDefined()
 		})
+
+		it("should set preserveReasoning when openAiR1FormatEnabled is on", () => {
+			const r1Handler = new OpenAiHandler({
+				...mockOptions,
+				openAiR1FormatEnabled: true,
+			})
+			const model = r1Handler.getModel()
+			expect(model.info).toEqual({ ...openAiModelInfoSaneDefaults, preserveReasoning: true })
```
edelauna, line 978: Both new tests leave `openAiCustomModelInfo` unset, so `info` is value-identical to the defaults here — the `{ ...info }` merge only gets exercised on the defaults path. A case combining custom model info (e.g. a distinct `contextWindow`) with the toggle would pin the merge for the llama.cpp / LM Studio / Ollama setups this PR targets. Worth adding?

src/api/providers/__tests__/openai.spec.ts
```diff
@@ -968,6 +968,20 @@ describe("OpenAiHandler", () => {
...
+		})
+
+		it("should not set preserveReasoning by default", () => {
+			const model = handler.getModel()
+			expect(model.info.preserveReasoning).toBeUndefined()
```
edelauna, line 983: This pins the flag, but the consumer it feeds — `buildCleanConversationHistory` in Task.ts (`preserveReasoning === true` at Task.ts:5020) — is only covered by tests that re-implement the branch (`reasoning-preservation.test.ts:223`, 290, 347, 398), so deleting the production gate would still pass everything. Would a regression test calling the real method be worth adding here?

edelauna (changes requested):
Thanks for the update - had one comment on how to handle deepseek

PR #1142, opened 2026-08-04
feat: support custom model info overrides

everyoneexe:
## Summary

- Add validated custom model metadata overrides for routed providers.
- Allow users to configure context window, maximum output tokens, image support, and prompt-cache support when a provider catalog is incomplete or a model is unlisted.
- Preserve configured model IDs for OpenRouter, Requesty, Unbound, Vercel AI Gateway, and Zoo Gateway when model discovery is unavailable.
- Add a cached-state settings UI with numeric validation, capability toggles, warnings, and reset behavior.
- Improve context progress handling so unavailable metadata does not show misleading percentages, while keeping the condense action available.

## Implementation details

- Introduces a strict `customModelInfo` schema and shared application helper in `@roo-code/types`.
- Applies overrides before derived request parameters are calculated in OpenRouter, Requesty, Unbound, and the supported RouterProvider implementations.
- Synthesizes model metadata for unknown models only when a valid context-window override is supplied; otherwise provider defaults and existing behavior are preserved.
- Keeps SettingsView inputs isolated in local cached state until the user explicitly saves.
- Covers the provider-specific behavior, model selection fallback, settings form validation, and context-header behavior with regression tests.

## Testing

- Type checks passed for the types package, extension host, and webview.
- Targeted provider tests passed: 120 tests.
- Types package tests passed: 27 tests.
- Webview test suite passed: 1,564 tests.
- Custom settings UI tests passed: 2 tests.
- ESLint passed for all changed files, with no suppression-count increase.
- Production webview build, extension bundle, and VSIX packaging completed successfully.
- Pre-push monorepo type-check passed across all 11 participating packages.

## Notes

- No changeset was added, following repository guidance that changesets are maintained separately.
- The branch also includes the design specification commit for the custom model settings override.

## Summary by CodeRabbit

* **New Features**
  * Added provider settings for overriding model context windows, maximum output tokens, image support, and prompt caching.
  * Overrides are validated, resettable, and applied across supported providers, including models missing from provider catalogs.
  * Preserved configured model selections when catalogs are incomplete or unavailable.

* **Bug Fixes**
  * Improved context usage displays for missing, unlimited, or exceeded context windows.
  * Prevented invalid metadata from affecting token calculations and controls.

* **Localization**
  * Added custom model metadata settings translations across supported languages.


packages/types/src/model.ts
```diff
@@ -183,6 +183,86 @@ export const modelInfoSchema = z.object({
...
+
+	if (typeof override.supportsPromptCache === "boolean") {
+		validOverride.supportsPromptCache = override.supportsPromptCache
+	}
+
+	if (info) {
+		return Object.keys(validOverride).length > 0 ? { ...info, ...validOverride } : info
+	}
+
+	if (!validOverride.contextWindow) {
```
edelauna, line 253: For an unlisted model (info is undefined), this returns undefined unless a contextWindow override is present — so a user who configures only maxTokens (or only image/cache capability) gets that override silently dropped via the `?? info` fallback at each call site. The settings UI doesn't require contextWindow. Should a maxTokens-only override for an unlisted model be honored (e.g. synthesized with a default window) or surfaced as a warning?

src/api/providers/router-provider.ts
```diff
@@ -58,6 +58,25 @@ export abstract class RouterProvider extends BaseProvider {
...
+	 * Only vercel-ai-gateway and zoo-gateway opt in here because the other
+	 * RouterProvider subclasses (openrouter, requesty, unbound) apply overrides
+	 * in their own `getModel()` methods — they need to merge with provider-
+	 * specific logic (e.g. specific-provider endpoints, tool preferences) that
+	 * runs before the overlay.  LiteLLM, Kenari, and OpenCode Go don't support
+	 * `customModelInfo` because they have their own discovery mechanisms and
+	 * are not exposed in the settings UI.
+	 */
+	private resolveModelInfo(info: ModelInfo | undefined, fallback: ModelInfo): ModelInfo {
+		if (this.name !== "vercel-ai-gateway" && this.name !== "zoo-gateway") {
```
edelauna, line 73: This hardcoded name check duplicates the CUSTOM_MODEL_INFO_PROVIDERS set in useSelectedModel.ts:67. Adding a 6th provider means editing both, and the two can silently drift (UI shows the override panel while the backend ignores it, or vice versa). Could both read one shared constant from @roo-code/types?

src/api/providers/__tests__/zoo-gateway.spec.ts
```diff
@@ -210,6 +210,18 @@ describe("ZooGatewayHandler", () => {
 			expect(result.info.supportsPromptCache).toBe(true)
 		})
 
+		it("applies custom metadata overrides to the discovered model", async () => {
```
edelauna, line 233: This test asserts the override landed on result.info, but not that it reaches the derived request param. The PR's claim is "applied before derived request params" — could we also capture the built request (e.g. mock createMessage) and assert max_completion_tokens reflects the override, matching the openrouter/requesty/unbound specs?

src/api/providers/__tests__/vercel-ai-gateway.spec.ts
```diff
@@ -165,6 +165,18 @@ describe("VercelAiGatewayHandler", () => {
 			expect(result.info.supportsPromptCache).toBe(true)
 		})
 
+		it("applies custom metadata overrides to the discovered model", async () => {
```
edelauna, line 173: Same as zoo-gateway: this asserts result.info.{contextWindow,maxTokens,supportsImages} but never the derived request param. ZooGateway/Vercel builders consume info.maxTokens directly for max_completion_tokens — should we capture the request and assert the override propagates end-to-end?

webview-ui/src/components/settings/__tests__/CustomModelInfoSettings.spec.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+		const maxTokensInput = screen.getByLabelText("settings:providers.customModelInfo.maxTokens.label")
+		fireEvent.input(maxTokensInput, { target: { value: "12abc" } })
+
+		expect(maxTokensInput).toHaveValue("12abc")
+		expect(maxTokensInput).toHaveAttribute("aria-invalid", "true")
+		expect(setApiConfigurationField).toHaveBeenLastCalledWith("customModelInfo", undefined)
+	})
+
+	it("updates capability overrides and warns when output exceeds the context window", () => {
```
edelauna: This test passes a static apiConfiguration that never re-renders, so each toggle rebuilds customModelInfo from the original prop (the comment at :96 even notes this). In SettingsView, setApiConfigurationField accumulates via setCachedState, so the edit-accumulation the cached-state claim depends on isn't exercised here. Could we drive the parent contract — rerender with the produced customModelInfo after each call — so a text edit + a toggle is asserted to accumulate?

edelauna (changes requested):
Nice, thanks for adding this, had a couple comments, mainly around test improvements.

Can you also include some playwright ct snapshots of the new panel (collapsed/expanded, the maxTokens-exceed-contextWindow warning, rest).

packages/types/src/model.ts
everyoneexe: Intentional — contextWindow is the anchor for synthesis. Without it there's no baseline for token tracking. The UI already warns when model info is unavailable.

src/api/providers/router-provider.ts
everyoneexe: Fixed — extracted customModelInfoProviders to @roo-code/types, added cross-reference comment.

src/api/providers/__tests__/zoo-gateway.spec.ts
everyoneexe: Fixed — added completePrompt + mockCreate assertion for max_completion_tokens.

src/api/providers/__tests__/vercel-ai-gateway.spec.ts
everyoneexe: Fixed — same pattern.

webview-ui/src/components/settings/__tests__/CustomModelInfoSettings.spec.tsx
everyoneexe: Fixed — test now uses rerender to verify toggle accumulation.

everyoneexe:
Added maxDiffPixelRatio: 0.02 to playwright-ct.config.ts — local (Arch) and CI (noble Docker) font rendering produces ~5px anti-aliasing differences. This threshold tolerates that without masking real regressions.

src/api/providers/router-provider.ts
```diff
@@ -52,6 +52,28 @@ export abstract class RouterProvider extends BaseProvider {
...
+	 * are not exposed in the settings UI.
+	 *
+	 * See also: `customModelInfoProviders` in @roo-code/types for the full set
+	 * of providers whose UI exposes the override panel.
+	 */
+	private resolveModelInfo(info: ModelInfo | undefined, fallback: ModelInfo): ModelInfo {
+		if (this.name !== "vercel-ai-gateway" && this.name !== "zoo-gateway") {
+			return info ?? fallback
+		}
+
```
edelauna, line 76: This applies the override for the zoo/vercel gateways, whose request builders then send `info.maxTokens` as `max_completion_tokens` verbatim -- no clamp -- while openrouter/requesty/unbound route the same value through `getModelMaxOutputTokens` (clamps to ~20% of contextWindow). The UI only warns on `maxTokens > contextWindow` but still persists it, so on a gateway that override reaches the request unclamped and the provider rejects it. Worth clamping here too, or hard-blocking `maxTokens > contextWindow` in the UI.

packages/types/src/__tests__/custom-model-info.test.ts
```diff
@@ -0,0 +1,75 @@
...
+		).toEqual({
+			...model,
+			contextWindow: 128_000,
+			maxTokens: 16_384,
+			supportsImages: true,
+			supportsPromptCache: true,
+		})
+	})
+
+	it("does not synthesize model info without a valid context window", () => {
```
edelauna: This info=undefined negative case uses `contextWindow: 0` + `maxTokens: -1` (both invalid), so the `if (!validOverride.contextWindow) return undefined` guard is never exercised with a *valid* maxTokens-only override -- a mutation loosening it to also accept maxTokens would survive. Same gap on the info-present side: no test feeds an *invalid* value alongside a defined `info`, so dropping an `isPositiveInteger` guard wouldn't fail either. Two small cases close both.

edelauna (changes requested):

src/api/providers/router-provider.ts
everyoneexe: Addressed in ac8416f9b. Gateway model resolution now clamps maxTokens to the effective context window, with Vercel AI Gateway and Zoo Gateway request-body regression tests.

packages/types/src/__tests__/custom-model-info.test.ts
everyoneexe: Addressed in ac8416f9b. Added separate regression coverage for a max-tokens-only override without a context window, and invalid numeric overrides alongside discovered model metadata.

ThreeScreenStudios:
Was wondering if there was any progress with merging this in? It would be nice to set custom context and etc limits, especially when a new model is released (e.g. Opus 5.5) because the defaults for custom models are very far off and not configurable.

PR #1141, opened 2026-08-04
refactor(webview): complete provider identifier migration

WebMad:
## Summary

Completes item 3 from the provider identifier migration plan in #944:

- canonicalizes provider keys and values in `MODELS_BY_PROVIDER` and `PROVIDERS`
- replaces remaining provider identity comparisons in `ApiOptions`
- canonicalizes provider model/service maps and custom-model UI collections
- migrates fallback provider arguments and `routerModels` keys in `useSelectedModel`
- migrates provider-specific settings filters, refresh requests, and lookups
- canonicalizes Zoo Gateway router-model synchronization
- adds focused assertions for canonical provider configuration keys

Serialized provider values and runtime behavior remain unchanged.

## Validation

- `npx vitest run` for 8 focused webview test files: 128 tests passed
- `pnpm check-types` in `webview-ui`
- ESLint with `--prune-suppressions --max-warnings=0` for all changed files
- repository pre-push type checks
- `git diff --check`

Related to #944.

## Summary by CodeRabbit

* **Refactor**
  * Standardized provider handling across settings, model selection, model discovery, and integrations.
  * Improved consistency for provider configuration, validation, refresh actions, and fallback behavior.
  * Preserved existing provider labels, model mappings, and user-facing behavior.

* **Bug Fixes**
  * Improved model refresh handling so updated provider models appear more reliably.
  * Improved handling for provider-specific settings, custom models, and model selection.

* **Tests**
  * Expanded coverage for provider configuration, model selection, refresh behavior, and provider-specific interactions.

PR #1143, opened 2026-08-04
refactor(webview): canonicalize provider and model message identifiers

WebMad:
## Summary

Canonicalizes provider settings identifiers and centralizes model-related webview message identifiers across the shared types package, extension host, and webview.

- replaces provider string literals with `providerIdentifiers` in provider settings, model lookups, refresh requests, OAuth callbacks, and response filters
- refreshes provider-scoped and shared router-model caches without invalidating unrelated providers
- introduces shared typed constants and schemas for router, OpenAI, Ollama, LM Studio, and VS Code LM model messages
- uses the shared message identifiers consistently in webview hooks/components, extension message contracts, and the extension-side message handler
- centralizes the shared all-provider router-model cache sentinel
- adds focused coverage for provider identifiers, model-message contracts, refresh flows, cache invalidation, OAuth routes, and error handling

Serialized provider values and webview message values remain unchanged.

## Validation

- provider-focused webview suites: 35 tests passed
- backend router-model and NanoGPT suites: 30 tests passed
- `@roo-code/types`: 364 tests passed
- `pnpm check-types` in `webview-ui`
- ESLint with `--prune-suppressions --max-warnings=0` for changed runtime and UI files
- `git diff --check`

Related to #944. Extracted from #1141.

### DIFF
--- a/webview-ui/src/components/settings/providers/LiteLLM.tsx
+++ b/webview-ui/src/components/settings/providers/LiteLLM.tsx
@@ -8,6 +8,8 @@ import {
 	type ExtensionMessage,
 	litellmDefaultModelId,
 	providerIdentifiers,
+	allRouterModelsProvider,
+	RouterModelsMessageType,
 } from "@roo-code/types"
 
 import { RouterName } from "@roo/api"
@@ -28,6 +30,13 @@ type LiteLLMProps = {
 	simplifySettings?: boolean
 }
 
+enum RefreshStatus {
+	Idle = "idle",
+	Loading = "loading",
+	Success = "success",
+	Error = "error",
+}
+
 export const LiteLLM = ({
 	apiConfiguration,
 	setApiConfigurationField,
@@ -38,31 +47,35 @@ export const LiteLLM = ({
 	const { t } = useAppTranslation()
 	const queryClient = useQueryClient()
 	const { routerModels } = useExtensionState()
-	const [refreshStatus, setRefreshStatus] = useState<"idle" | "loading" | "success" | "error">("idle")
+	const [refreshStatus, setRefreshStatus] = useState(RefreshStatus.Idle)
 	const [refreshError, setRefreshError] = useState<string | undefined>()
 	const litellmErrorJustReceived = useRef(false)
 
 	useEffect(() => {
 		const handleMessage = (event: MessageEvent<ExtensionMessage>) => {
 			const message = event.data
-			if (message.type === "singleRouterModelFetchResponse" && !message.success) {
+			if (message.type === RouterModelsMessageType.singleRouterModelFetchResponse && !message.success) {
 				const providerName = message.values?.provider as RouterName
 				if (providerName === providerIdentifiers.litellm) {
 					litellmErrorJustReceived.current = true
-					setRefreshStatus("error")
+					setRefreshStatus(RefreshStatus.Error)
 					setRefreshError(message.error)
 				}
-			} else if (message.type === "routerModels") {
+			} else if (message.type === RouterModelsMessageType.routerModels) {
 				// If we were loading and no specific error for litellm was just received, mark as success.
 				// The ModelPicker will show available models or "no models found".
-				if (refreshStatus === "loading") {
+				if (refreshStatus === RefreshStatus.Loading) {
 					if (!litellmErrorJustReceived.current) {
-						setRefreshStatus("success")
+						setRefreshStatus(RefreshStatus.Success)
 						// Refresh the provider-scoped cache used by useSelectedModel and the shared cache used by
 						// ApiOptions. Target both exact keys rather than the bare ["routerModels"] prefix, which
 						// would needlessly invalidate every other provider's query too.
-						void queryClient.invalidateQueries({ queryKey: ["routerModels", providerIdentifiers.litellm] })
-						void queryClient.invalidateQueries({ queryKey: ["routerModels", "all"] })
+						void queryClient.invalidateQueries({
+							queryKey: [RouterModelsMessageType.routerModels, providerIdentifiers.litellm],
+						})
+						void queryClient.invalidateQueries({
+							queryKey: [RouterModelsMessageType.routerModels, allRouterModelsProvider],
+						})
 					}
 					// If litellmErrorJustReceived.current is true, status is already (or will be) "error".
 				}
@@ -88,19 +101,22 @@ export const LiteLLM = ({
 
 	const handleRefreshModels = useCallback(() => {
 		litellmErrorJustReceived.current = false // Reset flag on new refresh action
-		setRefreshStatus("loading")
+		setRefreshStatus(RefreshStatus.Loading)
 		setRefreshError(undefined)
 
 		const key = apiConfiguration.litellmApiKey
 		const url = apiConfiguration.litellmBaseUrl
 
 		if (!key || !url) {
-			setRefreshStatus("error")
+			setRefreshStatus(RefreshStatus.Error)
 			setRefreshError(t("settings:providers.refreshModels.missingConfig"))
 			return
 		}
 
-		vscode.postMessage({ type: "requestRouterModels", values: { litellmApiKey: key, litellmBaseUrl: url } })
+		vscode.postMessage({
+			type: RouterModelsMessageType.requestRouterModels,
+			values: { litellmApiKey: key, litellmBaseUrl: url },
+		})
 	}, [apiConfiguration, setRefreshStatus, setRefreshError, t])
 
 	return (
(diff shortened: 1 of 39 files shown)


webview-ui/src/components/settings/providers/__tests__/KimiCode.spec.tsx
```diff
@@ -1,16 +1,68 @@
...
+	})
+
+	it("uses discovered Kimi Code models when the extension returns models", () => {
+		mockUseRouterModels.mockReturnValue({
+			data: { [providerIdentifiers.kimiCode]: { "discovered-model": {} } },
+			refetch: vi.fn(),
+			isFetching: false,
+		})
+
+		render(<KimiCode apiConfiguration={{ apiProvider: "kimi-code" }} setApiConfigurationField={vi.fn()} />)
```
WebMad: Should we use `providerIdentifiers.kimiCode` here?

webview-ui/src/components/settings/providers/__tests__/KimiCode.spec.tsx
```diff
@@ -1,16 +1,68 @@
...
+
+		expect(screen.getByTestId("kimi-code-model-picker")).toHaveAttribute(
+			"data-model-ids",
+			JSON.stringify(["discovered-model"]),
+		)
+	})
+
+	it("falls back to static Kimi Code models when no models are discovered", () => {
+		render(<KimiCode apiConfiguration={{ apiProvider: "kimi-code" }} setApiConfigurationField={vi.fn()} />)
```
WebMad: Should we use `providerIdentifiers.kimiCode` here?

WebMad: Please check for other occurrences in this PR and replace them where necessary.

webview-ui/src/components/settings/providers/__tests__/KimiCode.spec.tsx
WebMad: Updated this occurrence to `providerIdentifiers.kimiCode` in c6cac6d7e.

webview-ui/src/components/settings/providers/__tests__/KimiCode.spec.tsx
WebMad: Updated this occurrence to `providerIdentifiers.kimiCode`. I also checked the other provider literals introduced by this PR and replaced the remaining Kimi Code and Moonshot occurrences in c6cac6d7e.

WebMad:
A note on the two LiteLLM invalidations: there are currently two independent React Query entries for router models. `useSelectedModel()` uses the provider-scoped `["routerModels", providerIdentifiers.litellm]` key, while `ApiOptions` calls `useRouterModels()` without a provider and uses `["routerModels", "all"]`. The refresh message does not update either cache directly, so both exact keys must be invalidated to keep model IDs/metadata and the settings model list in sync. Using the broader `["routerModels"]` prefix would also refetch unrelated provider-scoped queries. This is the minimal fix without consolidating the router-model caches.

webview-ui/src/components/settings/providers/LiteLLM.tsx
```diff
@@ -57,12 +58,11 @@ export const LiteLLM = ({
...
-						// picks up the refreshed list. useSelectedModel reads LiteLLM under the
-						// compound key ["routerModels", "litellm"] (see useRouterModels), so we
-						// target that exact key rather than the bare ["routerModels"] prefix,
-						// which would needlessly invalidate every other provider's query too.
-						queryClient.invalidateQueries({ queryKey: ["routerModels", "litellm"] })
+						// Refresh the provider-scoped cache used by useSelectedModel and the shared cache used by
+						// ApiOptions. Target both exact keys rather than the bare ["routerModels"] prefix, which
+						// would needlessly invalidate every other provider's query too.
+						void queryClient.invalidateQueries({ queryKey: ["routerModels", providerIdentifiers.litellm] })
+						void queryClient.invalidateQueries({ queryKey: ["routerModels", "all"] })
```
edelauna: The `"all"` sentinel comes from `useRouterModels.ts:59` (`provider || "all"`) but is not exported from there. If that fallback string ever changes, this invalidation silently misses with no type error. Worth exporting a small `routerModelsQueryKey(provider?: string)` factory from the hook so both sides share the same source of truth?

webview-ui/src/components/settings/providers/__tests__/LiteLLM.spec.tsx
```diff
@@ -0,0 +1,145 @@
...
+						success: false,
+						values: { provider: providerIdentifiers.openrouter },
+						error: "OpenRouter unavailable",
+					},
+				}),
+			)
+		})
+
+		expect(screen.queryByText("OpenRouter unavailable")).not.toBeInTheDocument()
+		expect(screen.getByText("settings:providers.refreshModels.loading")).toBeInTheDocument()
```
edelauna, line 185: `Moonshot.spec.tsx` and `OpenCodeGo.spec.tsx` both have a test where an error response and a `routerModels` message are dispatched in the same `act()` block — to cover the `litellmErrorJustReceived` guard at `LiteLLM.tsx:58`. LiteLLM doesn't have the equivalent case. Worth adding a fourth test that asserts the error state is shown and `invalidateQueries` is not called when both arrive in the same tick?

webview-ui/src/components/settings/providers/Moonshot.tsx
```diff
@@ -37,7 +40,7 @@ export const Moonshot = ({ apiConfiguration, setApiConfigurationField, simplifyS
 			const message = event.data
 			if (message.type === "singleRouterModelFetchResponse" && !message.success) {
 				const providerName = message.values?.provider as RouterName
-				if (providerName === "moonshot" && refreshStatus === "loading") {
+				if (providerName === providerIdentifiers.moonshot && refreshStatus === "loading") {
```
edelauna: A few lines below (line 52), the success branch calls `queryClient.invalidateQueries({ queryKey: ["routerModels"] })` — a prefix match that busts every active provider's cache. The LiteLLM refresh added in this PR targets two specific keys and its comment explains why: "which would needlessly invalidate every other provider's query too." Should Moonshot follow the same targeted pattern?

webview-ui/src/components/settings/providers/Poe.tsx
```diff
@@ -49,7 +49,7 @@ export const Poe = ({
 			const message = event.data
 			if (message.type === "singleRouterModelFetchResponse" && !message.success) {
 				const providerName = message.values?.provider as RouterName
-				if (providerName === "poe") {
+				if (providerName === providerIdentifiers.poe) {
```
edelauna, line 61: A few lines below (line 63), the success branch calls `queryClient.invalidateQueries({ queryKey: ["routerModels"] })` — a prefix match that busts every active provider's cache. The LiteLLM refresh added in this PR uses two targeted keys to avoid that. Should Poe follow the same pattern?

edelauna (changes requested):
Thanks for the change! Had a comment around preventing busting the cache.

webview-ui/src/components/settings/providers/LiteLLM.tsx
```diff
@@ -57,12 +58,8 @@ export const LiteLLM = ({
...
 					if (!litellmErrorJustReceived.current) {
 						setRefreshStatus("success")
-						// Invalidate only the LiteLLM router-models query so useSelectedModel
-						// picks up the refreshed list. useSelectedModel reads LiteLLM under the
-						// compound key ["routerModels", "litellm"] (see useRouterModels), so we
-						// target that exact key rather than the bare ["routerModels"] prefix,
-						// which would needlessly invalidate every other provider's query too.
-						queryClient.invalidateQueries({ queryKey: ["routerModels", "litellm"] })
+						// Refresh the shared cache used by ApiOptions without invalidating unrelated queries.
+						void queryClient.invalidateQueries({ queryKey: ["routerModels", "all"] })
```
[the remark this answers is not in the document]
WebMad: Addressed in a6b23c3ac: LiteLLM now invalidates both the provider-scoped and shared router-model query keys with explicit void handling, with focused test coverage.

webview-ui/src/components/settings/providers/LiteLLM.tsx
WebMad: Addressed in a6b23c3ac. The shared sentinel is now exported as allRouterModelsProvider from @roo-code/types and used consistently by useRouterModels and all targeted invalidations/tests, so the value has a single source of truth.

webview-ui/src/components/settings/providers/__tests__/LiteLLM.spec.tsx
WebMad: Added the same-tick regression case in a6b23c3ac. It dispatches the LiteLLM failure and routerModels response in one act block, asserts the error remains visible, and verifies invalidateQueries is not called.

webview-ui/src/components/settings/providers/Moonshot.tsx
WebMad: Addressed in a6b23c3ac. Moonshot now invalidates only its provider-scoped key and the shared all-provider key; the test also verifies the broad routerModels prefix is not invalidated.

webview-ui/src/components/settings/providers/Poe.tsx
WebMad: Addressed in a6b23c3ac. Poe now invalidates only its provider-scoped key and the shared all-provider key; focused coverage verifies that the broad routerModels prefix is not invalidated.

edelauna (approved):

PR #1157, opened 2026-08-05
feat(cli): add direct headless extension API

taltas:
## Stack

Position **2 of 6** in the Zoo CLI stack.

- Parent/base: `fm/zoo-cli-contracts` (PR #1150)
- Child: `fm/zoo-cli-host-security` (planned)
- Native stack tracking remains `(main) <- contracts <- headless-api`; GitHub uses the approved classic fallback because native submit reported stacked PRs are not enabled.

## Scope

- Adds direct headless initialize/start/resume methods without focus, webview readiness, or webview message interpretation.
- Adds task-and-ask-ID-safe approve/reject/message responses.
- Adds canonical non-rehydrating cancellation that persists interrupted status before settlement.
- Adds typed ask, terminal-failure, and root-result events.
- Ignores child completion and verifies persisted root `completed` status before one authoritative result.
- Adds bounded failure semantics for unexpected abort, persistence failure, and shutdown.
- Preserves all existing VS Code API methods and inherited `apps/cli` behavior.

## Acceptance Evidence

- `pnpm --dir src test extension/__tests__/api-headless.spec.ts` (5 tests)
- `pnpm --dir packages/types check-types`
- Targeted suppression-pruning ESLint; suppression count reduced by five.
- Repository pre-commit lint and pre-push typecheck passed.

## Risk

The facade is additive, but it touches canonical task cancellation and ask identity. Existing cancellation keeps rehydration by default; only the explicit headless path disables replacement-task creation. Run overrides remain on the next stack layer and are not represented as persisted headless configuration semantics here.

```mermaid
sequenceDiagram
    participant H as zoo-host
    participant A as Headless API
    participant T as Canonical task tree
    participant S as Task history
    H->>A: startHeadlessTask
    A->>T: createTask directly
    T-->>A: typed asks/events
    H->>A: respond(taskId, askId)
    T-->>A: root TaskCompleted
    A->>S: verify root status=completed
    S-->>A: persisted
    A-->>H: exactly one root result
```

### DIFF
--- a/src/core/webview/ClineProvider.ts
+++ b/src/core/webview/ClineProvider.ts
@@ -247,7 +247,8 @@ export class ClineProvider
 				this.scheduleGlobalStateWriteThrough()
 			},
 		})
-		this.initializeTaskHistoryStore().catch((error) => {
+		this.taskHistoryInitialization = this.initializeTaskHistoryStore()
+		this.taskHistoryInitialization.catch((error) => {
 			this.log(`Failed to initialize TaskHistoryStore: ${error}`)
 		})
 
@@ -3054,6 +3055,16 @@ export class ClineProvider
 		return this.taskRegistry.current
 	}
 
+	private readonly taskHistoryInitialization: Promise<void>
+
+	public waitUntilReady(): Promise<void> {
+		return this.taskHistoryInitialization
+	}
+
+	public getTaskById(taskId: string): Task | undefined {
+		return this.taskRegistry.getById(taskId)
+	}
+
 	private logWebviewHiddenDiagnostics(): void {
 		const task = this.getCurrentTask()
 		if (!task || task.abort || task.abandoned) {
@@ -3221,7 +3232,7 @@ export class ClineProvider
 		return task
 	}
 
-	public async cancelTask(): Promise<void> {
+	public async cancelTask(options: { rehydrate?: boolean } = {}): Promise<void> {
 		const task = this.getCurrentTask()
 
 		if (!task) {
@@ -3229,10 +3240,10 @@ export class ClineProvider
 		}
 
 		console.log(`[cancelTask] cancelling task ${task.taskId}.${task.instanceId}`)
-		await this.cancelTaskInternal(task)
+		await this.cancelTaskInternal(task, options)
 	}
 
-	private async cancelTaskInternal(task: Task): Promise<void> {
+	private async cancelTaskInternal(task: Task, options: { rehydrate?: boolean }): Promise<void> {
 		let historyItem: HistoryItem | undefined
 		try {
 			const history = await this.getTaskWithId(task.taskId)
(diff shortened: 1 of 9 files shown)


src/core/webview/ClineProvider.ts
```diff
@@ -3315,6 +3326,11 @@ export class ClineProvider
 			return
 		}
 
+		if (!task.parentTaskId && historyItem.status !== "interrupted") {
```
edelauna, line 3329: The guard excludes only `interrupted`, not `completed`. `updateTaskHistory` runs `assertValidTransition` before writing and `VALID_TRANSITIONS.completed` is `[]`, so a completed root throws here -- unwrapped on the root branch, after `abortTask()` but before `createTaskWithHistoryItem`. A checkpoint-restore or cancel of a completed task (via `webviewMessageHandler` completionCheckpointRestore) then aborts it without rehydrating. Gate the write on the task actually being active.

src/extension/api.ts
```diff
@@ -167,6 +185,202 @@ export class API extends EventEmitter<RooCodeEvents> implements RooCodeAPI {
...
+	}
+
+	public async cancelHeadlessTask({
+		rootTaskId,
+	}: {
+		rootTaskId: string
+		reason: "user" | "signal" | "timeout"
+	}): Promise<HeadlessCancelSettlement> {
+		const run = this.headlessRuns.get(rootTaskId)
+		if (!run || run.result) throw new Error(`Root task ${rootTaskId} is not active`)
```
edelauna, line 261: `cancelCurrentTask` (L478) gained a completed-status guard but this path has none. Between `TaskCompleted` and `settleCompletedHeadlessRun` resolving (its 5s `pWaitFor`), `run.result` is still undefined so this guard passes and `getCurrentTask()` still returns the completed root; `cancelTask({rehydrate:false})` then throws on the completed->interrupted write and settles a spurious `cancel_failed`. Mirror the completed-guard here.

src/extension/api.ts
```diff
@@ -260,6 +474,8 @@ export class API extends EventEmitter<RooCodeEvents> implements RooCodeAPI {
 	}
 
 	public async cancelCurrentTask() {
+		const currentTask = this.sidebarProvider.getCurrentTask()
+		if (currentTask && this.sidebarProvider.taskHistoryStore.get(currentTask.taskId)?.status === "completed") return
```
edelauna, line 478: `saveClineMessages` preserves the sticky `completed` status, so a task resumed from a completed history item keeps `completed` while it runs again -- this early-return then makes `cancelCurrentTask` (a real IPC/CLI entry) a silent no-op and the live stream can't be stopped. Key the guard on whether the task is actually running, not the persisted status. (Also: this guard has no test -- add cancel-of-resumed-completed coverage.)

edelauna (changes requested):

PR #1158, opened 2026-08-05
feat(cli): add secure supervised Zoo host

taltas:
## Stack

Position **3 of 6** in the Zoo CLI stack.

- Parent/base: `fm/zoo-cli-headless-api` (PR #1157)
- Child: `fm/zoo-cli-automation` (planned)
- Native local ancestry is preserved; GitHub uses the documented classic fallback because native stacked PRs are disabled for this repository.

## Scope

- Adds private `@roo-code/zoo-host` with production extension bootstrap through the VS Code shim.
- Adds dedicated Node IPC transport, gapless sequencing, heartbeat, serialized dispatch, ACK plus DONE/ERROR, and immutable workspace identity.
- Adds explicit extension/workspace/storage/app roots.
- Adds injectable macOS Keychain/Linux Secret Service storage and no plaintext shim fallback.
- Adds true task-local provider/profile/model/mode/reasoning/approval overrides resolved read-only before eviction.
- Preserves canonical approval policy, including denials and destructive-command boundaries.
- Adds non-rehydrating headless cancellation while retaining existing UI cancellation behavior.

## Acceptance Evidence

- `pnpm --dir packages/zoo-host test` (4 tests), typecheck, and build
- `pnpm --dir packages/vscode-shim test src/__tests__/ExtensionContext.test.ts` (22 tests)
- `pnpm --dir src test core/webview/__tests__/ClineProvider.run-overrides.spec.ts extension/__tests__/api-headless.spec.ts` (8 tests)
- Relevant package and extension typechecks/lint passed.
- Repository pre-commit lint and pre-push typecheck passed.

## Risk

Provider-specific model fields are resolved through the canonical model-key map. OS-vault command availability remains a packaging/runtime prerequisite; unsupported platforms fail rather than downgrading to plaintext. The child currently provides an empty read-only history result pending the automation/session layer.

```mermaid
flowchart LR
    P[zoo parent] <-->|Node IPC only| H[zoo-host]
    H --> S[VS Code shim]
    S --> E[Production extension.js]
    E --> A[Direct headless API]
    V[OS credential vault] --> S
    R[Explicit workspace/storage roots] --> H
    H -->|ACK then DONE/ERROR| P
```

### DIFF
--- a/packages/zoo-host/src/dispatcher.ts
+++ b/packages/zoo-host/src/dispatcher.ts
@@ -1,0 +1,84 @@
+import type { RooCodeAPI } from "@roo-code/types"
+import { hostCommandSchema, type HostCommand } from "@roo-code/zoo-protocol"
+
+import { HostTransport } from "./transport.js"
+
+export class HostCommandDispatcher {
+	private queue = Promise.resolve()
+	private activeRootTaskId: string | undefined
+
+	constructor(
+		private readonly api: RooCodeAPI,
+		private readonly transport: HostTransport,
+		private readonly workspace: string,
+	) {}
+
+	public dispatch(input: unknown): Promise<void> {
+		const command = hostCommandSchema.parse(input)
+		const operation = this.queue.then(() => this.execute(command))
+		this.queue = operation.catch(() => undefined)
+		return operation
+	}
+
+	private async execute(command: HostCommand): Promise<void> {
+		await this.transport.send({ type: "command.ack", commandId: command.id })
+		try {
+			const data = await this.executeCommand(command)
+			await this.transport.send({ type: "command.done", commandId: command.id, data })
+		} catch (error) {
+			await this.transport.send({
+				type: "command.error",
+				commandId: command.id,
+				error: {
+					code: "task_failed",
+					kind: "runtime",
+					phase: command.type,
+					message: error instanceof Error ? error.message : String(error),
+				},
+			})
+		}
+	}
+
+	private async executeCommand(command: HostCommand) {
+		switch (command.type) {
+			case "task.start": {
+				if (command.workspace !== this.workspace) throw new Error("Host workspace identity cannot change")
+				const task = await this.api.startHeadlessTask({ text: command.prompt, overrides: command.overrides })
+				this.activeRootTaskId = task.rootTaskId
+				return { commandType: command.type, task }
+			}
+			case "task.resume": {
+				const task = await this.api.resumeHeadlessTask(command.taskId, command.overrides)
+				this.activeRootTaskId = task.rootTaskId
+				return { commandType: command.type, task }
+			}
+			case "task.input":
+				await this.api.sendMessage(command.text, command.images)
+				return { commandType: command.type, taskId: command.taskId }
+			case "ask.respond":
+				await this.api.respondToHeadlessAsk({
+					taskId: command.taskId,
+					askId: command.askId,
+					response:
+						command.response === "message"
+							? { response: "message", text: command.text! }
+							: { response: command.response },
+				})
+				return { commandType: command.type, taskId: command.taskId, askId: command.askId }
+			case "task.cancel":
+				await this.api.cancelHeadlessTask({ rootTaskId: command.rootTaskId, reason: command.reason })
+				return { commandType: command.type, rootTaskId: command.rootTaskId }
+			case "host.snapshot":
+				return {
+					commandType: command.type,
+					lastSeq: this.transport.lastSequence,
+					activeRootTaskId: this.activeRootTaskId,
+				}
+			case "host.shutdown":
+				await this.api.shutdownHeadless()
+				return { commandType: command.type }
+			case "history.list":
+				return { commandType: command.type, workspace: command.workspace, tasks: [] }
+		}
+	}
+}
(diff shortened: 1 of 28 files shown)


src/core/webview/ClineProvider.ts
```diff
@@ -1183,12 +1186,16 @@ export class ClineProvider
 			diffFuzzyThreshold,
 		} = await this.getState()
 
+		const resolvedRun = await this.resolveRunOverrides(options?.runOverrides, persistedApiConfiguration, {
```
edelauna, line 1189: `createTask` validates the resolved override against `organizationAllowList` (isProfileAllowed) before starting; this resume path never does -- `organizationAllowList` isn't even read here. `resumeHeadlessTask(id, {provider|profile|model})` runs against a provider the org policy forbids. Add the same `ProfileValidator.isProfileAllowed` gate before constructing the Task.

src/core/webview/ClineProvider.ts
```diff
@@ -3125,6 +3136,51 @@ export class ClineProvider
...
+			if (!Object.values(providerIdentifiers).includes(overrides.provider as ProviderName)) {
+				throw new Error(`Unknown provider override: ${overrides.provider}`)
+			}
+			apiConfiguration = { ...apiConfiguration, apiProvider: overrides.provider as ProviderName }
+			profile = undefined
+		}
+		if (overrides.model) {
+			const provider = apiConfiguration.apiProvider
+			if (isTypicalProvider(provider)) {
+				const key = modelIdKeysByProvider[provider]
```
edelauna, line 3164: For custom `openai` (and internal `vscode-lm`), `isTypicalProvider` is false so the override lands on `apiModelId` -- but the OpenAI handler reads `openAiModelId` (openai.ts:87,287) and ignores it, so `--model` is silently dropped. `getModelId` still reports the override, so UI/telemetry and the real request disagree. Reuse the canonical mapping in webview-ui/src/utils/validate.ts:209-219.

src/core/task/Task.ts
```diff
@@ -740,6 +757,43 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
 		provider.on(RooCodeEventName.ProviderProfileChanged, this.providerProfileChangeListener)
 	}
 
+	private async getEffectiveState() {
+		const state = await this.providerRef.deref()?.getState()
+		if (!state || !this.isolateRunConfiguration) return state
```
edelauna, line 762: Subtasks created via `delegateParentAndOpenChild` (ClineProvider.ts:3709) call `createTask` without `runOverrides`, so a child gets `isolateRunConfiguration=false` and this early-returns the live global state. A run started with `approval:"safe"`/`"interactive"` that spawns a subtask auto-executes per the machine's global `alwaysAllowExecute`, bypassing the requested gate. Thread the parent's overrides into subtasks.

packages/zoo-host/src/transport.ts
```diff
@@ -0,0 +1,38 @@
...
+		public readonly hostId: string,
+		private readonly sendIPC: SendIPC,
+	) {}
+
+	public get lastSequence(): number {
+		return this.sequence
+	}
+
+	public async send(event: OutboundHostEvent<HostEvent>): Promise<void> {
+		const message = hostEventSchema.parse({ v: 1, seq: ++this.sequence, hostId: this.hostId, ...event })
```
edelauna, line 22: `++this.sequence` commits before `await sendIPC`. If the send rejects (the heartbeat path calls this), the seq is consumed but nothing is delivered, leaving a permanent hole; the receiver enforces strict +1 (host-events.ts:267-268) and rejects the whole stream. Reserve-then-commit: advance `sequence` only after `sendIPC` resolves, and serialize sends.

packages/zoo-host/src/dispatcher.ts
```diff
@@ -0,0 +1,84 @@
...
+	private activeRootTaskId: string | undefined
+
+	constructor(
+		private readonly api: RooCodeAPI,
+		private readonly transport: HostTransport,
+		private readonly workspace: string,
+	) {}
+
+	public dispatch(input: unknown): Promise<void> {
+		const command = hostCommandSchema.parse(input)
```
edelauna, line 17: `dispatch` isn't `async`, so this `.parse` throws synchronously and the `void dispatcher.dispatch(msg).catch(...)` in child.ts:76 can't catch it -- a malformed IPC message escapes as an uncaught exception and crashes the host, defeating the `command.error("invalid")` handler. Make `dispatch` async, or move the parse inside the queued `execute`.

packages/zoo-host/src/security.ts
```diff
@@ -0,0 +1,112 @@
...
+				}
+			},
+		}
+	}
+	if (platform === "linux") {
+		return {
+			async get(account) {
+				try {
+					const { stdout } = await execFile("secret-tool", ["lookup", "service", service, "account", account])
+					return stdout.trimEnd() || undefined
```
edelauna, line 65: The blanket `catch { return undefined }` on the Linux `get` makes a missing `secret-tool`, a locked keyring, and "not found" indistinguishable, with no log. macOS only swallows exit 44 and rethrows real errors -- the platforms disagree, and both diverge from FileSecretStorage's never-throw contract. Distinguish not-found from failure and log real errors.

edelauna (changes requested):

PR #1159, opened 2026-08-05
feat(cli): add Zoo automation client

taltas:
## Stack

Position **4 of 6** in the Zoo CLI stack.

- Parent/base: `fm/zoo-cli-host-security` (PR #1158)
- Child: `fm/zoo-cli-interactive` (planned)
- Native local ancestry remains `(main) <- contracts <- headless-api <- host-security <- automation`.
- Native submit evidence: `gh stack submit --auto --open --remote origin` returned `Stacked PRs are not enabled for this repository`, so GitHub uses the approved classic base-branch fallback without flattening or retargeting local ancestry.

## Scope

- Adds the fresh public `@zoo-code/cli` application and `zoo` executable in `apps/zoo`; inherited `apps/cli` remains untouched.
- Implements explicit `run`, formatted `resume`, and workspace-scoped `sessions list` grammar with non-persisting overrides and safe unattended approval defaults.
- Supervises the private host over dedicated Node IPC with capability negotiation, readiness gating, ACK-before-DONE enforcement, heartbeat liveness, crash settlement, bounded shutdown, and redacted diagnostics.
- Adds one pure session reducer consumed by text, final JSON, and versioned NDJSON renderers.
- Handles whole-task deadlines, canonical cancellation, SIGINT/SIGTERM exit mapping, output closure, ephemeral roots, and exactly one accepted root result.
- Bridges canonical headless events and workspace history without interpreting webview messages.

## Acceptance Evidence

- `pnpm --dir apps/zoo test` (9 unit and packaged subprocess tests)
- `pnpm --dir apps/zoo check-types`
- `pnpm --dir packages/zoo-host test` (5 tests) and typecheck
- `pnpm --dir packages/zoo-protocol exec vitest run src/__tests__/contracts.test.ts` (60 tests)
- `pnpm --dir src exec vitest run extension/__tests__/api-headless.spec.ts` (6 tests)
- Targeted suppression-pruning ESLint passed; repository pre-commit lint passed.

## Risk

This layer establishes automation semantics before release artifact assembly. The principal risk is lifecycle drift between extension events and normalized events; strict negotiation, subprocess crash/timeout/signal tests, and one authoritative parent settlement limit that risk. Interactive Ink behavior and installed cross-platform artifact layout intentionally remain in later stack layers.

```mermaid
sequenceDiagram
    participant U as User or CI
    participant Z as zoo parent
    participant H as zoo-host
    participant E as Production extension
    U->>Z: zoo run / resume
    Z->>H: hello.select + task command
[...]

PR #1160, opened 2026-08-05
feat(cli): add interactive Zoo terminal client

taltas:
## Stack

Position **5 of 6** in the Zoo CLI stack.

- Parent/base: `fm/zoo-cli-automation` (PR #1159)
- Child: `fm/zoo-cli-release-docs` (planned)
- Native local ancestry is preserved; GitHub uses the recorded classic fallback because `gh stack submit` reports stacked PRs are not enabled.

## Scope

- Adds a focused Ink terminal frontend over the same pure `SessionProjection` used by automation.
- Supports initial prompts, follow-up input, approve-once/reject actions, cancellation, and idle exit.
- Renders ordered assistant/reasoning messages, tool state, approval cards, session status, and final outcome.
- Dynamically loads Ink only for the interactive root command so automation output and signal behavior remain isolated.
- Keeps all runtime/task behavior in the supervised production extension host.

## Acceptance Evidence

- `pnpm --dir apps/zoo test` (10 tests, including Ink rendering and packaged subprocess automation)
- `pnpm --dir apps/zoo check-types`
- Targeted suppression-pruning ESLint
- Repository pre-commit lint and pre-push typecheck

## Risk

Terminal input is intentionally focused rather than a broad control protocol. Ink is dynamically imported so its signal handling cannot affect `run`, JSON, or NDJSON automation paths.

```mermaid
flowchart LR
    U[TTY user] --> I[Ink client]
    I --> A[Input and approval actions]
    A --> H[Supervised zoo-host]
    H --> E[Production extension]
    E --> N[Normalized events]
    N --> R[SessionProjection reducer]
    R --> I
```

### DIFF
--- a/apps/zoo/package.json
+++ b/apps/zoo/package.json
@@ -19,12 +19,16 @@
 	},
 	"dependencies": {
 		"@roo-code/zoo-protocol": "workspace:^",
-		"commander": "^12.1.0"
+		"commander": "^12.1.0",
+		"ink": "^6.6.0",
+		"react": "^19.1.0"
 	},
 	"devDependencies": {
 		"@roo-code/config-eslint": "workspace:^",
 		"@roo-code/config-typescript": "workspace:^",
 		"@types/node": "22.20.1",
+		"@types/react": "18.3.31",
+		"ink-testing-library": "4.0.0",
 		"rimraf": "6.0.1",
 		"tsup": "8.5.1",
 		"vitest": "4.1.9"
--- a/apps/zoo/src/__tests__/interactive.test.tsx
+++ b/apps/zoo/src/__tests__/interactive.test.tsx
@@ -1,0 +1,24 @@
+import { render } from "ink-testing-library"
+import { describe, expect, it, vi } from "vitest"
+
+import { initialProjection } from "../projection.js"
+import { InteractiveSession } from "../interactive.js"
+
+describe("InteractiveSession", () => {
+	it("submits input and renders approval controls", () => {
+		const submit = vi.fn()
+		const projection = {
+			...initialProjection(),
+			pendingAsks: new Map([["ask-1", { taskId: "root", category: "tool", subject: "Write file?" }]]),
+		}
+		const view = render(
+			<InteractiveSession
+				projection={projection}
+				actions={{ submit, approve: vi.fn(), cancel: vi.fn(), exit: vi.fn() }}
+			/>,
+		)
+
+		expect(view.lastFrame()).toContain("Approval required")
+		expect(view.lastFrame()).toContain("Write file?")
+	})
+})
--- a/apps/zoo/src/index.ts
+++ b/apps/zoo/src/index.ts
@@ -48,6 +48,8 @@ function normalized(options: SharedOptions & { format: OutputFormat; quiet: bool
 	return { ...options, workspace: resolveWorkspace(options.cwd), timeout: parseDuration(options.timeout) }
 }
 
+program.argument("[prompt...]")
+
 automation(program.command("run [prompt...]").description("run one task without an interactive UI")).action(
 	async (words: string[] | undefined, options: SharedOptions & { format: OutputFormat; quiet: boolean }) => {
 		const positional = words?.join(" ").trim()
@@ -72,9 +74,13 @@ shared(
 	await listSessions({ ...options, workspace: resolveWorkspace(options.cwd) })
 })
 
-program.action(() => {
+program.action(async (words: string[] | undefined, options: SharedOptions) => {
 	if (!process.stdin.isTTY || !process.stdout.isTTY) throw new Error("Interactive Zoo requires TTY stdin and stdout")
-	throw new Error("Interactive Zoo is not available in this build")
+	const { runInteractive } = await import("./interactive.js")
+	process.exitCode = await runInteractive(words?.join(" ").trim(), {
+		...options,
+		workspace: resolveWorkspace(options.cwd ?? process.cwd()),
+	})
 })
 
 program.showSuggestionAfterError().showHelpAfterError()
--- a/apps/zoo/src/interactive.tsx
+++ b/apps/zoo/src/interactive.tsx
@@ -1,0 +1,172 @@
+import fs from "node:fs"
+import os from "node:os"
+import path from "node:path"
+import { fileURLToPath } from "node:url"
+
+import { Box, Text, render, useInput } from "ink"
+import { useState } from "react"
+
+import { exitCodeFor, type ZooRunResult } from "@roo-code/zoo-protocol"
+
+import { runOverrides, type SharedOptions } from "./options.js"
+import { initialProjection, reduceSession, type SessionProjection } from "./projection.js"
+import { defaultStorageRoot, HostClient } from "./supervisor.js"
+
+type InteractiveOptions = Omit<SharedOptions, "cwd" | "timeout"> & { workspace: string }
+
+type Actions = {
+	submit: (text: string) => void
+	approve: (approve: boolean) => void
+	cancel: () => void
+	exit: () => void
+}
+
+export function InteractiveSession({ projection, actions }: { projection: SessionProjection; actions: Actions }) {
+	const [input, setInput] = useState("")
+	const ask = [...projection.pendingAsks.entries()][0]
+
+	useInput((value, key) => {
+		if (key.ctrl && value === "c") return actions.cancel()
+		if (key.ctrl && value === "d" && input.length === 0) return actions.exit()
+		if (ask && (value.toLowerCase() === "y" || value.toLowerCase() === "n")) {
+			actions.approve(value.toLowerCase() === "y")
+			return
+		}
+		if (key.return) {
+			if (input.trim()) actions.submit(input.trim())
+			setInput("")
+			return
+		}
+		if (key.backspace || key.delete) return setInput((current) => current.slice(0, -1))
+		if (!key.ctrl && !key.meta && value) setInput((current) => current + value)
+	})
+
+	return (
+		<Box flexDirection="column" paddingX={1}>
+			<Box borderStyle="round" borderColor="cyan" paddingX={1} justifyContent="space-between">
+				<Text bold color="cyan">
+					Zoo Code
+				</Text>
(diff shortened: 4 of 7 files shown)


apps/zoo/src/interactive.tsx
```diff
@@ -0,0 +1,172 @@
...
+		exit: () => settle?.(undefined),
+	}
+	const App = () => {
+		const [state, setState] = useState(projection)
+		update = setState
+		return <InteractiveSession projection={state} actions={actions} />
+	}
+	const instance = render(<App />, { exitOnCtrlC: false })
+	if (initialPrompt) actions.submit(initialPrompt)
+	const result = await settled
```
edelauna, line 154: What happens if the host process dies after `task.start` succeeds but before a `task.result` event arrives? `client.failed` is never raced against this promise, so on a host crash (watchdog timeout, SIGKILL) `settle()` is never called and the CLI hangs here with no escape. `automation.ts` races `client.failed.catch(...)` in a `Promise.race` — should the interactive path do the same?

apps/zoo/src/interactive.tsx
```diff
@@ -0,0 +1,172 @@
...
+					.command({
+						type: "task.start",
+						workspace: options.workspace,
+						prompt: text,
+						overrides: runOverrides({ ...options, approval: "interactive" }),
+					})
+					.then((response) => {
+						if (response.data.commandType === "task.start") rootTaskId = response.data.task.rootTaskId
+					})
+					.catch(() => settle?.(undefined))
```
edelauna, line 126: If `task.start` rejects (bad credentials, host error, crash while the start is pending), this `.catch` settles with `undefined`, which `runInteractive` maps to exit 0 at line 158 — a task that never launched reports success, with no stderr message. Should this surface as a non-zero exit, or reset `starting` so the user can retry?

apps/zoo/src/interactive.tsx
```diff
@@ -0,0 +1,172 @@
...
+			: (result.error?.code ?? "task_failed")
+	return exitCodeFor(
+		result.outcome === "failed"
+			? { outcome: "failed", errorCode: failedCode }
+			: result.outcome === "cancelled"
+				? { outcome: "cancelled" }
+				: result.outcome === "timed_out"
+					? { outcome: "timed_out" }
+					: { outcome: result.outcome },
+	)
```
edelauna, line 171: This `failedCode` mapping mirrors `automation.ts:155-173`, and the two copies already diverge — automation passes a `signal` field for the cancelled outcome and keeps a timed-out sub-branch, this one omits both. Could this be a shared helper so the exit-code logic can't drift?

apps/zoo/src/index.ts
```diff
@@ -72,9 +74,13 @@ shared(
 	await listSessions({ ...options, workspace: resolveWorkspace(options.cwd) })
 })
 
-program.action(() => {
+program.action(async (words: string[] | undefined, options: SharedOptions) => {
```
edelauna, line 77: `shared()` is applied to the subcommands but not the root `program`, so none of `--provider/--model/--approval/--ephemeral/--debug` are registered here — at runtime `options` is `{}` and `zoo --ephemeral "prompt"` throws "unknown option". Should the root program call `shared(program)` before `.action(...)`?

apps/zoo/src/interactive.tsx
```diff
@@ -0,0 +1,172 @@
...
+	let projection = initialProjection()
+	let update: ((projection: SessionProjection) => void) | undefined
+	let rootTaskId: string | undefined
+	let currentTaskId: string | undefined
+	let settle: ((result: ZooRunResult | undefined) => void) | undefined
+	const settled = new Promise<ZooRunResult | undefined>((resolve) => (settle = resolve))
+	const client = new HostClient({
+		workspace: options.workspace,
+		storageRoot,
+		extensionRoot: process.env.ZOO_EXTENSION_PATH ?? fileURLToPath(new URL("../../../src/dist", import.meta.url)),
```
edelauna, line 100: Is `ZOO_EXTENSION_PATH` documented anywhere? It's the only way to point the CLI at a dev or custom extension root, but I don't see it in a README or `.env.example`.

apps/zoo/src/__tests__/interactive.test.tsx
```diff
@@ -0,0 +1,24 @@
+import { render } from "ink-testing-library"
+import { describe, expect, it, vi } from "vitest"
+
+import { initialProjection } from "../projection.js"
+import { InteractiveSession } from "../interactive.js"
+
+describe("InteractiveSession", () => {
+	it("submits input and renders approval controls", () => {
```
edelauna, line 8: The test renders the component but never drives `view.stdin.write(...)`, so the `useInput` branches (approve `y`/`n`, submit, cancel, exit) are unexercised — deleting `if (input.trim()) actions.submit(...)` or flipping the `y`/`n` mapping wouldn't fail it. The name also implies submit is covered, but it never triggers it. Could you simulate keypresses and assert the mocks are called? The empty-`pendingAsks` case (no "Approval required") looks uncovered too.

apps/zoo/src/interactive.tsx
```diff
@@ -0,0 +1,172 @@
...
+				taskId: pending[1].taskId,
+				askId: pending[0],
+				response: approve ? "approve" : "reject",
+			})
+		},
+		cancel() {
+			if (rootTaskId) void client.command({ type: "task.cancel", rootTaskId, reason: "user" })
+			else settle?.(undefined)
+		},
+		exit: () => settle?.(undefined),
```
edelauna, line 145: `exit` fires whenever input is empty, even with a task running or an ask pending — the in-flight outcome is discarded and `runInteractive` returns 0. The "exits when idle" hint at line 81 isn't enforced by any idle check. Should exit be gated when a task/ask is active?

apps/zoo/src/interactive.tsx
```diff
@@ -0,0 +1,172 @@
...
+			void client.command({
+				type: "ask.respond",
+				taskId: pending[1].taskId,
+				askId: pending[0],
+				response: approve ? "approve" : "reject",
+			})
+		},
+		cancel() {
+			if (rootTaskId) void client.command({ type: "task.cancel", rootTaskId, reason: "user" })
+			else settle?.(undefined)
```
edelauna, line 143: If Ctrl+C lands between `task.start` being sent (line 116, `starting = true`) and `rootTaskId` being populated (line 124), this takes the `settle(undefined)` → exit 0 branch while the host may still accept the side-effecting task. Should cancel await the start outcome when `starting` is set?

apps/zoo/src/interactive.tsx
```diff
@@ -0,0 +1,172 @@
...
+						prompt: text,
+						overrides: runOverrides({ ...options, approval: "interactive" }),
+					})
+					.then((response) => {
+						if (response.data.commandType === "task.start") rootTaskId = response.data.task.rootTaskId
+					})
+					.catch(() => settle?.(undefined))
+				return
+			}
+			if (currentTaskId) void client.command({ type: "task.input", taskId: currentTaskId, text })
```
edelauna, line 129: These follow-up `client.command(...)` calls (here and at ask.respond / cancel) have no `.catch()` — if the host dies, `supervisor.ts` rejects every pending command and this surfaces as an unhandled rejection, or a silently lost follow-up. Intentional?

apps/zoo/src/interactive.tsx
```diff
@@ -0,0 +1,172 @@
...
+		const [state, setState] = useState(projection)
+		update = setState
+		return <InteractiveSession projection={state} actions={actions} />
+	}
+	const instance = render(<App />, { exitOnCtrlC: false })
+	if (initialPrompt) actions.submit(initialPrompt)
+	const result = await settled
+	instance.unmount()
+	await client.stop()
+	if (options.ephemeral) fs.rmSync(storageRoot, { recursive: true, force: true })
```
edelauna, line 157: If `client.start()` throws at line 110, control leaves `runInteractive` before this `rmSync` — the ephemeral `mkdtemp` directory leaks. `automation.ts` wraps its cleanup in a `finally`; should this match that pattern?

apps/zoo/src/interactive.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+export async function runInteractive(initialPrompt: string | undefined, options: InteractiveOptions): Promise<number> {
+	const storageRoot = options.ephemeral
+		? fs.mkdtempSync(path.join(os.tmpdir(), "zoo-"))
+		: path.join(defaultStorageRoot(), "state")
+	fs.mkdirSync(storageRoot, { recursive: true })
+	let projection = initialProjection()
+	let update: ((projection: SessionProjection) => void) | undefined
+	let rootTaskId: string | undefined
+	let currentTaskId: string | undefined
```
edelauna, line 94: This `currentTaskId` is overwritten with `projection.currentTaskId` on every event and only read at line 129, so it always equals the projection field. Could the reads just use `projection.currentTaskId`?

apps/zoo/src/index.ts
```diff
@@ -72,9 +74,13 @@ shared(
...
 	if (!process.stdin.isTTY || !process.stdout.isTTY) throw new Error("Interactive Zoo requires TTY stdin and stdout")
```
edelauna, line 78: With this TTY guard in place and no interactive subprocess test, the default `zoo` path (onEvent → settle → unmount → exitCodeFor) has no packaged/CLI-layer coverage — `subprocess.test.ts` covers `run`/`resume`/`sessions` only. Worth a `ZOO_FORCE_INTERACTIVE` bypass and a fake-host subprocess test?

apps/zoo/package.json
```diff
@@ -19,12 +19,16 @@
...
-		"commander": "^12.1.0"
+		"commander": "^12.1.0",
+		"ink": "^6.6.0",
+		"react": "^19.1.0"
 	},
 	"devDependencies": {
 		"@roo-code/config-eslint": "workspace:^",
 		"@roo-code/config-typescript": "workspace:^",
 		"@types/node": "22.20.1",
+		"@types/react": "18.3.31",
```
edelauna, line 30: why not match the react version v19?

edelauna (changes requested):

PR #1165, opened 2026-08-05
refactor(providers): canonicalize model-router calls

WebMad:
## Summary

- use the shared `providerIdentifiers` registry for model-router provider options and cache calls
- canonicalize OpenRouter endpoint routing, LM Studio/Kimi Code/Requesty/Unbound/Poe cache access, and webview router candidates and filters
- preserve all existing serialized and wire provider values

## Validation

- `npx vitest run api/providers/__tests__/poe.spec.ts api/providers/fetchers/__tests__/modelEndpointCache.spec.ts core/webview/__tests__/webviewMessageHandler.routerModels.spec.ts` — 28 tests passed
- `pnpm run check-types` in `src` — passed
- repository pre-commit lint — passed
- repository pre-push type checks — passed

Related to #944 (remaining-work item 4).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency and reliability across AI provider and router integrations.
  * Enhanced model discovery, caching, filtering, and refresh behavior.
  * Improved handling of missing credentials, project settings, and custom endpoints.
  * Preserved existing provider behavior while reducing configuration-related issues.

* **Tests**
  * Expanded coverage for model loading, cache refreshes, unsaved credentials, provider filtering, error handling, and custom endpoints.

### DIFF
--- a/src/api/providers/__tests__/poe.spec.ts
+++ b/src/api/providers/__tests__/poe.spec.ts
@@ -1,6 +1,55 @@
-const mockStreamText = vitest.fn()
-const mockGenerateText = vitest.fn()
-const mockCreatePoe = vitest.fn()
+import { poeDefaultModelId, providerIdentifiers } from "@roo-code/types"
+
+import { PoeHandler } from "../poe"
+import { getModelsFromCache } from "../fetchers/modelCache"
+
+import { clearAllMocks } from "../../../test-utils/reset"
+
+const { mockStreamText, mockGenerateText, mockCreatePoe, mockGetModelsFromCache, mockCaptureException } =
+	vitest.hoisted(() => ({
+		mockStreamText: vitest.fn(),
+		mockGenerateText: vitest.fn(),
+		mockCreatePoe: vitest.fn(),
+		mockCaptureException: vitest.fn(),
+		mockGetModelsFromCache: vitest.fn(),
+	}))
+
+const cachedModels = {
+	"anthropic/claude-sonnet-4": {
+		maxTokens: 10_000,
+		contextWindow: 200_000,
+		supportsImages: true,
+		supportsPromptCache: true,
+		supportsReasoningBudget: true,
+		inputPrice: 3,
+		outputPrice: 15,
+	},
+	"openai/gpt-4o": {
+		maxTokens: 16_384,
+		contextWindow: 128_000,
+		supportsImages: true,
+		supportsPromptCache: false,
+		inputPrice: 2.5,
+		outputPrice: 10,
+	},
+	"openai/o3": {
+		maxTokens: 100_000,
+		contextWindow: 200_000,
+		supportsImages: true,
+		supportsPromptCache: false,
+		supportsReasoningEffort: ["low", "medium", "high"],
+		inputPrice: 10,
+		outputPrice: 40,
+	},
+}
+
+vitest.mock("@roo-code/telemetry", () => ({
+	TelemetryService: {
+		instance: {
+			captureException: (...args: unknown[]) => mockCaptureException(...args),
+		},
+	},
+}))
 
 vitest.mock("ai-sdk-provider-poe", () => ({
 	createPoe: (...args: unknown[]) => mockCreatePoe(...args),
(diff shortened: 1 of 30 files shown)


src/core/webview/__tests__/webviewMessageHandler.routerModels.spec.ts
```diff
@@ -283,19 +292,71 @@ describe("webviewMessageHandler - requestRouterModels provider filter", () => {
 		)
 
 		// flushModels should NOT have been called for litellm
-		const litellmFlushCalls = flushModelsMock.mock.calls.filter((c: any[]) => c[0] === "litellm")
+		const litellmFlushCalls = flushModelsMock.mock.calls.filter((c: any[]) => c[0] === providerIdentifiers.litellm)
```
edelauna: `c[0]` is an options object — comparing it to a string with `===` is always `false`, so `litellmFlushCalls` is always `[]` and this assertion cannot fail. The `getModels` filter two lines down already uses the correct pattern.

```suggestion
			const litellmFlushCalls = flushModelsMock.mock.calls.filter((c: any[]) => c[0]?.provider === providerIdentifiers.litellm)
```

src/api/providers/openrouter.ts
```diff
@@ -164,9 +165,9 @@ export class OpenRouterHandler extends BaseProvider implements SingleCompletionH
 	private async loadDynamicModels(): Promise<void> {
 		try {
 			const [models, endpoints] = await Promise.all([
-				getModels({ provider: "openrouter" }),
+				getModels({ provider: providerIdentifiers.openrouter }),
```
edelauna, line 168: This PR canonicalized the `getModels` / `getModelEndpoints` calls to use `providerIdentifiers.openrouter`. Should the `ApiProviderError` calls in `handleStreamingError`, `createMessage`, and `completePrompt` also use `providerIdentifiers.openrouter` instead of `this.providerName` (`"OpenRouter"`) for consistency in telemetry grouping? `poe.ts` was migrated in the same PR.

src/core/webview/webviewMessageHandler.ts
```diff
@@ -1212,28 +1238,28 @@ export const webviewMessageHandler = async (
...
 
 			candidates.push({
-				key: "nanogpt",
-				options: { provider: "nanogpt", apiKey: nanoGptApiKey },
+				key: providerIdentifiers.nanogpt,
+				options: { provider: providerIdentifiers.nanogpt, apiKey: nanoGptApiKey },
 			})
 
-			if (!providerFilter || providerFilter === "kimi-code") {
+			if (!providerFilter || providerFilter === providerIdentifiers.kimiCode) {
```
edelauna, line 1267: Does the routerModels spec have coverage for this branch? I do not see tests for `providerFilter === providerIdentifiers.kimiCode`, or for the api-key vs. OAuth dispatch and the null-token exclusion inside it.

src/api/providers/__tests__/poe.spec.ts
```diff
@@ -116,9 +134,19 @@ describe("PoeHandler", () => {
...
+			const options = {
+				poeApiKey: "key",
+				poeBaseUrl: "https://custom.poe.com/v1",
+				apiModelId: "anthropic/claude-sonnet-4",
+			}
+			const handler = new PoeHandler(options)
 			const result = handler.getModel()
 
+			expect(getModelsFromCache).toHaveBeenCalledWith({
+				provider: providerIdentifiers.poe,
```
edelauna, line 146: Since both the production code and this assertion import `providerIdentifiers.poe`, a regression where production reverts to the literal `"poe"` goes undetected here. Would using the string literal `"poe"` in the expectation be more mutation-resistant? (The constant's string value is already pinned in `provider-identifiers.test.ts`.)

edelauna (changes requested):

src/api/providers/__tests__/poe.spec.ts
WebMad: Thanks for raising this. Using the literal in the expectation would not make this behavioral test more mutation-resistant: `providerIdentifiers.poe` and `"poe"` evaluate to the same runtime value, so changing production code between them is an equivalent mutation that neither expectation can detect. The wire value is already pinned independently in `provider-identifiers.test.ts`, while this test verifies the cache call contract through the typed provider API. I have kept the canonical identifier here and opened #1294 as a focused follow-up to enforce this source-level convention with linting/static analysis.

src/core/webview/__tests__/webviewMessageHandler.routerModels.spec.ts
WebMad: Fixed. The filter now checks `c[0]?.provider === providerIdentifiers.litellm`, so the assertion correctly detects an unexpected LiteLLM flush.

src/api/providers/openrouter.ts
WebMad: Agreed and updated. All `ApiProviderError` instances in `handleStreamingError`, `createMessage`, and `completePrompt` now use `providerIdentifiers.openrouter`, and the telemetry assertions were updated accordingly.

src/core/webview/webviewMessageHandler.ts
WebMad: Added focused coverage for this branch: Kimi Code provider filtering with explicit API-key dispatch, OAuth token dispatch, and exclusion when OAuth returns a null token. The API-key test also verifies that the OAuth manager is not queried. The focused routerModels suite passes.

src/api/providers/openrouter.ts
```diff
@@ -197,7 +198,7 @@ export class OpenRouterHandler extends BaseProvider implements SingleCompletionH
 		const rawErrorMessage = parsedError || error?.message || "Unknown error"
 
 		const apiError = Object.assign(
-			new ApiProviderError(rawErrorMessage, this.providerName, modelId, operation, error?.code),
+			new ApiProviderError(rawErrorMessage, providerIdentifiers.openrouter, modelId, operation, error?.code),
```
edelauna, line 201: `this.providerName` is `"OpenRouter"` (L147) but `providerIdentifiers.openrouter` is `"openrouter"`, so this changes the emitted telemetry `provider` value (via `extractApiProviderErrorProperties`) and splits any Sentry/PostHog series or alert keyed on `"OpenRouter"`. It's the one non-value-preserving swap in this PR, and `handleOpenAIError(error, this.providerName)` below still emits `"OpenRouter"` -- the same error now carries two spellings. If the casing change is intended, align `handleOpenAIError` and note it in the PR; otherwise keep `this.providerName` here.

edelauna: Created #1302 to track this - so we don't block this pr.

edelauna (approved):

PR #1175, opened 2026-08-06
fix: stream reasoning_content in LM Studio provider

### LINKED ISSUE #1075: [BUG] "Thinking" content is not displayed in the "lm studio provider" + "lmstudio" + "qwen3.6-35b-a3b" environment.
### environment 
* windows 11 25H2 64bit
* vscode 1.131.0
* zoo code 3.72.0
* lm studio 0.4.20(build 1) + qwen3.6-35b-a3b

`Thinking` content is not displayed in the `lm studio provider` + `lmstudio` + `qwen3.6-35b-a3b` environment.
But, `Thinking` content is displayed in the `openai compatible provider` + `lmstudio` + `qwen3.6-35b-a3b` environment.

### `lm studio provider` + `lmstudio` + `qwen3.6-35b-a3b`
Image

### `openai compatible provider` + `lmstudio` + `qwen3.6-35b-a3b` environment.
Image
Image

### provider setting
```json

			"lmstudio/qwen3.6 35b a3b": {
				"modelTemperature": null,
				"enableReasoningEffort": false,
				"reasoningEffort": "medium",
				"lmStudioModelId": "qwen/qwen3.6-35b-a3b",
				"lmStudioBaseUrl": "http://127.0.0.1:1234",
				"lmStudioSpeculativeDecodingEnabled": false,
				"apiProvider": "lmstudio",
				"id": "....."
			},
			"lmstudio-openai/qwen-35b-a3b": {
				"modelTemperature": null,
				"enableReasoningEffort": true,
				"reasoningEffort": "medium",
				"openAiBaseUrl": "http://127.0.0.1:1234/v1",
				"openAiApiKey": "ff",
				"openAiModelId": "qwen/qwen3.6-35b-a3b",
				"openAiCustomModelInfo": {
					"maxTokens": -1,
					"contextWindow": 262144,
					"supportsImages": true,
					"supportsPromptCache": true,
					"inputPrice": 0,
					"outputPrice": 0,
					"reasoningEffort": "medium"
				},
				"openAiHeaders": {},
				"apiProvider": "openai",
				"id": "....."
			}

```

daewoongoh:
### Related GitHub Issue

Closes: #1075

### Description

Reasoning models served by LM Studio stream their thinking in a dedicated `reasoning_content` (or
`reasoning`) delta field. The LM Studio handler only read `delta.content` and passed it through
`TagMatcher` to extract `<think>`/`<thought>` tags, so the entire thought stream was silently
dropped for these models.

The repo already has a shared helper for exactly this — `extractReasoningFromDelta` in
`src/api/providers/utils/extract-reasoning.ts` — used by 13 providers (`openai`, `deepseek`,
`lite-llm`, `kenari`, `mimo`, `qwen-code`, `requesty`, `unbound`, `opencode-go`,
`base-openai-compatible-provider`, …). **LM Studio was the only one still missing it.** That is why
the reporter saw the same model + same LM Studio instance display thinking correctly through the
OpenAI Compatible provider but not through the LM Studio provider.

The fix mirrors the existing call site in `openai.ts` rather than introducing a new pattern:

```ts
const reasoningText = extractReasoningFromDelta(delta)
if (reasoningText) {
    reasoningOutput += reasoningText
    yield { type: "reasoning", text: reasoningText }
}
```

### Test Procedure

**Automated.** Three tests added to `src/api/providers/__tests__/lmstudio.spec.ts`, following the
existing patterns in `openai.spec.ts`:

- `streams reasoning chunks from delta.reasoning_content` — the issue reproduction; asserts both the
  `reasoning` chunk and the `text` chunk are emitted, so a regression that breaks `content` handling
  is also caught
- `falls back to delta.reasoning when reasoning_content is absent` — the helper's fallback path
- `still parses <think> tags embedded in content` — regression guard for models that inline reasoning
  as tags (this one passes before the fix too; it exists to pin existing behaviour)

Verified these are genuine regression tests, not just passing tests: with the handler reverted to its
pre-fix state the reasoning tests fail (`0` reasoning chunks emitted); with the fix they pass.

```bash
cd src
npx vitest run api/providers/__tests__/lmstudio.spec.ts
```

**Manual verification** was not possible — I don't have an LM Studio instance with a reasoning model
in this environment. The tests exercise the delta shapes the issue describes, but a reviewer with the
reporter's setup (LM Studio 0.4.20 + Qwen3.6 35B A3B) confirming thinking now renders would be
valuable before merge.

### Pre-Submission Checklist

- [x] **Visual Snapshot** (UI changes only): N/A — no UI changes; this is provider stream-parsing logic.

### Visual Snapshots

image

N/A

### Documentation Updates

### Additional Notes

N/A

### Get in Touch

hehegwk_23849

## Summary by CodeRabbit

* **Improvements**
  * Improved LM Studio streaming to reliably capture and display AI reasoning.
  * Added support for reasoning delivered through multiple response formats, including streamed reasoning updates.
  * Preserved support for reasoning embedded in `<think>` tags.
  * Reasoning is now included in output token usage counts, improving usage reporting during streamed responses.
  * Streaming responses now maintain more consistent reasoning and visible answer content, even when multiple reasoning formats are combined.


src/api/providers/__tests__/lmstudio.spec.ts
```diff
@@ -112,6 +112,59 @@ describe("LmStudioHandler", () => {
...
+						usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 },
+					},
+				]),
+			)
+
+			const chunks = await collectStream(handler.createMessage(systemPrompt, messages))
+
+			expect(chunks).toContainEqual({ type: "reasoning", text: "tagged thought" })
+			expect(chunks).toContainEqual({ type: "text", text: "visible" })
+		})
```
edelauna, line 196: Worth adding a case where both `reasoning_content` and `reasoning` are set in the same delta — similar to the test in `openai.spec.ts` at line 297? That would confirm exactly one reasoning chunk is emitted here (rather than two), which the `extractReasoningFromDelta` unit tests cover in isolation but this handler does not exercise end-to-end.

edelauna (changes requested):
Looks good - had 1 test nit

src/api/providers/__tests__/lmstudio.spec.ts
daewoongoh: Added in 45f4f8b — mirrors the openai.spec.ts test you pointed at. Asserts on the filtered reasoning-chunk array so the count is pinned to exactly one, not just membership.

edelauna (approved):
Thanks for the contribution.

Issue #575, opened 2026-06-11
[Story] Bm25Ranker for MCP tool search (pure-JS, no external deps)


edelauna:
## Context

A pure-JS BM25 ranker is the always-available baseline for dynamic tool loading. It works without any embedding provider configured, has no network dependency, and handles the keyword-heavy queries that dominate MCP tool search ("send slack message", "read jira ticket").

This is foundational for the dynamic loading feature (#577 ToolRouter, #578 threshold gate); it doesn't close a user-facing issue on its own.

## Developer Notes

- New \`src/services/tools/Bm25Ranker.ts\` implementing a \`Ranker\` interface.
- \`Ranker\` interface (in \`src/services/tools/types.ts\`): \`rank(query: string, items: ToolDoc[], k: number): ToolDoc[]\`.
- \`ToolDoc\` shape: \`{ serverName: string, toolName: string, description: string }\`. Ranker tokenizes the concatenation \`\${serverName} \${toolName} \${description}\`.
- Tokenizer: lowercase, split on non-word chars. Standard BM25 (k1=1.5, b=0.75). No external dependencies.
- Build the inverted index lazily on first \`rank()\` call; rebuild when the input set changes (detected by reference inequality).

## Acceptance Criteria

- [ ] \`Bm25Ranker\` implements \`Ranker\` interface
- [ ] Unit test: query "send slack message" ranks \`slack.postMessage\` above \`jira.createIssue\` above \`postgres.query\`
- [ ] Unit test: empty query → empty result; query with no matches → empty result
- [ ] No external npm dependencies added

PR #1207, opened 2026-08-09
feat(tools): add dependency-free BM25 ranker

edelauna:
## Related GitHub Issue

Relates to #575

## Description

Adds a dependency-free, pure-JavaScript BM25 ranker as the always-available baseline for MCP tool search. This PR is limited to the foundational ranker described in issue #575; it does not implement downstream routing or threshold-gating work.

### Implementation

- Adds the `ToolDoc` type with `serverName`, `toolName`, and `description` fields.
- Adds the `Ranker` interface with `rank(query, items, k)` returning the top-ranked `ToolDoc` results.
- Implements standard BM25 scoring with `k1 = 1.5` and `b = 0.75`, with no external npm dependencies.
- Tokenizes the combined server name, tool name, and description by lowercasing and splitting on non-word characters.
- Builds the inverted index lazily on the first `rank()` call and rebuilds it only when the input array reference changes.
- Handles empty queries, empty item sets, no-match queries, zero or limited `k`, empty descriptions, and stable input ordering for equal scores.

## Tests and Validation

- Adds focused unit tests covering BM25 relevance ordering, edge cases, top-`k` truncation, stable tie ordering, case-insensitive punctuation-aware tokenization, empty descriptions, and reference-based index rebuilding.
- Validation: `cd src && npx vitest run services/tools/__tests__/Bm25Ranker.spec.ts`

## Summary by CodeRabbit

* **New Features**
  * Added relevance-based search for available tools.
  * Search results now account for tool names and descriptions, with case- and punctuation-insensitive matching.
  * Results are limited to the requested number and consistently ordered when relevance is equal.
  * Empty searches and searches without matches return no results.

* **Bug Fixes**
  * Improved result accuracy when tool information changes or descriptions are unavailable.

### DIFF
--- a/src/services/tools/Bm25Ranker.ts
+++ b/src/services/tools/Bm25Ranker.ts
@@ -1,0 +1,101 @@
+import { Ranker, ToolDoc } from "./types"
+
+type IndexedDocument = {
+	item: ToolDoc
+	termFrequencies: Map<string, number>
+	length: number
+	index: number
+}
+
+export class Bm25Ranker implements Ranker {
+	private readonly k1 = 1.5
+	private readonly b = 0.75
+	private indexedItems: ToolDoc[] | undefined
+	private documents: IndexedDocument[] = []
+	private documentFrequency = new Map<string, number>()
+	private averageDocumentLength = 0
+
+	rank(query: string, items: ToolDoc[], k: number): ToolDoc[] {
+		const queryTerms = tokenize(query)
+
+		if (queryTerms.length === 0 || items.length === 0 || k <= 0) {
+			return []
+		}
+
+		this.ensureIndex(items)
+
+		const scores = this.documents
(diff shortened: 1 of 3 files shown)


src/services/tools/Bm25Ranker.ts
```diff
@@ -0,0 +1,101 @@
...
+			score += inverseDocumentFrequency * ((termFrequency * (this.k1 + 1)) / (termFrequency + normalization))
+		}
+
+		return score
+	}
+}
+
+function tokenize(value: string): string[] {
+	return value.toLowerCase().split(/\W+/).filter(Boolean)
+}
```
taltas, line 90: Could we treat underscores as separators here so common MCP tool names such as `create_issue` match a query like `create issue`, and add a regression test?

PR #1219, opened 2026-08-11
feat(friendli): fetch model list dynamically from /v1/models

Lee-Si-Yoon:
Friendli's `/v1/models` is now the source of truth for the model list. This narrows the static seed to the default model (GLM-5.2) and lets everything else come from the live fetcher at runtime.

- `FriendliModelId` narrowed to `"zai-org/GLM-5.2"`; static entries for GLM-5.1, DeepSeek-V3.2, MiniMax-M2.5 removed (served dynamically now)
- Static seed values corrected to match the live API (maxTokens/contextWindow: 1,048,576)
- Static seed is only a cold-start fallback — the dynamic fetcher populates the rest within seconds of handler construction
- Added `src/api/providers/fetchers/friendli.ts` to parse `/v1/models` (pricing, reasoning options, tool-calling flags, deprecation status)

Supersedes #1152 and #1028 — Friendli changed the `/v1/models` response shape after those were opened.

Tests: `pnpm vitest run api/providers/__tests__/friendli.spec.ts api/providers/fetchers/__tests__/friendli.spec.ts` (from `src/`), `pnpm vitest run src/components/settings/providers/__tests__/Friendli.spec.tsx src/components/ui/hooks/__tests__/useSelectedModel.spec.ts` (from `webview-ui/`), `pnpm check-types`, `pnpm lint`, `pnpm build`

## Summary by CodeRabbit

* **New Features**
  * Added dynamic Friendli model discovery with pricing, image support, reasoning capabilities, deprecation status, and token limits.
  * Added Friendli models to router results and settings-based model selection.
  * Added a Friendli model picker with dynamically loaded models and static fallback options.

* **Bug Fixes**
  * Improved fallback behavior when model loading is empty or unsuccessful.
  * Preserved selected models while dynamic data loads.
  * Improved handling of reasoning options and request token limits.


Lee-Si-Yoon:
Rebased onto current main (was showing conflicting against a stale fork base — actual base `upstream/main` had no conflicts). Fixed one test file that needed a `friendli` router-models mock entry after the rebase. check-types, lint, and targeted vitest suites all pass.

Ready for review when you have a chance.

src/api/providers/fetchers/friendli.ts
```diff
@@ -0,0 +1,253 @@
...
+
+	// supportsPromptCache is true when the API exposes cache pricing at all —
+	// even a zero write price indicates the provider honors cached reads.
+	const supportsPromptCache = typeof cacheWritesPrice !== "undefined" || typeof cacheReadsPrice !== "undefined"
+
+	const supportsImages = Array.isArray(model.input_modalities) ? model.input_modalities.includes("image") : false
+
+	const modelInfo: ModelInfo = {
+		maxTokens: model.max_completion_tokens ?? 0,
+		contextWindow: model.context_length ?? 0,
```
edelauna: If the live API returns a model with `max_completion_tokens` but no `context_length`, this produces `contextWindow: 0`. `getModelMaxOutputTokens` then computes `Math.min(maxTokens, Math.ceil(0 * 0.2)) = 0`, so `createStream` sends `max_tokens: 0` — which the API rejects with a 400.

Minimal fix:
```suggestion
		contextWindow: model.context_length ?? model.max_completion_tokens ?? 0,
```

webview-ui/src/components/ui/hooks/useSelectedModel.ts
```diff
@@ -385,9 +385,26 @@ function getSelectedModel({
...
+					return { id: savedId, info: staticInfo }
+				}
+				return { id: defaultModelId, info: friendliModels[defaultModelId as keyof typeof friendliModels] }
+			}
+
+			const availableModels = { ...friendliModels, ...routerModels.friendli! }
+			const id = getValidatedModelId(apiConfiguration.apiModelId, availableModels, defaultModelId)
+			const routerInfo = routerModels.friendli?.[id]
+			const staticInfo = friendliModels[id as keyof typeof friendliModels]
+			return { id, info: routerInfo ?? staticInfo }
```
edelauna, line 414: The `routerInfo ?? staticInfo` priority looks right, but is it covered by a test where the same model ID is in *both* maps with different info? Since GLM-5.1 was removed from the static seed, flipping to `staticInfo ?? routerInfo` would produce identical results for all current Friendli tests (staticInfo would be undefined in either ordering). A case where `"zai-org/GLM-5.2"` appears in both maps with distinct info would pin this down.

webview-ui/src/components/ui/hooks/useSelectedModel.ts
```diff
@@ -385,9 +385,26 @@ function getSelectedModel({
...
+			// the default, so the user's selection survives the loading window.
+			const hasFriendliRouter = routerModels.friendli && Object.keys(routerModels.friendli).length > 0
+
+			if (!hasFriendliRouter) {
+				const savedId = apiConfiguration.apiModelId
+				const staticInfo = savedId ? friendliModels[savedId as keyof typeof friendliModels] : undefined
+				if (savedId && staticInfo) {
+					return { id: savedId, info: staticInfo }
+				}
+				return { id: defaultModelId, info: friendliModels[defaultModelId as keyof typeof friendliModels] }
```
edelauna, line 407: If the user has a dynamic-only model configured (not in `friendliModels`) and `routerModels.friendli` is still loading, this returns GLM-5.2 as the active selection. The server's `getModel()` does the opposite — it preserves `requestedId` via `openAiModelInfoSaneDefaults` during the same window.

Would returning `{ id: savedId, info: undefined }` here keep the UI in sync with the server while the dynamic list loads?

webview-ui/src/components/settings/providers/__tests__/Friendli.spec.tsx
```diff
@@ -60,4 +80,26 @@ describe("Friendli provider settings", () => {
...
+		expect(screen.getByTestId("friendli-model-picker-mock")).toBeInTheDocument()
+		expect(mockModelPickerProps).toMatchObject({
+			defaultModelId: "zai-org/GLM-5.2",
+			modelIdKey: "apiModelId",
+			serviceName: "Friendli",
+			serviceUrl: "https://friendli.ai",
+			errorMessage: "test error",
+		})
+		// When routerModels is not provided, ModelPicker should receive the static fallback
+		expect(mockModelPickerProps.models).toEqual(friendliModels)
```
edelauna, line 103: This only tests the no-`routerModels` fallback path. The branch in `Friendli.tsx` that passes `routerModels["friendli"]` to `ModelPicker` when it's non-empty is never exercised here — removing that key-presence check would pass all tests. Worth adding a case that passes a non-empty `routerModels.friendli` and asserts `models` equals the router catalog?

edelauna (changes requested):
Thanks for updating the merge conflict - noticed one bug where context can be 0 and had some additional comments.

src/api/providers/fetchers/friendli.ts
Lee-Si-Yoon: Good catch — fixed. Falls back to `max_completion_tokens` when `context_length` is missing so `contextWindow` is never 0, plus a regression test. (9edff1e4c)

webview-ui/src/components/ui/hooks/useSelectedModel.ts
Lee-Si-Yoon: Good point — now returns `{ id: savedId, info: undefined }` during the loading window instead of substituting the default, matching what `FriendliHandler.getModel()` already does server-side. (f9bd1eb43)

webview-ui/src/components/ui/hooks/useSelectedModel.ts
Lee-Si-Yoon: Added a case with `zai-org/GLM-5.2` in both maps with distinct info, asserting router wins. (f9bd1eb43)

webview-ui/src/components/settings/providers/__tests__/Friendli.spec.tsx
Lee-Si-Yoon: Added a case with a non-empty `routerModels.friendli` asserting ModelPicker receives the router catalog, not the static fallback. (5113411e7)

Lee-Si-Yoon:
@edelauna All 4 comments addressed and replied inline — mind taking another look?

- `contextWindow` fallback fixed + regression test (9edff1e4c)
- loading-window fallback now matches server-side behavior (f9bd1eb43)
- router-vs-static priority test added (f9bd1eb43)
- non-empty routerModels branch test added (5113411e7)

Static seed intentionally left at just GLM-5.2.

src/api/providers/fetchers/friendli.ts
```diff
@@ -0,0 +1,256 @@
...
+	const supportsPromptCache = typeof cacheWritesPrice !== "undefined" || typeof cacheReadsPrice !== "undefined"
+
+	const supportsImages = Array.isArray(model.input_modalities) ? model.input_modalities.includes("image") : false
+
+	const modelInfo: ModelInfo = {
+		maxTokens: model.max_completion_tokens ?? 0,
+		// Fall back to max_completion_tokens when context_length is missing —
+		// leaving this at 0 makes getModelMaxOutputTokens clamp maxTokens to
+		// 0 (20% of a 0 window), which the API then rejects with a 400.
+		contextWindow: model.context_length ?? model.max_completion_tokens ?? 0,
```
edelauna: When `context_length` is absent, both `contextWindow` and `maxTokens` end up equal to `max_completion_tokens`. `getModelMaxOutputTokens` then returns `min(maxTokens, ceil(contextWindow × 0.2))` — so a model with `max_completion_tokens: 8000` would get `max_tokens: 1600` instead of `8000`. Would multiplying by 5 as the contextWindow fallback keep the clamp a no-op?

```suggestion
		contextWindow: model.context_length ?? (model.max_completion_tokens != null ? model.max_completion_tokens * 5 : 0),
```

webview-ui/src/components/ui/hooks/__tests__/useSelectedModel.spec.ts
```diff
@@ -1323,7 +1347,134 @@ describe("useSelectedModel", () => {
...
+			const apiConfiguration: ProviderSettings = {
+				apiProvider: providerIdentifiers.friendli,
+			}
+
+			const wrapper = createWrapper()
+			const { result } = renderHook(() => useSelectedModel(apiConfiguration), { wrapper })
+
+			expect(result.current.id).toBe(friendliDefaultModelId)
+			expect(result.current.info).toEqual(friendliModels[friendliDefaultModelId])
+		})
```
edelauna: This fixture looks identical to the test at line 1376 — same provider, no `apiModelId`, empty `friendli` catalog, same two assertions. Is the `if (savedId)` branch at `useSelectedModel.ts:403` reachable from this fixture? The dynamic-only-id path (non-empty `apiModelId`, empty catalog) is already covered by the test at line 1422.

src/api/providers/fetchers/friendli.ts
Lee-Si-Yoon: Confirmed — `ceil(M*0.2) < M` for any `M>0`, so the clamp silently cut `max_completion_tokens` to 20% (8000 → 1600). Applied your `*5` fallback in 74af128d4, updated the existing test's expected `contextWindow` (8000→40000), and added a regression test that asserts `getModelMaxOutputTokens` returns the unclamped value end-to-end.

webview-ui/src/components/ui/hooks/__tests__/useSelectedModel.spec.ts
Lee-Si-Yoon: Confirmed — traced it to f9bd1eb43, which dropped `apiModelId` from this test (to accommodate the new preserve-dynamic-id behavior) and added the removed case as a new, correctly-named test right after it. That left this one byte-identical to its neighbor. Removed in 74af128d4.

webview-ui/src/components/settings/ApiOptions.tsx
```diff
@@ -698,6 +698,10 @@ const ApiOptions = ({
 						<Friendli
 							apiConfiguration={apiConfiguration}
 							setApiConfigurationField={setApiConfigurationField}
+							routerModels={routerModels}
+							organizationAllowList={organizationAllowList}
+							modelValidationError={modelValidationError}
+							simplifySettings={fromWelcomeView}
```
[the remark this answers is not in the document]
Lee-Si-Yoon: Fixed and verified.

- The empty-record -> static fallback guard landed in commit e333a8869 (`Object.keys(routerModels["friendli"]).length > 0` in `webview-ui/src/components/settings/providers/Friendli.tsx`), since `requestRouterModels` posts `{}` on a failed/empty fetch and `{}` is non-nullish.
- The missing empty-record UI test case is now added in 723931815 ("falls back to static models when routerModels.friendli is an empty record"). All three cases are covered: missing (`passes correct props...`, routerModels not provided), empty (new test), and populated ("passes the router catalog...").
- Mutation check: removing the guard makes the new test fail (1 failed / 5 passed), confirming it is red-capable, then the guard was restored byte-identical.

All Friendli settings tests: 6/6 passed.

src/api/providers/__tests__/friendli.spec.ts
```diff
@@ -536,9 +483,96 @@ describe("FriendliHandler — Friendli-specific reasoning params", () => {
...
+		mockGetModels.mockRejectedValue(new Error("Network error"))
+
+		const handler = new FriendliHandler({
+			friendliApiKey: "test-key",
+		})
+
+		// After rejection, getModel() falls back to the default model.
+		await vi.waitFor(() => {
+			expect(handler.getModel().id).toBe(friendliDefaultModelId)
+		})
```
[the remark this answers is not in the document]
Lee-Si-Yoon: Valid finding — fixed exactly as suggested in commit 723931815.

The handler is now constructed with `apiModelId: "friendli-only/future-model"` (dynamic-only, not in the static map), so while the load is pending `getModel()` preserves the requested id and the `waitFor` assertion can only pass once the rejection handler marks the load settled and the id falls back to `friendliDefaultModelId`.

Verified: the rejection test now observes the fallback transition — handler suite 28/28 passed.

PR #1218, opened 2026-08-10
feat(scm): generate commit messages from the Source Control panel

Rafael-Silva-Oliveira:
### Related GitHub Issue

Part of: #145

This PR implements most of the #145 sub-issue tree in one piece:

| Issue | Status in this PR |
| --- | --- |
| #282 Git context collector | Closes -- `getCommitContext()` in `src/utils/git.ts` |
| #283 Commit-message support prompt type | Closes -- `COMMIT_MESSAGE` in `src/shared/support-prompt.ts` |
| #284 Commit-message generator service | Closes -- `src/services/commit-message/` |
| #285 Dedicated API profile setting | Closes -- `commitMessageApiConfigId` |
| #286 Register command in VS Code SCM | Closes -- `zoo-code.generateCommitMessage` + `scm/title` |
| #287 Resolve Git repository from SCM context | Closes -- matches `SourceControl.rootUri` against the git API |
| #288 Insert message into commit input box | Closes -- writes `repository.inputBox.value` |
| #289 Progress, cancellation, and error UX | **Partial** -- progress and error handling done; no cancel button (see Additional Notes) |
| #290 Commit-message prompt settings UI | Closes -- inherited from the support-prompt system |
| #291 Commit-message API profile settings UI | Closes -- `CommitMessageModelSelect` in Providers |
| #292 Privacy-safe telemetry | **Not addressed** -- left for a follow-up |

Happy to split this along those lines if that's easier to review -- see "Relationship
to the existing stack" below.

### Description

Adds a Zoo Code button to the Source Control title bar. Click it and the current
changes are summarized into a commit message, written straight into the commit
input box.

The implementation leans on things Zoo Code already has, so the new surface area is
small: one git helper, one ~110-line service, one React `<Select>`.

**Where the changes come from.** `getCommitContext()` in `src/utils/git.ts` reads
staged changes first, since that's what a commit will actually contain. When nothing
is staged it falls back to the whole working tree, so the button still does something
useful before you've staged. The fallback path uses `git status --short` rather than a
diff, because untracked files don't appear in any diff and would otherwise be
invisible to the model. Oversized diffs go through the existing `truncateOutput`
helper, and `--unified=1` keeps the prompt lean without hiding what changed.

**The prompt is editable.** `COMMIT_MESSAGE` is registered in
`src/shared/support-prompt.ts` alongside `ENHANCE` and the rest, which means it picks
up the existing Prompts settings UI for free -- textarea, reset button, and
[...]

PR #301, opened 2026-05-24
[4/4] Add commit message settings, profiles, and attribution

Mirrowel:
### Related GitHub Issue

Closes: #290, #291

Part of: #145

Replaces remaining parts of: #146

Depends on: #298, #299, #300

### Description

This is part 4 of the split commit-message generation stack. It adds the user-facing settings, profile management, attribution footer support, state plumbing, and localization needed to make the commit-message feature configurable.

This PR intentionally comes last because it touches the widest surface area: shared settings schemas, extension state, webview settings UI, profile normalization, attribution generation, and locale files. The preceding PRs keep the collector, generator, and SCM integration independently reviewable before adding this configurability layer.

Key implementation details:

- Adds commit-message prompt editing to the existing Settings > Prompts support prompt UI.
- Adds a dedicated commit-message API profile selector with a "use current config" option.
- Adds configurable Git context options, including diff context lines, diff stats, current branch, recent commits, recent commit bodies, recent commit stats, recent commit diffs, and recent commit diff count.
- Adds multiple commit-message profiles with per-profile prompt, API config, Git context, and attribution settings.
- Adds profile normalization helpers and fallback behavior from the earlier single-profile settings keys.
- Adds deterministic AI attribution footer support for generated commit messages.
- Adds attribution template placeholders for agent/tool/provider/model values.
- Wires new values through `packages/types`, `ClineProvider`, `ExtensionStateContext`, and `SettingsView`.
- Preserves the `SettingsView` cached-state pattern so settings inputs edit local cached state until Save.
- Adds localized strings for the complete commit-message settings and attribution UI.
- Adds focused tests for profile normalization, attribution formatting, and settings UI cached-state updates.

Reviewers should focus on SettingsView cached-state correctness, profile normalization/fallback behavior, attribution footer behavior, and whether the settings UI scope is acceptable as the final configurability layer of the feature.

This aligns with Zoo Code's UX roadmap by making the feature configurable for different team conventions and model-cost preferences while preserving reliable settings save behavior.

### Test Procedure

Automated verification run before opening the split PR stack:

- `pnpm --dir src exec vitest run services/commit-message/__tests__/profileSettings.spec.ts services/commit-message/__tests__/attribution.spec.ts services/commit-message/__tests__/CommitMessageGenerator.spec.ts`
- `pnpm --dir webview-ui exec vitest run src/components/settings/__tests__/CommitMessagePromptSettings.spec.tsx`
[...]

PR #298, opened 2026-05-24
[1/4] Add Git context collector for commit messages

Mirrowel:
### Related GitHub Issue

Closes: #282

Part of: #145

Replaces part of: #146

### Description

This is part 1 of the split commit-message generation stack. It adds the reusable Git context collection layer that later PRs use to generate commit messages from the user's actual Git state.

This PR adds `GitContextCollector` under `src/services/git-context` and keeps it independent from AI generation, VS Code SCM command wiring, and settings UI. The collector uses Git as the source of truth rather than Zoo file ignore filters, because commit-message generation should describe what the user staged or modified for Git.

Key implementation details:

- Adds `GitContextCollector` plus exported Git context/change/result types.
- Parses staged changes from `git diff --name-status --cached -z`.
- Parses unstaged and untracked changes from `git status --porcelain=v1 -z --untracked-files=all`.
- Handles modified, added, deleted, renamed, copied, unknown, and untracked files.
- Handles untracked text files by creating synthetic new-file diffs.
- Handles binary files without embedding binary payloads into prompt context.
- Supports diff stats, configurable diff context lines, current branch, recent commits, recent commit stats, and recent commit diffs.
- Returns non-fatal warnings for supplemental repository context failures while preserving required diff failures.
- Adds focused tests for parsing, untracked files, binary handling, warnings, and context output.

Reviewers should focus on path-safe Git parsing, rename/copy handling, binary/untracked behavior, and whether the collector contract is reusable enough for the following PRs.

This aligns with Zoo Code's reliability-first roadmap by isolating the Git-source-of-truth behavior in a tested service before connecting it to UI or model calls.

### Test Procedure

Automated verification run before opening the split PR stack:

- `pnpm --dir src exec vitest run services/git-context/__tests__/GitContextCollector.spec.ts`
- Split verification agent checked that this PR only contains `src/services/git-context/**` and has no generator, SCM command, package manifest, state, or UI wiring.
- Pre-push typecheck completed successfully through the repository hook.

Reviewer verification:

- Inspect `src/services/git-context/GitContextCollector.ts` and `src/services/git-context/__tests__/GitContextCollector.spec.ts`.
- Confirm staged name-status parsing handles rename/copy entries.
- Confirm untracked files are expanded via `--untracked-files=all` and represented in context.
- Confirm binary files are summarized without binary payload.

### Pre-Submission Checklist

- [x] **Issue Linked**: This PR is linked to an approved GitHub Issue (see "Related GitHub Issue" above).
[...]

Issue #145, opened 2026-05-16
[ENHANCEMENT] Add AI commit message generation


Mirrowel:
### Problem (one or two sentences)

# Summary

Add AI-assisted Git commit message generation to Zoo Code, porting the existing Kilo Code commit generation workflow into the Zoo Code extension.

This would let users generate a conventional commit message directly from the VS Code Source Control UI by analyzing the current Git changes. The feature should prefer staged changes, since those are the exact changes the user intends to commit, and fall back to unstaged changes when nothing is staged.

## Problem

Writing accurate commit messages is a repeated workflow step for developers. Even when the code changes are already clear, summarizing the right scope, type, and intent can add friction before every commit.

Zoo Code already has the model configuration, prompt customization, and extension integration needed to help with this. Kilo Code previously had a commit message generation feature, but Zoo Code does not currently expose that workflow. 

Since Kilo was a fork of Roo, it is only Natural we give back to the Roo community.

## User Benefit

- Reduces friction in the commit workflow.
- Helps users produce clearer, more consistent commit messages.
- Reuses Zoo Code's existing provider and prompt customization capabilities.
- Improves daily developer experience from the VS Code Source Control panel.

### Context (who is affected and when)

-

## Proposed Solution

Port and adapt Kilo Code's commit message generation feature into Zoo Code with a Zoo-specific VS Code implementation.

The feature should:

- Add a Source Control command to generate a commit message with Zoo.
- Analyze staged Git changes first.
- Fall back to unstaged Git changes when no staged changes exist.
- Generate a conventional commit message using the user's configured AI provider.
- Allow users to customize the commit message prompt through the existing support prompt settings UI.
- Allow users to select a dedicated API profile for commit message generation, similar to prompt enhancement configuration.
- Write the generated message directly into the VS Code Git commit input box.

## Implementation Notes

This should be a focused feature addition around SCM commit message generation. The implementation should avoid using Zoo's normal file-access ignore filters for this workflow because Git state is the source of truth: if a user staged or modified a file for commit, that file should be considered for commit message generation.

The Zoo Code implementation should also be simpler than the legacy Kilo implementation where possible. The initial port can be adapted to the VS Code extension only, rather than preserving abstractions for non-VS Code IDE integrations that Zoo Code does not currently use.

## Acceptance Criteria

- Users can trigger commit message generation from the VS Code Source Control UI.
- The generated commit message is inserted into the Git commit input box.
- Staged changes are used when present.
[...]

PR #1227, opened 2026-08-12
[1/4] feat(git): add commit context collector

### LINKED ISSUE #282: [ENHANCEMENT] Add Git context collector for commit-message generation

Zoo needs a safe, scriptable way to inspect Git changes before AI commit generation can work.

Part of #145.

Developers may have staged, unstaged, renamed, copied, deleted, binary, or untracked files when preparing a commit from VS Code Source Control.

Zoo can collect commit-relevant Git context without using Zoo file ignore filters, preferring staged changes and falling back to unstaged changes only when nothing is staged.

Use stable, machine-readable Git output and path-safe parsing. Avoid shell-string interpolation for file paths.

### Request checklist

_No response_

---

- Staged changes are used when present.
- Unstaged and untracked changes are used only when no staged changes exist.
- Added, modified, deleted, renamed, copied, binary, and untracked files are represented safely.
- Zoo ignore filters are not applied to this Git-source-of-truth workflow.
- Unit tests cover parsing and staged/unstaged precedence.

Add a Git context collector independent of VS Code UI and AI generation. Prefer Git porcelain/name-status formats and test the parser directly.

Path parsing is the main risk, especially renamed/copied paths and paths containing whitespace or unusual characters.

Rafael-Silva-Oliveira:
### Related GitHub Issue

Closes: #282

Part of: #145 · **Stack 1 of 4** · Replaces the all-in-one #1218

### Description

Adds `getCommitContext()`, the Git-reading half of AI commit-message generation.
Nothing consumes it yet — that arrives in stack 2. This PR is self-contained and
introduces no user-facing behavior.

**Staged first, working tree as fallback.** Staged changes are what a commit will
actually contain, so they take priority. When nothing is staged the collector falls
back to the working tree, so a caller still has something to summarize before the
user has staged anything.

**The fallback reads `git status --short`, not a diff.** Untracked files appear in no
diff, so a diff-only fallback would silently omit brand-new files — usually the most
interesting thing in the change.

**No `HEAD` in the fallback diff.** `git diff` is used rather than `git diff HEAD`.
The index is known to be empty on that path so the two are equivalent, but `HEAD` does
not resolve in a repository without an initial commit, where it fails outright. This
is covered by a regression test.

Reuses the existing `checkGitInstalled`, `checkGitRepo`, and `truncateOutput` helpers
rather than adding new ones. `maxBuffer` is raised past Node's 1 MB `exec` default,
which real diffs routinely exceed.

Scope note: this deliberately shells out to `git diff` and passes the output through.
It does not parse rename/copy status codes or summarize binary files. #298 takes a
much more thorough approach to the same problem — see "Relationship to #298-#301" below.

### Test Procedure

`src/utils/__tests__/git.spec.ts` covers: staged path, working-tree fallback,
untracked-only repository with no initial commit, clean tree returning `null`, git not
installed, and not-a-repository.

There is also a test asserting the diff argument string contains no shell
metacharacters. That guards a real bug found during development: an earlier revision
used `:(exclude)` pathspecs to skip lockfiles, which unit tests happily passed because
they mock `exec` — but real git rejected the command, since `exec` runs through
`cmd.exe` on Windows and does not strip the single quotes those pathspecs require. The
mocked tests cannot validate git syntax, so that class of bug needs the guard.

Verified against real repositories, not just mocks:

- Repo with staged + unstaged + untracked changes → correct staged-first context.
- Repo with nothing staged → fallback context including the untracked file.
- Freshly `git init`'d repo with no commits → returns context instead of throwing.

Local checks: `pnpm lint`, `pnpm check-types` (11/11 packages), full `src` suite
(7379 passed, 37 skipped), `node scripts/find-missing-translations.js`.

### Pre-Submission Checklist

- [x] **Visual Snapshot** (UI changes only): not applicable — no UI in this PR.

### Visual Snapshots

Not applicable. This PR adds a service function with no UI.

### Documentation Updates

- [ ] Yes, documentation updates are required.

Nothing user-facing lands until stack 3.

### Additional Notes

**Review order:** 1 → #1228 → #1229 → #1230. Each targets `main` because GitHub cannot
base a cross-fork PR on another fork's branch, so later PRs show cumulative diffs until
their parents merge. **The Commits tab shows only that PR's own commit** — that is the
reviewable unit.

**Relationship to #298-#301.** @Mirrowel has an open stack covering this same feature,
untouched since 2026-06-30. I built this independently before finding it. Where they
overlap, that stack is more thorough: #298 is ~996 lines handling rename/copy status
codes, `-z` null-delimited parsing, synthetic diffs for untracked files, and binary-file
summarization. This is ~190 lines and does none of that.

The tradeoff is size against completeness. If #298 is revived I would rather see that
land, and I am happy to close this. If it stays stale, this is ready now.

## Summary by CodeRabbit

- **New Features**
  - Added structured Git context for staged and working-tree changes, including branch details and recent commit summaries.
  - Supports file statuses, renames, copies, unusual paths, binary files, and untracked files.
  - Displays untracked file contents as additions and keeps diff excerpts concise with output limits.

- **Bug Fixes**
  - Improved handling of clean repositories, missing Git installations, non-repository paths, repositories without commits, and command failures.
  - Provides clearer results when commit context cannot be generated.

### DIFF
--- a/src/utils/git.ts
+++ b/src/utils/git.ts
@@ -1,7 +1,7 @@
 import * as vscode from "vscode"
 import * as path from "path"
 import { promises as fs } from "fs"
-import { exec } from "child_process"
+import { exec, execFile } from "child_process"
 import { promisify } from "util"
 
 import type { GitRepositoryInfo, GitCommit } from "@roo-code/types"
@@ -10,7 +10,30 @@ import { truncateOutput } from "../integrations/misc/extract-text"
 
 const execAsync = promisify(exec)
 
+// Used for the commit-context commands: arguments are passed as an array, so no shell is
+// involved and paths never need quoting.
+const execFileAsync = promisify(execFile)
+
 const GIT_OUTPUT_LINE_LIMIT = 500
+
+// A line limit alone is not a bound: one minified or generated file can be a single line of
+// several megabytes. This caps the payload regardless of how it is distributed across lines.
+const GIT_OUTPUT_CHARACTER_LIMIT = 100 * 1024
+
+// Node's default `exec` buffer is 1MB, which real-world diffs routinely exceed.
+const GIT_DIFF_MAX_BUFFER = 10 * 1024 * 1024
+
+// A commit message needs to know what changed, not every line of how. One line of surrounding
+// context per hunk is enough to tell the model where an edit landed, and shrinking the prompt is
+// the one latency factor we control without affecting the model's output.
+const COMMIT_DIFF_ARGS = ["--unified=1"]
+
+// Untracked files have no diff, so their contents are read directly. Enough to tell the model
+// what a new file is for, not enough for a large one to crowd out the rest of the context.
+const UNTRACKED_FILE_BYTE_LIMIT = 2 * 1024
+const UNTRACKED_TOTAL_CHARACTER_LIMIT = 20 * 1024
+
+const RECENT_COMMIT_COUNT = 5
 
 /**
  * Extracts git repository information from the workspace's .git directory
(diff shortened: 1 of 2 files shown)


src/utils/git.ts
```diff
@@ -346,6 +357,52 @@ export async function getWorkingState(cwd: string): Promise<string> {
...
+		return truncateOutput(output, GIT_OUTPUT_LINE_LIMIT)
+	}
+
+	// Nothing staged - describe the working tree instead. `git status --short` is used rather than
+	// `--stat` here because it also lists untracked files, which no diff would show.
+	const { stdout: status } = await execAsync("git status --short", options)
+
+	if (!status.trim()) {
+		return null
+	}
```
taltas, line 575: Could we include bounded contents for untracked files rather than only their paths? Without the contents, an untracked-only change does not give the model enough information to write an accurate commit message.

taltas, line 575: Could we use NUL-delimited, machine-readable Git output and parse it into structured file entries? The current human-readable output can ambiguously represent renames, binary files, and filenames containing unusual characters.

src/utils/git.ts
```diff
@@ -346,6 +357,52 @@ export async function getWorkingState(cwd: string): Promise<string> {
...
+
+	const options = { cwd, maxBuffer: GIT_DIFF_MAX_BUFFER }
+
+	const { stdout: stagedSummary } = await execAsync("git diff --cached --stat", options)
+
+	if (stagedSummary.trim()) {
+		const { stdout: stagedDiff } = await execAsync(`git diff --cached ${COMMIT_DIFF_ARGS}`, options)
+		const output = `Staged changes:\n\n${stagedSummary.trim()}\n\n${stagedDiff.trim()}`
```
taltas: Could we also pass a finite character limit to `truncateOutput()`? A minified or generated file can contain a multi-megabyte single line that bypasses the current 500-line limit.

src/utils/git.ts
```diff
@@ -346,6 +357,52 @@ export async function getWorkingState(cwd: string): Promise<string> {
...
+	}
+
+	const isRepo = await checkGitRepo(cwd)
+	if (!isRepo) {
+		return null
+	}
+
+	const options = { cwd, maxBuffer: GIT_DIFF_MAX_BUFFER }
+
+	const { stdout: stagedSummary } = await execAsync("git diff --cached --stat", options)
```
taltas: Could we handle failures from the status and diff commands through a clear nullable or typed-error contract? Right now an expected Git failure, such as exceeding `maxBuffer`, rejects even though this function is documented as returning context or `null`.

src/utils/git.ts
Rafael-Silva-Oliveira: Untracked files now have a bounded head of their contents included, labelled --- New file: <path> ---.
only the first 2KB of each file is read (via fs.open + a single sized read, so an big untracked file costs nothing rather than being read fully and then sliced), 20KB in total across all of them, and anything containing a NUL byte in that head is skipped as binary

src/utils/git.ts
Rafael-Silva-Oliveira: GIT_OUTPUT_CHARACTER_LIMIT = 100 * 1024 is now passed alongside the line limit on every path

src/utils/git.ts
Rafael-Silva-Oliveira: now returns CommitContextResult: { ok: true, context } or { ok: false, reason } where reason is git-missing | not-a-repo | no-changes | failed, with the message attached on failed

Branch and recent commit subjects are collected here too (context for the prompt placeholders in the next PR), and they're deliberately failure-tolerant: git log fails outright on an unborn HEAD, so it must not take the whole collection down — git branch --show-current still works there, which is why it's used instead of rev-parse --abbrev-ref HEAD.

src/utils/git.ts
Rafael-Silva-Oliveira: the collector no longer reads human-readable output at all. The index is read with git diff --cached --name-status -z and the working tree with git status --porcelain=v1 -z --untracked-files=all, both parsed into typed GitFileChange entries (status, path, oldPath)

Mirrowel:
You might wanna focus on either staged or unstaged. No reason to include unstaged if you have staged.
But if you don't have staged - use unstaged the same way you'd do staged.

Rafael-Silva-Oliveira:
> You might wanna focus on either staged or unstaged. No reason to include unstaged if you have staged. But if you don't have staged - use unstaged the same way you'd do staged.

Yeah, focusing on staged only, if unstaged it will show a message "Stage the changes you want to commit, then generate the message."

If they were unstaged, for some reason ollama cloud models would stay stuck and not generate any message. However, local models would still be able to generate the commit message for all of the unstaged files (0 staged). So decided to ditch the whole unstage files and the models will now only look for the staged, throwing that message if none are staged (also may avoid behavior such as commiting a bunch of files that include different implementations, which should be separated in different commits/branches anyways). I have also added a cancel commit button + commit message generation timeout (changeable)

taltas:
As shown in the attached screenshot, commit-message generation fails with HTTP 400 `Stream must be set to true` because `OpenAiCodexHandler.completePrompt()` sends `stream: false` while the Codex subscription endpoint requires streaming.
Screenshot 2026-08-13 at 8 58 15 pm


Can this path use the existing streaming Codex transport and combine its text chunks into one final string?

src/utils/git.ts
```diff
@@ -346,6 +364,203 @@ export async function getWorkingState(cwd: string): Promise<string> {
...
+
+	try {
+		const staged = parseNameStatus(await runGit(["diff", "--cached", "--name-status", "-z"], cwd))
+
+		if (staged.length > 0) {
+			const diff = await runGit(["diff", "--cached", ...COMMIT_DIFF_ARGS], cwd)
+			return { ok: true, context: await buildContext(cwd, staged, diff) }
+		}
+
+		// Only the index is described, so an empty one has nothing to summarize. Whether the
```
taltas: Could this empty-index path return bounded working-tree context for unstaged and untracked changes, as issue #282 requires?

src/utils/git.ts
```diff
@@ -346,6 +364,203 @@ export async function getWorkingState(cwd: string): Promise<string> {
...
+	if (!(await checkGitInstalled())) {
+		return { ok: false, reason: "git-missing" }
+	}
+
+	if (!(await checkGitRepo(cwd))) {
+		return { ok: false, reason: "not-a-repo" }
+	}
+
+	try {
+		const staged = parseNameStatus(await runGit(["diff", "--cached", "--name-status", "-z"], cwd))
```
taltas: Would you make rename and copy detection explicit in this Git command so classification does not vary with each user's Git configuration?

Mirrowel:
> As shown in the attached screenshot, commit-message generation fails with HTTP 400 `Stream must be set to true` because `OpenAiCodexHandler.completePrompt()` sends `stream: false` while the Codex subscription endpoint requires streaming. Screenshot 2026-08-13 at 8 58 15 pm
> 
> Can this path use the existing streaming Codex transport and combine its text chunks into one final string?

Is this not using existing profiles? Seems strange.

Rafael-Silva-Oliveira:
> > As shown in the attached screenshot, commit-message generation fails with HTTP 400 `Stream must be set to true` because `OpenAiCodexHandler.completePrompt()` sends `stream: false` while the Codex subscription endpoint requires streaming. Screenshot 2026-08-13 at 8 58 15 pm
> > Can this path use the existing streaming Codex transport and combine its text chunks into one final string?
> 
> Is this not using existing profiles? Seems strange.

@taltas This indeed uses the already existing profiles

I dont use codex, how are you setting it up?

src/utils/git.ts
Rafael-Silva-Oliveira: Done. Empty index falls back to the working tree now. Staged still wins outright when anything is staged, so the two never mix.

src/utils/git.ts
Rafael-Silva-Oliveira: Added. Contents are inlined now, bounded three ways: 10 files max, 8KB per file, NUL byte marks it binary instead of pasting it in. Anything past the limit still gets listed by path.

src/utils/git.ts
Rafael-Silva-Oliveira: It was inheriting `diff.renames`, so a rename showed up as delete + add for anyone with it off. Passes `--find-renames --find-copies` explicitly on both calls now.

Rafael-Silva-Oliveira:
Fixed. It was sending `stream: false` to an endpoint that only takes streaming. Now runs the same streaming path as chat and joins the text chunks, so it also picks up the OAuth retry and SSE fallback. Reasoning chunks dropped so they can't land in a commit message.

The spec asserted `stream: false`, so the test was holding the bug in place. Inverted now. Also `abortSignal` was never wired to the request here, so cancelling did nothing. Fixed too.

Issue #1242, fix in #1243.

taltas (dismissed):

src/utils/git.ts
Rafael-Silva-Oliveira: Done in 518a8fb2c. When the index is empty it now describes the working tree, unstaged and untracked alike, with the same bounds as the staged path.

src/utils/git.ts
Rafael-Silva-Oliveira: Done. Rename and copy detection is passed explicitly on the command now, so it does not depend on whatever the user has in their git config.

src/utils/git.ts
```diff
@@ -346,6 +375,270 @@ export async function getWorkingState(cwd: string): Promise<string> {
...
+ * Bounded on every axis that can grow without limit - how many files are read, how much of each is
+ * read, and whether the bytes are text at all - because an untracked listing is whatever happens to
+ * be sitting in the working directory.
+ */
+async function readUntrackedFiles(cwd: string, files: GitFileChange[]): Promise<string> {
+	const untracked = files.filter((file) => file.status === "untracked")
+	const blocks: string[] = []
+
+	for (const file of untracked.slice(0, UNTRACKED_FILE_LIMIT)) {
+		blocks.push(`--- /dev/null\n+++ b/${file.path}\n${await readUntrackedFile(cwd, file.path)}`)
```
edelauna: The untracked file body is inlined here with no `+` line prefixes, so it isn't a valid unified-diff hunk. A new file whose contents contain markdown list items (`- x`), YAML front-matter (`---`), or any diff-like line is read by the model as deletions/headers, and the generated message describes changes that never happened. Prefix each content line of `readUntrackedFile`'s output with `+`. (Distinct from the raw-path escaping already flagged on this line.)

edelauna (changes requested):
Nice excited to see this - noticed a small bug in how the diff is processed.

src/utils/git.ts
Rafael-Silva-Oliveira: You're right, fixed in 93e35ee2d. Content lines are prefixed with `+` now, with a test pinning the exact block for a markdown file full of `---` and `- item` lines. Left the binary/unreadable markers unprefixed since they annotate the block rather than claim to be its content.

PR #1228, opened 2026-08-12
[2/4] feat(commit-message): add prompt template and generator service

Rafael-Silva-Oliveira:
### Related GitHub Issue

Closes: #283
Closes: #284
Closes: #285
Closes: #290

Part of: #145 · **Stack 2 of 4** · Depends on #1227 · Replaces the all-in-one #1218

### Description

Adds the model-facing half of commit-message generation: a customizable prompt, a
dedicated model setting, and the service that turns git context into a message. Still
no command or picker UI, so nothing is reachable by a user yet.

**The prompt is editable, for free.** `COMMIT_MESSAGE` is registered in
`support-prompt.ts` alongside `ENHANCE`. The Prompts settings tab iterates the
support-prompt registry, so the template immediately gets a textarea, a reset button,
and `customSupportPrompts` persistence with no new UI code — which is why #290 is
closed here rather than in stack 4.

**The default prompt tells the model to account for every changed file.** That
instruction is load bearing rather than decorative. Without it, a model given four
staged files across three concerns will confidently describe the largest one and
silently drop the rest. Observed repeatedly in testing before the wording was added.

**`commitMessageApiConfigId` mirrors `enhancementApiConfigId` exactly**, including the
`listApiConfigMeta.find(...)` guard before `getProfile()` — that call throws on an
unknown id, so a profile deleted after being selected must fall back to the active
configuration rather than break the button.

**Response cleanup.** Models wrap answers in code fences and quotes despite being told
not to, so the result is stripped before use.

Progress is reported at `ProgressLocation.Window`. `SourceControl` was the obvious
choice but silently drops the title, leaving an unlabelled spinner. No location renders
a cancel button that would do anything useful — see Additional Notes.

### Test Procedure

`src/services/commit-message/__tests__/generateCommitMessage.spec.ts` covers: writing
the cleaned message to the input box, fence/quote stripping, using the dedicated
profile when configured, falling back when the configured profile no longer exists,
selecting the right repository from `SourceControl.rootUri` in a multi-root workspace,
the no-changes path leaving the input box untouched, the missing-git-extension path,
and generation failures surfacing rather than throwing.

Also asserts progress is reported at a location that actually renders the title, so a
regression back to `SourceControl` cannot silently produce an unlabelled spinner.

Translations for all 17 locales are included for the new keys, not left as English
fallbacks — `find-missing-translations.js` gates on this.

Local checks: `pnpm lint`, `pnpm check-types` (11/11 packages), full `src` suite
[...]

PR #1229, opened 2026-08-12
[3/4] feat(scm): add Source Control button for commit message generation

### LINKED ISSUE #286: [ENHANCEMENT] Register Zoo commit-message command in VS Code SCM

Users need to trigger commit-message generation from the VS Code Source Control UI.

Part of #145.

VS Code exposes SCM menu contribution points such as scm/title and scm/input for source-control actions.

Zoo contributes a Generate Commit Message command for Git SCM providers.

The command should appear only for Git SCM providers.

### Request checklist

_No response_

---

- A Zoo command id is added for commit-message generation.
- The command is contributed in src/package.json.
- Menu entries are added for scm/title and scm/input.
- The menu entries use a Git-specific when clause.
- The command is registered by the extension and can be invoked in tests.

Implement the contribution point and command registration without requiring the full generation workflow in this issue.

Incorrect menu contribution placement could make the command noisy or invisible in multi-SCM workspaces.

### LINKED ISSUE #287: [ENHANCEMENT] Resolve Git repository from SCM command context

Multi-root VS Code workspaces may have multiple Git repositories, so Zoo must target the correct repository.

Part of #145.

A user can invoke the commit-message command from a specific Source Control repository or input box.

When invoked from Source Control, Zoo finds the matching Git repository for that SCM context.

Use the built-in VS Code Git extension API where possible and avoid guessing when the target repository is ambiguous.

### Request checklist

_No response_

---

- The built-in vscode.git extension is activated or accessed safely.
- The repository matching the SCM argument rootUri is selected when available.
- Fallback occurs only when the target is unambiguous.
- A clear error is shown when no Git repository is available.
- Unit tests mock multiple repositories and verify selection behavior.

Add a small repository resolution helper used by the SCM command handler before generation or input-box mutation.

Wrong repository selection would write messages into the wrong commit input in multi-root workspaces.

### LINKED ISSUE #288: [ENHANCEMENT] Insert generated message into Git commit input box

A generated commit message is only useful if it lands in the user's actual Git commit message field.

Part of #145.

VS Code SCM exposes the Git commit message field through the SourceControl input box.

Zoo writes the generated commit message into the selected Git repository's commit input box.

Zoo must not commit automatically.

### Request checklist

_No response_

---

- The generated message is assigned to the matching repository inputBox.value.
- Behavior is defined for an existing non-empty commit input.
- No commit is created automatically.
- Tests verify the selected repository input is updated and unrelated repositories are untouched.

Use the repository resolved from SCM context and update only its SourceControl input box.

Silent overwrite of a user's existing commit text would be surprising; the UX should be explicit.

Rafael-Silva-Oliveira:
### Related GitHub Issue

Closes: #286
Closes: #287
Closes: #288

Part of: #145 · **Stack 3 of 4** · Depends on #1227, #1228 · Replaces the all-in-one #1218

### Description

Wires the generator to a button in the Source Control title bar. This is the PR where
the feature becomes reachable: click the Zoo Code icon next to the commit box and a
message appears in it.

**Repository resolution.** The `scm/title` menu passes the clicked `SourceControl`,
whose `rootUri` is matched against the git extension's repository list. That is what
makes the button correct in a multi-root workspace rather than always targeting the
first repository. Falls back to the first repository when no match is found.

**Uses the registered provider, not the visible one.** `getVisibleProviderOrLog` would
return nothing when the Zoo Code sidebar is closed — which is the common case when
someone is working in the Source Control panel. The command takes the provider passed
into `registerCommands` instead, so the button works regardless of sidebar state.

**The icon reuses the existing zebra mark** (`panel_light.png` / `panel_dark.png`),
already shipped for the tab icon, so no new art is added.

**One change outside the feature.** That themed icon forced a fix in `packages/build`:
`commandsSchema.icon` was declared `z.string().optional()`, accepting only a codicon
string. A `{light, dark}` pair would have thrown in `contributesSchema.parse()` during
the nightly manifest build. Worth flagging because that path only runs under
`vsix:nightly` — the failure would not appear in normal development or in most CI runs,
and would have surfaced as a broken nightly.

### Test Procedure

`src/activate/__tests__/registerCommands.spec.ts` gains a test asserting the command
forwards the clicked `SourceControl` to the generator and uses the registered provider
rather than the visible instance.

Packaging verified with `pnpm --filter ./src vsix`, then the resulting VSIX inspected to
confirm both icon files ship at the paths the manifest references and that the packaged
`contributes` block contains the command and the `scm/title` menu entry. That is the
check that would have caught the `packages/build` schema problem.

Manual verification in a scratch repository:

1. Open Source Control — the Zoo Code icon appears in the title bar, in both light and
   dark themes.
2. Click it — the message lands in the commit input box.
3. Unstage everything — click again, fallback path still produces a message.
4. Commit everything — click again, "No changes to commit", input box untouched.

Local checks: `pnpm lint`, `pnpm check-types` (11/11 packages), full `src` suite
(7389 passed, 37 skipped), `node scripts/find-missing-translations.js`, `pnpm knip`.

### Pre-Submission Checklist

- [ ] **Visual Snapshot** (UI changes only): see below.

### Visual Snapshots

The new UI is a VS Code menu contribution, not a webview component, so the Playwright CT
harness cannot capture it — `webview-ui` snapshots only cover the webview. Screenshot
below is a review aid.

**Generated message in the commit input box.** Four files staged across three concerns;
note `test.py` is called out as unrelated rather than forced under `feat(auth)`:

> [!NOTE]
> _Screenshot to be attached._

### Documentation Updates

- [x] Yes, documentation updates are required.

This is the PR that makes the feature visible, so the docs repo will want a short page
on the Source Control button.

### Additional Notes

Aligns with the roadmap's **Enhanced User Experience** goal — it removes a small,
repeated friction point and gives Zoo Code a presence in a panel where it previously had
none.

**Review order:** #1227 → #1228 → 3 → #1230. All target `main` because GitHub cannot base a
cross-fork PR on another fork's branch, so this diff appears cumulative until the
parents merge. **The Commits tab shows only this PR's own commit.**

## Summary by CodeRabbit

- **New Features**
  - Generate commit messages directly from the Source Control view using repository changes and history.
  - Preserve existing drafts and receive clear status or error notifications.
  - Choose a dedicated commit-message configuration, customize the generation prompt, and configure a 10–600 second timeout.
  - Command icons support light and dark theme-specific images.

- **Localization**
  - Added translated commit-message commands, prompts, statuses, and errors across supported languages.

- **Tests**
  - Expanded coverage for repository detection, configuration, generation, persistence, timeouts, and theme-aware icons.


src/services/commit-message/index.ts
```diff
@@ -0,0 +1,137 @@
...
+
+	if (clickedPath) {
+		const match = repositories.find((repo) => repo.rootUri.fsPath === clickedPath)
+
+		if (match) {
+			return match
+		}
+	}
+
+	return repositories[0]
```
taltas: What should happen when the clicked root has no match, or when multiple repositories exist without a supplied target? Would returning no repository be safer than silently operating on `repositories[0]`?

src/services/commit-message/index.ts
```diff
@@ -0,0 +1,137 @@
...
+		// an `abortSignal`, but nearly every provider (Ollama included) ignores the argument, so
+		// the underlying request cannot actually be interrupted today.
+		await vscode.window.withProgress(
+			{
+				location: vscode.ProgressLocation.Window,
+				title: t("common:info.commit_message_generating"),
+			},
+			async () => {
+				const message = await singleCompletionHandler(configToUse, prompt)
+				repository.inputBox.value = cleanCommitMessage(message)
```
taltas: How should we protect an existing draft or edits made while generation is running? This unconditional assignment can silently destroy user-written commit text.

src/package.json
```diff
@@ -265,6 +274,13 @@
 					"group": "overflow@2",
 					"when": "activeWebviewPanelId == zoo-code.TabPanelProvider"
 				}
+			],
+			"scm/title": [
```
taltas, line 284: Should this also add the intended SCM input-box contribution, or should that acceptance criterion remain open? Issue #286 requires both locations, while this manifest currently contributes the command only to `scm/title`.

src/shared/support-prompt.ts
```diff
@@ -240,6 +241,19 @@ Please provide:
...
+
+Follow the Conventional Commits specification: \`type(scope): description\`, where type is one of feat, fix, docs, style, refactor, perf, test, build, ci, chore, or revert. Keep the description under 72 characters and in the imperative mood.
+
+Account for every changed file. The subject line describes the change as a whole, so do not let the largest file speak for the rest. When the changes touch more than one file or concern, follow the subject with a blank line and one \`- \` bullet per distinct change, naming the file or area it affects. Use a subject line on its own only when it genuinely covers everything that changed.
+
+If the changes are unrelated to one another, say so plainly rather than inventing a single scope that hides some of them.
+
+Reply with ONLY the commit message - no explanation, no markdown code fences, no surrounding quotes.
+
+\${gitContext}`,
```
taltas: Can we mark and delimit `${gitContext}` as untrusted data rather than placing it directly beside model instructions? Repository content can otherwise inject directions that override the commit-message prompt.

packages/build/src/types.ts
```diff
@@ -31,7 +31,8 @@ const commandsSchema = z.array(
 		command: z.string(),
 		title: z.string(),
 		category: z.string().optional(),
```
taltas, line 33: Would a package test with both `light` and `dark` icon paths make sense here? The current string-codicon fixture would not catch this schema being narrowed again.

Rafael-Silva-Oliveira: packages/build/src/__tests__/types.test.ts covers a codicon string, a {light, dark} pair, and an incomplete pair that should be rejected. It reaches the commands schema through contributesSchema.shape.commands, so the test stays focused on the icon field without needing a whole valid contributes object

src/package.json
Rafael-Silva-Oliveira: Added

src/package.json
```diff
@@ -265,6 +280,24 @@
...
```
taltas, line 284: Could you add the required SCM input-box action here? Can the manifest test cover it too?

src/core/webview/ClineProvider.ts
```diff
@@ -2852,6 +2854,7 @@ export class ClineProvider
 			customModePrompts: stateValues.customModePrompts ?? {},
 			customSupportPrompts: stateValues.customSupportPrompts ?? {},
 			enhancementApiConfigId: stateValues.enhancementApiConfigId,
+			commitMessageApiConfigId: stateValues.commitMessageApiConfigId,
```
taltas, line 2859: Why is the persisted timeout omitted from the state returned here? Could you return it and add set/unset state-path tests so configured values reach generation?

src/package.json
Rafael-Silva-Oliveira: Added, with a manifest test. Left the stop command off it since that menu holds one action rather than the pair the title bar swaps between, so there's no slot to pin. Can add it if you'd rather have both.

src/core/webview/ClineProvider.ts
Rafael-Silva-Oliveira: It was never returned from either state path, so the generator always read `undefined` and fell back to the 60s default, so it was doing nothing. Returned now with set/unset tests on both paths

Moved the fix down into this PR too, it was landing in 4/4, which meant 3/4 shipped a setting that couldn't work. `config.spec.ts` stubs the provider so it never caught this, test had to go at the ClineProvider level.

src/services/commit-message/index.ts
Rafael-Silva-Oliveira: Handled. It only operates on a repository it actually matched, rather than falling back to repositories[0] when the clicked root does not match or there are several.

src/services/commit-message/index.ts
Rafael-Silva-Oliveira: Fixed. The draft is captured before the request and compared after it, so if the box changed while generation was running the user's text wins and nothing is overwritten.

src/package.json
Rafael-Silva-Oliveira: Added in 119eccf5e, see src/package.json line 296. Both locations are contributed now.

src/shared/support-prompt.ts
Rafael-Silva-Oliveira: Done. gitContext is gone, replaced by delimited branch, recent_commits, changed_files and diff blocks, with the prompt saying they hold repository data and are never to be acted on. The fields are also neutralized so their content cannot close a block early.

packages/build/src/types.ts
Rafael-Silva-Oliveira: Added, packages/build/src/__tests__/types.test.ts. Three cases: a codicon string, a light and dark pair, and an icon pair missing a theme which should be rejected. That last one is what would catch the schema being narrowed again.

src/package.json
Rafael-Silva-Oliveira: Both done in 119eccf5e. The scm/inputBox action is in the manifest and contributions.spec.ts covers it at lines 49 and 59.

src/core/webview/ClineProvider.ts
Rafael-Silva-Oliveira: Good catch, it was only missing from that path. It is returned from both now, ClineProvider.ts lines 2465, 2625 and 2860, with set and unset tests for each in ClineProvider.spec.ts around line 1271.

Issue #292, opened 2026-05-24
[ENHANCEMENT] Add privacy-safe telemetry for commit-message generation


edelauna:
### Problem (one or two sentences)

Maintainers need to understand adoption and failure patterns for commit-message generation without collecting source content.

Part of #145.

### Context (who is affected and when)

Git diffs, file paths, branch names, and generated commit messages may contain sensitive project information.

### Desired behavior (conceptual, not technical)

Zoo records privacy-safe commit-message generation events.

### Constraints / preferences (optional)

Telemetry must not include branch names, file paths, diffs, or generated commit text.

### Request checklist

- [x] I've searched existing Issues and Discussions for duplicates
- [x] This describes a specific problem with clear context and impact

---
Optional (for contributors): You can stop here if you're just proposing the improvement.

### Acceptance criteria (optional)

- A telemetry event type is added for commit-message generation.
- Successful generation is captured.
- Failure or cancellation tracking is either added with coarse categories or explicitly deferred.
- No branch name, file path, diff content, or generated message content is sent.
- Type checks or tests cover the new telemetry event.

### Proposed approach (optional)

Add a minimal event aligned with existing Zoo telemetry patterns and keep payload fields content-free.

### Trade-offs / risks (optional)

Overly detailed telemetry could leak source or workflow information.

Issue #289, opened 2026-05-24
[ENHANCEMENT] Add progress, cancellation, and error UX for commit generation


edelauna:
### Problem (one or two sentences)

AI commit-message generation can take time and can fail.

Part of #145.

### Context (who is affected and when)

Users trigger generation from Source Control and need feedback while the operation runs.

### Desired behavior (conceptual, not technical)

Zoo shows cancellable Source Control progress and concise error feedback for failures.

### Constraints / preferences (optional)

Avoid noisy notifications for expected or non-fatal states.

### Request checklist

- [x] I've searched existing Issues and Discussions for duplicates
- [x] This describes a specific problem with clear context and impact

---
Optional (for contributors): You can stop here if you're just proposing the improvement.

### Acceptance criteria (optional)

- Generation uses vscode.window.withProgress with ProgressLocation.SourceControl.
- User cancellation aborts generation where practical.
- Timeout or provider failure produces a clear error message.
- Non-fatal Git warnings are surfaced sparingly and do not spam users.
- Tests cover cancellation/failure handling where practical.

### Proposed approach (optional)

Wrap the generation flow in VS Code progress and route failures through concise user-facing messages.

### Trade-offs / risks (optional)

Long-running generation without cancellation can make the extension feel stuck.

PR #1230, opened 2026-08-12
[4/4] feat(settings): add commit message model picker

### LINKED ISSUE #291: [ENHANCEMENT] Add commit-message API profile settings UI

Users need a visible way to configure the dedicated API profile used for commit-message generation.

Part of #145.

Prompt-related quick actions are configured from Settings > Prompts.

Settings lets users choose either the current API config or a saved API config for commit-message generation.

The control must save through the normal Settings save flow and must not reuse enhancementApiConfigId.

### Request checklist

_No response_

---

- The selector lists existing API profiles.
- A Use current config option is available.
- Selection persists as commitMessageApiConfigId.
- The control binds through SettingsView cached state.
- Tests cover rendering and state update where practical.

Mirror the enhancement API profile selector behavior, but store the value in commitMessageApiConfigId.

Using the wrong setting key would couple commit generation to prompt enhancement unexpectedly.

Rafael-Silva-Oliveira:
### Related GitHub Issue

Closes: #291

Part of: #145 · **Stack 4 of 4** · Depends on #1227, #1228, #1229 · Replaces the all-in-one #1218

### Description

Adds the **Commit Message Model** picker to Settings → Providers, so the profile used
for commit messages can be chosen without editing configuration by hand. The setting
itself (`commitMessageApiConfigId`) landed in stack 2; this is the control that sets it.

**Why Providers rather than Prompts.** The enhancement equivalent lives in the Prompts
tab, so there is a reasonable argument either way. This sits in Providers because it
selects a provider profile, directly below the profile it relates to — the prompt text
lives in Prompts, the model choice lives with the models. Happy to move it if you would
rather keep the two pickers together.

**Binds to `cachedState`, saves with the Save button**, per the SettingsView pattern in
`AGENTS.md` — not an immediate-persist `postMessage`. The value is included in the
`updateSettings` payload and round-trips through `getStateToPostToWebview`, so the
selection survives a reload rather than visually reverting.

**Accessibility.** A sibling `<label>` does not name a Radix trigger, so the label and
trigger are linked with `aria-labelledby` and the relationship is covered by a test
rather than assumed.

### Test Procedure

`CommitMessageModelSelect.spec.tsx` covers: rendering the profile list alongside the
fallback option, the `"-"` sentinel mapping to `""`, showing a saved profile, storing a
selected id, and the label-to-trigger accessibility relationship.

`SettingsView.spec.tsx` gains a test asserting the setting survives the `cachedState`
round trip into the `updateSettings` payload — the failure mode being a control that
renders and appears to save but is dropped from the payload.

Translations for all 17 locales are included, not left as English fallbacks.

Manual verification: set the picker to a specific profile → Save → reload the window →
selection persists → generating a message uses the chosen provider.

Local checks: `pnpm lint`, `pnpm check-types` (11/11 packages), full `webview-ui` suite
(1679 passed), full `src` suite (7389 passed, 37 skipped),
`node scripts/find-missing-translations.js`.

### Pre-Submission Checklist

- [ ] **Visual Snapshot** (UI changes only): see below.

### Visual Snapshots

Check #1218

### Documentation Updates

- [x] Yes, documentation updates are required.

Worth a line in the docs alongside the Source Control button from stack 3, noting that a
small fast profile is usually the right choice here.

### Additional Notes

This completes the stack. Of the #145 sub-issues, #289 (progress/cancellation UX) is only
partially addressed — progress and error handling are done, but there is no cancel
button, because no progress location renders one that would work while nearly every
provider ignores `completePrompt`'s abort signal. #292 (telemetry) is untouched. Both
remain open for follow-up.

**Review order:** #1227 → #1228 → #1229 → 4. All target `main` because GitHub cannot base
a cross-fork PR on another fork's branch, so this diff appears cumulative until the
parents merge. **The Commits tab shows only this PR's own commit.**

## Summary by CodeRabbit

- **New Features**
  - Generate commit messages directly from the Source Control view.
  - Preserve drafts and provide progress, timeout, cancellation, and error feedback.
  - Configure a dedicated model or use the active configuration, with a 10–600 second timeout.
  - Support light and dark command icons.
- **Localization**
  - Added translated commit-message commands, settings, prompts, and status messages.
- **Bug Fixes**
  - Improved handling of staged, unstaged, untracked, renamed, binary, and unusually named files.
  - Added safer repository detection and configuration fallback behavior.


webview-ui/src/components/settings/CommitMessageModelSelect.tsx
```diff
@@ -0,0 +1,65 @@
...
+		<SearchableSetting
+			settingId="commit-message-model"
+			section="providers"
+			label={t("settings:providers.commitMessageModel.label")}
+			className="mt-4">
+			<label id={LABEL_ID} className="block font-medium mb-1">
+				{t("settings:providers.commitMessageModel.label")}
+			</label>
+			<Select
+				value={commitMessageApiConfigId || USE_CURRENT_CONFIG}
```
taltas: What should the picker display when `commitMessageApiConfigId` no longer exists in `listApiConfigMeta`? Can it use the fallback value so Radix always has a matching item to display?

webview-ui/src/i18n/locales/ca/settings.json
```diff
@@ -356,6 +356,11 @@
 	},
 	"providers": {
 		"providerDocumentation": "Documentació de {{provider}}",
+		"commitMessageModel": {
+			"label": "Model per als missatges de comissió",
```
taltas: Should this use “missatges de commit” instead of “missatges de comissió”? “Comissió” means “commission,” not the Git operation.

webview-ui/src/components/settings/__tests__/SettingsView.spec.tsx
```diff
@@ -373,6 +373,30 @@ describe("SettingsView - Sound Settings", () => {
 		)
 	})
 
+	it("includes the commit message model in the saved settings", async () => {
```
taltas: Would this test be stronger if it selected a different profile before saving? That would cover the complete user-selection → cached-state → `updateSettings` path rather than only resaving the initial value.

webview-ui/src/i18n/locales/ca/settings.json
Rafael-Silva-Oliveira: fixed

Rafael-Silva-Oliveira:
Ready for review

webview-ui/src/components/settings/CommitMessageModelSelect.tsx
Rafael-Silva-Oliveira: Fixed. A saved id that matches nothing in listApiConfigMeta falls back to the use current option, so Radix always has a matching item and the trigger never renders blank.

webview-ui/src/i18n/locales/ca/settings.json
Rafael-Silva-Oliveira: You are right, that was commission rather than the git operation. Changed to missatges de commit.

webview-ui/src/components/settings/__tests__/SettingsView.spec.tsx
Rafael-Silva-Oliveira: Agreed, done. It selects a different profile before saving now, so it covers selection through cached state to updateSettings rather than just resaving the initial value.

PR #1243, opened 2026-08-14
fix(openai-codex): complete prompts over the streaming transport

### LINKED ISSUE #1242: [BUG] OpenAI Codex completePrompt sends stream:false and fails with HTTP 400
### What happened

Any one-shot completion through the OpenAI Codex provider fails with HTTP 400 `Stream must be set to true`.

`OpenAiCodexHandler.completePrompt()` posts its own request body with `stream: false`, but the Codex subscription endpoint (`/backend-api/codex/responses`) only accepts streaming requests and rejects the call outright.

This affects every caller that goes through `singleCompletionHandler` — commit-message generation, prompt enhancement and context condensing — while normal chat is unaffected, since `createMessage` streams and hard-codes `stream: true`.

Reported by @taltas in https://github.com/Zoo-Code-Org/Zoo-Code/pull/1227#issuecomment-5288200138, with a screenshot of the 400.

### Steps to reproduce

1. Sign in with an OpenAI Codex (ChatGPT subscription) profile.
2. Trigger anything that uses a single completion rather than a chat stream — e.g. Enhance Prompt.
3. The request fails with `400 ... Stream must be set to true`.

### Expected behavior

A one-shot completion returns text, like every other provider.

### Additional context

Two related problems in the same area:

- The provider spec asserted `expect(body.stream).toBe(false)`, so the test suite pinned the bug in place rather than catching it.
- `executeRequest` always creates its own `AbortController` and never links `metadata?.abortSignal`, so cancellation never reaches the wire on this provider.

The fix is to run the existing streaming transport and join the text chunks, which also inherits the OAuth refresh-and-retry and the SDK→SSE fallback instead of duplicating them.

Rafael-Silva-Oliveira:
### Related GitHub Issue

Closes: #1242

### Description

`completePrompt()` built its own request body with `stream: false`, which the Codex subscription endpoint rejects with `400 Stream must be set to true`. Rather than flip the flag and hand-roll SSE parsing, this runs the request through `handleResponsesApiMessage` — the path `createMessage` already uses — and joins the text chunks into one string.

Going through the existing path means the OAuth refresh-and-retry, the SDK→SSE fallback, the Luna body and the service tier all apply without being duplicated. The hand-built body is gone, along with the now-redundant `notAuthenticated` check.

Only `text` chunks are accumulated. Reasoning is deliberately dropped — commit-message generation writes this result straight into the Source Control input box.

Two things worth flagging for review:

- **The spec asserted the bug.** `expect(body.stream).toBe(false)` meant the test suite held the broken behavior in place. That assertion is inverted, and the two `completePrompt` tests now drive real streams.
- **`abortSignal` was dead on this provider.** `executeRequest` always made its own `AbortController` and never linked `metadata?.abortSignal`, so cancelling never reached the wire. The caller's signal is now linked to that controller, which also makes the commit-message stop button actually cancel the request.

Behavior change worth knowing: the old code returned the *first* `output_text` block; concatenating returns all of them. For a one-shot completion that is more correct, and `cleanCommitMessage` already post-processes the result.

An SSE-path failure now produces two telemetry events, since `makeCodexRequest` already captures one of its own. Left alone as it is not worth restructuring the error handling for.

#### Follow-up from review (979269b01)

Two defects @taltas spotted in the above, both fixed:

- **The stream could be replayed after it had already produced output.** The SDK stream and the loop consuming it sat in the same `try`, so an error raised part-way through was handled as "the SDK could not be used at all" and the whole request was replayed over SSE — appending a second generation to text the caller already had. `executeRequest` now tracks whether an SDK event has arrived and only falls back while none has. The flag is set *before* `processEvent` runs, since that mutates response state too, so a throw from it must not replay either. This fixes the chat path as well, since both go through `executeRequest`.
- **An abort resolved as if it had completed.** Both transports end quietly on cancellation — they break out of their loops rather than throwing — so `completePrompt` returned whatever partial text had arrived and callers read a cancelled generation as a finished one. It now rejects with an `AbortError`, and a cancellation passes straight through the catch rather than being captured by telemetry or relabelled as a completion error, since stopping is the caller's own doing.

#### Second review round (c68dae15f)

CodeRabbit caught the two remaining ways back into a request the service had already accepted.

- **The OAuth retry loop was the other replay route.** Closing the SSE fallback left `handleResponsesApiMessage` free to refresh the token and resend after a mid-stream error that reads as an auth failure, so `completePrompt` could still concatenate two generations and the chat path could repeat streamed effects. I had flagged this as knowingly left open; it is now closed. `sawSdkEvent` moved onto the handler as `sawSdkEventInCurrentResponse`, reset per request alongside the other response state, and the retry is skipped once it is set. A refresh *before* any event still retries exactly as before, which is what that loop exists for. Went with a plain field rather than the suggested typed marker, since it carries the same information without an error type to thread across the generator boundary and match on.
- **An abort still reached the SSE fallback.** The SDK rejects when the caller cancels, which read as a transport failure, so it spent a second request on an already-aborted signal and reported the cancellation as a connection error. The fallback now rethrows instead.

#### Third review round (bb2676a81)

A review note about the cancellation tests turned out to have a real bug behind it.

- **The abort result was not consistent.** A stream that ended quietly threw an `AbortError`, but a transport that rejected on abort had its own error passed straight through the catch. What a cancelled `completePrompt` rejected with therefore depended on how far the request had got, which is nothing a caller can key off. An abort is now restated as an `AbortError` unless it already is one. Both cancellation tests fail without the change and pass with it.
- **The in-flight case had no coverage.** Cancelling *after* the request is away but before any event is the case that only works if the caller's signal is genuinely linked to the internal controller, so it is what would catch that link breaking. It has to wait for the request to actually be in flight before aborting, since the token lookup and the listener are both async and a synchronous abort fires before anything is listening.

One suggestion from that round was skipped: asserting that an already-aborted signal never calls `create`. That contradicts the implementation and an existing passing test. `executeRequest` aborts its internal controller and *then* still calls `create` with that aborted signal, which is what makes the SDK reject it.

### Test Procedure

`npx vitest run api/providers/__tests__/openai-codex.spec.ts api/providers/__tests__/openai-codex-native-tool-calls.spec.ts` — 54 pass. The full provider suite passes at 1267. `check-types` and `lint` are clean.

New coverage: joining multiple text deltas, reasoning excluded from the result, tool calls and usage excluded, OAuth retry reaching `completePrompt`, error wrapping when both transports fail, and from the review round — rejection on mid-stream cancellation and on an already-aborted signal, plus a mid-stream SDK failure asserting that SSE is never reached, no refreshed-token retry after the SDK has emitted, the pre-event refresh path left untouched, and no SSE fallback when the SDK fails because the caller aborted.

The pre-existing tests that force `create` to reject *before* any event and then assert the SSE path still pass, which is what proves the fallback only closed for the mid-stream case.

Manually: with a Codex profile selected, run Enhance Prompt or generate a commit message and confirm text comes back instead of a 400. Pressing stop mid-generation leaves the input box as it was, with no error notification.

### Pre-Submission Checklist

- [x] **Scope**: My changes are focused on the linked issue.
- [x] **Documentation Impact**: No user-facing documentation change needed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming response handling for more reliable text generation.
  * Prevented duplicate output when retrying after partial responses.
  * Improved cancellation behavior, including requests aborted before or during processing.
  * Added safer fallback handling when streaming responses fail.
  * Prevented refusal messages from appearing in one-shot completions while preserving them in chat streams.
* **Tests**
  * Expanded coverage for streaming, retries, cancellation, fallback responses, service tiers, refusal handling, and error scenarios.

### DIFF
--- a/src/api/providers/__tests__/openai-codex.spec.ts
+++ b/src/api/providers/__tests__/openai-codex.spec.ts
@@ -471,7 +471,42 @@ describe("OpenAiCodexHandler.completePrompt streaming", () => {
 		const mockFetch = vitest.fn()
 		vitest.stubGlobal("fetch", mockFetch)
 
-		await expect(handler.completePrompt("Hello", { abortSignal: AbortSignal.abort() })).rejects.toThrow()
+		await expect(handler.completePrompt("Hello", { abortSignal: AbortSignal.abort() })).rejects.toMatchObject({
+			name: "AbortError",
+		})
+		expect(mockFetch).not.toHaveBeenCalled()
+	})
+
+	// The abort arrives while the request is still in flight, before any event, which is the case
+	// that only works if the caller's signal is genuinely linked to the internal controller. A
+	// broken link would leave the request hanging on a signal that never fires.
+	it("aborts the in-flight request when the caller cancels before any event", async () => {
+		const handler = createHandler()
+		const controller = new AbortController()
+		let signalDuringRequest: AbortSignal | undefined
+
+		const create = vitest.fn().mockImplementation(
+			(_body: unknown, options: { signal: AbortSignal }) =>
+				new Promise((_resolve, reject) => {
+					signalDuringRequest = options.signal
+					// Reject the way the SDK does once the signal it was handed aborts.
+					options.signal.addEventListener("abort", () => reject(new Error("Request was aborted")), {
+						once: true,
+					})
+				}),
+		)
+		Reflect.set(handler, "client", { responses: { create } })
+		const mockFetch = vitest.fn()
+		vitest.stubGlobal("fetch", mockFetch)
+
+		const completion = handler.completePrompt("Hello", { abortSignal: controller.signal })
+		// The token lookup and the listener that links the two signals are both async, so aborting
+		// synchronously here would fire before anything is listening.
+		await vitest.waitFor(() => expect(create).toHaveBeenCalled())
+		controller.abort()
+
+		await expect(completion).rejects.toMatchObject({ name: "AbortError" })
+		expect(signalDuringRequest!.aborted).toBe(true)
 		expect(mockFetch).not.toHaveBeenCalled()
 	})
 
--- a/src/api/providers/__tests__/openai-codex.spec.ts
+++ b/src/api/providers/__tests__/openai-codex.spec.ts
@@ -247,28 +247,182 @@ describe("OpenAiCodexHandler.completePrompt service tier", () => {
 	it.each<[string, OpenAiCodexServiceTier | undefined, typeof OpenAiCodexServiceTier.Priority | undefined]>([
 		["Fast", OpenAiCodexServiceTier.Priority, OpenAiCodexServiceTier.Priority],
 		["Standard", undefined, undefined],
-	])("uses the %s preference in non-streaming requests", async (_mode, configuredTier, expectedTier) => {
+	])("uses the %s preference in completion requests", async (_mode, configuredTier, expectedTier) => {
 		const handler = new OpenAiCodexHandler({
 			apiModelId: "gpt-5.6-sol",
 			...(configuredTier ? { [OPEN_AI_CODEX_SERVICE_TIER_KEY]: configuredTier } : {}),
 		})
 		vitest.spyOn(openAiCodexOAuthManager, "getAccessToken").mockResolvedValue("test-token")
 		vitest.spyOn(openAiCodexOAuthManager, "getAccountId").mockResolvedValue("acct_test")
-		const mockFetch = vitest.fn().mockResolvedValue({
-			ok: true,
-			json: vitest.fn().mockResolvedValue({ text: "Complete" }),
-		})
-		vitest.stubGlobal("fetch", mockFetch)
+		const create = vitest.fn().mockResolvedValue(
+			asyncStreamFrom([
+				{ type: "response.output_text.delta", delta: "Complete" },
+				{ type: "response.completed", response: { id: "r1", status: "completed", output: [] } },
+			]),
+		)
+		Reflect.set(handler, "client", { responses: { create } })
 
 		await expect(handler.completePrompt("Hello")).resolves.toBe("Complete")
 
-		const body = JSON.parse(mockFetch.mock.calls[0][1].body)
-		expect(body.stream).toBe(false)
+		const body = create.mock.calls[0][0]
+		// The Codex subscription endpoint rejects `stream: false` outright.
+		expect(body.stream).toBe(true)
 		if (expectedTier) {
 			expect(body[SERVICE_TIER_KEY]).toBe(expectedTier)
 		} else {
 			expect(body).not.toHaveProperty(SERVICE_TIER_KEY)
 		}
+	})
+})
+
+describe("OpenAiCodexHandler.completePrompt streaming", () => {
+	function createHandler() {
+		const handler = new OpenAiCodexHandler({ apiModelId: "gpt-5.6-sol" })
+		vitest.spyOn(openAiCodexOAuthManager, "getAccessToken").mockResolvedValue("test-token")
+		vitest.spyOn(openAiCodexOAuthManager, "getAccountId").mockResolvedValue("acct_test")
+		return handler
+	}
+
+	function injectStream(handler: OpenAiCodexHandler, events: unknown[]) {
+		const create = vitest.fn().mockResolvedValue(asyncStreamFrom(events))
+		Reflect.set(handler, "client", { responses: { create } })
+		return create
+	}
+
+	afterEach(() => {
+		vitest.restoreAllMocks()
+		vitest.unstubAllGlobals()
+	})
+
+	it("joins consecutive text deltas into one string", async () => {
+		const handler = createHandler()
+		injectStream(handler, [
+			{ type: "response.output_text.delta", delta: "feat: " },
+			{ type: "response.output_text.delta", delta: "add commit messages" },
+			{ type: "response.completed", response: { id: "r1", status: "completed", output: [] } },
+		])
+
+		await expect(handler.completePrompt("Hello")).resolves.toBe("feat: add commit messages")
+	})
+
+	// A commit message is written straight into the Source Control box, so reasoning must never
+	// become part of it.
+	it("omits reasoning from the completion", async () => {
+		const handler = createHandler()
+		injectStream(handler, [
+			{ type: "response.reasoning_summary_text.delta", delta: "Thinking about the diff" },
+			{ type: "response.output_text.delta", delta: "fix: correct the parser" },
+			{ type: "response.completed", response: { id: "r1", status: "completed", output: [] } },
+		])
+
+		const result = await handler.completePrompt("Hello")
+
+		expect(result).toBe("fix: correct the parser")
+		expect(result).not.toContain("Thinking")
+	})
+
+	it("omits usage and tool calls from the completion", async () => {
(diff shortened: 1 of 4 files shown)


src/api/providers/openai-codex.ts
```diff
@@ -1256,98 +1271,38 @@ export class OpenAiCodexHandler extends BaseProvider implements SingleCompletion
...
-
-			if (reasoningEffort) {
-				baseRequestBody.reasoning = {
-					effort: reasoningEffort,
-					summary: "auto" as const,
+			// Only the answer is wanted. Reasoning is deliberately dropped rather than concatenated:
+			// callers such as commit-message generation write this straight into the editor.
+			let text = ""
+
+			for await (const chunk of this.handleResponsesApiMessage(
```
taltas, line 1321: If the SDK yields text and then throws, the shared transport replays the request over SSE and this loop appends both generations; can we allow the fallback only before the first SDK event and add a mid-stream failure test?

src/api/providers/openai-codex.ts
```diff
@@ -1256,98 +1271,38 @@ export class OpenAiCodexHandler extends BaseProvider implements SingleCompletion
...
-					}
-				}
-			}
-
-			if (responseData?.text) {
-				return responseData.text
-			}
-
-			return ""
+			return text
```
taltas, line 1345: An abort currently reaches this return with empty or partial text, so callers treat a cancelled generation as complete; can we reject when the caller's signal is aborted and test both pre-aborted and mid-stream cancellation?

taltas (commented):
A few comments

src/api/providers/openai-codex.ts
Rafael-Silva-Oliveira: fixed in 979269b01.

src/api/providers/openai-codex.ts
Rafael-Silva-Oliveira: fixed in 979269b01.

src/api/providers/openai-codex.ts
```diff
@@ -438,10 +439,23 @@ export class OpenAiCodexHandler extends BaseProvider implements SingleCompletion
...
+		// abort through `this.abortController`. Without this the signal never reaches the wire.
+		const abortFromCaller = () => this.abortController?.abort()
+
+		if (abortSignal) {
+			if (abortSignal.aborted) {
+				this.abortController.abort()
+			} else {
+				abortSignal.addEventListener("abort", abortFromCaller, { once: true })
+			}
+		}
```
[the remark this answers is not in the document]
Rafael-Silva-Oliveira: Valid, fixed in c68dae15f. The fallback now rethrows when the controller is already aborted, so a cancellation no longer spends a second request on a dead signal or comes back as a connection error. I rethrow rather than return, since completePrompt needs the rejection to tell a cancelled generation apart from a finished one.

src/api/providers/openai-codex.ts
```diff
@@ -479,18 +498,27 @@ export class OpenAiCodexHandler extends BaseProvider implements SingleCompletion
...
 						}
 						yield outChunk
 					}
 				}
-			} catch (_sdkErr) {
+			} catch (sdkErr) {
+				if (sawSdkEvent) {
+					throw sdkErr
+				}
+
```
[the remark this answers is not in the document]
Rafael-Silva-Oliveira: Correct, and fixed in c68dae15f. This was the one gap I had flagged as knowingly left open, so thanks for pushing on it.

Went with a plain handler field rather than a typed marker. sawSdkEvent moved onto the handler as sawSdkEventInCurrentResponse, reset per request next to the other response state, and the OAuth retry is skipped once it is set. Same information, no error type to thread through the generator boundary and match on.

A refresh before any event still retries exactly as before, which is what that loop is actually for. Tests cover both sides of it, plus the SSE fallback case.

src/api/providers/__tests__/openai-codex.spec.ts
```diff
@@ -272,6 +275,216 @@ describe("OpenAiCodexHandler.completePrompt service tier", () => {
...
+	it("does not fall back to SSE when the SDK fails because the caller aborted", async () => {
+		const handler = createHandler()
+		const create = vitest.fn().mockRejectedValue(new Error("Request was aborted"))
+		Reflect.set(handler, "client", { responses: { create } })
+		const mockFetch = vitest.fn()
+		vitest.stubGlobal("fetch", mockFetch)
+
+		await expect(handler.completePrompt("Hello", { abortSignal: AbortSignal.abort() })).rejects.toThrow()
+		expect(mockFetch).not.toHaveBeenCalled()
+	})
```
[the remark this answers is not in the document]
Rafael-Silva-Oliveira: Two of the three hold up, and chasing the first turned up a real bug behind it. Fixed in bb2676a81.

The AbortError assertion was right, but not for the reason given. The two existing cancellation tests at lines 405 and 419 already assert name AbortError, so only the newest one was using a bare rejects.toThrow. Tightening it made it fail, which exposed the actual problem: a stream that ends quietly throws an AbortError, but a transport that rejects on abort had its own error passed straight through the catch. So what a cancelled completePrompt rejected with depended on how far the request had got, which is not something a caller can key off. An abort is now restated as an AbortError unless it already is one. Both cancellation tests fail without that change and pass with it, so it is actually guarded.

The in flight test was a genuine gap and is added. Worth noting it needs to wait for the request to be in flight before aborting, since the token lookup and the signal linking are both async and a synchronous abort fires before anything is listening. My first attempt at it hung for exactly that reason.

Skipped the third: asserting that an already aborted signal does not call create. That contradicts the implementation and the passing test at line 423. executeRequest aborts its internal controller and then still calls create with that aborted signal, which is what makes the SDK reject it. Asserting create is never called would fail today, and changing the code to short circuit before the request would be a behaviour change not worth making for a test.

Full provider suite passes at 1267, types and lint clean.

src/api/providers/openai-codex.ts
```diff
@@ -1256,99 +1294,56 @@ export class OpenAiCodexHandler extends BaseProvider implements SingleCompletion
...
+				model,
+				"",
+				[{ role: "user", content: prompt }],
+				// `taskId` is required, and resolves to the same session id this used to send
+				// directly, so `prompt_cache_key` is unchanged.
+				{ taskId: this.sessionId },
+				options?.abortSignal,
+			)) {
+				if (chunk.type === "text") {
+					text += chunk.text
```
edelauna, line 1334: Model refusals are emitted by `processEvent` as `{type:'text', text:'[Refusal] ...'}`, so this loop accumulates them and `completePrompt` returns the refusal string -- where the old `stream:false` path read only `output_text` and returned `''`. That undocumented change surfaces `[Refusal] ...` directly in the prompt-enhancer input box; consider skipping refusal chunks or returning `''` to match prior behavior. (The comment above also names 'commit-message generation', but the actual caller is the prompt enhancer.)

edelauna (dismissed):
Noticed a small regression in how messages handle refusals.

src/api/providers/openai-codex.ts
Rafael-Silva-Oliveira: Good catch, fixed in 78adb9a3f. The refusal prefix is a shared constant now and completePrompt skips those chunks, so it returns "" like the old output_text read did — any real text alongside a refusal still comes through. Fixed the stale comment too, the prompt enhancer is indeed the only caller.

taltas (approved):

PR #1257, opened 2026-08-16
fix(webview): render expanded task header text as markdown with consistent scrollbar

### LINKED ISSUE #1256: [BUG] Expanded task header renders prompt as raw text (no markdown) and shows a second, inconsistently-styled scrollbar
## Summary

Two related issues in the expanded task header (the collapsible "Task" panel at the top of an open task):

1. **Prompt is not rendered as markdown.** The collapsed title uses plain-text rendering, which is fine for a single-line summary, but the *expanded* view also used the same plain-text component, so prompts containing markdown syntax display it verbatim — e.g. bold markers, backtick code spans, lists and headings appear as literal characters while assistant messages in the same conversation render formatted markdown via MarkdownBlock.

2. **Inconsistent / doubled scrollbars.** The expanded prompt box (overflow-auto max-h-80) uses a default always-visible Chromium scrollbar, while the message list below (Virtuoso) uses the VS Code-style .scrollable hover-reveal scrollbar from index.css. For long prompts both bars are visible stacked in the same column with different styling.

## Steps to Reproduce

1. Start a task whose prompt contains markdown (e.g. paste a markdown document).
2. Expand the task header (click the title / chevron).
3. Observe:
   - the prompt text shows raw markdown source instead of rendered formatting;
   - for long prompts, two scrollbars appear in the same column — one default-style (prompt box) and one VS Code-style (message list).

## Expected Behavior

- The expanded task header renders the prompt through the same MarkdownBlock pipeline as chat messages.
- The prompt box uses the shared .scrollable class so its scrollbar matches the message list (transparent until hover/focus).

## Suggested Fix

In webview-ui/src/components/chat/TaskHeader.tsx:

- Replace the plain-text Mention component in the expanded state with MarkdownBlock and drop the now-redundant whitespace-pre-wrap class.
- Add the .scrollable class to the prompt box so it inherits the VS Code scrollbar styling.

## Additional Information

- Regression tests added in webview-ui/src/components/chat/__tests__/TaskHeader.spec.tsx (markdown rendering + scrollable class).
- Verified: vitest 23/23 passing, ESLint clean, tsc -b passes.

easonLiangWorldedtech:
Fixes #1256

## Summary

Two small fixes in the expanded task header (the collapsible "Task" panel):

1. **Render the prompt as markdown.** The expanded view used the plain-text `Mention` component, so prompts with markdown syntax displayed it verbatim while assistant messages render formatted markdown via `MarkdownBlock`. Now rendered through the same pipeline; the collapsed one-line title keeps plain text (appropriate for an ellipsized summary).
2. **Consistent scrollbar.** The prompt box used a default always-visible Chromium scrollbar while the message list uses the VS Code-style `.scrollable` hover-reveal style, so long prompts showed two differently-styled bars stacked in the same column. Added the shared `scrollable` class to match.

## Security hardening (required by pre-merge review)

Rendering arbitrary task markdown opened a new path for a prompt such as `[open](../../.env)` to become a clickable anchor that posts `openFile` to the extension. This is addressed on both ends, per the CodeRabbit "Trust And Persistence Invariants" finding:

- **Webview (`MarkdownBlock.tsx`):** the link anchor now validates before posting `openFile`:
  - only local/file targets are handled (existing scheme filter);
  - path traversal is rejected — any `..` segment (e.g. `[x](../../.env)`) is dropped before the message is posted;
  - surviving requests are tagged `values.fromMarkdown: true`, marking them as markdown-sourced (untrusted input);
  - existing relative-prefix behavior is unchanged.
- **Extension (`webviewMessageHandler.ts` `openFile` case):** defense in depth — for markdown-sourced requests (`values.fromMarkdown === true`) the handler resolves the requested path against the current workspace cwd, enforces workspace containment via `isPathOutsideWorkspace`, and surfaces a localized "cannot access path / outside workspace" error (new i18n key `path_outside_workspace` in all 18 locales) instead of blindly opening.
  - **Why scoped to tagged requests:** first-party `openFile` post sites (e.g. the slash-commands settings view) legitimately open files *outside* the workspace — global slash-command files live in a user directory, not the project — and keep their legacy behavior. Only untrusted, markdown-sourced targets get the strict containment, so no production flow regresses.

## Mutation testing (Stryker 10.0.0 note)

The `mutation-diff` gate blocks on every Survived/NoCoverage mutant on changed lines. `MarkdownBlock.tsx` needed mutator-specific `// Stryker disable next-line` directives for 35 mutants, each verified equivalent or defensive (the directive states the concrete reason):

- **14 module-scope static mutants** (placeholder char/regex, mask node-type set) are **false survivors** caused by a Stryker 10.0.0 + vitest-runner interaction: whenever `testFiles` is set, the planner switches `mutantActivation` to `runtime`, and the runner only enables `activeMutant` inside `beforeAll` — module scope has already executed by then, so static mutants never activate and every test passes. The affected constants are pinned by the mask tests in `MarkdownBlock.spec.tsx`.
- **21 per-test mutants** are equivalent (the guards only defend against shapes the remark/hast pipeline cannot produce: non-numeric position offsets, a text/element node as tree root, a mask cell one past the last mention, an orphan placeholder index) or defensive (the code/link skip guards in `rehypeMentions` are a second line of defense behind the source-region masking in `prepareMentions`).
- Two styling mutants (the mention span's Tailwind classes) were killed with new `className` assertions in `MarkdownBlock.spec.tsx` instead of being disabled.
- `TaskHeader.tsx` already carries 2 directives for equivalent click-target backstops.

## Regression evidence (lowest layer, per pre-merge review)

- **`MarkdownBlock.spec.tsx` (lowest layer, `mentions` enabled):** image alt text, inline and block math, and raw HTML each containing a mention pattern — asserts zero `mention-context-highlight` spans and that no placeholder control character leaks into the rendered output; fenced/inline code cases; file-anchor validation tests (traversal click posts nothing, external URLs render inert, relative paths still post `./...`); the anchor posts `values.fromMarkdown: true`; mention span styling pinned via `className`; empty-markdown behavior (no empty `<p>`, stable re-render deps).
- **`TaskHeader.spec.tsx`:** existing markdown/link/mention/empty-text/scrollbar coverage plus an empty-prompt text assertion.
- **`webviewMessageHandler.openFile.spec.ts` (new, 7 tests):** tagged relative-in-workspace open, tagged `../../.env` traversal rejected with the localized error, tagged absolute-outside-workspace rejected, tagged absolute-in-workspace open, **untagged caller keeps legacy behavior for outside-workspace paths**, missing-text no-op, missing-cwd error.
- **`ClineProvider.spec.ts`:** untagged `openFile` posts still take the legacy path (existing suite, unchanged).
- Playwright component snapshots for the expanded markdown and overflow UI.

## Verification

- vitest: TaskHeader + MarkdownBlock + openFile handler + ClineProvider suites all passing
- ESLint `--max-warnings=0`: clean (src and webview-ui)
- tsc --noEmit: passes (src and webview-ui)
- Visual regression: `webview-visual` green in CI
- Mutation gate: `scripts/stryker-diff.mjs ci` (base `0dbd5846f`) clean for both packages

## Line budget

This PR is **1445 changed lines** (32 files, 1431 added / 14 deleted) against the 1000-line hard cap. The overage was accepted explicitly and is documented here per the agreed decision. Where the lines come from:

| Block | Lines | Why |
|---|---|---|
| `MarkdownBlock.tsx` (+309/−5) | 314 | Link validation + `fromMarkdown` tagging, mention masking, memoized pipeline, and 35 Stryker directives — core of the hardening |
| `MarkdownBlock.spec.tsx` (+559/−2) | 561 | Lowest-layer regression evidence demanded by the pre-merge review (masking ×5 node shapes, anchor validation ×4, empty markdown ×3, code fences, className pin) |
| `webviewMessageHandler.openFile.spec.ts` (+192) | 192 | New handler-containment suite (7 tests incl. untagged-legacy scoping) — the extension-side half of the invariant |
| `TaskHeader.spec.tsx` (+175) | 175 | TaskHeader regression tests (markdown, links, mentions, scrollbar, empty prompt) |
| Playwright story + visual spec + 4 binary baselines (stories.tsx +60, TaskHeader.visual.tsx +43) | 103 | Required Story Gallery coverage for the new expanded UI |
| `TaskHeader.tsx` (+9/−3) | 12 | The actual feature change (MarkdownBlock render + scrollable class + 2 Stryker directives) |
| `webviewMessageHandler.ts` (+37/−4) + 18× i18n `common.json` (+1 each) | 59 | Scoped handler containment + localized error |
| `pnpm-lock.yaml` (26) / `webview-ui/package.json` (3) | 29 | Merged-in main updates (not authored in this branch) |

The two smallest possible diffs (the original two fixes) are ~60 lines; the remaining ~1385 lines are the review-mandated security, scoping, and regression coverage, which is why the budget could not be met while satisfying the pre-merge checks.


webview-ui/src/components/chat/TaskHeader.tsx
```diff
@@ -324,13 +328,13 @@ const TaskHeader = ({
...
 								ref={textRef}
-								className="overflow-auto max-h-80 whitespace-pre-wrap break-words break-anywhere cursor-text py-0.5"
+								className="scrollable overflow-auto max-h-80 break-words break-anywhere cursor-text py-0.5"
 								style={{
 									display: "-webkit-box",
 									WebkitLineClamp: "unset",
 									WebkitBoxOrient: "vertical",
 								}}>
-								<Mention text={task.text} />
+								<MarkdownBlock markdown={task.text ?? ""} />
```
taltas: This removes the existing `openMention` click behavior for context tokens such as `@/src/file.ts`, `@problems`, and `@terminal` whenever the header is expanded. Can we preserve context-mention rendering in the markdown path and add an expanded-mention interaction test?

easonLiangWorldedtech: Fixed. The expanded markdown path now wraps context mentions with the same clickable mention styling/behavior and posts `openMention` for `@/src/file.ts`, `@problems`, and `@terminal`.

I also added an expanded TaskHeader interaction regression test to verify the mentions remain rendered and clickable after expanding the header.

webview-ui/src/components/common/MarkdownBlock.tsx
```diff
@@ -415,7 +494,7 @@ const MarkdownBlock = memo(({ markdown }: MarkdownBlockProps) => {
 						}
 					},
 				]}
-				rehypePlugins={[rehypeKatex as any]}
+				rehypePlugins={[rehypeMentions, rehypeKatex as any]}
```
edelauna: `rehypeMentions` is added to the shared `MarkdownBlock`, so every `@mention` in assistant reasoning, MCP output, and todo lists -- not just the task header -- now renders as a clickable span that posts `openMention`. Those were inert text before. Is app-wide clickable mentions intended? If not, gate the plugin behind a prop passed only where text is user-authored.

webview-ui/src/components/common/MarkdownBlock.tsx
```diff
@@ -6,12 +6,91 @@ import rehypeKatex from "rehype-katex"
...
+	return (tree: any) => {
+		visit(tree, "text", (node: any, index, parent) => {
+			if (parent?.tagName === "span" && parent.properties?.className?.includes("mention-context-highlight")) {
+				return
+			}
+
+			// Code is literal content: never interactive-ify mention patterns inside it.
+			// Wrapping them would also corrupt the CodeBlock text extraction, which only
+			// keeps string children (the mention text would silently disappear).
+			if (parent?.tagName === "code") {
```
edelauna: The guard skips `code` but not `a`, so a mention inside link text (`[open @/src/main.ts](/path)`) splices this span into the `<a>`. The span sets `role=button` and `tabIndex=0`; per the WHATWG HTML standard an `<a>` allows no interactive-content or tabindex descendant, so this is invalid nested interactive content (the W3C validator flags `role=button` under `a`). Its `stopPropagation` also blocks the anchor's own `openFile` handler. https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element

```suggestion
			if (parent?.tagName === "code" || parent?.tagName === "a" || parent?.tagName === "pre") {
```

edelauna (changes requested):
Nice! This is a cool quality of life improvement - thank you for this. Had a question about how we should handle @ mentions?

webview-ui/src/components/common/MarkdownBlock.tsx
easonLiangWorldedtech: Good catch - app-wide clickable mentions were not the intent. I've gated `rehypeMentions` behind a new `mentions` prop on `MarkdownBlock` (off by default): assistant messages, reasoning, MCP output, and todo lists keep mention patterns as inert text again. Only the expanded `TaskHeader` prompt (user-authored) passes `mentions`. Regression test asserts mentions stay inert by default. (commit 84b22e36c)

webview-ui/src/components/common/MarkdownBlock.tsx
easonLiangWorldedtech: Fixed in 84b22e36c - the guard now skips `a` and `pre` in addition to `code`, so a mention inside link text is no longer spliced into a nested `role=button` span. Added a regression test using your exact case: `[open @/src/main.ts](/src/main.ts)` renders plain inside the anchor, and clicking the anchor still posts `openFile` (no `stopPropagation` interception, no invalid nested interactive content).

webview-ui/src/components/chat/TaskHeader.tsx
```diff
@@ -324,13 +328,13 @@ const TaskHeader = ({
 							className="text-vscode-font-size overflow-y-auto break-words break-anywhere relative">
 							<div
 								ref={textRef}
-								className="overflow-auto max-h-80 whitespace-pre-wrap break-words break-anywhere cursor-text py-0.5"
+								className="scrollable overflow-auto max-h-80 break-words break-anywhere cursor-text py-0.5"
```
edelauna, line 333: Dropping `whitespace-pre-wrap` here while switching to `MarkdownBlock` collapses single newlines in plain-text prompts -- the remark pipeline has no `remark-breaks`/`breaks:true`, so per CommonMark soft breaks become spaces and a multi-line prompt reflows into one paragraph (line-initial `#`/`-`/`>` and 4-space indents also get reinterpreted). Add `remark-breaks` (or the `breaks` option) so newlines render as `<br>` while keeping markdown.

webview-ui/src/components/common/MarkdownBlock.tsx
```diff
@@ -6,12 +6,94 @@ import rehypeKatex from "rehype-katex"
...
+			// span would be invalid nested interactive content (WHATWG) and its
+			// stopPropagation would block the anchor's own openFile handler; inside
+			// code it would corrupt the CodeBlock text extraction, which only keeps
+			// string children (the mention text would silently disappear).
+			if (parent?.tagName === "code" || parent?.tagName === "pre" || parent?.tagName === "a") {
+				return
+			}
+
+			const originalValue = String(node.value)
+			const matches = Array.from(originalValue.matchAll(mentionRegexGlobal))
```
edelauna: This matches mentions on already-tokenized text nodes, so a path with markdown-active chars truncates: remark splits `@/src/__init__.py` into `@/src/`+`<strong>init</strong>`+`.py`, this matches only `@/src/`, and the span posts `openMention` for `/src/` (wrong path, silently). The old `<Mention text={task.text}/>` matched the raw string, so this regresses `__init__.py`-style paths. Consider matching on the mdast before tokenization; no test covers a mention with `*`/`_`.

edelauna (changes requested):

webview-ui/src/components/chat/TaskHeader.tsx
easonLiangWorldedtech: Fixed in bd93019c2 — added a `breaks` prop to `MarkdownBlock` that enables `remark-breaks`. Single newlines in user-authored prompts now render as a structural `<br>` (the prop is off by default, so assistant messages, reasoning, MCP output and todos keep the CommonMark soft-break behavior). The expanded `TaskHeader` prompt is the only place that opts in.

One wrinkle found while implementing: `mdast-util-to-hast`'s `hardBreak` handler emits the `<br>` **and** a lone `\n` text node after it. Since this webview styles `p` with `white-space: pre-wrap`, that literal newline would render as a *second* line break — a blank line between every soft-broken line. I added a small rehype plugin (`rehypeStripBreakNewlines`) that drops that node when `breaks` is on, so each soft break produces exactly one line break.

Tests: TaskHeader now asserts a plain two-line prompt renders a single `<br>` inside one paragraph, and that headings/lists still parse while inner newlines are kept; MarkdownBlock covers `<br>` output, the default off behavior, and blank-line paragraph breaks.

webview-ui/src/components/common/MarkdownBlock.tsx
easonLiangWorldedtech: Fixed in bd93019c2 — mention matching now runs on the raw markdown string **before** remark tokenizes it, which is also how the collapsed `<Mention>` component matches, so the expanded view restores that behavior: `@/src/__init__.py` stays a single mention and posts the full path to `openMention` (no more silent truncation to `/src/`).

Implementation: `prepareMentions()` first does a throwaway mdast parse with the same extensions as the render pipeline and masks the literal / non-text regions (code, inline code, links, images, raw HTML, math) to spaces using the mdast position offsets, so a mention inside any of those regions is never rewritten. The raw string is then matched with `mentionRegexGlobal` and each match is replaced by an indexed \u0001 placeholder; the rehype plugin swaps the placeholders for the clickable spans. The existing rehype guards (skip `code`/`pre`/`a`) are kept so the plugin stays safe on any tree.

Tests: `@/src/__init__.py` and `@/src/*files*` each render as one span and post the full path on click (and no stray `<strong>` for the `__init__` part), plus a combined mentions+breaks case and a no-placeholder-leak case for non-matching `@` text.

easonLiangWorldedtech:
@coderabbitai resume

webview-ui/src/components/common/MarkdownBlock.tsx
```diff
@@ -3,15 +3,208 @@ import ReactMarkdown from "react-markdown"
...
+						},
+						// Keyboard parity with the click handler (a role=button span is not a
+						// native button, so Enter/Space must be handled explicitly).
+						onKeyDown: (event: React.KeyboardEvent<HTMLSpanElement>) => {
+							if (event.key !== "Enter" && event.key !== " ") {
+								return
+							}
+							event.stopPropagation()
+							vscode.postMessage({ type: "openMention", text: mentionValue })
+						},
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Fixed in 630309c18 — the keydown handler now calls `event.preventDefault()` before posting for Enter and Space, so Space no longer scrolls the expanded task panel while a mention has focus. The keyboard test now asserts both keys are default-prevented (via the `false` return of the cancelable `fireEvent.keyDown`) and that an unrelated key (`a`) neither posts nor prevents the default.

webview-ui/src/components/common/MarkdownBlock.tsx
```diff
@@ -3,15 +3,208 @@ import ReactMarkdown from "react-markdown"
...
+
+	const mentions: string[] = []
+	let preparedMarkdown = ""
+	let lastIndex = 0
+	for (const match of masked.join("").matchAll(mentionRegexGlobal)) {
+		const start = match.index!
+		preparedMarkdown += markdown.slice(lastIndex, start)
+		mentions.push(markdown.slice(start, start + match[0].length))
+		preparedMarkdown += `${MENTION_PLACEHOLDER_CHAR}${mentions.length - 1}${MENTION_PLACEHOLDER_CHAR}`
+		lastIndex = start + match[0].length
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Good catch — fixed in 630309c18. Matching now runs on the raw markdown string and a match is discarded only when its range intersects a masked literal region (the mask is now a `boolean[]` of the same length as the input, built from the same mdast positions). Verified against the shared regex directly: `[file](/src/a.ts)@problems` matches nothing in the raw text (the start boundary rejects `)`), and the masked-string approach was the only reason it became actionable. Added regression tests for both the directly-adjacent case (stays inert, 0 spans) and the space-separated case (still actionable, 2 spans posting `problems`).

taltas:
Nice QOL improvement, thanks for the contribution

taltas (dismissed):

webview-ui/src/components/common/MarkdownBlock.tsx
```diff
@@ -3,15 +3,224 @@ import ReactMarkdown from "react-markdown"
...
+		// rules must see: masking would turn a preceding `)` or backtick into a
+		// space and make a non-mention actionable (e.g. `[file](/src/a.ts)`@problems``).
+		// Discard a match only when its range lands inside a masked literal region.
+		if (isMasked.slice(start, end).some(Boolean)) {
+			continue
+		}
+		preparedMarkdown += markdown.slice(lastIndex, start)
+		mentions.push(markdown.slice(start, end))
+		preparedMarkdown += `${MENTION_PLACEHOLDER_CHAR}${mentions.length - 1}${MENTION_PLACEHOLDER_CHAR}`
+		lastIndex = end
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Fixed in d9bc16d9b — `definition` is now included in `MENTION_MASK_NODE_TYPES`, so a reference link destination is masked before `prepareMentions()` runs. Without it, a destination such as `[docs]: @/docs/readme.md` was rewritten to an indexed mention placeholder, and that placeholder then became the reference link's href — control characters leaking into the rendered output instead of the destination staying inert.

Regression test added (`keeps reference link destinations inert and preserves the original href`): the reference link keeps its original href, the destination never becomes a mention span, a real mention in the body stays actionable, and no placeholder control characters leak into the output. Verified the test fails against the pre-fix code and passes now; the full MarkdownBlock and TaskHeader suites are green.

webview-ui/src/components/common/MarkdownBlock.tsx
```diff
@@ -3,15 +3,236 @@ import ReactMarkdown from "react-markdown"
...
+const MENTION_MASK_NODE_TYPES = new Set([
+	"code",
+	"inlineCode",
+	"link",
+	"image",
+	"html",
+	"inlineMath",
+	"math",
+	"definition",
+])
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Fixed in 203f76b2a — `linkReference` and `imageReference` are now included in `MENTION_MASK_NODE_TYPES`, so a reference link's label and an image reference's alt text are masked before mention replacement.

Confirmed the failure mode before the fix: in `[the @problems summary][docs]` the `@problems` pattern matched on the raw string (its position inside the label sat outside every previously masked node type), the placeholder was spliced into the label, and because `rehypeMentions()` skips `<a>` the raw placeholder (control character plus index) rendered verbatim inside the link text — the new test captured the literal leaked placeholder in `anchor.textContent`. Image references corrupted the `alt` attribute the same way.

Two regression tests added: a mention inside a reference link label and one inside an image reference alt both stay inert, the label text and alt attribute are preserved verbatim, no mention span is rendered, and no placeholder control characters leak into the output. Both fail against the pre-fix code and pass now; the full MarkdownBlock and TaskHeader suites are green.

easonLiangWorldedtech:
## Fix summary: mask all reference-style link regions before mention rewriting

Addressed the latest two CodeRabbit findings about mention masking in reference-style links with two commits:

**`d9bc16d9b` — mask reference link definitions**

`definition` was missing from `MENTION_MASK_NODE_TYPES`. A reference definition's destination renders as the reference link's href, so a mention pattern like `[docs]: @/docs/readme.md` was rewritten to an indexed mention placeholder — and that placeholder became the link's href, leaking control characters into the rendered output instead of keeping the destination inert.

**`203f76b2a` — mask reference links and image references**

`linkReference` and `imageReference` were likewise missing. A mention inside a reference link's label (e.g. `[the @problems summary][docs]`) matched on the raw string, the placeholder was spliced into the label, and because `rehypeMentions()` skips anchors, the raw placeholder (control character plus index) rendered verbatim inside the link text. Image references corrupted the `alt` attribute the same way. Verified this baseline failure before the fix: the new test captured the literal leaked placeholder in `anchor.textContent`.

**Verification**

- Added three regression tests (definition destination, reference link label, image reference alt): mentions in each region stay inert, the href / label text / alt attribute are preserved verbatim, no mention span is rendered, and no placeholder control characters (``) leak into the output.
- Each new test fails against the pre-fix code and passes now (baseline-checked via stash).
- `MarkdownBlock` + `TaskHeader` suites: 69 passed; ESLint `--max-warnings=0` clean; pre-push `check-types` passed.

Head of `fix/task-header-markdown`: `203f76b2a`.

webview-ui/src/components/chat/TaskHeader.tsx
```diff
@@ -324,13 +328,13 @@ const TaskHeader = ({
...
 								style={{
 									display: "-webkit-box",
 									WebkitLineClamp: "unset",
 									WebkitBoxOrient: "vertical",
 								}}>
-								<Mention text={task.text} />
+								<MarkdownBlock markdown={task.text ?? ""} mentions breaks />
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Done in 62e7e93c8 — the expanded TaskHeader now has a Story Gallery visual case per webview-ui/AGENTS.md:

- New task-header-markdown story on the shared gallery harness (#1446 is already in the branch — it sits on top of current main, so no rebase was needed): an expanded TaskHeader with representative user-authored markdown (heading, bold, inline code, bullet list, external link, soft breaks, and the path / @problems / @terminal mention kinds) plus enough content to overflow the max-h-80 prompt box, so the .scrollable surface is captured.
- New TaskHeader.visual.tsx establishes the deterministic expanded state (heading, 4 list items, 3 mention spans with exact text, br soft breaks, and an assertion that the prompt box genuinely overflows) and snapshots all four VS Code themes.
- Baselines generated in the pinned Docker Playwright container (the test:visual:docker:update flow, same image as CI) and committed under __screenshots__: task-header-markdown-{dark,light,high-contrast,high-contrast-light}.png.
- Existing Vitest tests for semantic behavior are unchanged and still pass (TaskHeader + MarkdownBlock suites green).

Verification in the same container: the new tests pass in verify mode against the committed baselines; the rest of the CT suite also passes, with the exception of the four RenderedContentContrast snapshots that show a ~1% pixel drift only under fully-parallel load — that fixture renders the async shiki CodeBlock (not MarkdownBlock) and passes in isolation on both clean main and this branch, so it is a pre-existing timing flake unrelated to this change.

easonLiangWorldedtech:
@coderabbitai approve

easonLiangWorldedtech:
## Pre-merge checks addressed (latest commit `f451cee9f`)

Both CodeRabbit pre-merge checks have been addressed in the latest commit, and the CI `mutation-diff` gate was re-run locally with the same `scripts/stryker-diff.mjs ci` gate.

### 1. Trust And Persistence Invariants (Error)

The expanded TaskHeader now renders task markdown through `MarkdownBlock`, which opens a new path where a prompt like `[open](../../.env)` becomes a clickable anchor that posts `openFile` to the extension. This is handled with two layers of defense:

- **Webview side (`MarkdownBlock.tsx`)**: the anchor click is validated before `openFile` is posted — only local/file targets are handled (the existing scheme filter), and any path containing a `..` segment (e.g. `../../.env`) is dropped and never posted. Validated requests carry the `values.fromMarkdown: true` tag, marking them as markdown-sourced (i.e. untrusted input). The existing `./` prefix behavior is unchanged.
- **Extension side (the `openFile` case in `webviewMessageHandler.ts`)**: defense in depth — **for tagged requests only** (`fromMarkdown: true`), the handler resolves the requested path against the current workspace cwd and runs the `isPathOutsideWorkspace` containment check; out-of-workspace paths are rejected with a localized error (new i18n key `path_outside_workspace`, added to all 18 locales).
- **Why it is scoped to tagged requests**: first-party `openFile` posters (e.g. the slash-commands settings view) legitimately open files outside the workspace — global slash command files live in the user directory, not the project — so unconditional containment would break a production flow and they keep the legacy behavior. Only markdown-sourced untrusted targets are strictly contained.

Supporting tests:
- `webviewMessageHandler.openFile.spec.ts` (new, 7 tests): tagged relative path inside the workspace opens; tagged `../../.env` traversal is rejected with the full localized message asserted; tagged absolute path outside the workspace is rejected; tagged absolute path inside the workspace opens; **untagged callers keep legacy behavior** (outside-workspace paths still open); no text is a no-op; missing cwd errors.
- `MarkdownBlock.spec.tsx` adds file anchor validation tests: a traversal click posts nothing, external https renders inert, relative paths still post `./...`, and anchors post `values.fromMarkdown: true`.
- The existing untagged `openFile` posts in `ClineProvider.spec.tsx` still take the legacy path (existing suite unchanged).

### 2. Regression Evidence (Warning)

Masking regression tests were added at the lowest layer (`MarkdownBlock`, `mentions` enabled):

- image alt text containing a mention pattern → asserts 0 `mention-context-highlight` spans and no placeholder control-character leakage in the rendered output;
- inline / block math containing a mention → same assertions;
- raw HTML containing a mention → same assertions;
- plus fenced / inline code containing mentions, and empty-markdown behavior (no empty `<p>` rendered, re-render deps stable).

### mutation-diff

Re-ran the mutation diff for both the webview and extension packages locally with `scripts/stryker-diff.mjs ci --base 0dbd5846f`; every blocking mutant (Survived / NoCoverage) on changed lines has been dealt with:

- Killable mutants were all killed with lowest-layer tests (the new tests above), including the two Tailwind-class styling mutants on the mention span — killed with `className` assertions (`text-[0.9em]` / `cursor-pointer`), not disabled — and the LogicalOperator mutant on `values ?? {}` — killed by a new line-anchor test (clicking `[line 12](src/main.ts:12)` must post `values` that keeps both `line: 12` and `fromMarkdown: true`); that test also pins the line number from being dropped when posting.
- The remaining mutants were each verified equivalent or defensive and silenced with mutator-specific `// Stryker disable next-line` directives, each with a concrete reason:
  - `MarkdownBlock.tsx` has 35 directives in total: **14 are module-scope static mutants that are false survivors of a Stryker 10.0.0 + vitest-runner interaction** — as soon as `testFiles` is set, the planner switches `mutantActivation` to `runtime`, and the runner only enables `activeMutant` in `beforeAll`, after module scope has already executed, so static mutants never activate and all tests pass. The affected constants (placeholder character / regex, mask node-type set) are pinned by the mask tests in `MarkdownBlock.spec.tsx`; **the other 21 per-test mutants** are equivalent (guards only defend against shapes the remark/hast pipeline never emits: non-numeric position offsets, a text/element node as the tree root, a mask cell running one past the last mention, an orphan placeholder index) or defensive (the code/link skip guards inside `rehypeMentions` are a second line of defense after `prepareMentions` source-region masking).
  - `TaskHeader.tsx` has 2 pre-existing directives (the click-target backstop after `closest("img")` / `closest("a")` never fires independently).

### Line budget

The PR ends up at **1445 changed lines** (32 files, 1431 added / 14 deleted), over the 1000-line hard cap. The overage was explicitly accepted, and the PR description breaks down the line count and reason for each block (the bulk is the security hardening, `fromMarkdown` scoping, and lowest-layer regression tests demanded by the pre-merge review; the two original fixes themselves are ~60 lines).

easonLiangWorldedtech:
@coderabbitai review

easonLiangWorldedtech:
Superseded by the 2-way split (both based on current main):

- #1762 ??extension-side: workspace containment for markdown-sourced openFile requests
- #1763 ??webview-side: markdown task header rendering + link/mention hardening

The original red check (`verify:coverage-contract`) came from the hardcoded coverage population baseline, which #1644 replaced with the shared coverage-contract merge. Both split PRs are based on main after #1644, so the failure cannot recur and no baseline edit is needed.

Merge order: #1762 before #1763 (defense-in-depth; the two PRs are independent and share no files).

PR #1261, opened 2026-08-16
fix(task-history): atomic per-task merge and drop shared index file (#1231)

### LINKED ISSUE #1231: [BUG][regression] Global _index.json full rewrite is unsafe under concurrent tasks (real corruption under JetBrains multi-agent)
## Summary
Task history persistence uses a global `_index.json` updated via
read → mutate entire entries array → atomic full-file rewrite (`safeWriteJson`).
That is safe for a **single** writer, not for concurrent agents.

We are already seeing **production corruption** with concurrent agents on the
JetBrains plugin (not observed on last Roo Code CE with serial usage).

This blocks trustworthy parallelizable tasks [](https://github.com/Zoo-Code-Org/Zoo-Code/issues/355):
runtime pieces (TaskRegistry, TaskScheduler, TaskSemaphore, status machine)
increase concurrent writers; storage still assumes one writer.

## Failure mode
1. Task A reads index (version V).
2. Task B reads index (version V).
3. A writes V+A (atomic rename — OK in isolation).
4. B writes V+B and **drops A’s updates**.

`safeWriteJson` prevents half-written JSON; it does **not** prevent lost updates.

(Real-life situation is probably more complex, as ui_message.json and history_item.json also get corrupted if one opens a task in the UI!)

## Observed corruption
- Placeholder task names / wiped fields
- `size=0` while task dirs have real UI/ACH data
- Broken delegation graph (parent/child status, delegatedToId, childIds)
- Invalid transitions when one writer restores stale status

## Environment
- IDE: PhpStorm 2026.2.1  / WebStorm 2026.2.1 (build WS-262.9437.145)
- Java: 25.0.3
- JCEF Support: Yes
- Zoo Code JetBrains plugin: 3.76.0 (5b0bcd0c)
- OS: Windows 11 10.0 (amd64)
- Multiple agents: yes (One session in Phpstorm, another session in Webstorm, they do NOT both need to actually be active)

## What changed vs Roo CE (hypothesis)
Zoo added richer lifecycle persistence and parallel-task infrastructure:
- TaskStatus transitions (active/delegated/completed/interrupted)
- More parent/child fields written on delegate/return/cancel
- TaskRegistry / scheduler / semaphore → more concurrent activity

Storage model remained a single global mutable JSON document.

### P0 – stop the bleeding
- Optimistic concurrency (occ): record file identity on read (size/mtime/inode);
  refuse write if changed; retry read-patch-write for single-entry updates.
- Process-wide mutex around index updates (same extension host).
- Prefer updating `tasks/<id>/history_item.json` as source of truth; treat
  `_index.json` as a rebuildable cache where possible.

Possible implementation: [`saveFile`](https://github.com/metaworx/zoo-code-history-repair/blob/master/src/lib/file.ts#L111)

### P1 – reduce contention
- Workspace-scoped indexes (no single file for all projects).
- Narrow writes: patch one entry, not conceptual “rewrite the world” without merge.

### P2 – durable design for #355
- Append-only event log (JSONL or line protocol) or SQLite (WAL) task catalog.
- Repository interface shared by GUI, CLI, and repair tools.

## Evidence / tooling
- Repair CLI that rebuilds index from per-task files and detects corruption.
- Snapshot-checked writer prototype (mtime/ctime/size/inode) that turns
  silent clobber into `Concurrent modification detected`.

## Ask
Treat concurrent history writes as a **release blocker for parallel tasks**,
not a rare edge case. JetBrains multi-agent already [proves](https://github.com/Zoo-Code-Org/Zoo-Code-Jetbrains/issues/6) the race in the wild.

edelauna:
### Related GitHub Issue

Closes #1231

### Description

Multiple extension hosts sharing the same task-history storage directory could corrupt task data. Each host rebuilt and overwrote the shared `tasks/_index.json` from its own partial cache, silently dropping entries written by other hosts. Per-task `history_item.json` files were also vulnerable: a host with a stale cache could overwrite fields that another host had updated on disk.

This change:

1. **Removes `_index.json` entirely.** The shared index file was derived state and the sole source of cross-process clobbering. `initialize()` now scans task directories directly via `reconcile({ forceRefresh: true })`. For the task counts in this system (tens to hundreds), the directory scan is sub-millisecond.

2. **Adds atomic per-task read-modify-write.** `safeWriteJson` gains a `merge` option: a callback that reads the current file under the already-held advisory lock and lets the caller merge before writing. `writeTaskFile` uses this to compute a diff-delta (only fields the caller actually changed) and apply it to the disk version, so fields updated by another host are preserved rather than reverted from a stale cache.

3. **Fixes cross-host delete detection.** `reconcile()` now checks for `history_item.json` existence (not just directory presence) when deciding whether a task is live. A `delete()` that removes only the file is correctly detected by peer hosts on their next reconciliation.

Same-field conflicts remain last-writer-wins by design.

### Test Procedure

```bash
cd src
pnpm exec vitest run \
  core/task-persistence/__tests__/TaskHistoryStore.spec.ts \
  core/task-persistence/__tests__/TaskHistoryStore.crossInstance.spec.ts \
  core/task-persistence/__tests__/TaskHistoryStore.reconciliation.spec.ts \
  utils/__tests__/safeWriteJson.test.ts
pnpm check-types
pnpm exec eslint . --ext=ts --max-warnings=0
```

- Full test suite: 7436 tests pass, 0 failures.
- `tsc --noEmit`: clean.
- ESLint with zero warnings: clean.

Cross-instance tests cover:
- Two hosts writing different tasks without conflict
- Reconciliation detecting tasks created or deleted by a peer
- Per-task diff-delta preserving a peer's status change on full-object upsert
- Same-field last-writer-wins behavior (documented, not a bug)
- Delete detection when only `history_item.json` is removed (directory remains)

### Pre-Submission Checklist

- [x] **Visual Snapshot** (UI changes only): Not applicable; this PR has no UI changes.

### Visual Snapshots

Not applicable; this PR has no UI changes.

### Documentation Updates

- [ ] Yes, documentation updates are required.

### Get in Touch

GitHub: @edelauna

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved task history reliability by preserving concurrent updates and preventing valid changes from being overwritten.
  - Improved recovery when task history files are missing, invalid, or removed.
  - Invalid stale status changes are now rejected safely.
  - Fixed delegation state repair and cleanup after successful recovery.
  - Roo history imports now refresh task history correctly and provide more consistent state updates.

- **Improvements**
  - Task history now loads directly from available task records for greater resilience.
  - Improved handling of simultaneous history updates and incomplete or invalid records.


martin-rueegg:
Thank you @edelauna for this PR. 

In my honest opinion, this is taking the wrong route!

Firstly, It pretends to "close" the original issue, while it only addresses the surface of the general design flaw laid out in the [original issue](1231).

But maybe more importantly, it just shifts the problem of concurrency away from the global `_index.php` to the individual task's `history_item.json`. And while doing so, it not only increases the disk IO massively, it also increases the chance of corruption!

If I have understood the solution correctly, it does:
- get the lock of the global index
- read *ALL* tasks' history_item.json, notabene *without* locking them.
- combine the result of that read, including the own newly written history_item.json
- writing the combined index back to disk
- releasing the lock.

While in theory, the tasks' history_item.json is written atomically, there is a fraction window, where the file does NOT exist during write (1. rename existing->backckup, 2. rename new->existing, 3. delete backup). In such a case the whole process may fail or the task at hand being ignored (I have not totally traced through the exception handling).

While this is recoverable, as the process which is just updating that history file will also eventually update the global index and the item will be re-inserted, it still is a potential point for future failures.

But also, I'm not sure if a directory scan of ALL files/dirs in tasks directory, the reading and parsing of ALL history_item.jsons it the right approach.

**Possible alternative:**

if we do a re-read of ALL tasks during every update (with 5 second window of gathering local changes), would it not be much more efficient to simply drop the global index altogether and scan the directories the few times we really need to read it (namely when displaying history index in ui)? That is not happening as often as every 5 seconds with working tasks.

These are my two cents. But I do hope we find a better solution than the one suggested here.

Nonetheless, thanks again for taking the time to resolve this issue!

edelauna:
> Thank you @edelauna for this PR.
> 
> In my honest opinion, this is taking the wrong route!
> ...

Thanks for the feedback @martin-rueegg - saying this closes #1231 was maybe an over reach - I'll think on the feedback provided as better solution to the issue.

src/core/task-persistence/TaskHistoryStore.ts
```diff
@@ -1105,16 +1005,23 @@ export class TaskHistoryStore {
 			const mergedFirst = { ...first, ...updatedFirst }
 			const mergedSecond = { ...second, ...updatedSecond }
 
+			// Compute actual diffs against cached state, mirroring upsertCore.
+			const deltaFirst = Object.fromEntries(
+				Object.entries(updatedFirst).filter(([k, v]) => !deepEqual(v, (first as Record<string, unknown>)[k])),
```
martin-rueegg: to have a single point of truth, AND make it testable in isolation, would it not be good to extract the three delta calculations into one local private method?
same for line 1013 and 241?

src/core/task-persistence/TaskHistoryStore.ts
```diff
@@ -845,96 +805,36 @@ export class TaskHistoryStore {
...
+	 * process are preserved. Without a delta the full item is written
+	 * as-is (used by administrative repair paths that are authoritative).
 	 */
-	private async writeTaskFile(item: HistoryItem): Promise<void> {
+	private async writeTaskFile(item: HistoryItem, delta?: Partial<HistoryItem>): Promise<void> {
 		const filePath = await this.getTaskFilePath(item.id)
-		await safeWriteJson(filePath, item)
+		if (delta) {
+			await safeWriteJson(filePath, item, {
+				merge: (existing, incoming) => {
```
martin-rueegg, line 908: I wonder if it would be good to extract the merge functionality into a local function of it's own to make it testable in isolation

martin-rueegg:
Thank you, @edelauna, for having considered my feedback. I think the current solution is much cleaner and safer. Awesome work!

It is still worth considering in a next phase, to maybe have a per-workspace index, only including the task-ids. this could allow the "workspace" filter field to be populated easily and speed up the index scan, if only a specific workspace is of interest. But that is an addition, rather than a fix. So maybe we could create a follow-up PR once the fix is reviewed and merged.

Thank you for the effort. 🙏 - Having regular issues with my index, I'm looking forward to get this patch ASAP.

src/core/task-persistence/TaskHistoryStore.ts
```diff
@@ -257,13 +246,11 @@ export class TaskHistoryStore {
 		// Merge: preserve existing metadata unless explicitly overwritten
 		const merged = existing ? { ...existing, ...item } : item
 
-		// Write per-task file (source of truth)
-		await this.writeTaskFile(merged)
+		const delta = existing ? ({ id: item.id, ...this.computeDelta(existing, item) } as HistoryItem) : undefined
+		await this.writeTaskFile(merged, delta)
```
taltas: What prevents a stale host that cached `active` from applying a `delegated` status delta after another host has already written terminal `completed`? Please validate status transitions against the record read under the file lock, and cache/publish the actual merged record so a preserved peer status is not left stale in memory.

taltas: Nit: can the nearby `upsert()` comment that still says this schedules a debounced index write, plus the reconciliation comment that says it scans directories “vs index,” be updated now that this PR removes the index and its scheduler?

taltas (changes requested):
The shared-index removal is a strong simplification, but the per-task merge still validates lifecycle transitions against stale host-local state. The inline blocker below reproduces a completed task being moved back to delegated; please fix that before merging. I also left one documentation nit.

src/utils/safeWriteJson.ts
martin-rueegg, line 45: I wonder if we could change the signature to return `Promise<SafeWriteResult>` where

```typescript
interface SafeWriteJsonResult {
	/**
	 * Number of bytes written to the new file.
	 * @see fs.WriteStream.bytesWritten
	 */
	bytesWritten: number

	/**
	 * The data as ist was passed to JSON.stringify.
	 * If `SafeWriteJsonOptions.merge` callback was passed, this is the result of that function call. The incoming `data` otherwise.
	 * @see fs.WriteStream.bytesWritten
	 */
	dataWritten: any

	/**
	 * A future possible addition:
	 *
	 * Path to the `.bak_*.tmp` file if a backup was created on overwrite.
	 * Null if the target file did not exist (first write).
	 * The caller is responsible for renaming or deleting this file.
	 */
	backupPath: string | null
}
```

this would allow consumers to update their in-memory cache.

Maybe that could be a solution to https://github.com/Zoo-Code-Org/Zoo-Code/pull/1261/changes#r3809557026, if I understood that comment correctly.

PS: the backupPath is an example, since I use this in my repairTool after extending `SafeWriteJsonOptions` with

```typescript
	/**
	 * When true, the backup file is NOT deleted on success — its path is returned
	 * in the result so the caller can rename or delete it.
	 * When false or undefined (default), the backup is deleted as in the original implementation.
	 * @default false
	 */
	keepBackup?: boolean
```

src/core/task-persistence/__tests__/TaskHistoryStore.crossInstance.spec.ts
```diff
@@ -164,4 +164,64 @@ describe("TaskHistoryStore cross-instance safety", () => {
...
+		expect(taskDirs.sort()).toEqual(["task-a", "task-b"])
+
+		// Reconcile rebuilds the full picture from authoritative per-task files.
+		await storeA.reconcile({ forceRefresh: true })
+		expect(storeA.get("task-a")).toBeDefined()
+		expect(storeA.get("task-b")).toBeDefined()
+		expect(storeA.getAll()).toHaveLength(2)
+
+		// A post-reconcile flush writes the complete set.
+		await storeA.flushIndex()
```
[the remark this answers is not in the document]
edelauna: Addressed — the shared `_index.json` was removed entirely in an earlier revision. The current approach uses per-task `history_item.json` files as the source of truth, with `reconcile()` scanning task directories. The concurrent lost-update scenario is now covered by the `mergeWithDisk` merge callback under the advisory lock.

src/core/task-persistence/TaskHistoryStore.ts
```diff
@@ -880,17 +890,44 @@ export class TaskHistoryStore {
...
+				const prev = existing as HistoryIndex | null
+				const next = incoming as HistoryIndex
+				if (!prev || prev.version !== 1 || !Array.isArray(prev.entries)) {
+					return next
+				}
+				const ourIds = new Set(next.entries.map((e) => e.id))
+				const peerEntries = prev.entries.filter((e) => e.id && !ourIds.has(e.id) && onDiskIds.has(e.id))
+				return {
+					...next,
+					entries: [...next.entries, ...peerEntries].sort((a, b) => b.ts - a.ts),
```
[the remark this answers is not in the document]
edelauna: Addressed — `reconcile()` now keys liveness on `readTaskFile` returning a valid record with `item.id === taskId`, not on directory existence. Deleted tasks whose `history_item.json` was unlinked are correctly evicted even when the task directory remains.

src/utils/safeWriteJson.ts
```diff
@@ -78,6 +88,19 @@ async function safeWriteJson(filePath: string, data: any, options?: SafeWriteJso
...
+	if (options?.merge) {
+		let existing: unknown = null
+		try {
+			const raw = await fs.readFile(absoluteFilePath, "utf8")
+			existing = JSON.parse(raw)
+		} catch {
+			// No readable file yet, so the merge receives null.
+		}
+		data = options.merge(existing, data)
+	}
```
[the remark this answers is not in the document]
edelauna: Addressed in commit a066d55 — the merge block was moved inside the `try`/`finally` that releases `releaseLock`, so a throwing merge callback still releases the lock.

src/core/task-persistence/TaskHistoryStore.ts
edelauna: Done — extracted a `buildDelta(id, cached, incoming)` helper that combines `id` + `computeDelta()`. All three call sites now use it.

src/core/task-persistence/TaskHistoryStore.ts
edelauna: The merge logic lives in the module-level `mergeWithDisk()` function, which is already isolated and testable. `writeTaskFile` delegates to it via a closure. The cross-instance spec now exercises it end-to-end (including the `DeltaRejectedError` throw path).

src/utils/safeWriteJson.ts
```diff
@@ -83,6 +93,19 @@ async function safeWriteJson(filePath: string, data: any, options?: SafeWriteJso
...
+		// If a merge callback was provided, read the current file under the lock
+		// and let the caller merge before we write. Must be inside try/finally
+		// so a throwing merge still releases the lock.
+		if (options?.merge) {
+			let existing: unknown = null
+			try {
+				existing = JSON.parse(await fs.readFile(absoluteFilePath, "utf8"))
+			} catch {
+				// No readable file yet, so the merge receives null.
+			}
```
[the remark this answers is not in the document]
edelauna: Addressed in commit a066d55 — the catch block now distinguishes `ENOENT` and `SyntaxError` (treated as null) from other filesystem errors like `EACCES`/`EIO` (rethrown). A regression test covers the `EIO` case.

src/core/task-persistence/TaskHistoryStore.ts
edelauna: Addressed — three layers of protection now:

1. `mergeWithDisk` validates the transition against the disk record under the advisory lock. If the disk status rejects the transition, it throws `DeltaRejectedError` (the entire delta is dropped, not just the status field).
2. `upsertCore` re-reads disk when the cache-side check fails, so a stale cache cannot block a valid completion.
3. `writeTaskFile` returns the actual merged record and callers cache that value, so the in-memory cache always reflects what was persisted.

A cross-instance regression test verifies that a stale host attempting `completed → delegated` gets `DeltaRejectedError` and the disk/cache both retain `completed`.

edelauna: Done — updated four stale doc comments:
- `initialize()`: "Load index" → "Scan task files"
- `upsert()`: removed "schedules a debounced index write"
- `reconcile()`: "vs index" → "between disk and cache"
- `reconcileDelegationStateCore()`: removed "index" reference

src/utils/safeWriteJson.ts
edelauna: Acknowledged — filed as #1304 for a follow-up. For now, writeTaskFile captures the merge result via a closure and returns it, which lets callers update their cache. Changing safeWriteJson to return SafeWriteResult would be a cleaner API but is a broader change across all consumers.

src/core/task-persistence/TaskHistoryStore.ts
```diff
@@ -34,12 +34,33 @@ export function assertValidTransition(from: HistoryItemStatus | undefined, to: H
...
+					)
+					return disk
+				}
+			}
+		}
+		const merged = { ...disk, ...delta }
+		if (delta.childIds && disk.childIds) {
+			merged.childIds = [...new Set([...disk.childIds, ...delta.childIds])]
+		}
+		return merged
```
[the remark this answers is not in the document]
edelauna: Addressed — `mergeWithDisk` now throws `DeltaRejectedError` instead of silently returning `disk`. `upsertCore` catches it, re-reads disk to update the cache, and re-throws. `atomicUpdatePair` propagates the error: if the first write is rejected, the second never runs; if the second is rejected after the first succeeded, the first record's cache is updated before the error propagates. `reconcileDelegationState` repairs the pair on next startup.

src/core/task-persistence/TaskHistoryStore.ts
```diff
@@ -250,20 +256,25 @@ export class TaskHistoryStore {
...
-				assertValidTransition(existing.status, item.status)
+				try {
+					assertValidTransition(existing.status, item.status)
+				} catch {
+					// Cache may be stale from a peer write. Re-read disk
+					// under the store lock before rejecting the transition.
+					const diskItem = await this.readTaskFile(item.id)
+					assertValidTransition(diskItem?.status, item.status)
+				}
 			}
```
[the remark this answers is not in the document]
edelauna: Addressed — the catch block now rethrows the original cache error when `readTaskFile` returns null, preventing a missing file from normalizing to `active` and accepting an invalid transition.

src/core/task-persistence/TaskHistoryStore.ts
```diff
@@ -250,20 +256,25 @@ export class TaskHistoryStore {
...
-		await this.writeTaskFile(merged)
+		const delta = existing ? this.buildDelta(item.id, existing, item) : undefined
+		const written = await this.writeTaskFile(merged, delta)
 
-		// Update in-memory cache
-		this.cache.set(merged.id, merged)
-		// Schedule debounced index write
-		this.scheduleIndexWrite()
+		// Update in-memory cache with what was actually persisted
+		this.cache.set(written.id, written)
```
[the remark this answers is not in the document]
edelauna: Addressed — `upsertCore` now always passes a delta to `writeTaskFile`, even without a cache entry (`{ ...item }` as the delta). This routes first-write-after-eviction through the merge path, preserving peer-written fields.

src/core/task-persistence/TaskHistoryStore.ts
```diff
@@ -369,31 +374,35 @@ export class TaskHistoryStore {
...
+					// held for the entire write, so its presence means a
+					// write is in progress — keep the task live.
+					try {
+						const lockPath = (await this.getTaskFilePath(taskId)) + ".lock"
+						await fs.access(lockPath)
+						liveIds.add(taskId)
+					} catch {
+						// No lock file — file is genuinely absent
+					}
 				}
```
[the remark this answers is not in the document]
edelauna: Addressed — the lock-file liveness check now compares `lockStat.mtimeMs` against `Date.now()` with a 31-second threshold (matching `safeWriteJson`'s stale lock window). Older lock files are treated as crash residue and ignored.

taltas (approved):

martin-rueegg (approved):
All my comments have been resolved. Thank you for the great work!

navedmerchant (approved):

PR #1265, opened 2026-08-17
test(e2e): cover auto-flatten inline behavior across the extension host

easonLiangWorldedtech:
Part 3/8 of the task-tree series (builds on parts 1-2). Two commits: feature + e2e coverage. When a new_task call would exceed maxNestingDepth and autoFlattenOnLimit is set, the subtask is flattened into the current conversation instead of opening a new tab; nested new_task calls during an active phase are rejected with guidance to complete the inline work first. Commit 2 adds e2e coverage across the real extension host: a three-level new_task chain where the depth-2 task's own call is flattened - asserts exactly 3 tasks exist, the flattened work completes in-conversation, parents resume correctly, and no fourth tab opens (hand-written aimock fixtures, no API key needed).


easonLiangWorldedtech:

PR #16, opened 2026-05-04
Add DeepSeek V4 models and dynamic model fetching

capitanfeeder:
Captura de pantalla 2026-05-04 155058
DeepSeek recently released V4 models (v4-pro and v4-flash) with 1M context windows and significantly better pricing. The old V3.2 models (deepseek-chat and deepseek-reasoner) are being deprecated on 2026/07/24.

Changes:
- Add deepseek-v4-pro and deepseek-v4-flash to the model definitions with correct specs (1M context, 384K output, updated pricing)
- Mark deepseek-chat and deepseek-reasoner as deprecated
- Move DeepSeek from static to dynamic provider so models are fetched from the API endpoint instead of being hardcoded — this way new models will show up automatically without code changes
- Create a fetcher (src/api/providers/fetchers/deepseek.ts) that calls GET /models and merges the response with known specs as fallback
- Update the DeepSeek handler to detect thinking mode via the preserveReasoning flag instead of string matching on model ID, so V4 models get thinking mode enabled automatically
- Update the DeepSeek settings component to use ModelPicker with dynamic models from the router
- Default model changed from deepseek-chat to deepseek-v4-pro

### Related GitHub Issue

Closes: # 

### Pre-Submission Checklist

- [ ] **Issue Linked**: This PR is linked to an approved GitHub Issue (see "Related GitHub Issue" above).
- [ ] **Scope**: My changes are focused on the linked issue (one major feature/fix per PR).
- [ ] **Self-Review**: I have performed a thorough self-review of my code.
- [ ] **Testing**: New and/or updated tests have been added to cover my changes (if applicable).
- [ ] **Documentation Impact**: I have considered if my changes require documentation updates (see "Documentation Updates" section below).
- [ ] **Contribution Guidelines**: I have read and agree to the [Contributor Guidelines](/CONTRIBUTING.md).

## Summary by CodeRabbit

* **New Features**
  * DeepSeek added as a dynamic provider with live model fetching, cache support, and inclusion in router model responses
  * New DeepSeek v4 models (v4-pro, v4-flash); default switched to v4-pro; legacy models marked deprecated
  * Model Picker for DeepSeek added to settings with additional validation props

* **Tests**
  * Updated tests to explicitly cover DeepSeek model selection, defaults, fallbacks, and router-model inclusion

* **UX**
  * DeepSeek uses a custom model selection UI and validates dynamic model choices before falling back to defaults

PR #1268, opened 2026-08-17
feat(task): surface nesting depth and parent task id to the model (#16)

easonLiangWorldedtech:
Part 6/8 of the task-tree series (builds on parts 1-5). The per-request environment_details now includes a Task Context block with the task's nesting depth (root = 0) and, for children, its parent task id - so a subtask can see its own position in the delegation tree. Parent title resolution is deliberately avoided to keep every environment refresh free of history-store lookups.


easonLiangWorldedtech:

Issue #18, opened 2026-05-05
[Refactor] Extract token & tool usage tracking into UsageTracker from Task.ts


doctarock:
## Parent Issue
Part of #8 — Code Refactor on Monolith Files

## Summary
Extract the token and tool usage accumulation cluster from `src/core/task/Task.ts` into a dedicated `UsageTracker` module.

## Scope
**Methods to extract:**
- `getTokenUsage()`
- `recordToolUsage()`
- `recordToolError()`
- `emitFinalTokenUsageUpdate()`

**State owned:**
- `tokenUsageSnapshot?`
- `tokenUsageSnapshotAt?`
- `toolUsageSnapshot?: ToolUsage`
- `debouncedEmitTokenUsage` (debounced throttle function)
- `TOKEN_USAGE_EMIT_INTERVAL_MS = 2000`

**Estimated size:** ~150–200 lines

## Notes
Simple accumulator pattern with a clear boundary. No circular dependency risk. Good candidate for unit testing once isolated.

## Constraints (from #8)
- Preserve method names and call signatures
- Run type checks after extraction
- One extraction per commit

PR #1270, opened 2026-08-17
Fix delegation link loss across interrupt/resume; type NewTaskTool provider call (#18)

easonLiangWorldedtech:
Part 8/8 of the task-tree series (builds on parts 1-7). Resuming an interrupted delegated child no longer loses its parent delegation link: reestablishDelegationLinkOnResume restores the legal active->delegated transition after markDelegatedChildInterrupted, skipping cancelled children and never clobbering a live other-child delegation. Also types the NewTaskTool provider call (removes the as-any cast).


easonLiangWorldedtech:

PR #1008, opened 2026-07-24
feat(api): introduce RequestConfigBuilder for SDK-agnostic abort signal support

easonLiangWorldedtech:
# feat(api): introduce RequestConfigBuilder for SDK-agnostic abort signal support

## Problem

Currently each provider manually constructs its request config by spreading `{ signal: metadata?.abortSignal, headers: ... }` — a repetitive pattern that:

1. **Duplicates logic** across every provider implementation
2. **Is error-prone** — easy to forget or mistype the spread
3. **Doesn't handle SDK differences well** — OpenAI uses `queryParams`, Bedrock uses `body`, Anthropic uses `apiVersion`
4. **Lacks immutability guarantees** — direct spreading can accidentally mutate original config objects

## Proposed Solution

Introduce `RequestConfigBuilder` — a generic, SDK-agnostic request configuration builder that provides a fluent API for building type-safe request configurations.

### Usage Example

```typescript
const config = new RequestConfigBuilder()
    .addAbortSignal(metadata)
    .addHeaders({ "X-Custom-Header": "value" })
    .setOption("modelId", "claude-3-opus")
    .build()
```

### Key Features

- **Fluent chainable API** — concise, readable code
- **Generic type safety** — `TOptions extends Record<string, any>` ensures compile-time correctness per SDK
- **Zero runtime overhead** — Generics erased at compile time; `mergeAbortSignals()` returns primary directly when no secondary provided
- **Immutability guaranteed** — `build()` returns shallow copy
- **Defensive by default** — undefined/empty values are silently skipped, never thrown

#### Instance Methods (all chainable)
- `addAbortSignal(metadata?)` — Extracts abort signal from metadata and adds it to config
- `addHeaders(headers?)` — Merges custom headers (empty objects skipped, no default param overhead)
- `setOption(key, value)` — Type-safe single option setter
- `getOption(key)` — Get an option by key
- `build()` — Returns shallow copy for immutability; `undefined` if no options set
- `addMergedSignal(internalController, metadata?, timeoutMs?)` — Merge internal controller + external signal with timeout support (for providers that maintain their own AbortController)

#### Static Methods
- `fromMetadata(metadata?, extraOptions?)` — Quick factory for simple signal + options scenarios
- `mergeAbortSignals(primary, secondary?)` — Combines two signals (returns primary directly when no secondary — zero overhead)

## Files to Add/Modify

| File | Description |
|------|-------------|
[...]

PR #901, opened 2026-07-15
feat(api): add CompletePromptOptions parameter to completePrompt method

easonLiangWorldedtech:
## Summary

Fixes #615 — Core plumbing for abort signal in `completePrompt` path.

This PR adds an optional second parameter (`CompletePromptOptions`) to all provider `completePrompt` methods, enabling future abort signal and timeout support without breaking existing callers.

### New: `CompletePromptOptions` interface
```ts
export interface CompletePromptOptions extends Pick<ApiHandlerCreateMessageMetadata, "abortSignal"> {
    /** Optional timeout override (ms) — falls back to provider default if omitted */
    timeoutMs?: number
}
```

**Design note:** Rather than passing the full `ApiHandlerCreateMessageMetadata` as issue #615 described, this PR introduces a minimal interface that only exposes `abortSignal` and `timeoutMs`. This keeps the API surface clean while still enabling abort signal pass-through. The `metadata` pattern used in the stream path (via `ApiHandlerCreateMessageMetadata`) is already handled separately.

### Updated files
- **`src/api/index.ts`** — Added `CompletePromptOptions`, updated `SingleCompletionHandler.completePrompt` signature
- **25+ provider implementations** — All now accept `options?: CompletePromptOptions` (backward compatible, optional)
- **`src/utils/single-completion-handler.ts`** — Forwards options to the handler

### Tests (`Task.spec.ts`)
- Updated existing assertions from `toBeInstanceOf(AbortSignal)` → `toBe(task.currentRequestAbortController!.signal)` for signal identity verification
- Added new test: "should create a fresh AbortController for each sequential request" confirming controllers are not reused across requests

## Acceptance criteria (from #615)
- [x] `SingleCompletionHandler.completePrompt` interface accepts optional metadata/options
- [x] `single-completion-handler.ts` forwards options to the provider
- [x] `Task.ts` attaches abortSignal to metadata (already wired in #434, this PR completes the interface layer)
- [x] Tests verify signal identity and fresh-controller-per-request

## Related
- Part of #404 — Stop does not abort underlying HTTP request
- Depends on #434 (abort signal core plumbing already merged)

## Summary by CodeRabbit

* **New Features**
  * Added optional structured controls for non-streaming prompt completion via a shared `options` parameter, including `abortSignal` and optional `timeoutMs`.
  * Updated `completePrompt` across providers and the single-completion flow to accept and forward these options.

* **Tests**
[...]

PR #1290, opened 2026-08-19
feat(api): abort signal support for openai-codex (completePrompt + createMessage)

easonLiangWorldedtech:
Adds abort signal + timeout support to the OpenAI Codex provider. completePrompt now uses a request-local signal built from options.abortSignal/timeoutMs (merged via the shared mergeAbortSignalAndTimeout util, imported directly from utils/abort-signal like Bedrock), and createMessage bridges metadata.abortSignal into the provider's internal request AbortController using the Bedrock pattern, covering both the OpenAI SDK streaming path and the manual SSE fetch fallback.

- Provider: src/api/providers/openai-codex.ts
  - completePrompt: throwIfAborted(options?.abortSignal) fast-fail before the first await (a pre-aborted request never spends the OAuth token/account setup or an SDK request); request-local signal via mergeAbortSignalAndTimeout(options?.abortSignal, options?.timeoutMs) imported directly from utils/abort-signal (the Bedrock pattern); top-of-loop abort break on the streaming consumer loop so a buffered post-abort chunk is never joined into the completion; catch normalizes via isRequestAborted(error, requestSignal) to the shared abort contract (name === "AbortError"); handler-wide this.abortController no longer used on the fetch path
  - createMessage/executeRequest: metadata param added; external abortSignal bridged into the internal controller (pre-aborted guard + { once: true } listener)
- Tests:
  - src/api/providers/__tests__/openai-codex.spec.ts: ported reference completePrompt suite (request body/headers, timeoutMs=0 no-timeout, abortSignal and abortSignal+timeoutMs merging, empty/text-fallback outputs, unauthenticated and non-ok error paths, reasoning config, ChatGPT-Account-Id header cases), plus focused tests that a pre-aborted signal and an in-flight abort both reject with name === "AbortError"; a fast-fail test (pre-aborted signal rejects before the deferred OAuth resolution and the SDK call) and a structural kill test for the top-of-loop guard (pull-count assertion - the abort rides in on the in-flight chunk after executeRequest's check has passed)
  - src/api/providers/__tests__/openai-codex-native-tool-calls.spec.ts: createMessage abort bridge test (external signal propagates to the internal controller signal) and pre-aborted test (internal signal already aborted before the request starts)


easonLiangWorldedtech:

**Status: adoption commit in flight on this branch.** A mechanical call-site refactor routing the openai-codex abort wiring through `RequestConfigBuilder` is being pushed to this PR before merge; this flag is resolved by that commit.

easonLiangWorldedtech:

Part of the abort-signal series addressing #404 (builds on #674, #901, #1008). openai-codex abort wiring + config-builder retrofit.


- **Final head:** `0ca6c2327` (rebased onto main `252c69b52`)
- **Work in this round:** request-local abort bridging in `createMessage` (per-request `AbortController`, named abort listener removed in `finally` on both paths — never the class-field controller) and `completePrompt`; CodeRabbit minor (primary-signal coverage) addressed.
- **Config builder:** `completePrompt` now routes through `RequestConfigBuilder.mergeAbortSignalAndTimeout`. This commit adds the two builder statics (`mergeAbortSignalAndTimeout` / `mergeAbortSignals`) delegating to `utils/abort-signal.ts`, plus the shared spec block — byte-identical to #1291's additions, so either merge order is conflict-free.
- **Changed-line coverage:** 33/33 executable changed lines covered (100%), including the new builder static bodies. 111 provider/builder tests green.

easonLiangWorldedtech:
@coderabbitai approve

src/api/providers/openai-codex.ts
```diff
@@ -438,10 +439,21 @@ export class OpenAiCodexHandler extends BaseProvider implements SingleCompletion
...
 
+		// Bridge the external abort signal into the internal controller (Bedrock pattern)
+		const externalAbortSignal = metadata?.abortSignal
+		if (externalAbortSignal) {
+			if (externalAbortSignal.aborted) {
+				this.abortController.abort()
+			} else {
+				externalAbortSignal.addEventListener("abort", () => this.abortController?.abort(), { once: true })
+			}
+		}
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Fixed in 714754cfa: executeRequest now owns a request-local AbortController; the caller-signal bridge listener captures it by closure, removes itself with the exact same reference in the finally block, and the shared field is only cleared when it still references this request's controller.

src/api/providers/openai-codex.ts
```diff
@@ -1354,12 +1367,15 @@ export class OpenAiCodexHandler extends BaseProvider implements SingleCompletion
 			const apiError = new ApiProviderError(errorMessage, this.providerName, errorModel.id, "completePrompt")
 			TelemetryService.instance.captureException(apiError)
 
+			// Re-throw abort errors as-is so callers can detect cancellation by the "AbortError" name
+			if (error instanceof Error && error.name === "AbortError") {
+				throw error
+			}
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Fixed in 714754cfa: completePrompt merges the caller signal and timeout via RequestConfigBuilder.mergeAbortSignalAndTimeout and normalizes an aborted request to createAbortError(this.providerName) on both the quiet path and the catch path, so no raw DOMException/SDK abort error leaks out.

src/api/providers/__tests__/openai-codex-native-tool-calls.spec.ts
```diff
@@ -523,4 +523,109 @@ describe("OpenAiCodexHandler native tool calls", () => {
...
+
+			// Wait until the request has started; the bridge listener is registered before
+			// the SDK call, so aborting now lands mid-flight
+			await vi.waitFor(() => expect(mockCreate).toHaveBeenCalled())
+
+			// Abort the external signal mid-flight; the bridge must abort the request-local controller
+			controller.abort()
+
+			const chunks = await collected
+			expect(chunks.length).toBeGreaterThan(0)
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Fixed in 714754cfa: the bridge test now asserts exactly one pre-abort delta chunk ([{ type: "text", text: "test" }]) and the pre-aborted test asserts an empty chunk list — the mock transport pauses on the captured signal, so the abort can only block events after the first delta.

src/api/providers/__tests__/openai-codex-native-tool-calls.spec.ts
```diff
@@ -523,4 +523,113 @@ describe("OpenAiCodexHandler native tool calls", () => {
...
+			// the SDK call, so aborting now lands mid-flight
+			await vi.waitFor(() => expect(mockCreate).toHaveBeenCalled())
+
+			// Abort the external signal mid-flight; the bridge must abort the request-local controller
+			controller.abort()
+
+			const chunks = await collected
+			// Exactly the pre-abort delta: the completed event only arrives once the abort resolves
+			// the transport's pause, so the loop must break before processing it.
+			expect(chunks).toEqual([{ type: "text", text: "test" }])
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Fixed in 1aa5fcd58: the active-abort test's mock now yields an output-bearing post-abort delta (`response.text.delta` "post-abort") after the abort resolves the transport's pause, and the test asserts the emitted chunks contain exactly the pre-abort delta — the assertion fails if the SDK loop guard processes the completed event or the post-abort delta. Mutation-diff gate is green on this head (run 33953145011).

src/api/providers/__tests__/openai-codex.spec.ts
```diff
@@ -591,6 +592,72 @@ describe("OpenAiCodexHandler.completePrompt streaming", () => {
...
+		})
+		vitest.stubGlobal("fetch", mockFetch)
+
+		const completion = handler.completePrompt("Hello", { abortSignal: controller.signal })
+		await vitest.waitFor(() => expect(mockFetch).toHaveBeenCalled())
+		controller.abort()
+
+		await expect(completion).rejects.toMatchObject({ name: "AbortError" })
+		expect(captureException).not.toHaveBeenCalled()
+	})
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Fixed in 1aa5fcd58: a new `createMessage` abort-bridging describe block drives the transports directly (completePrompt re-normalizes any error into the shared abort contract once the caller's signal has fired, which would mask regressions in the loop checks): (1) fetch now returns a streaming `ReadableStream` SSE body — "stops reading the fallback stream once the request aborts" delivers the first chunk, aborts between two reads (exactly where the loop's guard runs), and asserts no post-abort output; (2) "hands back the abort contract when the fallback stream tears down while the caller cancels" covers a reader failure during abort → AbortError with no telemetry; (3) "wraps a torn-down fallback stream as a stream error when nothing was aborted" covers a reader failure without abort → provider error with telemetry (the double `captureException` count is the witness that the abort path was not taken). Additionally, `handleStreamResponse`'s `abortSignal` parameter is now required (`AbortSignal` instead of optional), removing the optional-chaining variants of the guard. Mutation-diff gate is green on this head (run 33953145011).

src/api/providers/config-builder/request-config-builder.ts
```diff
@@ -163,4 +163,19 @@ export class RequestConfigBuilder<TOptions extends RequestConfigOptionsBase = Re
...
+	 * the timeout; native AbortSignal.timeout self-manages its timer) stay
+	 * single-sourced.
+	 */
+	static mergeAbortSignalAndTimeout(externalSignal?: AbortSignal, timeoutMs?: number): AbortSignal | undefined {
+		return mergeAbortSignalAndTimeout(externalSignal, timeoutMs)
+	}
+
+	static mergeAbortSignals(primarySignal: AbortSignal, secondarySignal?: AbortSignal): AbortSignal {
+		return mergeAbortSignals(primarySignal, secondarySignal)
+	}
```
edelauna, line 180: Does `mergeAbortSignals` have any production callers? Looking at the codebase it only appears in this file's own spec. If nothing calls it in production, could we drop it and let callers import `mergeAbortSignals` directly from `utils/abort-signal` when they need it?

src/api/providers/openai-codex.ts
```diff
@@ -1329,6 +1370,10 @@ export class OpenAiCodexHandler extends BaseProvider implements SingleCompletion
 	 * from having to be duplicated here.
 	 */
 	async completePrompt(prompt: string, options?: CompletePromptOptions): Promise<string> {
+		// Merge an optional timeout into the caller's abort signal so a timeout cancels the
+		// completion the same way an external abort does (timeoutMs <= 0 disables it).
+		const requestSignal = RequestConfigBuilder.mergeAbortSignalAndTimeout(options?.abortSignal, options?.timeoutMs)
```
edelauna, line 1375: Bedrock reaches `mergeAbortSignalAndTimeout` via a direct import from `./utils/abort-signal` rather than through `RequestConfigBuilder`. Would it make sense to do the same here — that would also let us drop the `RequestConfigBuilder` import on line 32?

```suggestion
		const requestSignal = mergeAbortSignalAndTimeout(options?.abortSignal, options?.timeoutMs)
```

edelauna (changes requested):
thanks - had 2 nits

src/api/providers/config-builder/request-config-builder.ts
easonLiangWorldedtech: Yes — fixed in 0183b943a. Once openai-codex stopped routing through the builder, the static pass-throughs had no production callers left: `mergeAbortSignalAndTimeout` had exactly one (completePrompt, which the follow-up below switches to a direct import) and `mergeAbortSignals` had none. Both static methods are dropped and `request-config-builder.ts` + its spec are reverted to main, so this PR no longer touches them (diff goes 5 files -> 3). The merge-helper behavior stays covered by the util's own spec, `src/api/providers/utils/__tests__/abort-signal.spec.ts`, which the remaining callers (`addMergedSignal`, `bedrock`, and now `openai-codex.completePrompt`) import directly from `utils/abort-signal`.

src/api/providers/openai-codex.ts
easonLiangWorldedtech: Done — 0183b943a applies exactly that: `completePrompt` now calls `mergeAbortSignalAndTimeout(options?.abortSignal, options?.timeoutMs)` imported directly from `./utils/abort-signal` (the Bedrock pattern), and the `RequestConfigBuilder` import is gone.

While in this area I also closed the gaps the series alignment check flags in `completePrompt` (introduced by this PR):

- **73b2b2374** — `throwIfAborted(options?.abortSignal)` fast-fail before the first await, matching the sibling units: a pre-aborted request no longer spends the OAuth token/account setup or an SDK request on a dead completion (the regression test defers the OAuth resolution to prove the fast-fail does not wait on it); a top-of-loop abort break on the streaming consumer loop so a buffered post-abort chunk is never joined into the completion; and catch normalization via `isRequestAborted(error, requestSignal)` like the sibling units.
- **641400265** — the structural kill test for that top-of-loop guard: it asserts the pull count (event2's `delta` getter fires the abort after `executeRequest`'s own check has passed and before `completePrompt`'s; the post-loop check throws the same `AbortError` either way, so the pull count is the observable the guard is measured against).

Gates on the final head (`641400265`): focused vitest suites pass (openai-codex 61, native-tool-calls, request-config-builder, abort-signal util), `zdt align check` exits 0 with no divergences, and the local Stryker mutation-diff preflight is green (30 valid / 30 killed / 0 survived / 0 noCoverage).

easonLiangWorldedtech:
Addressing the two nits in this push, plus follow-up commits that round out the `completePrompt` contract:

- **0183b943a** — the nits: `completePrompt` imports `mergeAbortSignalAndTimeout` directly from `utils/abort-signal` (Bedrock pattern, `RequestConfigBuilder` import dropped), and the two static pass-through wrappers are removed from `request-config-builder.ts` (no production callers remained, so the builder files revert to main and the PR diff goes 5 files -> 3).
- **73b2b2374** — closes the gaps the series alignment check flags in `completePrompt`: `throwIfAborted` fast-fail before the first await (a pre-aborted request no longer spends the OAuth setup or an SDK request on a dead completion; regression test defers the OAuth resolution to prove it), a top-of-loop abort break on the streaming consumer loop, and catch normalization via `isRequestAborted(error, requestSignal)` to match the sibling units.
- **641400265** — structural kill test for the top-of-loop guard (pull-count assertion; the post-loop check throws the same `AbortError` either way).

Gates on `641400265`: focused vitest suites pass, `zdt align check` exit 0 (no divergences), local Stryker mutation-diff preflight green (30 valid / 30 killed / 0 survived / 0 noCoverage). Details in the thread replies.

src/api/providers/openai-codex.ts
```diff
@@ -250,7 +257,29 @@ export class OpenAiCodexHandler extends BaseProvider implements SingleCompletion
...
 
 		// Get access token from OAuth manager
-		let accessToken = await openAiCodexOAuthManager.getAccessToken()
+		// The lookup can await credential loading, token refresh, and persistence: race it against
+		// the caller's signal so a cancelled request settles immediately instead of hanging on a
+		// lookup that no longer matters.
+		let accessToken: string | null
+		try {
+			if (abortSignal) {
+				accessToken = await rejectOnAbort(
```
easonLiangWorldedtech, line 266: Addressed in `a27305b5c` (code) and `2ce9e64d4` + `bf39b3cb9` (regressions + mutation-gate pinning).

Every OAuth setup await now races the request signal through `rejectOnAbort` (the shared `utils/abort-signal.ts` series helper), and the race's settle handler removes its abort listener:

- `getAccessToken()` in `handleResponsesApiMessage()` (L266): raced against the caller's signal. The catch normalizes an abort landing while the lookup is pending — or a lookup failure racing the abort — to the shared abort contract via `isRequestAborted(error, abortSignal)`, so a cancelled request settles immediately with `AbortError` instead of hanging on credential loading/refresh/persistence or surfacing as an auth/model-fetch failure.
- `getAccountId()` in `executeRequest()` (L521): raced against the request-local controller (already bridged to the caller's signal); the abort propagates through the existing catch → retry-loop normalization.
- `getAccountId()` in `makeCodexRequest()` (L700, SSE fallback): raced against the request-local signal, so a cancelled fallback settles before the fallback fetch starts.

Regressions keep the lookup pending and assert the cancellation wins:
- pending `getAccessToken` + caller abort mid-await → the stream rejects with `AbortError` before the SDK is called; the late token resolution is ignored.
- pending `getAccountId` on the SDK path (macrotask to the account race) and on the fallback's own lookup (SDK down), with the late account resolution ignored in both.
- `completePrompt` with `{ timeoutMs }` and a pending token lookup → the timeout rejects with the shared abort contract before the lookup resolves (no abort signal: the timeout is the only cancellation source, so it must race the lookup itself).
- a non-abort token failure with no cancellation involved propagates as-is (pins the catch's normalization conditional).

One `ConditionalExpression` mutant on the catch's `isRequestAborted` is excluded with a mutator-specific directive: an abort landing in that catch already rejected the race with `createAbortError(providerName)`, which is byte-identical to the normalized re-throw, so removing the conditional changes no observable outcome (proof in the PR body; the `true` mutant is pinned by the non-abort-failure regression above).

src/api/providers/openai-codex.ts
```diff
@@ -513,6 +559,11 @@ export class OpenAiCodexHandler extends BaseProvider implements SingleCompletion
 					this.sawSdkEventInCurrentResponse = true
 
 					for await (const outChunk of this.processEvent(event, model)) {
+						// The top-of-loop check covers the wire; this one covers the buffered chunks
+						// processEvent still emits for an event whose processing raced the abort. A break would let
+						// the event loop pull the stream once more before its top check sees the abort, so hand
+						// back the abort contract immediately instead.
+						if (abortController.signal.aborted) throw createAbortError(this.providerName)
```
easonLiangWorldedtech, line 566: Addressed in `a27305b5c` (code) and `2ce9e64d4` + `bf39b3cb9` (regressions + mutation-gate pinning).

Cancellation is now checked immediately before every reachable `yield`:

- `processEvent()` consumption on both transports (SDK at L566, SSE delegate at L870): a guard at the loop top hands back the abort contract immediately. A `break` there would not be enough: Node dispatches the `abort` event on a microtask, so a `break` lets the `for await` continuation pull the wire stream one more time before the loop's top check sees the abort — and `createMessage` consumers have no consumer-side guard, so that extra chunk would reach them. The throw is the only defense on that path.
- Every direct buffered SSE result in `handleStreamResponse()` (complete-response text/reasoning/usage blocks, legacy `choices`/`item`/`usage` shapes, plain-JSON lines) is now preceded by an `abortSignal.aborted` check.
- Scope note: the typed `else if (parsed.type === "response.*")` branches of the SSE line loop are unreachable — every `response.*` type is captured by the `coreHandledEventTypes` delegate (which is itself guarded) — so guards were added on the reachable shape-based yields only, not the dead typed branches.

Regressions (on `createMessage`, which has no consumer-side guard, so the transport checks are the defense):
- an event whose `delta` getter fires the abort while processing → its buffered "post-abort" chunk never reaches the caller; the stream rejects with `AbortError` (SDK path, asserted via manual pulls).
- a second SSE line buffered in the handler's line loop when the abort lands between lines → never yielded; the stream rejects with `AbortError` (SSE delegate path).
- a complete-response line with two text items → the first streams; the abort lands; the rest never streams.
- a content test pinning, in order, every direct SSE chunk shape the fallback emits.
- table-driven per-shape regressions: for each direct chunk shape (complete-response reasoning/usage, legacy choices/item/usage, plain JSON), one line streams before the abort and the target line is buffered in the line loop; the target's pre-yield guard breaks before its chunk streams — asserted as a quiet end with nothing after the first chunk.
- an SDK stream that fails at the moment the caller cancels (the cancellation lands mid-pull, after the top-of-loop check has passed) → settles via the retry-catch's abort check with `AbortError`, no auth refresh, no SSE fallback.

One `ConditionalExpression` mutant is excluded with a mutator-specific directive on the `completePrompt` consumer guard (L1455): every yield site in both transports now checks the signal (or throws the abort contract) before yielding, and the check and the yield are synchronous, so a post-abort chunk can no longer reach that consumer — the pre-abort window the guard documents is closed by the per-yield checks, and removing the check changes no observable outcome (proof in the PR body; the `true` mutant is pinned by the multi-chunk happy paths).

PR #1291, opened 2026-08-19
feat(api): abort signal support for openai-native and openai-compatible (completePrompt + createMessage)

easonLiangWorldedtech:
## Purpose

Adds abort-signal support to the openai-native and openai-compatible providers: `completePrompt` honors `CompletePromptOptions.abortSignal`/`timeoutMs`, and `createMessage` bridges the task's external `metadata.abortSignal` into in-flight requests so task cancellation actually cancels the provider request.

## Changes

- **openai-native.ts**
  - `completePrompt`: request-local signal via `mergeAbortSignalAndTimeout(options?.abortSignal, options?.timeoutMs)` (falls back to a fresh controller signal) instead of clobbering the streaming `this.abortController`; `AbortError` is rethrown as-is so callers can identify cancellations.
  - `createMessage` paths (`executeRequest` and the `makeResponsesApiRequest` fetch fallback): Bedrock-pattern bridging of `metadata.abortSignal` into the internal controller (pre-aborted guard + `{ once: true }` listener); abort errors rethrown as-is in the fallback path.
- **openai-compatible.ts**
  - `completePrompt`: merged signal from `mergeAbortSignalAndTimeout` forwarded to the AI SDK `generateText` `abortSignal` option.
  - `createMessage`: `metadata.abortSignal` forwarded to `streamText` so in-flight streams abort on cancellation.

## Tests

- **openai-native.spec.ts** (extended): abort signal passthrough, timeout abort, streaming-controller isolation, merged signal abort, pre-aborted `AbortError`, fallback fetch pre-aborted + mid-request abort, non-Error rethrow, gpt-5.1 request-body coverage (service tier / reasoning / verbosity / prompt cache retention), response id and encrypted-content accessors.
- **openai-compatible.spec.ts** (new): completePrompt signal/timeout passthrough, `timeoutMs <= 0` disabled, pre-aborted `AbortError`, error propagation; createMessage abortSignal bridging (pass-through, absent metadata, pre-aborted, mid-request abort).


easonLiangWorldedtech:

**Status: adoption commit in flight on this branch.** A mechanical call-site refactor routing the openai-native and openai-compatible abort wiring through `RequestConfigBuilder` is being pushed to this PR before merge; this flag is resolved by that commit.

easonLiangWorldedtech:

Part of the abort-signal series addressing #404 (builds on #674, #901, #1008). openai-native + openai-compatible abort wiring + config-builder retrofit.


- **Final head:** `470273876` (rebased onto main `252c69b52`)
- **Work in this round:** abort bridging in both `openai-native` `executeRequest` paths (request-local controller; named listener removed in `finally` both paths) and in `openai-compatible` (both `createMessage` forwarding + `completePrompt`); CodeRabbit minors addressed (incl. the primary-signal spec case).
- **Config builder:** both `completePrompt` paths route through `RequestConfigBuilder.mergeAbortSignalAndTimeout`; builder statics + shared spec block are byte-identical to #1290's (verified by blob hash) so either merge order is conflict-free. The `openai-compatible` `createMessage` pass-through is intentionally kept direct — the builder writes the `signal` key while the Vercel AI createMessage options shape expects `abortSignal` (documented in code).
- **Changed-line coverage:** 40/40 executable changed lines covered (100%), including the builder statics. 152 provider/builder tests green.

easonLiangWorldedtech:
@coderabbitai approve

src/api/providers/openai-native.ts
```diff
@@ -644,6 +692,11 @@ export class OpenAiNativeHandler extends BaseProvider implements SingleCompletio
 			// Handle streaming response
 			yield* this.handleStreamResponse(response.body, model)
 		} catch (error) {
+			// Re-throw abort errors as-is so callers can identify cancellations
+			if (error instanceof Error && error.name === "AbortError") {
```
edelauna, line 706: If abort fires while `reader.read()` is pending inside `handleStreamResponse`, undici rejects with a `DOMException { name: 'AbortError' }` — but `handleStreamResponse`'s own outer catch (~line 1185) wraps it as a plain `Error` before it propagates here. So when this guard runs, `error.name` is `'Error'` not `'AbortError'`. Would that cause two `captureException` calls per user-triggered stop during SSE streaming?

src/api/providers/openai-native.ts
```diff
@@ -560,8 +589,27 @@ export class OpenAiNativeHandler extends BaseProvider implements SingleCompletio
...
+		// - { once: true }: removes the listener after the first abort to avoid leaks
+		// The listener is removed in the finally block when the request completes.
+		let abortListener: (() => void) | undefined
+		const externalAbortSignal = metadata?.abortSignal
+		if (externalAbortSignal) {
+			if (externalAbortSignal.aborted) {
+				requestController.abort()
+			} else {
+				abortListener = () => requestController.abort()
+				externalAbortSignal.addEventListener("abort", abortListener, { once: true })
```
edelauna: This bridge setup (pre-abort guard + `addEventListener({ once: true })`) is identical to the one in `executeRequest` (line 432). Would it make sense to extract a private helper so a future change to the bridge pattern only needs one edit?

src/api/providers/config-builder/request-config-builder.ts
```diff
@@ -163,4 +163,19 @@ export class RequestConfigBuilder<TOptions extends RequestConfigOptionsBase = Re
...
+	 * surface providers used before the builder existed. Delegate to the shared
+	 * utils in ../utils/abort-signal so timeout semantics (timeoutMs <= 0 disables
+	 * the timeout; native AbortSignal.timeout self-manages its timer) stay
+	 * single-sourced.
+	 */
+	static mergeAbortSignalAndTimeout(externalSignal?: AbortSignal, timeoutMs?: number): AbortSignal | undefined {
+		return mergeAbortSignalAndTimeout(externalSignal, timeoutMs)
+	}
+
+	static mergeAbortSignals(primarySignal: AbortSignal, secondarySignal?: AbortSignal): AbortSignal {
```
edelauna: Does anything in production call `RequestConfigBuilder.mergeAbortSignals`? I couldn't find a non-test caller — only the spec exercises it.

src/api/providers/__tests__/openai-native.spec.ts
```diff
@@ -483,6 +625,197 @@ describe("OpenAiNativeHandler", () => {
...
+			const result = await handler.completePrompt("Test prompt", { timeoutMs: 50 })
+
+			expect(result).toBe("response")
+			expect(requestSignal).toBeInstanceOf(AbortSignal)
+			expect(requestSignal?.aborted).toBe(true)
+		})
+
+		it("completePrompt should not replace an active streaming abort controller", async () => {
+			const activeStreamingController = new AbortController()
+			handler["abortController"] = activeStreamingController
```
edelauna: Since `completePrompt` in production never writes `this.abortController`, won't this assertion pass trivially regardless of what `completePrompt` does? Does this test actually catch any regression?

src/api/providers/__tests__/openai-compatible.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+		it("should pass the caller signal unchanged when timeoutMs is negative", async () => {
+			mockGenerateText.mockResolvedValue({ text: "response" })
+
+			const controller = new AbortController()
+			await handler.completePrompt("test prompt", { abortSignal: controller.signal, timeoutMs: -1 })
+
+			expect(mockGenerateText.mock.calls[0][0].abortSignal).toBe(controller.signal)
+		})
+
+		it("should reject with AbortError when abortSignal is already aborted before request", async () => {
```
edelauna: The mock checks `options.abortSignal?.aborted` itself and manufactures the `AbortError` — so this verifies the signal is forwarded, but would it catch a mutation that forwards the signal and then swallows the resulting error? A mock that injects a real `DOMException('aborted', 'AbortError')` might cover more ground.

edelauna (changes requested):

src/api/providers/openai-native.ts
easonLiangWorldedtech: Fixed in ba72b77c10d855a9f96dad59aacc7463299bc3d4. `handleStreamResponse` now receives the request-local `AbortController` and checks `requestController.signal.aborted` at the top of its catch, before any error-shape inspection. When it is aborted (external stop or request timeout), it throws `new DOMException("The OpenAI Native request was aborted", "AbortError")` directly, so the error keeps its `AbortError` name, is re-thrown as-is by the `makeResponsesApiRequest` guard, and is never captured at provider level - a user stop surfaces exactly one error, with no double `captureException`. Regression tests: "should surface the contract AbortError once when the fallback stream read rejects on external abort" (fetch body `read()` stays pending, external signal aborts mid-stream, asserts the name and message plus zero `captureException` calls) and "should not convert a non-abort stream error into an AbortError" (pins the wrapped behavior for genuine stream errors).

src/api/providers/openai-native.ts
easonLiangWorldedtech: Fixed in ba72b77c10d855a9f96dad59aacc7463299bc3d4. The duplicated bridge (pre-aborted guard + `addEventListener("abort", listener, { once: true })` + finally removal) is extracted from `executeRequest` and `makeResponsesApiRequest` into one private helper `attachExternalAbort(externalAbortSignal, requestController)`. It returns the detach function (or `undefined` when no external signal exists), so the finally blocks are now `detach?.()`. The helper early-returns when the signal is absent, so its `removeEventListener` is a plain call on the narrowed non-nullish signal - the two `OptionalChaining` mutants that previously required Stryker directives are no longer representable, and both directives are removed. Behavior is unchanged: all 11 abort-signal bridging specs pass unmodified.

src/api/providers/config-builder/request-config-builder.ts
easonLiangWorldedtech: Fixed in ba72b77c10d855a9f96dad59aacc7463299bc3d4. The unused `RequestConfigBuilder.mergeAbortSignals` static is deleted together with its spec block - it had zero production callers (only the in-class util usage and the spec itself referenced it). Its sibling `mergeAbortSignalAndTimeout` is kept: it still has production callers in `openai-native.ts:1575` (this branch) and `openai-codex.ts:1317` (#1290), and the shared `utils/abort-signal` implementation is untouched.

src/api/providers/__tests__/openai-native.spec.ts
easonLiangWorldedtech: Fixed in ba72b77c10d855a9f96dad59aacc7463299bc3d4. The trivial assertion is removed - `completePrompt` never writes `this.abortController`, so the test passed by construction. The bridging describe already pins the observable isolation pieces ("should register a once-only abort listener on the external signal and detach it when the SDK request completes" and "should detach the external abort listener on completion so a late abort cannot abort the request signal"), but nothing pinned abort isolation between concurrent requests, so the test is rewritten as an observable one: "should not let an earlier request's external abort affect a later request on the same handler" - request 1 streams under external signal A, request 2 starts under external signal B, A aborts while both are in flight; only request 1's controller aborts and request 2 completes normally with its own content.

src/api/providers/__tests__/openai-compatible.spec.ts
easonLiangWorldedtech: Fixed in ba72b77c10d855a9f96dad59aacc7463299bc3d4. The test now emulates what the real AI SDK does: the mock rejects with a real `new DOMException("The operation was aborted.", "AbortError")` when the forwarded `abortSignal` is already aborted, and the assertion checks both `name` and the exact message. `openai-compatible.ts` intentionally has no error-normalization layer (`completePrompt` and `createMessage` have no try/catch), so the fetch-stack `DOMException` is expected to pass through unchanged - that contract is what the test now verifies.

src/api/providers/__tests__/openai-native.spec.ts
```diff
@@ -375,6 +380,660 @@ describe("OpenAiNativeHandler", () => {
...
+				})
+				const requireController = (index: number): ReadableStreamDefaultController<Uint8Array> => {
+					const entry = openStreams[index]
+					if (!entry?.controller) {
+						throw new Error("expected fallback fetch to have started")
+					}
+					return entry.controller
+				}
+				return { openStreams, mockFetch, requireController }
+			}
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Fixed in 0488a39b7: hoisted `makeOpenStreamFetchMock()` into the shared `describe("createMessage")` scope, deleted the duplicated `OpenStream` type and `requireController` implementation, and replaced the inline fixture in "should not let a late abort from an earlier request cancel a later request" with `const { openStreams, mockFetch, requireController } = makeOpenStreamFetchMock()`. All fallback-path tests now share the single fixture; behavior is unchanged (vitest 106/106, tsc, eslint, prettier clean).

src/api/providers/__tests__/openai-native.spec.ts
```diff
@@ -375,6 +380,660 @@ describe("OpenAiNativeHandler", () => {
...
+					// { once: true }, and both detach their own listener on completion.
+					expect(addSpy).toHaveBeenCalledTimes(2)
+					expect(removeSpy).toHaveBeenCalledTimes(2)
+					for (const call of addSpy.mock.calls) {
+						expect(call[0]).toBe("abort")
+						expect(call[2]).toEqual({ once: true })
+					}
+					for (const call of removeSpy.mock.calls) {
+						expect(call[0]).toBe("abort")
+					}
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Fixed in 0488a39b7: the fallback-path test now asserts exact-reference listener identity — the two listeners registered via `addEventListener("abort")` are the exact two functions passed to `removeEventListener`, each detached exactly once (set-based equality), matching the SDK-path assertion style.

PR #1303, opened 2026-08-20
feat(api): abort signal support for gemini, mistral, lite-llm (completePrompt + createMessage)

easonLiangWorldedtech:
Add abort-signal support to the Gemini, Mistral, and LiteLLM providers.

## Changes

- **`gemini.ts`**
  - `completePrompt`: forwards `CompletePromptOptions.abortSignal` to `GenerateContentConfig.abortSignal` and `timeoutMs` to `httpOptions.timeout` (`httpOptions` is omitted entirely when nothing is set). On catch, a user-initiated abort re-throws as a standard `DOMException` with `name = "AbortError"`.
  - `createMessage`: bridges `metadata.abortSignal` into a request-local `AbortController` passed to the SDK via `config.abortSignal`. A pre-aborted signal rejects immediately with `AbortError`; the abort listener is stored in a named const and removed in `finally`.
- **`mistral.ts`**
  - `completePrompt`: forwards `abortSignal` via `fetchOptions.signal` and `timeoutMs` to the Mistral SDK `RequestOptions` (options arg omitted when empty, preserving the legacy 1-arg call shape). Abort normalization in catch as above.
  - `createMessage`: same request-local controller bridging as Gemini; the stream call only receives `{ fetchOptions: { signal } }` when a signal is present.
- **`lite-llm.ts`**
  - `completePrompt`: forwards `abortSignal` via `OpenAI.RequestOptions.signal`; `timeoutMs` is only forwarded when `> 0` because the OpenAI SDK treats a `0` timeout as an immediate abort.
  - `createMessage`: same bridging pattern; the in-flight `chat.completions.create(...).withResponse()` call receives the request signal alongside the existing `X-Zoo-Session-ID` header.
- **`vertex.ts`**: unchanged — `VertexHandler` inherits the new behavior from `GeminiHandler`.

## Tests

- Ported the reference `completePrompt` request-options coverage (gemini, vertex, gemini-handler, mistral, lite-llm specs).
- New `createMessage` bridging regression tests per provider: pre-aborted signal rejects immediately with `name = "AbortError"` (no SDK call), and a mid-flight external abort propagates into the in-flight request and surfaces as `AbortError` on the stream.
- Mistral spec additionally covers `timeoutMs: 0` forwarding (valid for the Mistral SDK, which uses a truthy check) and no-signal call-shape preservation.
- Vertex spec verifies the inherited bridging behavior.
- All 5 specs green; `tsc --noEmit` and per-file ESLint (`--max-warnings=0`) clean.

## Review feedback addressed (2026-09-11)

Per @edelauna's review, this revision leverages the shared abort-signal work already on main:

- **Catch normalization**: all six catch sites (gemini / mistral / lite-llm, `createMessage` + `completePrompt`) now use `isRequestAborted(error, signal)` — which also catches SDK-native abort errors that surface before the signal flag propagates — and throw via `createAbortError(<provider>)`.
- **Signal bridging**: the per-provider manual `AbortController` + `addEventListener`/`removeEventListener` + `finally` cleanup is replaced with `RequestConfigBuilder.addMergedSignal` (#1008) — `AbortSignal.any` internally, no manual listener management.
- **Tests**: new per-provider regression tests pin the SDK-native-abort branch; the listener-identity assertions now assert that no manual listeners are attached to the external signal.

---

## Series alignment check (zdt align check)

exit 0 — 0 ERROR. Findings recorded, not refactored:

- 3 WARN — statically unguarded first-yield sites (gemini fallback-text / lite-llm reasoning /
  mistral string-content yields). A guard before the FIRST yield of a loop iteration is dead code:
  the top-of-loop break runs without a suspension point before it, so it can never throw. The
  loop-top check + post-loop AbortError cover those sites; the structural kill-tests pin them.
- 2 DIVERGENCE — per-method mechanism signature across the series (bridging builder + signal
  discovery: gemini/mistral vs bare for lite-llm createMessage/completePrompt). Recorded as a
  design decision; aligning would widen this PR beyond its unit.


easonLiangWorldedtech:

**Status: migration in the post-merge adoption PR.** The refactor is mechanical (call-site substitution through the builder with a typed `TOptions` variant) and is deliberately kept out of this PR to preserve its already-green CI and review state.

easonLiangWorldedtech:

Part of the abort-signal series addressing #404 (builds on #674, #901, #1008). gemini / mistral / lite-llm abort wiring + shared timeout helper.


- **Final head:** `4a1307248` (rebased onto main `252c69b52`)
- **Work in this round:** abort bridging in all three providers (request-local controllers, listener cleanup in `finally`, catch normalization to `AbortError`) plus the shared `utils/request-timeout.ts` helper (`getRequestTimeoutMs`) implementing the series-wide `timeoutMs > 0` guard so a zero/undefined timeout never reaches the SDKs.
- **Config builder:** migration of the call sites to `RequestConfigBuilder` is scheduled for the post-merge adoption PR (see the config-builder status comment on this PR).
- **Changed-line coverage:** 96/96 executable changed lines covered (100%) after adding 4 focused regression tests (three completePrompt aborted-signal `AbortError` re-throws — gemini / lite-llm / mistral — and the mistral usage-chunk yield). 142 spec tests green.

easonLiangWorldedtech:
@coderabbitai review

easonLiangWorldedtech:
@coderabbitai approve

src/api/providers/gemini.ts
```diff
@@ -477,6 +560,11 @@ export class GeminiHandler extends BaseProvider implements SingleCompletionHandl
 				}
 			}
 		} catch (error) {
+			// User-initiated abort: surface a standard AbortError rather than the
+			// wrapped provider error so callers can distinguish cancellation.
+			if (metadata?.abortSignal?.aborted) {
```
edelauna: The `signal?.aborted` check misses cases where the SDK itself throws with `error.name === "AbortError"` before the signal flag propagates. The existing `isRequestAborted(error, signal)` helper in `abort-signal.ts` covers both — signal flag _and_ SDK-native abort error names. Same pattern at `mistral.ts:186` and `lite-llm.ts:367`.

```suggestion
			if (isRequestAborted(error, metadata?.abortSignal)) {
```

(Also add `isRequestAborted` to the import from `./utils/abort-signal`.)

src/api/providers/gemini.ts
```diff
@@ -613,6 +723,11 @@ export class GeminiHandler extends BaseProvider implements SingleCompletionHandl
 
 			return text
 		} catch (error) {
+			// User-initiated abort: surface a standard AbortError rather than the
+			// wrapped provider error so callers can distinguish cancellation.
+			if (options?.abortSignal?.aborted) {
```
edelauna: Same question as the `createMessage` catch — `isRequestAborted(error, signal)` also catches SDK-level abort errors (`error.name === "AbortError"`, `"APIUserAbortError"`, etc.) that may arrive before `signal.aborted` is true. Also applies at `mistral.ts:265` and `lite-llm.ts:430`.

```suggestion
			if (isRequestAborted(error, options?.abortSignal)) {
```

src/api/providers/gemini.ts
```diff
@@ -344,7 +404,30 @@ export class GeminiHandler extends BaseProvider implements SingleCompletionHandl
 			}
 		}
 
-		const params: GenerateContentParameters = { model, contents, config }
+		// Bridge the external abort signal from Task (metadata.abortSignal) into a
+		// request-local controller so the in-flight generateContentStream request
+		// can be cancelled. The @google/genai SDK merges this signal with its own
+		// timeout handling, which is preserved rather than replaced.
+		// A pre-aborted signal rejects immediately with an AbortError.
+		const externalAbortSignal = metadata?.abortSignal
```
edelauna: This 15-line bridge block is structurally identical in `mistral.ts:108` and `lite-llm.ts:274`. PR #1008 introduced `RequestConfigBuilder.addMergedSignal(internalController, metadata, timeoutMs)` specifically to reduce per-provider abort boilerplate — it uses `AbortSignal.any()` internally, so no `addEventListener`/`removeEventListener` cleanup is needed in a `finally` block. Is there a reason this PR does not use it?

edelauna (changes requested):
Why are you not leveraging previous work?

src/api/providers/gemini.ts
easonLiangWorldedtech: Fixed in b428b8132: all six catch sites (gemini / mistral / lite-llm, createMessage + completePrompt) now use `isRequestAborted(error, signal)` and normalize the throw via `createAbortError(<provider>)`. The new per-provider regression tests pin the SDK-native-abort branch you described (the SDK rejects with an abort error before the signal flag propagates), so a regression to `signal?.aborted`-only would fail the suite.

src/api/providers/gemini.ts
easonLiangWorldedtech: Same fix — the `completePrompt` catch now uses `isRequestAborted(error, signal)` + `throw createAbortError(...)`, with its own regression test pinning the SDK-native-abort branch (see the `createMessage` thread for details).

src/api/providers/gemini.ts
easonLiangWorldedtech: Fixed in b428b8132: the three manual bridges (AbortController + addEventListener/removeEventListener + finally cleanup) are replaced with `RequestConfigBuilder.addMergedSignal(new AbortController(), metadata)` — `AbortSignal.any` internally, so no manual listener management and the finally blocks are gone. The specs now assert (a) no manual listeners are attached to the external signal, and (b) the request always runs against a live request-local merged signal (the provider's own abort handle) even when no external signal is provided.

src/api/providers/lite-llm.ts
```diff
@@ -267,9 +270,20 @@ export class LiteLLMHandler extends RouterProvider implements SingleCompletionHa
...
+		// inside RequestConfigBuilder), so external aborts cancel the in-flight
+		// request without manual listener management.
+		const requestBuilder = new RequestConfigBuilder<{ signal?: AbortSignal }>()
+		requestBuilder.addMergedSignal(new AbortController(), metadata)
+		const requestSignal = requestBuilder.getOption("signal")
+
 		try {
 			const { data: completion } = await this.client.chat.completions
-				.create(requestOptions, { headers: requestHeaders })
+				.create(requestOptions, { headers: requestHeaders, signal: requestSignal })
```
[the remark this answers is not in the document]
easonLiangWorldedtech: Fixed in f0f1fa5b9 — the fix goes one step beyond the pre-check this finding suggests, for the reason the finding itself points to:

- The `fetchModel()` await in both `createMessage` and `completePrompt` now races against the request signal via `rejectOnAbort` (the shared series helper in `utils/abort-signal.ts` from #1290). An already-aborted signal rejects immediately at the race, and a signal that fires during cold-cache discovery rejects the request the moment the abort lands — the request no longer waits on `getModels()`/`refreshModels()` completion.
- The catch below normalizes an abort that lands while discovery fails via `isRequestAborted(error, signal)` → `createAbortError("LiteLLM")`, so a discovery failure during abort surfaces as the standard `AbortError` instead of a model-fetch error.
- Why not thread the signal into `fetchModel()` itself: `RouterProvider.fetchModel()` is a shared single-flight lookup, so a caller's signal there would let one request's cancellation affect other in-flight callers (the signal-isolation contract #1538 relies on). The underlying discovery keeps running and still populates the cache for concurrent callers; a cancelled request simply stops awaiting it. Same `rejectOnAbort`/`isRequestAborted` pattern #1290 applies to its OAuth setup.

Evidence:
- `lite-llm.spec.ts`: 66 tests green — including the regression that keeps model discovery pending, aborts mid-await, and asserts the stream rejects with `name: "AbortError"` while the discovery resolves harmlessly late.
- `abort-signal.spec.ts`: 25 tests green (incl. the `rejectOnAbort` contract: immediate reject on a pre-aborted signal, listener detached on settle).
- `tsc --noEmit` and per-file ESLint (`--max-warnings=0`) clean.
- Local Stryker preflight on the unit delta (base `0ec250e3d` → head `f0f1fa5b9`): 226 valid mutants, 226 killed, PASS.

PR #934, opened 2026-07-18
fix(skills): safely serialize skill frontmatter

taltas:
## Summary

- Safely serialize generated skill frontmatter as YAML so descriptions containing quotes or other YAML-sensitive characters remain valid.
- Surface visible diagnostics for malformed skills instead of silently omitting them or showing a misleading missing-name error.

## Root cause and user impact

Skill creation interpolated description text directly into double-quoted YAML frontmatter. Descriptions containing unescaped double quotes could therefore produce invalid `SKILL.md` files, causing frontmatter parsing to fail even when the required `name` field was present. The parsing failure was then obscured by misleading or absent diagnostics, leaving affected skills unavailable without explaining the actual malformed YAML.

This change uses safe YAML serialization for generated frontmatter and preserves actionable parse diagnostics so users can identify and correct malformed skill files.

## Testing and validation

- Backend skill test suites: 74 tests passed.
- Webview skill test suites: 20 tests passed.
- Package typechecks passed.
- Lint and formatting checks passed.
- No-mistakes run `01KXSJBKY4H2SA0KV63NCNQ210` completed with no findings.
- Push-time repository typechecks passed.

Fixes #859

## Summary by CodeRabbit

- **New Features**
  - Added diagnostics for malformed `SKILL.md` files, including the file path plus optional line/column and error details.
  - Malformed skills no longer block loading valid skills.
  - Skills settings now shows a warning panel listing affected files and guidance to fix them.
- **Bug Fixes**
  - Improved skill creation/frontmatter generation to reliably preserve descriptions and mode settings.
  - Improved handling of YAML-sensitive characters/formatting in skill metadata.

PR #1321, opened 2026-08-21
fix(skills): report YAML frontmatter parse errors and serialize SKILL.md safely

### LINKED ISSUE #859: [Bug] SKILL.md YAML parsing fails silently when description contains unescaped double quotes

SKILL.md frontmatter parsing fails silently when the `description` field contains unescaped double quotes (e.g., `"TDD"`, `"test-driven development"`), causing gray-matter/js-yaml to throw a YAMLException. The error message shown is misleading: "missing required 'name' field" — but the name IS present; it just couldn't be parsed because of the YAML syntax error.

Affects any user or AI creating skills via the UI (`createSkill` tool), skill-writer mode, or manual file creation when the description contains double quotes. The skill silently fails to load — no visible indication in the Skills UI that parsing failed (only shown in Developer Tools console). This is a UX issue because users have no way to discover the real error without opening DevTools.

## Reproduction steps

1. Create a skill file at `~/.roo/skills/test-skill/SKILL.md` with this content:
```yaml
---
name: test-skill
description: Use when implementing features. Triggers on: "TDD", "test-driven development"
---

# Test Skill

Instructions here.
```

2. Open Zoo Code → Settings → Skills tab
3. The skill does not appear in the list (even though `SKILL.md` exists)
4. Check Developer Tools console (`Help → Toggle Developer Tools`) — see:
   ```
   [Extension Host] Skill at <path> is missing required 'name' field
   ```

Programmatic reproduction:
```bash
cd /c/work/Zoo-Code
node -e "const matter = require('./node_modules/.pnpm/gray-matter@4.0.3/node_modules/gray-matter'); const fs = require('fs'); const content = fs.readFileSync('C:/Users/eason/.roo/skills/test-driven-development/SKILL.md', 'utf-8'); console.log(JSON.stringify(matter(content).data, null, 2));"
```

Output:
```
YAMLException: incomplete explicit mapping pair; a key node is missed; or followed by a non-tabulated empty line at line 3, column 128
```

## Expected result

Either:
- The skill loads successfully (gray-matter handles unescaped quotes gracefully), OR
- A clear error message explains the YAML parsing failure and points to the exact issue (e.g., "description contains unescaped double quotes")

## Actual result

The skill silently fails to load. The console shows a misleading error: `missing required 'name' field` even though the name IS present in the frontmatter. Users/AI have no way to discover the real cause without opening DevTools.

## Variations tried (optional)

- Wrapping description in single quotes (`description: 'Use when ... "TDD" ...'`) — works ✅
- Using backticks instead of double quotes (`Triggers on: \`TDD\``) — works ✅
- Removing all quotes from description — works ✅

## App Version

Current main (v3.58.x+)

## API Provider

Not Applicable / Other

## Relevant logs or errors

```
[Extension Host] Skill at C:\Users\eason\.roo\skills\test-driven-development is missing required 'name' field
```

**Root cause:** In `src/services/skills/SkillsManager.ts` line 102-115, when `matter(fileContent)` throws a YAML parsing error (from js-yaml@3.14.1), it's caught by the outer try/catch block at line 173. The specific YAML syntax error is lost — only "missing required 'name' field" is logged.

## Proposed approach

**Option A: Separate gray-matter parsing error (recommended)**

In `loadSkillMetadata()`, catch the matter() parse error separately before checking fields:

```typescript
// Before (line ~102-115)
const fileContent = await fs.readFile(skillMdPath, "utf-8")
const { data: frontmatter, content: body } = matter(fileContent)

if (!frontmatter.name || typeof frontmatter.name !== "string") {
    console.error(`Skill at ${skillDir} is missing required 'name' field`)
    return
}

// After
try {
    const fileContent = await fs.readFile(skillMdPath, "utf-8")
    let parsedResult
    try {
        parsedResult = matter(fileContent)
    } catch (parseError) {
        console.error(`Failed to parse SKILL.md at ${skillDir}: YAML syntax error - ${parseError.message}`)
        return
    }
    
    const frontmatter = parsedResult.data
    if (!frontmatter.name || typeof frontmatter.name !== "string") {
        console.error(`Skill at ${skillDir} is missing required 'name' field`)
        return
    }
    // ... rest of validation
} catch (error) {
    console.error(`Failed to load skill at ${skillDir}:`, error)
}
```

**Option B: Auto-wrap description in `createSkill()` tool**

In the `handleCreateSkill` method, automatically wrap the description value in single quotes if it contains double quotes or other special YAML characters. This prevents the issue for skills created via the UI tool.

## Trade-offs / risks

- **Option A**: Minimal risk — only changes error message clarity. No behavior change for valid SKILL.md files.
- **Option B**: Could introduce unexpected formatting (e.g., single quotes in description might look odd). Requires careful escaping logic.
- Both options are backward-compatible — existing valid skills continue to work.

## Related

N/A — new issue.

easonLiangWorldedtech:
Fixes #859

**Stacked on #934** — that PR (also fixing #859) adds the structured `SkillDiagnostic` collection (`SkillsManager.diagnostics` / `recordDiagnostic()` / `getSkillDiagnostics()`) and the settings-page diagnostics panel. This PR keeps only the remaining gaps on top of it, so the diff stays minimal if #934 merges first and gets squashed.

## Problem (remainder after #934)

1. **Misleading error still hides the common cause.** #934's `recordDiagnostic()` reports the YAML exception, but the single most common trigger — unescaped double quotes in the `description` — still requires the user to read the raw YAML error and open the file. A targeted hint points at the exact line and the fix.
2. **Diagnostics can disappear on re-scan (gray-matter cache poisoning).** gray-matter keeps a global content-keyed cache that it populates *before* parsing. A frontmatter that throws on first parse is cached with an empty `data` object, and every later parse of the same content silently returns that empty object instead of re-throwing — so a re-scan (watcher event, manual refresh) resurfaces the misleading "missing required 'name' field" symptom and drops the diagnostic.
3. **Overlapping discovery scans can interleave.** File-watcher events start concurrent `discoverSkills()` runs; an older scan finishing after a newer one can append a stale diagnostic for an already-repaired skill.
4. **Frontmatter serialization reflows on every mode change.** `updateSkillModes()` rewrites the file with js-yaml's default `lineWidth: 80`, reflowing long plain descriptions into folded block scalars (`>-1`) on every mode toggle — noisy diffs and format churn. `createSkill()` had the same exposure.
5. **Settings crashes when `skillDiagnostics` is absent.** #934 made the field optional on `ExtensionState`, but `SkillsSettings` read it without a fallback — any render without it (and 14 webview tests in the unit-test CI job) crashed with `Cannot read properties of undefined (reading 'length')`.
6. **Missing translations.** #934's new `settings:skills.diagnostics.*` i18n keys were only added to `en`, which fails the `check-translations` CI job (all 17 non-English locales).
7. **Test coverage.** The new `handleUpdateSkillModes` postMessage path, the malformed-load branches, `updateSkillModes`' serialization, and the extension-host → watcher → webview diagnostics flow had no tests.

## Changes

**1. Deterministic frontmatter parsing + unescaped-quote hint (load path)** — `src/services/skills/SkillsManager.ts`

- `loadSkillMetadata()` now parses with explicit empty options (`matter(fileContent, {})`), bypassing gray-matter's global content cache so a malformed skill reports the same parse error on every scan instead of the first throw being cached as an empty data object.
- When the parse throws, a small module helper (`getFrontmatterLine()`) extracts the raw top-level `description:` line with its file line number; the "unescaped double quotes" `console.error` hint is only emitted when the parser error's mark is located on that exact line — so a valid quoted description plus an unrelated YAML error elsewhere does not produce a misleading hint.

**2. Serialized discovery scans** — `src/services/skills/SkillsManager.ts`

- `discoverSkills()` is a non-async serializer chaining each run onto a `discoveryChain` promise; the body moved to `performDiscovery()`. Overlapping watcher-triggered scans no longer interleave, so an older scan can no longer append a stale diagnostic after a newer scan has observed the repaired file.

**3. Stable SKILL.md serialization (create/update paths)**

- Both `createSkill()` and `updateSkillModes()` now pass `lineWidth: -1` (via a small typed alias, since gray-matter's bundled typings predate its js-yaml dump-options passthrough), keeping long plain values on a single line.
- Output for plain descriptions stays byte-identical to the previous format; values with YAML special characters (double quotes, booleans like `yes`) are quoted/escaped automatically and always round-trip.

**4. Settings crash fix** — `webview-ui/src/components/settings/SkillsSettings.tsx`

- `skillDiagnostics` falls back to `[]` (same pattern as the existing `skills` handling), so the component no longer crashes when the optional `ExtensionState` field is absent. Fixes the 14 webview test failures in the unit-test CI job.

**5. i18n completion** — `webview-ui/src/i18n/locales/*/settings.json`

- Adds `settings:skills.diagnostics.title` / `description` to all 17 non-English locales (translated per locale), fixing the `check-translations` CI failure.

**6. Tests**

- `SkillsManager.spec.ts` (gray-matter `vi.mock`ed with the real parser as default implementation): a serialization regression test (a delayed older scan cannot append stale diagnostics), a **gray-matter cache-poisoning regression test** (a re-scan of unchanged malformed content must keep reporting the parse failure), the exact issue #859 content (hint + diagnostic logged, misleading "missing required 'name' field" absent), a no-false-hint case with a valid quoted description and an error on another line, a non-Error parse failure exercising `recordDiagnostic`'s defensive fallbacks, unterminated frontmatter with a dangling quote (diagnostic recorded, hint absent), create-path round-trip cases, and an `updateSkillModes` serialization round-trip case.
- `skillsMessageHandler.spec.ts`: new `handleUpdateSkillModes` suite (success, empty-slug clearing, omitted `newSkillModeSlugs` → `undefined` with a non-empty diagnostics list forwarded, missing fields, manager unavailable, rejected promise).
- `ExtensionStateContext.spec.tsx`: the skills-message test now asserts the transition that clears stored skills/diagnostics, including a message that omits `skills` entirely.
- `api-get-skills-state.spec.ts`: unit coverage for the new test-only `getSkillsState()` accessor (skills + diagnostics returned; empty arrays when the manager is unavailable).
- **`apps/vscode-e2e/src/suite/skills-diagnostics.test.ts`** (new): a real extension-host smoke test of the diagnostics flow — writes a healthy and a malformed (double-quoted description with unescaped inner quotes, the exact #859 failure mode) `SKILL.md` into the workspace's `.roo/skills` on real disk via atomic sidecar+rename writes so the watcher only observes complete files, waits for the extension host's file watcher to re-discover, asserts the malformed skill is omitted with a diagnostic pointing at it while the healthy skill is unaffected, then repairs the frontmatter and asserts the watcher clears the diagnostic and loads the fixed skill. Teardown removes only the skill directories this suite created.

## Verification

```
src subset (services/skills, skillsMessageHandler, api-get-skills-state): 101/101 passed
webview (SkillsSettings + ExtensionStateContext specs): 45/45 passed
e2e (USE_MOCK=true, TEST_FILE=skills-diagnostics.test.js): 1 passing
tsc --noEmit (src + webview via check-types, 11/11): clean
eslint (CI command): clean — suppression counts unchanged
node scripts/find-missing-translations.js: ✅ all areas complete
```

## Merge order

Merge #934 first, then this PR (no rebase needed: this branch already sits on #934's head plus upstream/main).

## Summary by CodeRabbit

* **New Features**
  * Added skill-loading diagnostics with clear messages and optional line and column locations.
  * Added a Settings warning panel listing affected files and recommended fixes.
  * Added API support for retrieving skill metadata and diagnostics.
  * Improved skill creation and updates to preserve formatting and handle special values safely.

* **Bug Fixes**
  * Valid skills continue loading when others contain errors.
  * Overlapping or repeated scans now report reliable, up-to-date diagnostics.

* **Documentation**
  * Added translated diagnostic messages across supported languages.

* **Tests**
  * Added coverage for malformed skills, recovery, diagnostics, and formatting preservation.


easonLiangWorldedtech:
## CodeRabbit review findings — all addressed in `5153db6`

- **e2e fixture**: now a double-quoted description with unescaped inner quotes (the exact #859 failure mode), so it reproduces the unescaped-quote parse failure instead of failing earlier on the `: ` sequence.
- **e2e teardown**: only removes the skill directories this suite created; pre-existing fixtures under `.roo/skills` are left intact.
- **api-get-skills-state.spec**: the partial doubles need a documented double-assertion; added a `beforeEach` comment explaining why it is a last resort.
- **serialize overlapping discovery runs** (major): `discoverSkills()` is now a non-async serializer chaining each run onto a `discoveryChain` promise; a delayed older scan can no longer append a stale diagnostic after a newer scan observed the repaired file (regression test added).
- **quote hint scoped to the failing line**: the "unescaped double quotes" hint is only emitted when the parser error mark is on the `description` line itself (regression test with a valid quoted description plus an error on another line).
- **state transition that clears diagnostics**: the `skills` message test now asserts stored skills/diagnostics are cleared, including a message that omits `skills` entirely.
- **nitpick** (typed message fixture / non-empty diagnostics assertion): the omitted-`newSkillModeSlugs` test declares a `WebviewMessage` instead of an `as` cast, and asserts a concrete non-empty diagnostic list is forwarded.

**Additional root-cause fix** found while making the e2e deterministic: gray-matter keeps a global content-keyed cache that it populates *before* parsing, so a frontmatter that throws on first parse was cached with an empty data object and every later parse of the same content silently returned that object instead of re-throwing (resurfacing the misleading "missing required name field" symptom). `loadSkillMetadata()` now parses with explicit empty options to bypass the cache, with a unit-level regression test that a re-scan of unchanged malformed content keeps reporting the parse failure.

```
src subset: 101/101 | webview: 45/45 | e2e (skills-diagnostics): 1 passing
check-types 11/11 clean | eslint clean, suppression counts unchanged
```

easonLiangWorldedtech:
@coderabbitai approve

easonLiangWorldedtech:
## CI status update (2026-09-14)

Merged the latest main (`ba46d1f34`) to advance the merge-base. The previous `mutation-diff` failure was base drift, not a defect in this PR: the gate's diff (merge-base → PR merge-ref) included ~a week of main's own changed lines, and main's code is not covered by this PR's tests. Upstream #1610 has also made surviving mutants advisory since.

The diff now contains only this PR's own changes; expecting a green re-run of the mutation gate on the new head.

Conflict resolution: add/add conflicts in `packages/types/src/api.ts` and `src/extension/api.ts` (this PR's `getSkillsState` vs upstream #1452's `hasTaskApiConversationHistorySequence`) — resolved by keeping both.

PR #1339, opened 2026-08-22
fix(zoo-gateway): stop inventing UI cost from default model prices

JamesRobert20:
## Summary
- Refetch router models when the selected id is missing from a stale in-memory map, so new catalog models get real pricing metadata.
- When falling back to `defaultModelInfo` for a non-default model id, zero $/token rates so the panel cannot invent Sonnet-priced costs.
- Negative-cache missing (and empty-catalog) lookups for 5 minutes so every `createMessage` does not re-hit `getModels`.
- Before recording a miss, call `refreshModels` so non-auth-scoped providers bust a stale shared catalog instead of blocking newly listed models for that window.
- When gateway streaming usage omits `cost`, emit `totalCost: 0` instead of inventing a local estimate.

## Test plan
- [x] `pnpm --dir src exec vitest run api/providers/__tests__/zoo-gateway.spec.ts` (48 passed)
- [x] `pnpm run check-types`
- [ ] Manual: select a newly added Zoo Gateway model (e.g. `alibaba/qwen3.8-max`), run a short task, confirm API Cost matches server `usage.cost` / credit deduction
- [ ] Manual: with models already loaded for another id, switch to a model not yet in the map and confirm a refetch occurs (no Sonnet-rate estimate)

## Summary by CodeRabbit

- **Bug Fixes**
  - Streaming usage now handles missing gateway cost data correctly by defaulting charges to zero.
  - Configured models that are unavailable now show zero pricing instead of inaccurate values.
  - Model catalogs no longer refresh repeatedly when a configured model cannot be found.

- **Improvements**
  - Missing models are rechecked periodically, allowing newly available models to be detected.
  - Successful model discovery clears previous unavailable-model status.
  - Model availability updates are deduplicated to reduce unnecessary refresh requests.

### DIFF
--- a/src/api/providers/__tests__/kenari.spec.ts
+++ b/src/api/providers/__tests__/kenari.spec.ts
@@ -34,6 +34,10 @@ vitest.mock("../fetchers/modelCache", () => ({
 			},
 		}),
 	),
+	refreshModels: vitest.fn(async (options) => {
+		const { getModels } = await import("../fetchers/modelCache")
+		return getModels(options)
+	}),
 	getModelsFromCache: vitest.fn().mockReturnValue(undefined),
 }))
 
--- a/src/api/providers/__tests__/kimi-code.spec.ts
+++ b/src/api/providers/__tests__/kimi-code.spec.ts
@@ -17,7 +17,10 @@ vi.mock("../../../integrations/kimi-code/oauth", () => ({
 	},
 }))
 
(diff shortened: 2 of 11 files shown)


JamesRobert20:
@coderabbitai review

taltas (approved):
Clean PR, nice work

PR #1345, opened 2026-08-22
feat(ollama): add reasoning effort selectors and gate on Enable Thinking

ViX3L:
### Related GitHub Issue

Closes: # 

### Description
- Chat bar and settings page share option computation via reasoning-effort.ts; both hide when Enable Thinking is unticked
- Ollama settings now renders its own thinking checkbox + effort dropdown; generic ThinkingBudget is skipped for it
- Fetcher advertises low/medium/high/max for thinking-capable models
- Handler sends think: "max" verbatim (previously clamped to "high") and maps "none" to think: false so the None option disables thinking

### Pre-Submission Checklist

- [ ] **Issue Linked**: This PR is linked to an approved GitHub Issue (see "Related GitHub Issue" above).
- [ ] **Scope**: My changes are focused on the linked issue (one major feature/fix per PR).
- [x] **Visual Snapshot** (UI changes only): If a user would notice this change at a glance (layout, theme tokens, brand elements, empty/error states), I've added or updated a `*.visual.tsx` snapshot in `webview-ui/`. See `webview-ui/AGENTS.md` → "When a UI change needs a snapshot".
- [ ] **Documentation Impact**: I have considered if my changes require documentation updates (see "Documentation Updates" section below).

### Get in Touch

**Before**
image

**After**
image

## New Features
- Added model-aware reasoning effort controls in chat and provider settings.
- Ollama models now show only supported options, including Disable, Low, Medium, High, and Max where applicable.
- Added keyboard-accessible reasoning selection in the chat toolbar.

## Improvements
- Reasoning settings automatically adjust when switching models with different capabilities.
- Ollama uses provider-specific reasoning controls instead of the generic thinking budget.

## Bug Fixes
- Ollama’s Max setting is now passed through correctly instead of being reduced to High.


src/api/providers/fetchers/ollama.ts
```diff
@@ -59,6 +59,16 @@ export const parseOllamaModel = (rawModel: OllamaModelInfoResponse): ModelInfo |
 		// Inherit the sane default (4096) from ollamaDefaultModelInfo instead.
 	})
 
+	// Models that advertise the "thinking" capability expose a native
+	// reasoning-effort control. Ollama Cloud accepts low/medium/high/max and
+	// rejects "xhigh", so advertise exactly those values (matching the user's
+	// API verification) so the reasoning selector shows the model's real options
+	// instead of falling back to the generic low/medium/high defaults.
+	if (rawModel.capabilities?.includes("thinking")) {
+		modelInfo.supportsReasoningEffort = ["low", "medium", "high", "max"]
```
taltas: Ollama documents GPT-OSS as accepting only `low`, `medium`, or `high`, so this advertises an unsupported Max choice for that model. Can we make the capability list model-specific and add a GPT-OSS regression test?

webview-ui/src/components/settings/providers/Ollama.tsx
```diff
@@ -39,6 +40,32 @@ export const Ollama = ({ apiConfiguration, setApiConfigurationField }: OllamaPro
...
+	// ollama native `think` parameter supports so the dropdown still works.
+	// "none" is always prepended for Ollama because ollama's native `think`
+	// parameter accepts it as an explicit reasoning level alongside low/med/
+	// high/max, and users want to disable reasoning without removing the
+	// enableReasoningEffort flag.
+	const { info: selectedModelInfo } = useSelectedModel(apiConfiguration)
+	const reasoningModelInfo = selectedModelInfo?.supportsReasoningEffort
+		? ({
+				...selectedModelInfo,
+				supportsReasoningEffort: Array.isArray(selectedModelInfo.supportsReasoningEffort)
```
taltas: This prepends None even when an explicit capability array intentionally omits it; for GPT-OSS, `think: false` is ignored and reasoning cannot be disabled. Can we preserve exact arrays, model off-support explicitly, and reuse the same normalization in the chat selector?

webview-ui/src/components/chat/ReasoningEffortSelector.tsx
```diff
@@ -0,0 +1,176 @@
...
+				sideOffset={4}
+				container={portalContainer}
+				className="p-0 overflow-hidden w-[180px]">
+				<div className="flex flex-col w-full">
+					<div className="p-3 border-b border-vscode-dropdown-border">
+						<p className="text-xs text-vscode-descriptionForeground m-0">{title}</p>
+					</div>
+					<div className="max-h-[300px] overflow-y-auto py-1">
+						{availableOptions.map((option) => (
+							<div
```
taltas: These options are not focusable or keyboard-operable, so keyboard users cannot select an effort. Can we render native buttons or the shared selectable-item primitive and add a keyboard interaction test?

webview-ui/src/components/chat/ChatTextArea.tsx
```diff
@@ -1319,6 +1320,7 @@ export const ChatTextArea = forwardRef<HTMLTextAreaElement, ChatTextAreaProps>(
 							lockApiConfigAcrossModes={!!lockApiConfigAcrossModes}
 							onToggleLockApiConfig={handleToggleLockApiConfig}
 						/>
+						<ReasoningEffortSelector triggerClassName="min-w-[28px] text-ellipsis overflow-hidden flex-shrink" />
```
taltas, line 1315: This adds a visible control to the compact chat toolbar, but the PR adds no `*.visual.tsx` baseline even though `webview-ui/AGENTS.md` requires one for at-a-glance layout changes. Can we add a representative chat-input snapshot, including the narrow-width state, and commit its Docker-generated baseline?

webview-ui/src/components/chat/ReasoningEffortSelector.tsx
```diff
@@ -0,0 +1,176 @@
...
+		(option: ReasoningEffortOption) => {
+			if (!currentApiConfigName || !apiConfiguration) {
+				return
+			}
+
+			// Write only `reasoningEffort`. The Enable Thinking checkbox in the
+			// settings page owns `enableReasoningEffort` independently, so the chat
+			// selector never flips it. Picking "None" in chat keeps
+			// enableReasoningEffort as-is and stores reasoningEffort: "none",
+			// which `getOllamaThinkParam()` translates to `think: true` with
```
taltas: The handler and tests map `none` to `think: false`, with no separate `reasoning: "none"` field. Can we correct this comment and the matching native-None claim in the Ollama settings component?

webview-ui/src/components/settings/__tests__/ReasoningModeSelector.spec.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+import { render, screen, fireEvent } from "@/utils/test-utils"
+
+import type { ModelInfo } from "@roo-code/types"
+
+import { ReasoningModeSelector } from "../ReasoningModeSelector"
+
+// Mock the Select primitives so we can drive selection without Radix deps.
+vi.mock("@/components/ui", () => ({
+	Select: ({ children, value, onValueChange }: any) => (
```
taltas: These new mocks and fixtures use `any`/`as any`, so schema or component-prop changes will not be checked here. Can we use small typed mock-prop interfaces and typed `ProviderSettings` fixtures instead?

taltas (commented):
some questions

src/api/providers/fetchers/ollama.ts
ViX3L: Previously reasoning selector had none - high for every models, thanks for noting about for gpt-oss and `https://docs.ollama.com/capabilities/thinking#python-2` according to ollama docs, changes has been made to reflect the reasoning selector for gpt-oss. So, Detection matches on model id, family, or architecture, hyphen/underscore-normalized so "gpt-oss" / "gptoss" / "gpt_oss" all match — Ollama reports family/architecture as gptoss (no hyphen) while the model id keeps the hyphen (gpt-oss:20b/gpt-oss:120b), so checking all three is robust. The supportsReasoningEffort is always an explicit array (never bare true), so the selector respects it verbatim.

Added 3 regression tests in ollama.test.ts: gpt-oss advertises exactly ["low","medium","high"] with no max/disable; detection by architecture alone (gptoss); detection by model id alone.

webview-ui/src/components/settings/providers/Ollama.tsx
ViX3L: Removed the UI-side `none` prepend entirely. Off-support is now modeled explicitly in the capability array at the source: the fetcher includes `disable` for models that honor `think: false` (qwen3 → `["disable","low","medium","high","max"])` and omits it for GPT-OSS `(["low","medium","high"])`. Both surfaces now route through one shared helper, `getOllamaReasoningModelInfo()` in `webview-ui/src/utils/reasoning-effort.ts`, which passes the advertised array through verbatim and only synthesizes a `["disable","low","medium","high"]` fallback when no model info has loaded. The settings page (`Ollama.tsx`) and chat selector (`ReasoningEffortSelector.tsx`) both call it, so they can't drift. `getReasoningEffortSelection` already respected explicit arrays as-is (it only auto-adds "disable" for `supportsReasoningEffort === true`), so Ollama arrays now flow through unchanged. Tests updated: fallback asserts `disable,low,medium,high`; a new GPT-OSS test asserts `low,medium,high` with no `disable/none`.

webview-ui/src/components/chat/ReasoningEffortSelector.tsx
ViX3L: Each row is now a native `<button type="button" role="option" aria-selected={...}>` (focusable, Enter/Space-activated by default, announced as an option) inside a `<div role="listbox" aria-label>` container whose `onKeyDown` handles ArrowUp/ArrowDown/Home/End to move focus between options. Added three keyboard-interaction tests: one asserting the listbox/option structure (native buttons, `role="option"`, `aria-selected`, selected row marked `true`), one asserting an option is activatable (the path Enter/Space takes on a native button), and one asserting `ArrowDown` calls `preventDefault` so the listbox owns navigation. Note: this repo's `vitest.setup.ts` stubs `HTMLElement.prototype.focus` (for FAST/JSDOM compatibility), so `document.activeElement` never moves in JSDOM — the tests assert structure and `preventDefault` instead, which is the durable proof of keyboard-operability; real focus behavior is exercised by the new visual test in a real browser.

webview-ui/src/components/chat/ChatTextArea.tsx
ViX3L: Added `webview-ui/src/components/chat/__tests__/ChatTextArea.visual.tsx` + `ChatTextArea.visual.fixture.tsx`.

webview-ui/src/components/chat/ReasoningEffortSelector.tsx
ViX3L: Rewrote the comment in `ReasoningEffortSelector.tsx` to state the truth: the chat selector writes only `reasoningEffort` and never flips `enableReasoningEffort`; picking "`None`" (the "`disable`" option) stores `reasoningEffort: "disable"`; there is no separate `reasoning: "none"` field; Ollama has no native string "`none`" level and `"disable"` maps to `think: false` via `getOllamaThinkParam()`. Corrected the parallel comment in `Ollama.tsx` (the checkbox `onChange` block) to clarify the handler reads the stored `reasoningEffort` (undefined → no think param, not `think: false`) and that `"disable"` is a UI sentinel mapping to `think: false`. The native-ollama spec already asserts `none` → `think: false` and `disable` → `think: false`.

webview-ui/src/components/settings/__tests__/ReasoningModeSelector.spec.tsx
ViX3L: Replaced the `any`-typed Select mock props with typed interfaces (`SelectMockProps`, `SelectTriggerMockProps`, `SelectValueMockProps`, `SelectContentMockProps`, `SelectItemMockProps)` covering `children`, `value`, `onValueChange`, `placeholder`, and item-specific values; used `React.isValidElement` guards for the `cloneElement` calls. Replaced `{} as any` / `{ … } as any` fixtures with typed `ProviderSettings` and `ModelInfo` objects (both are Zod objects with all-optional fields, so an empty `{}` is a valid `ProviderSettings` — no cast needed). A prop rename on the real `Select` components now surfaces as a compile error here. File typechecks and lints clean.

Issue #630, opened 2026-06-14
feat(webview): incremental clineMessages delivery for focused task


edelauna:
# Context

Sub-issue of #355 (Milestone 2, Story 3.2e). Builds on Stories 3.2c (focus gating) and 3.2d (per-task seq + taskId on messageUpdated).

The throttle fix in #629 reduces push **frequency** but doesn't address payload **size**. A single focused task with 3,525+ messages still serializes the full `clineMessages` array (~7MB) on every state push. At 10,000+ messages (long coding sessions), even throttled pushes will OOM the webview.

# Developer Notes

## Problem

`postStateToWebviewWithoutTaskHistory` sends the entire `clineMessages` array on every call. The lightweight `{ type: "messageUpdated" }` path already exists for partial message updates (`Task.ts:1048`) but new message additions still go through the full-array path (`Task.ts:1015`).

## Proposed Approach

Replace full-array state pushes with append/update messages for the focused task:

1. **New message type:** `{ type: "clineMessageAppended", message: ClineMessage }` — sent from `addToClineMessages` instead of a full state push
2. **Webview handler:** `ExtensionStateContext` appends to existing `clineMessages` array instead of replacing it
3. **Full sync retained for:** task switch, initial load, checkpoint restore, `overwriteClineMessages`
4. **Sequencing:** Use existing `clineMessagesSeq` to detect missed appends and trigger a full resync

## Depends On

- Story 3.2c — focus gating (only send messages for the focused task)
- Story 3.2d — per-task `clineMessagesSeq` and `taskId` on `messageUpdated`

Without focus gating, appends from background tasks would corrupt the focused task's message list.

## Impact

Payload per update drops from O(N) to O(1) — a single message (~2KB) instead of the full array (~7MB at 3,525 messages). Combined with #629's throttle, this eliminates the gray screen OOM for any message count.

## Acceptance Criteria

- [ ] New messages appear via append, not full-array replacement
- [ ] Full sync fires on task switch, initial load, and checkpoint restore
- [ ] Missed-append detection triggers automatic full resync
- [ ] No regression in message ordering or partial message streaming
- [ ] Works correctly with focus gating from 3.2c

edelauna:
[...]

PR #1360, opened 2026-08-24
fix(webview) searchFiles memory leak / WebUI Gray Screen

Gh0st352:
> **Current status (head 1eec683efe8c0242fd2308e302c6d1169c4a4f27):** compilation and Linux/Windows tests are fixed; current-head executable CI passes apart from the disclosed mutation-scope check. See the [current-head status report](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1360#issuecomment-5745910155) and prerequisite [#1703](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1703). This PR remains at 567 extension executable lines versus the unchanged 500 cap until the prerequisite actually merges and this branch is updated. The prerequisite fits scope but has [inconclusive mutation advisories](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1703#issuecomment-5745959486). Review acceptance remains separate. Neither PR has been merged. Older validation sections below are historical, not current-head results.

### Related GitHub Issue

Closes: # 630

### Description

This PR completes the incremental transcript-delivery work proposed in #630 and builds on the state-push throttling from #1078. 

Throttling reduced how often large task state was sent, but every update and hydration could still serialize and transfer the complete transcript. For long-running tasks, that payload remains large enough to exhaust the webview renderer and produce a gray screen.

The implementation introduces a dedicated, task-scoped transcript transport:

- Sends new and changed transcript entries as sequenced append/update deltas instead of replacing the full `clineMessages` array.
- Removes transcript data from generic extension-state messages so unrelated state updates cannot repeatedly retransmit or overwrite chat history.
- Hydrates and resynchronizes transcripts using atomic snapshots split into chunks of 200 messages, avoiding one unbounded initial-load payload.
- Serializes transcript posts through a provider-level queue and invalidates stale queued work when the focused task changes.
- Uses task IDs and monotonic per-task sequence numbers to reject stale, duplicate, out-of-order, or background-task messages.
- Detects sequence gaps, malformed or incomplete snapshots, and updates for unknown messages, then requests one authoritative resync from the extension host.
- Routes initial webview launch, task switches, checkpoint rewinds, transcript overwrites, and legacy unsequenced updates through the snapshot/resync path.
- Updates task, provider, message-handler, context, and shared test helpers to use the new protocol while preserving message persistence and event ordering.

The steady-state payload is now O(1) per append/update rather than O(N) in transcript length. Full recovery remains available, but it is transferred in bounded chunks and applied only after the complete snapshot has been validated.

This aligns with Zoo Code's **Reliability First** roadmap goal by keeping long-running chats responsive and making transcript synchronization deterministic and self-healing across webview reloads and task switches.

Reviewer focus areas:

- Ordering and invalidation behavior when task changes race with queued transcript posts.
- Sequence-gap recovery and atomic snapshot application in the webview.
- Transcript/message event ordering relied on by task consumers.

### Test Procedure

**Current published head: [190b186fdbf60155f4e6272af3a1c06ecbb3369e](https://github.com/Zoo-Code-Org/Zoo-Code/commit/190b186fdbf60155f4e6272af3a1c06ecbb3369e). Published with the disclosed mutation-budget blocker; cap unchanged; not merged.**

The [head-specific validation/status comment](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1360#issuecomment-5642531758) is the current source of truth. Full local checks passed on byte-identical reviewed source: 12,355 Vitest tests passed / 40 skipped plus one Node test file, 10/10 fresh test tasks, 11/11 fresh type and lint tasks each, all seven models, and 32/32 fresh mutation-runner tests. No local E2E/visual run is claimed.

**Overall mutation validation is blocked, not green.** Fresh canonical full-PR-diff local gating against base 01c7357a72d6095363304548a4f3d4dec0548171 generated **434 extension mutants versus the unchanged 400 cap**. The final 861-test baseline passed, but extension mutation execution **did not start**; killed/survived/no-coverage/timeout/error counts are unavailable, not zero. Webview passed with 391 valid mutants all killed and nine pre-existing ignored; that is not an overall gate pass. No cap/configuration/exclusion was relaxed.

The four source findings are fixed and resolved independently of overall merge readiness. New-head executable CI and review acceptance are tracked separately in the status comment. GitHub generated merge result 951533e6937911fadc39109d4380d11799cdc44e with advanced main parent ebcd1a0032ab5abc5251bc235f5fe672e144adc9; do not confuse that integration result or its CI with the locally audited base/head. No merge action was performed.

#### Historical validation - superseded for the current head

The following retained evidence applies only to the old head; its green mutation result does **not** validate 190b186fdbf60155f4e6272af3a1c06ecbb3369e.

Historical validated head: **f7d2abb5bf06d9d439d133ef49c1cd925f11e863**, based on **01c7357a72d6095363304548a4f3d4dec0548171**. These are local results; remote CI status is reported by the checks on this PR.

- Run the test, type-check, and lint tasks in the [workspace scripts](https://github.com/Zoo-Code-Org/Zoo-Code/blob/f7d2abb5bf06d9d439d133ef49c1cd925f11e863/package.json). All 10 test tasks passed, including fresh runs of 8,412 extension tests and 1,955 webview tests; unchanged package test tasks used their existing cache. All 11 type-check tasks and all 11 lint tasks passed.
- Run the [aggregate model checker](https://github.com/Zoo-Code-Org/Zoo-Code/blob/f7d2abb5bf06d9d439d133ef49c1cd925f11e863/package.json#L16). All seven models passed. The production-backed transcript model explores 32,591 states and 48,094 transitions, reaches all 16 actions and 14 landmarks, and detects all 12 injected faults. Its [architecture note](https://github.com/Zoo-Code-Org/Zoo-Code/blob/f7d2abb5bf06d9d439d133ef49c1cd925f11e863/docs/architecture/transcript-transport-model.md) describes the bounds and limitations.
- Run the [changed-code mutation gate](https://github.com/Zoo-Code-Org/Zoo-Code/blob/f7d2abb5bf06d9d439d133ef49c1cd925f11e863/scripts/stryker-diff.mjs) against the base and head above. Extension: 358 valid mutants, 356 killed, two timeouts within the existing allowance, zero surviving or uncovered. Webview: 343 valid mutants, all killed, zero timeouts, surviving, or uncovered. No new exclusions or relaxed thresholds. The local Windows launch adapter preserved CI's selectors, related-test discovery, configuration, budgets, and evaluator.
- Run suppression-pruning ESLint for changed source and test files. All passed; the suppression ledger is unchanged.

Focused regressions cover queue invalidation while physical sends are held, exactly-once caller settlement, captured nested message contents, snapshot-before-resume ordering, immediate streaming updates, final-message ordering, timestamp-index maintenance, exact chunk payloads, and atomic timeout recovery. The final focused run passed 360 tests.

Manual verification for reviewers (not claimed as completed by the automated runs):

- Open or restore a task with a large transcript.
- Confirm the transcript appears after chunked hydration and the panel remains responsive.
- Continue the task and verify new and streaming messages appear once, in order, without full transcript replacement.
- Switch rapidly between tasks and confirm messages from the previous task do not appear in the focused task.
- Restore a checkpoint or edit/delete history and verify the transcript is replaced by one complete, ordered snapshot.
- Reload the webview and confirm the active task transcript reconstructs without duplication or a gray screen.

### Pre-Submission Checklist

- [X] **Visual Snapshot** (UI changes only): Not applicable. This changes transcript transport and synchronization behavior, not a static rendered UI state.

### Visual Snapshots
N/A

N/A

### Documentation Updates

Does this PR necessitate updates to user-facing documentation?

### Additional Notes

- This PR addresses transcript transport cost and synchronization. It does not compact the persisted ui_messages.json representation or remove large fields from individual messages; those are complementary storage concerns outside this issue's scope.
- Generic state messages intentionally remain metadata-only. Transcript snapshots are assembled off-state and committed atomically after validation, so partially received snapshots never replace the visible conversation.
- No changeset or changelog entry is included, per repository policy.
- AI assistance materially contributed to implementation and PR preparation. I reviewed and understand the meaningful changes and verified them with the tests and checks listed above.

### Get in Touch

- Discord: gh0st352  (Gh0st)
- GitHub: https://github.com/Gh0st352


Gh0st352:
"New Task" button malfunction found resulting from patch; working fix.

src/core/webview/ClineProvider.ts
```diff
@@ -1413,13 +1418,176 @@ export class ClineProvider
...
+			return this.postStateToWebviewWithoutTaskHistory()
+		}
+
+		const seq = taskId
+			? options.bumpSeq
+				? this.bumpClineMessagesSeq(taskId)
+				: this.getClineMessagesSeq(taskId)
+			: 0
+		const messages = structuredClone(currentTask?.clineMessages ?? [])
+		const snapshotId = `${taskId ?? "none"}:${++this.nextClineMessagesSnapshotId}`
```
edelauna, line 1580: This clone runs before the closure is enqueued. If a resync bumps the generation before the closure runs, the clone is discarded. Move the clone inside the closure, after the `isCurrent()` guard, to avoid the wasted allocation.

src/core/task/Task.ts
```diff
@@ -1077,11 +1067,12 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
...
 			}
 		}
+		await this.providerRef.deref()?.postClineMessagesSnapshot(this.taskId, { bumpSeq: true })
 	}
 
 	private async updateClineMessage(message: ClineMessage) {
 		const provider = this.providerRef.deref()
-		await provider?.postMessageToWebview({ type: "messageUpdated", clineMessage: message })
+		await provider?.postClineMessageUpdated(this.taskId, message)
 		this.emit(RooCodeEventName.Message, { action: "updated", message })
```
edelauna, line 1339: Each `updateClineMessage` call enqueues one sequenced post here. During partial streaming, `say()` fires this once per token. The throttled state push it replaced collapsed rapid updates. Add a trailing debounce for `partial: true` messages. Bypass the debounce when `partial` is false or absent.

webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -403,26 +527,142 @@ export const ExtensionStateContextProvider: React.FC<{
...
+					}
+					if (message.snapshotTotal !== snapshot.total || snapshot.messages.length !== snapshot.total) {
+						activeSnapshotRef.current = null
+						retryClineMessagesResync(seq)
+						break
+					}
+
+					activeSnapshotRef.current = null
+					clearClineMessagesResync()
+					clineMessagesRef.current = snapshot.messages
```
edelauna, line 690: The test at line 919 covers several `snapshotEnd` error paths. The one gap: `Start`+`Chunk` received but `End` is dropped in transit. In that case `activeSnapshotRef.current` holds partial data. Add a test that sends `Start`+`Chunk` without `End` and asserts the prior `clineMessages` is unchanged.

Gh0st352:
Update on the long term testing: 

- On a 'stress-test' task (Text collection with 200+ files changed, 220k+ lines written/modified, mix of execution and generation), letting it run overnight.

- - Pre-Patch: Gray screen after 5-10min of execution
- - Post-Patch: Gray Screen after 7 hours of non-stop execution

- searchFiles memory leak fully fixed, confirmed by hnbdr (HoneyBadger) on Discord.

PR Ready for review.

edelauna (changes requested):
Nice! Had a couple implementation questions.

src/core/webview/ClineProvider.ts
Gh0st352: Addressed in [6927d0b](https://github.com/Zoo-Code-Org/Zoo-Code/commit/6927d0bdbef9e77a81f84fa34dc6744f0d793167). Snapshot cloning now occurs only after the queued generation/focus guard passes, so stale snapshots are discarded before incurring the allocation. I also added regression coverage confirming that stale queued snapshots are neither cloned nor posted.

src/core/task/Task.ts
Gh0st352: Addressed in [6927d0b](https://github.com/Zoo-Code-Org/Zoo-Code/commit/6927d0bdbef9e77a81f84fa34dc6744f0d793167). Streaming partial transcript updates now use a 500 ms trailing-only debounce. Complete and non-partial updates bypass the debounce and cancel stale pending work, pending work is canceled during disposal, and delayed posting failures are handled. Focused tests cover coalescing, timing resets, latest-revision delivery, immediate bypasses, disposal, and rejected delayed posts.

webview-ui/src/context/ExtensionStateContext.tsx
Gh0st352: Addressed in [6927d0b](https://github.com/Zoo-Code-Org/Zoo-Code/commit/6927d0bdbef9e77a81f84fa34dc6744f0d793167). I added regression coverage for receiving snapshot start and chunk messages without a matching end message. The test verifies that the previously visible transcript remains unchanged when the incomplete snapshot is abandoned.

easonLiangWorldedtech:
@coderabbitai approve

src/core/task/Task.ts
```diff
@@ -2897,7 +2906,7 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
 			} satisfies ClineApiReqInfo)
 
 			await this.saveClineMessages()
-			await this.providerRef.deref()?.postStateToWebviewWithoutTaskHistory()
+			await this.updateClineMessage(this.clineMessages[lastApiReqIndex])
```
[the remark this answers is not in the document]
Gh0st352: Addressed in e5dfc01ae: the request row is re-read after the awaited save and updated only when it still exists; added a regression that truncates the transcript during persistence and verifies no stale delta is posted.

src/core/webview/ClineProvider.ts
```diff
@@ -2458,7 +2628,7 @@ export class ClineProvider
 	 */
 	async postStateToWebviewWithoutClineMessages(): Promise<void> {
 		const state = await this.getStateToPostToWebview({ includeTaskHistory: false })
-		const { clineMessages: _omitMessages, taskHistory: _omitHistory, ...rest } = state
+		const { taskHistory: _omitHistory, ...rest } = state
```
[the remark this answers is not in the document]
Gh0st352: Addressed in e5dfc01ae: postStateToWebviewWithoutClineMessages now delegates to the canonical lightweight state-post method, with documentation and focused delegation coverage.

src/core/webview/webviewMessageHandler.ts
```diff
@@ -584,7 +588,7 @@ export const webviewMessageHandler = async (
 			const customModes = await provider.customModesManager.getCustomModes()
 			await updateGlobalState("customModes", customModes)
 
-			await provider.postStateToWebview()
+			await provider.syncFocusedTaskToWebview({ includeTaskHistory: true })
```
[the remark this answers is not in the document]
Gh0st352: Addressed in e5dfc01ae: added direct webviewDidLaunch coverage asserting syncFocusedTaskToWebview receives includeTaskHistory true.

src/core/task/Task.ts
```diff
@@ -1192,6 +1194,7 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
 		if (persist) {
 			await this.saveClineMessages(false)
 		}
+		await this.providerRef.deref()?.postClineMessagesSnapshot(this.taskId, { bumpSeq: true })
```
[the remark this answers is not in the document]
Gh0st352: Fixed in e5750682e0cbd20c3b77ee48197b334cb84b04a1. The pending partial-update debounce is now cancelled before replacement hydration or persistence begins. Fake-timer regression tests cover both persistence modes, a delayed save, no stale update after the replacement snapshot, and successful delivery of subsequent updates for the replacement transcript. The focused 477-test regression run and full workspace test run passed; lint, type checks, and lifecycle model checks also passed.

Gh0st352:
Addressed both findings from the latest CodeRabbit review in e5750682e0cbd20c3b77ee48197b334cb84b04a1. The inline stale-partial-update thread has a fix explanation and is resolved. The outside-diff finding, Publish the hydrated transcript before the resume ask, is also fixed: a sequence-bumped snapshot is awaited after both histories are hydrated and pending-action reconciliation completes, before either the generic resume prompt or pending-action replay. Cancellation is checked before and after delivery. Regression coverage includes active/completed history resume, pending-action replay, delayed reads and delivery, cancellation and abandonment, failed snapshot delivery, and a missing provider reference. No standalone history write was added. Validation passed: 477 focused tests; the full workspace test run with 11,906 Vitest tests passing plus the ESLint configuration test; lifecycle model checks; repository lint and type-check hooks. CI is rerunning on the new commit. CodeRabbit chat reported an organization-members-only restriction, so automated chat confirmation may require a maintainer.

Gh0st352:
Follow-up: mutation CI found one surviving readiness-flag mutation before pending-action replay. Fixed the test gap in 05df08169 by asserting that replay starts only after initialization. I verified the exact reported mutation fails on that assertion, restored the unchanged production code, and reran the task/persistence suites: 170 tests passed. Type checking and repository hooks passed. The follow-up is test-only; both CodeRabbit fixes remain in e5750682e.

src/core/task/Task.ts
```diff
@@ -2218,16 +2227,23 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
...
+
 			if (this.abort || this.abandoned) {
 				return
 			}
 
+			if (this.pendingAction) {
+				this.isInitialized = true
+				await this.resumePendingTaskAction(this.pendingAction)
+				return
+			}
```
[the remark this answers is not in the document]
Gh0st352: Addressed in 05df0816976c58859a411e7f3d89504d8927305d, pushed before this review completed. The pending-action replay mock now asserts readiness at invocation time. I temporarily applied the exact reported Boolean mutation and verified the test fails with expected false to be true, then restored production code and confirmed all 170 task/persistence tests plus type checking and lint pass. The latest CI run includes this test-only fix.

webview-ui/src/components/ui/__tests__/AccessibilityContrast.visual.tsx
```diff
@@ -3,6 +3,59 @@ import { expectContrast } from "../../../../playwright/contrast"
...
+			),
+		).toBe(0)
+		await expectContrast(input, { background: input, label: `${theme.name} settled input text` })
+	}
+
+	expect(
+		await page.evaluate(() =>
+			document
+				.getAnimations()
+				.some((animation) => animation instanceof CSSAnimation && animation.playState === "running"),
```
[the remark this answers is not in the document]
Gh0st352: Fixed in dfc7e1ee1. The assertion now inspects only the intended fixture element and requires the injected animation name plus a running state, instead of accepting any running animation in the document. Validation: the two Chromium theme-transition tests passed three consecutive runs and passed again after negative-control testing. Moving the animation to another element, renaming it, and pausing it each caused the strengthened assertion to fail as intended; all temporary mutations were restored. Formatting, ESLint with no added suppressions, and repository type checks also pass. No screenshot baselines or user-facing content changed.

src/core/webview/ClineProvider.ts
```diff
@@ -1468,13 +1473,176 @@ export class ClineProvider
...
+			}
+			await this.postMessageToWebview({
+				type: "clineMessagesSnapshotEnd",
+				taskId,
+				clineMessagesSeq: seq,
+				snapshotId,
+				snapshotTotal: messages.length,
+			})
+		})
+	}
```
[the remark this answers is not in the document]
Gh0st352: Fixed in 0cd3bf3d4. Snapshot payloads are now deeply cloned synchronously alongside their sequence before entering the transport queue. Existing generation/current-task guards and chunked start/chunk/end delivery are preserved. Added deterministic blocked-queue regressions for normal and sequence-bumped snapshots at the 200-message boundary, including an intervening append and in-place text/nested-image update. Both cases failed on the previous code (201 messages in the old-sequence snapshot, followed by a duplicate append) and pass with the fix. Invalidation-before-first-post coverage now checks both focus and generation changes without requiring delayed cloning. Validation: 461 focused tests, full repository tests (10/10 tasks; 8,259 extension tests passed), all five lifecycle/model checks, repository lint and type checks. No lint suppression counts increased.

src/core/webview/ClineProvider.ts
```diff
@@ -1468,13 +1473,177 @@ export class ClineProvider
...
+		const run = this.clineMessagesPostQueue.then(operation, operation)
+		this.clineMessagesPostQueue = run.catch((error) => {
+			this.log(`[clineMessages] transport failure: ${error instanceof Error ? error.message : String(error)}`)
+		})
+		return run
+	}
+
+	private invalidateClineMessagesTransport(): number {
+		return ++this.clineMessagesTransportGeneration
+	}
```
edelauna, line 1511: Does bumping the generation also need to reset `clineMessagesPostQueue = Promise.resolve()`? Each discarded closure still holds a promise node in the chain until it executes — over a long conversation that accumulates O(N) retained nodes, which works against the O(1) goal.

src/core/task/Task.ts
```diff
@@ -2332,16 +2341,23 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
...
 
-			if (this.pendingAction) {
-				this.isInitialized = true
-				await this.resumePendingTaskAction(this.pendingAction)
+			if (this.abort || this.abandoned) {
 				return
 			}
 
+			// Publish the transcript after both histories hydrate, before any resume prompt or pending-action replay.
+			await this.providerRef.deref()?.postClineMessagesSnapshot(this.taskId, { bumpSeq: true })
```
edelauna, line 2349: Is there a test that pins the ordering of this snapshot relative to `resumePendingTaskAction`? If the two lines were swapped, the webview would show the resume dialog before the transcript is visible, with no test failing.

src/core/webview/ClineProvider.ts
```diff
@@ -2474,9 +2646,7 @@ export class ClineProvider
 	 *   (cloud auth, org settings, profiles, etc.) without interfering with task message streaming.
 	 */
 	async postStateToWebviewWithoutClineMessages(): Promise<void> {
-		const state = await this.getStateToPostToWebview({ includeTaskHistory: false })
-		const { clineMessages: _omitMessages, taskHistory: _omitHistory, ...rest } = state
-		await this.postMessageToWebview({ type: "state", state: rest })
+		await this.postStateToWebviewWithoutTaskHistory()
```
edelauna, line 2649: `postStateToWebviewWithoutClineMessages` just delegates to `postStateToWebviewWithoutTaskHistory`. Any reason to keep both names rather than migrating the ~10 call sites to the canonical one?

src/core/task/Task.ts
```diff
@@ -656,6 +658,16 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
...
+		this.debouncedPostPartialMessageUpdate = debounce((message: ClineMessage) => {
+			const provider = this.providerRef.deref()
+			if (!provider) {
+				return
+			}
+
+			void provider.postClineMessageUpdated(this.taskId, message).catch((error) => {
+				console.error("[Task#updateClineMessage] incremental post failed:", error)
+			})
+		}, PARTIAL_MESSAGE_UPDATE_DEBOUNCE_MS)
```
edelauna, line 670: The adjacent `debouncedEmitTokenUsage` uses `{ leading: true, trailing: true, maxWait }`. Without `leading: true` here, during fast LLM streaming the webview sees no update for the first 500 ms and can appear frozen.

```suggestion
		}, PARTIAL_MESSAGE_UPDATE_DEBOUNCE_MS, { leading: true, trailing: true, maxWait: PARTIAL_MESSAGE_UPDATE_DEBOUNCE_MS })
```

src/core/webview/ClineProvider.ts
```diff
@@ -1468,13 +1473,177 @@ export class ClineProvider
...
+		if (this.getCurrentTask()?.taskId !== taskId) {
+			return Promise.resolve()
+		}
+		if (process.env.ROO_CLI_RUNTIME === "1") {
+			return this.postStateToWebviewWithoutTaskHistory()
+		}
+
+		const seq = this.bumpClineMessagesSeq(taskId)
+		const generation = this.clineMessagesTransportGeneration
+		const clonedMessage = structuredClone(message)
```
edelauna, line 1523: `structuredClone` runs at enqueue time, before the generation guard fires inside the closure. After a task switch, every queued delta pays the clone cost before being discarded. Would moving this inside the closure (after the `generation !==` guard) be safe?

webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -336,13 +340,176 @@ export const ExtensionStateContextProvider: React.FC<{
...
+			if (seq !== clineMessagesSeqRef.current + 1) {
+				requestClineMessagesResync(seq)
+				return
+			}
+
+			let nextMessages: ClineMessage[]
+			if (operation === "append") {
+				nextMessages = [...clineMessagesRef.current, clineMessage]
+			} else {
+				const index = findLastIndex(clineMessagesRef.current, (item) => item.ts === clineMessage.ts)
```
edelauna, line 451: For a 1000-message transcript this is a linear scan on every `clineMessageUpdated`. Since `ts` is unique, would a `Map<ts, index>` (updated on append and snapshot commit) make sense here?

webview-ui/src/context/__tests__/ExtensionStateContext.spec.tsx
```diff
@@ -399,6 +433,1478 @@ describe("ExtensionStateContext", () => {
...
+			})
+
+			expect(postMessage).toHaveBeenCalledTimes(1)
+			expect(postMessage).toHaveBeenCalledWith({
+				type: "requestClineMessagesResync",
+				taskId: "task-1",
+				expectedSeq: 2,
+				receivedSeq: 2,
+			})
+			expect(vi.getTimerCount()).toBe(1)
```
edelauna, line 676: After the timer fires, does the test verify that `clineMessages` is still the pre-snapshot value? If the code accidentally committed partial chunk data on timeout, the current assertions would not catch it.

src/core/webview/__tests__/ClineProvider.spec.ts
```diff
@@ -860,6 +847,711 @@ describe("ClineProvider", () => {
...
+			expect(posts.map(({ type }) => type)).toEqual([
+				"clineMessagesSnapshotStart",
+				"clineMessagesSnapshotChunk",
+				"clineMessagesSnapshotChunk",
+				"clineMessagesSnapshotChunk",
+				"clineMessagesSnapshotEnd",
+			])
+			expect(posts.map(({ clineMessagesSeq }) => clineMessagesSeq)).toEqual([1, 1, 1, 1, 1])
+			expect(posts.slice(1, 4).map(({ snapshotStartIndex }) => snapshotStartIndex)).toEqual([0, 200, 400])
+			expect(posts.slice(1, 4).map(({ clineMessages }) => clineMessages?.length)).toEqual([200, 200, 1])
```
edelauna, line 1059: The test verifies chunk sizes and start indices but not the actual message objects. Would `.toEqual(messages.slice(0, 200))` etc. per chunk catch off-by-one or shuffle bugs?

packages/types/src/vscode-extension-host.ts
```diff
@@ -138,7 +143,13 @@ export interface ExtensionMessage {
 		isActive: boolean
 		path?: string
 	}>
+	taskId?: string
 	clineMessage?: ClineMessage
+	clineMessages?: ClineMessage[]
+	clineMessagesSeq?: number
+	snapshotId?: string
+	snapshotStartIndex?: number
+	snapshotTotal?: number
```
edelauna, line 152: The seven new transport fields (`taskId`, `clineMessage`, `clineMessages`, `clineMessagesSeq`, `snapshotId`, `snapshotStartIndex`, `snapshotTotal`) have no JSDoc. The PR description mentions protocol docs embedded in JSDoc — are these intended to go here?

packages/types/src/vscode-extension-host.ts
```diff
@@ -646,8 +660,11 @@ export interface WebviewMessage {
 		| "openRuleFile"
 		| "openRulesDirectory"
 		| "themeFixtureProbeResponse"
+		| "requestClineMessagesResync"
 	text?: string
 	taskId?: string
+	expectedSeq?: number
+	receivedSeq?: number
```
edelauna, line 667: Which of `expectedSeq`/`receivedSeq` is "last applied" and which is "just received"? A brief comment here would help the next person writing a handler.

src/core/webview/ClineProvider.ts
```diff
@@ -1468,13 +1473,177 @@ export class ClineProvider
...
+			})
+		})
+	}
+
+	public resyncClineMessagesToWebview(taskId?: string): Promise<void> {
+		if ((this.getCurrentTask()?.taskId ?? undefined) !== taskId) {
+			return Promise.resolve()
+		}
+		const generation = this.invalidateClineMessagesTransport()
+		return this.postClineMessagesSnapshot(taskId, { generation })
```
edelauna, line 1631: This invalidates the generation and fires a full snapshot, but nothing is logged. Would a log entry here help diagnose resync storms (repeated triggers from the webview)?

src/core/webview/webviewMessageHandler.ts
```diff
@@ -574,6 +575,9 @@ export const webviewMessageHandler = async (
 	}
 
 	switch (message.type) {
+		case "requestClineMessagesResync":
+			await provider.resyncClineMessagesToWebview(message.taskId)
```
edelauna, line 579: `message.expectedSeq` and `message.receivedSeq` are typed on the message but not forwarded or logged here. Are these intended purely for future diagnostic use, or should they influence the resync behavior?

edelauna (changes requested):
Awesome! Thanks so much for your continued work on this - had some implementation comments.

This PR introduces a state machine (generation counter, promise queue, snapshot commit protocol) that has the same shape as the existing lifecycle:model-check callers. Would you be able to add a model-check script for the transport layer and wire it into that command, following the same convention as cleanup-protocol:model-check and parser-scope:model-check? The core invariants are fairly contained - generation monotonicity, no stale commit after invalidation, seq monotonicity - so the script should be small.

src/core/webview/ClineProvider.ts
Gh0st352: Good catch. I replaced the promise chain with an explicit queue so invalidation immediately drops the waiting jobs and their payloads, and settles their callers. I kept a single barrier for a send that has already started: resetting that barrier would let the new generation overlap the old send. The regression holds a snapshot post open through repeated resyncs, checks that 26 waiting jobs are released, and verifies that sends stay serialized and no stale snapshot end is started afterward. The new model also catches both the old retention behavior and the overlapping-send variant. Fixed in 7346fbb38.

src/core/task/Task.ts
Gh0st352: Yes, this is pinned now. I strengthened the resume test with a deliberately held snapshot and an explicit event order. It checks that history is hydrated but neither the pending action nor a generic resume prompt runs until the snapshot finishes, then checks that the action replays exactly once. I also tried both regressions locally—removing the wait and swapping the order—and each fails the assertions. Added in d87c5abe5.

src/core/webview/ClineProvider.ts
Gh0st352: No good reason to keep both now. I removed the alias and migrated the callers and test doubles to the existing canonical method. There are no remaining references to the old name, and the browser transcript filtering and CLI behavior are unchanged. Cleaned up in 7346fbb38.

src/core/task/Task.ts
Gh0st352: Agreed. The first partial update now goes out immediately, with trailing coalescing and a 500 ms maximum wait for sustained streaming. I added coverage for the immediate update, continued streaming, cancellation, and final-message ordering so a trailing update cannot overwrite the final one. Fixed in d87c5abe5.

src/core/webview/ClineProvider.ts
Gh0st352: I checked this, and moving the clone to delivery time is not safe with the current mutable messages. Text and nested arrays can change while a post is queued, so an earlier sequence could end up carrying later content. I kept the deep copy at admission and added a test that mutates both text and images before releasing the queue. Stale-generation and wrong-task requests are now rejected before cloning or allocating a sequence, and invalidation releases already-queued copies immediately. That avoids the safe-to-skip work without changing what each sequence represents. Covered in 7346fbb38.

webview-ui/src/context/ExtensionStateContext.tsx
Gh0st352: Yes, I added a timestamp-to-index lookup. It is initialized once, maintained on append, and rebuilt only when the transcript is replaced or cleared, including snapshot commits and task switches. The tests cover first/middle/last updates and replacement/clear paths, and check that updates do not rescan message timestamps or rebuild the index on render. Lookup is now constant-time on average; the immutable array copy is still linear. Added in 6e137b9f4.

webview-ui/src/context/__tests__/ExtensionStateContext.spec.tsx
Gh0st352: It does now. The test starts with two visible messages at sequence 7, receives only part of a replacement at sequence 10, and checks that both the original transcript and applied sequence are unchanged after the timeout. It also verifies one recovery request, no extra retry, and a subsequent update against the preserved transcript. That catches an accidental partial commit rather than just checking that a timer fired. Added in 6e137b9f4.

src/core/webview/__tests__/ClineProvider.spec.ts
Gh0st352: Added. The 401-message test now compares each of the three chunk payloads with its exact slice of the original messages, alongside the existing offset, size, sequence, and snapshot identity checks. A skipped, duplicated, or reordered message will fail the payload assertions. Updated in 7346fbb38.

packages/types/src/vscode-extension-host.ts
Gh0st352: Yes, these belonged on the shared contract. I added JSDoc for all seven fields covering task scope, per-task revision semantics, snapshot identity and chunk offsets, the complete-snapshot commit boundary, and the empty-task and legacy cases. I also added an architecture note for the queue ownership model and its limits. Updated in 7346fbb38.

packages/types/src/vscode-extension-host.ts
Gh0st352: The expected value is actually the next sequence the webview wants—last applied plus one—not the last applied value itself. The received value is the sequence on the message that triggered recovery. I documented both, including when the received value is absent, and made their diagnostic-only role explicit. Updated in 7346fbb38.

src/core/webview/ClineProvider.ts
Gh0st352: Added a log for each accepted resync with the task ID, old/new generation, authoritative sequence, and any valid expected/received sequence diagnostics. It contains no transcript or image content. Wrong-task requests are still ignored before invalidation, and the tests cover both accepted and ignored requests. This should make repeated resyncs much easier to trace. Added in 7346fbb38.

src/core/webview/webviewMessageHandler.ts
Gh0st352: They are now forwarded to the provider and included in the resync log. I kept them diagnostic-only: recovery still uses the provider's current task and authoritative sequence, never a client-supplied sequence. Tests cover forwarding, missing or malformed values, and confirm that even extreme client values do not change the recovery revision. Updated in 7346fbb38.

Gh0st352:
@edelauna, I’ve pushed the latest fixes and replied to each of your inline comments. The branch is rebased onto the current target and GitHub reports it as mergeable.

I also added the transport model you suggested. The [checker](https://github.com/Zoo-Code-Org/Zoo-Code/blob/f7d2abb5bf06d9d439d133ef49c1cd925f11e863/scripts/check-transcript-transport.ts) runs through the [aggregate lifecycle command](https://github.com/Zoo-Code-Org/Zoo-Code/blob/f7d2abb5bf06d9d439d133ef49c1cd925f11e863/package.json#L16), alongside the existing models. It exercises the actual production reducer and frame conversion, rather than a second implementation of the queue.

It explores 32,591 states and 48,094 transitions across queued updates, repeated resync, task switching/clearing, and failed-send recovery. It checks generation and sequence ordering, immediate release of obsolete work, exactly-once caller settlement, a single physical send, and complete atomic snapshots with bounded chunk ranges. All 16 action classes and 14 named landmarks are reached, and all 12 injected faults produce the expected counterexample. The state and depth budgets fail closed rather than reporting success after truncation.

One important distinction: I kept the barrier around an already-started send. Resetting it would allow old and new generations to overlap. An end marker started before invalidation can still finish; the guarantee is that no new stale post is started afterward. I documented that boundary, along with the model’s delivery and liveness assumptions, in the [architecture note](https://github.com/Zoo-Code-Org/Zoo-Code/blob/f7d2abb5bf06d9d439d133ef49c1cd925f11e863/docs/architecture/transcript-transport-model.md#L85).

Local validation is complete on f7d2abb5b:

- Full workspace tests passed, including fresh runs of 8,412 extension tests and 1,955 webview tests. Unchanged package test tasks used their existing cache.
- Full type checks and lint passed, with no suppression-count increase.
- All seven aggregate model checks passed.
- The unchanged mutation gate passed: extension 356 killed, 2 timeouts within the allowed bound, and no survivors or uncovered mutants; webview 343 killed, no timeouts, survivors, or uncovered mutants. No new exclusions or relaxed limits. The local run used Windows-compatible launchers with the same selectors, related-test discovery, and report evaluator as CI.

Update: all build, test, coverage, security, translation, and visual checks now pass on this head, including the [remote mutation run](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34631880856). The first mutation event carried a stale base SHA and stopped at scope selection; the fresh event used the current base and passed without changing code or relaxing the gate. Automated review is still running, and maintainer approval remains separate. Thanks for the detailed feedback—the queue ownership and model invariants are much clearer now.

src/core/webview/ClineProvider.ts
```diff
@@ -1407,6 +1413,7 @@ export class ClineProvider
 
 			// Perform preparation tasks and set up event listeners
 			await this.performPreparationTasks(task)
+			await this.syncFocusedTaskToWebview()
 
 			this.log(
 				`[createTaskWithHistoryItem] rehydrated task ${task.taskId}.${task.instanceId} in-place (flicker-free)`,
```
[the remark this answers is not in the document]
Gh0st352: Fixed in published head [190b186fdbf60155f4e6272af3a1c06ecbb3369e](https://github.com/Zoo-Code-Org/Zoo-Code/commit/190b186fdbf60155f4e6272af3a1c06ecbb3369e).

- The original producer instance now travels from the Task through provider admission and every transcript frame; late old producers are rejected rather than relabeled as current. [Provider admission](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/src/core/webview/ClineProvider.ts#L1540-L1579).
- Same-ID replacement publishes dedicated focus before awaited cleanup/preparation, and the final generic-state post rejects captured stale identity. The receiver ignores old-instance frames. This retains admission-time deep cloning and the single physical-send barrier; it does not pretend an already-initiated send can be revoked. [Replacement ordering](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/src/core/webview/ClineProvider.ts#L1390-L1402), [state guard and focus publication](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/src/core/webview/ClineProvider.ts#L1488-L1538).
- Passing [real Task/provider regressions](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/src/core/task/__tests__/Task.persistence.spec.ts#L306-L531) hold all five frame types, reject late producers using the current generation, and recover through the new instance. [Receiver regressions](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/webview-ui/src/context/__tests__/ExtensionStateContext.spec.tsx#L505-L977) and [CLI regressions](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/apps/cli/src/ui/__tests__/transcript-focus.test.tsx#L7-L57) also preserve setup/hydration and resume behavior.

The remote head/tree was verified before this reply. Resolving this source finding is separate from merge readiness: the extension mutation gate remains blocked at 434 generated mutants versus the unchanged 400 cap. No merge performed.

src/core/webview/transcriptTransport.ts
```diff
@@ -0,0 +1,280 @@
...
+
+export function transcriptFrameMessage(frame: TranscriptFrame, messages: readonly ClineMessage[]): ExtensionMessage {
+	const { job, phase } = frame
+	const common = { taskId: job.taskId, clineMessagesSeq: job.seq }
+	if (phase === "append" || phase === "update") {
+		return {
+			...common,
+			type: phase === "append" ? "clineMessageAppended" : "clineMessageUpdated",
+			clineMessage: messages[0],
+		}
```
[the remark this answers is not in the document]
Gh0st352: Fixed in published head [190b186fdbf60155f4e6272af3a1c06ecbb3369e](https://github.com/Zoo-Code-Org/Zoo-Code/commit/190b186fdbf60155f4e6272af3a1c06ecbb3369e). The [driver guard](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/src/core/webview/transcriptTransport.ts#L244-L264) rejects empty append/update payloads before request capture, cloning, or allocation, and the [reducer guard](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/src/core/webview/transcriptTransport.ts#L98-L107) independently prevents protocol-state changes. Empty snapshots remain valid.

Passing [reducer regressions](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/src/core/webview/__tests__/transcriptTransport.spec.ts#L51-L69) and [driver regressions](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/src/core/webview/__tests__/transcriptTransport.spec.ts#L221-L297) check unchanged state/sequence, no cloning or frame insertion, then successful valid-delta and empty-snapshot recovery. The remote reviewed tree is confirmed. This source fix does not clear the separately reported unchanged mutation-budget blocker.

src/extension.ts
```diff
@@ -191,7 +191,7 @@ export async function activate(context: vscode.ExtensionContext) {
 			// Push the new vscode.env.isTelemetryEnabled value to the webview too, so its
 			// own PostHog client (gated separately in TelemetryClient.ts) can't keep
 			// sending events after the global toggle flips off mid-session.
-			void ClineProvider.getVisibleInstance()?.postStateToWebviewWithoutClineMessages()
+			void ClineProvider.getVisibleInstance()?.postStateToWebviewWithoutTaskHistory()
```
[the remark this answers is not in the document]
Gh0st352: Fixed in published head [190b186fdbf60155f4e6272af3a1c06ecbb3369e](https://github.com/Zoo-Code-Org/Zoo-Code/commit/190b186fdbf60155f4e6272af3a1c06ecbb3369e). The [telemetry listener](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/src/extension.ts#L188-L201) now catches and logs refresh rejection while preserving fire-and-forget behavior and the absent-provider path.

The passing [activation regressions](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/src/__tests__/extension.spec.ts#L438-L510) exercise success, no visible provider, and both Error and non-Error rejection values with exact logging assertions. The fix is verified on the remote reviewed tree; overall mutation readiness remains separately blocked by the unchanged cap.

webview-ui/src/context/__tests__/ExtensionStateContext.spec.tsx
```diff
@@ -15,6 +16,13 @@ import {
 } from "@roo-code/types"
 
 import { ExtensionStateContextProvider, useExtensionState, mergeExtensionState } from "../ExtensionStateContext"
+import { vscode } from "@/utils/vscode"
+
+const dispatchExtensionMessage = (message: ExtensionMessage) => {
+	window.dispatchEvent(new MessageEvent("message", { data: message }))
+}
```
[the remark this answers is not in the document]
Gh0st352: Fixed in published head [190b186fdbf60155f4e6272af3a1c06ecbb3369e](https://github.com/Zoo-Code-Org/Zoo-Code/commit/190b186fdbf60155f4e6272af3a1c06ecbb3369e). The [receiver test imports](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/webview-ui/src/context/__tests__/ExtensionStateContext.spec.tsx#L1-L10) now reuse the exported dispatch helper; the local duplicate is removed. The receiver suite passed all 197 tests, and the full webview suite passed 2,034 tests on the identical reviewed source. Remote head/tree verified before replying.

Gh0st352:
## Published review fixes â€” mutation budget still blocking

Published **[190b186fdbf60155f4e6272af3a1c06ecbb3369e](https://github.com/Zoo-Code-Org/Zoo-Code/commit/190b186fdbf60155f4e6272af3a1c06ecbb3369e)** by ordinary fast-forward push to the contributor fork. GitHub's PR head and Git tree independently match the reviewed 17-file commit. **Publication with the disclosed blocker was expressly authorized; the cap remains unchanged and no merge was performed. Source-review resolution is not overall merge readiness.**

### Review disposition

The four open findings were answered with head-pinned code/test evidence and then resolved:

| Finding | Published reply |
| --- | --- |
| Original producer-instance isolation across same-ID replacement; dedicated focus and stale-state protection | [Evidence and resolution](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1360#discussion_r3994578474) |
| Reject empty deltas while retaining empty snapshots | [Evidence and resolution](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1360#discussion_r3994578555) |
| Catch/log telemetry state-refresh rejection | [Evidence and resolution](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1360#discussion_r3994578710) |
| Reuse the exported dispatch helper | [Evidence and resolution](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1360#discussion_r3994578831) |

**35/35 existing threads resolved, zero open at the post-resolution read.** Earlier replies were not reposted. Bot chat-restriction notices are not approval.

The two assertion findings repeated in the [latest pre-push review](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1360#pullrequestreview-5182267802) are also fixed in this commit; their historically resolved threads were not reopened or given duplicate replies:

- [Exact winning generation](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1360#discussion_r3931859430): the [race regression](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/src/core/webview/__tests__/ClineProvider.spec.ts#L1892-L1920) requires exactly one snapshot with the pre-race generation plus two and the original current instance, both before and after releasing the older post.
- [Exact recovery requests](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1360#discussion_r3931859467): the [receiver assertions](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/webview-ui/src/context/__tests__/ExtensionStateContext.spec.tsx#L2644-L2655) require all five complete ordered requests, expected revision 2 and received revisions 2, 4, 5, 6, 7, plus unchanged visible state. Wrong-task frames are ignored, not resynchronized.

### Local validation of the exact published source

Full checks ran before committing; reviewed file hashes and committed blobs were verified unchanged after normal hooks and again before/after publication. These are source-identity-backed results, not a claim that full checks were rerun after the commit:

- **12,355 Vitest tests passed / 40 skipped**, plus one passing Node test file; **10/10 fresh test tasks**. Extension: **8,471 passed / 37 skipped**. Webview: **2,034 passed**.
- Types and lint: **11/11 successful tasks each, zero cached top-level tasks**; suppression ledger unchanged. Nested dependency builds included cached outputs; normal TypeScript incremental behavior remained enabled.
- All **seven aggregate models passed**. Transport: **36,927 states, 56,078 transitions, 23 actions, 26 landmarks, 20/20 injected faults detected**. [Model scope and limitations](https://github.com/Zoo-Code-Org/Zoo-Code/blob/190b186fdbf60155f4e6272af3a1c06ecbb3369e/docs/architecture/transcript-transport-model.md).
- Mutation-runner tests: **32/32 fresh passes**. No local extension-host E2E or visual run is claimed. Existing Node-version/tool warnings remain disclosed, with no dependency workaround.

### Unchanged extension mutation-budget blocker

Fresh canonical **full-PR-diff** local gate: base **01c7357a72d6095363304548a4f3d4dec0548171**, head **190b186fdbf60155f4e6272af3a1c06ecbb3369e**.

| Package | Changed executable lines | Generated mutants | Final baseline | Mutation outcome |
| --- | ---: | ---: | --- | --- |
| Extension | 365 | **434 against cap 400** | **861 tests passed** | **Execution did not start: budget blocked** |
| Webview | 391 | 400 | 197 tests passed | 391 valid, all killed; nine pre-existing ignored; zero survivors, no-coverage, timeouts, or errors |

The canonical gate **exited 1**. Extension killed/survived/no-coverage/timeout/error counts are **unavailable, not zero**; no extension execution report exists. Webview passed twice, but that is **not an overall gate pass**. Independent read-only analysis reproduced the 434 generated extension mutants. No small justified cap-clearing refactor was established or attempted during publication.

No cap, threshold, exclusion, runner/plugin, selector, test configuration, dependency, or suppression budget was changed by this publication. The earlier green mutation result for **f7d2abb5bf06d9d439d133ef49c1cd925f11e863** is historical only and does **not** validate this head.

### Fresh remote CI â€” distinct from local validation

The new run metadata identifies PR head **190b186fdbf60155f4e6272af3a1c06ecbb3369e**. GitHub's newly generated merge result is **951533e6937911fadc39109d4380d11799cdc44e**, with parents **ebcd1a0032ab5abc5251bc235f5fe672e144adc9** (advanced main) and the published PR head. The PR API still reports base **01c7357a72d6095363304548a4f3d4dec0548171**. These are recorded separately: the merge result is not the contributor commit, and local old-base results are not relabeled as validation of advanced main. The mutation workflow's actual base/head command will be reported from its completed logs.

Snapshot: **2026-09-12 01:27:39 UTC**. Results below are fresh new-head checks; none is borrowed from the old head.  | Check | Status | Workflow | | --- | --- | --- |
| [Zoo Code / reconcile PR review state](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664754708/job/103474333682) | SUCCESS | Label PR review state |
| [CodeQL](https://github.com/Zoo-Code-Org/Zoo-Code/runs/103474113863) | SUCCESS |  |
| [compile](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607296/job/103473892885) | SUCCESS | Code QA Roo Code |
| [knip](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607296/job/103473892669) | SUCCESS | Code QA Roo Code |
| [dependency-review](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607296/job/103473892863) | SUCCESS | Code QA Roo Code |
| [platform-unit-test (ubuntu-latest)](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607296/job/103473892814) | IN_PROGRESS | Code QA Roo Code |
| [webview-visual](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607228/job/103473892809) | SUCCESS | Visual Regression |
| [mutation-diff](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607219/job/103473892553) | IN_PROGRESS | Changed-code mutation testing |
| [invisible-chars](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607296/job/103473892811) | SUCCESS | Code QA Roo Code |
| [theme-fixtures](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607228/job/103473892578) | SUCCESS | Visual Regression |
| [check-translations](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607296/job/103473892952) | SUCCESS | Code QA Roo Code |
| [extension-host-visual](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607228/job/103473892640) | SUCCESS | Visual Regression |
| [validate-release](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607581/job/103473893447) | SUCCESS | Release Validation |
| [Analyze (javascript-typescript)](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607258/job/103473892470) | SUCCESS | CodeQL Advanced |
| [Build test VSIX](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607296/job/103473892817) | SUCCESS | Code QA Roo Code |
| [e2e-mock](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607206/job/103473892340) | IN_PROGRESS | E2E Tests (Mocked) |
| [platform-unit-test (windows-latest)](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664607296/job/103473892862) | IN_PROGRESS | Code QA Roo Code |
| [Zoo Code / reconcile PR review state](https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/34664605678/job/103473888534) | SUCCESS | Label PR review state |
| [Zoo Code / PR review gate](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1360) | PENDING |  |
| CodeRabbit | SUCCESS |  |

Review decision: **
CHANGES_REQUESTED
**. Pending executable work and the review gate are not passes. This single head-specific comment will retain its local evidence while receiving a bounded final CI update.

The PR remains **open and unmerged**. The mutation-budget failure and required review acceptance remain explicit; thread resolution alone clears neither.

Gh0st352:
## CI fix and validation for 1679bf3a6e42cf8d4854a99f6cddeb75285ad080

Pushed the existing local fixes after validation. The Linux coverage failure and mutation related-test discovery failure both came from the combined exhaustive transport-model test exceeding its unchanged 20-second timeout. The searches now run as separate test cases while retaining every scenario, injected fault, shortest-counterexample check, and exploration bound. Shared producer admission and frame dispatch remove duplicated decisions while retaining task/instance guards, capture-time cloning, and CLI behavior.

### Local validation on the exact committed source
- 12,375 Vitest tests passed, 40 skipped; all 10 workspace test tasks ran fresh.
- 378 focused tests passed with coverage.
- All seven aggregate model checks passed.
- All 11 type-check tasks and all 11 lint tasks passed fresh; suppression counts did not increase.
- 32/32 mutation-runner tests passed.
- Strict full-PR mutation gate against base ebcd1a0032ab5abc5251bc235f5fe672e144adc9 passed: extension 388 valid (387 killed, 1 permitted timeout); webview 391 valid (391 killed, 9 pre-existing ignored). No survivors or uncovered mutants. All discovery, preflight, and mutation processes exited successfully within the existing limits.
- The Windows-only launcher adaptation canonicalized filesystem path casing and invoked the installed tools through Node; no tracked runner/configuration, cap, exclusion, timeout, or evaluation rule was changed. The committed blobs and full-PR mutation selectors were verified against the validated tree.

### Remote status
All executable Actions workflows completed successfully on this head, including Linux and Windows coverage, visual tests, mocked E2E, security, translations, release validation, and the mutation workflow.

Important mutation qualification: current upstream CI uses an advisory mutation workflow and tested generated merge result de306d2ba3cebdf4de1fe22fc472b5a7b8b5d4f2 with newer base parent 745656a5070d7a4763aab8bfc462a28b38d4809d, while its comparison BASE_SHA remained ebcd1a0032ab5abc5251bc235f5fe672e144adc9. Its scope therefore includes additional upstream changes. The remote extension report contains 389 killed mutants and 1 timeout; webview preflight reported 426 mutants against the 400 cap, so remote webview mutation execution did not proceed. The green advisory status is not a strict integration mutation-gate pass. The separate exact-head local strict pass above remains the applicable evidence for this fix.

webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -336,13 +346,214 @@ export const ExtensionStateContextProvider: React.FC<{
...
+	)
+
+	const applyClineMessagesDelta = useCallback(
+		(message: ExtensionMessage, operation: "append" | "update") => {
+			const seq = message.clineMessagesSeq as number
+			const clineMessage = message.clineMessage
+			if (
+				message.taskId !== activeTaskIdRef.current ||
+				message.taskInstanceId !== activeTaskInstanceIdRef.current
+			) {
```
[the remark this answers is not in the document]
Gh0st352: Addressed in https://github.com/Zoo-Code-Org/Zoo-Code/commit/d258df9132cebffb9b7922b9e1439ec1e2b95435. All four delta/snapshot scope guards now require an active task before checking the existing task and instance identities. Authoritative state/focus clearing still clears the transcript; later unscoped frames are ignored without changing messages, sequence numbers, timers, or recovery requests. Added 12 regression cases covering all five frame types and a complete nonempty snapshot after both state-based and focus-based clearing, including same-event-batch delivery and clearing a buffered snapshot. These reproduced the bug before the production fix. Existing context expectations and four chat fixtures now declare an active task before valid hydration; welcome fixtures explicitly declare no task. Local validation: all 2,046 webview tests across 162 files passed, including 209 context tests; coverage exercises every new guard. Full webview lint, type checking, formatting, and whitespace checks passed with no added suppressions. Commit/push hooks passed all 11 workspace lint and 11 type-check tasks, with 10 unchanged tasks cached in each run. Windows tests used canonical filesystem path casing without tracked configuration changes. The fix is pushed and new-head CI is running. No fresh local full-PR mutation or E2E pass is claimed for this commit; review acceptance and overall merge readiness remain separate.

Gh0st352:
@edelauna Ready for Review.

docs/architecture/transcript-transport-model.md
```diff
@@ -0,0 +1,128 @@
...
+
+[`TranscriptRequest.taskInstanceId`](../../src/core/webview/transcriptTransport.ts:6) is optional for legacy fixtures, while every [`TranscriptJob`](../../src/core/webview/transcriptTransport.ts:11) retains its originating instance, including an absent value. The constructor preserves its first three arguments and adds a fourth focused-instance callback defaulting to an absent value. Identity comparisons are exact: an unscoped request cannot adopt a live instance, and an identified request cannot match absent focus. Production must provide the actual instance callback and the originating instance on requests. Every wire frame, including append/update and snapshot start/chunk/end, copies the descriptor's instance through [`transcriptFrameMessage()`](../../src/core/webview/transcriptTransport.ts:185); it never derives identity from later focus.
```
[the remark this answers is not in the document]
Gh0st352: Addressed in 9e5a02fe0ac07415a7a7e2a47598b78765d811f4. Corrected the two production-boundary declaration anchors to their verified current lines (223 and 194), leaving the surrounding documentation unchanged. Formatting and the transcript transport model check passed.

webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -336,13 +346,215 @@ export const ExtensionStateContextProvider: React.FC<{
...
+	)
+
+	const retryClineMessagesResync = useCallback(
+		(receivedSeq?: number) => {
+			clearClineMessagesResync()
+			requestClineMessagesResync(receivedSeq)
+		},
+		// Stryker disable next-line ArrayDeclaration: both dependencies are stable callbacks; omitting them cannot alter callback identity or captured values.
+		[clearClineMessagesResync, requestClineMessagesResync],
+	)
```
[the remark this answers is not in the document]
Gh0st352: Addressed in 9e5a02fe0ac07415a7a7e2a47598b78765d811f4. The retry callback now retires pending recovery only when abandoning an accepted snapshot. Orphaned invalid starts/chunks/ends share the next pending request instead of repeatedly clearing its guard; a newly accepted snapshot can still retry immediately if it fails, and lost-response timeout recovery is preserved. Regression coverage exercises eight failure modes followed by bursts of orphaned frames, replacement failure, and successful recovery, plus the unchanged retry deadline. All 2,063 webview tests and 315 provider/transport tests passed, as did lint, type checking, and the transport model.

webview-ui/src/utils/test-utils.tsx
```diff
@@ -37,6 +37,67 @@ export const makeExtensionState = (overrides: Partial<ExtensionState> = {}): Par
...
+export const dispatchExtensionMessage = (message: ExtensionMessage) => {
+	window.dispatchEvent(new MessageEvent("message", { data: message }))
+}
+
+export const hydrateExtensionState = (
+	state: Partial<ExtensionState>,
+	options: { taskId?: string; clineMessagesSeq?: number } = {},
+) => {
+	const { clineMessages, clineMessagesSeq: stateSeq, ...metadataState } = state
+	const taskId = options.taskId ?? metadataState.currentTaskId ?? undefined
```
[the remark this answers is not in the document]
Gh0st352: Addressed in 9e5a02fe0ac07415a7a7e2a47598b78765d811f4. The hydration helper now derives task-instance scope from the explicit option or state metadata and forwards it on start, chunk, and end frames. Added regression tests for metadata-derived and option-provided scopes, non-empty and empty snapshots, and explicit override precedence. All 2,063 webview tests passed.

Gh0st352:
## CodeRabbit findings addressed

Published commit: 9e5a02fe0ac07415a7a7e2a47598b78765d811f4.

All three requested findings are fixed, regression-tested, replied to, and resolved: coalesced recovery from orphaned snapshot frames; instance-scoped hydration helpers; corrected documentation anchors. No unresolved review threads remained at verification.

### Validation

- Full webview suite: 2,063 tests passed across 162 files.

- Focused provider/transport suites: 315 tests passed.

- Webview lint and types, formatting, and transcript transport model passed. Repository hooks passed all 11 lint and 11 type-check tasks (9 cached in each).

- Fix-commit-only mutation check, base 847165c2190d3287e0cc898a39047ad5b76a1a6c to the published head: 5 valid mutants killed, 0 survived, 0 uncovered, 0 timeouts; 2 existing equivalent dependency-array mutants ignored. The Windows CLI wrapper could not launch discovery, so this diagnostic used the same Stryker configuration and automatically selected changed lines with the focused 221-test suite. It is not a full-PR mutation result.

- Required remote CI passed, including Windows/Linux tests, mocked E2E, both visual jobs, compilation, and patch coverage.

### Remaining gates / limitations

The fresh CodeRabbit review is still in progress and the PR review gate remains pending. No merge was performed.

The full-PR mutation job reports success because it is advisory, but its validation is INCOMPLETE: extension mutation execution exceeded 12 minutes; webview preflight generated 411 mutants versus the unchanged 400 cap. No configuration, budget, exclusion, or review policy was relaxed. See https://github.com/Zoo-Code-Org/Zoo-Code/actions/runs/35288648572 .

Gh0st352:
## Current-head fixes and prerequisite split

Published head: 1eec683efe8c0242fd2308e302c6d1169c4a4f27. Upstream base: c5b585565be1bfbf53ace90dac33b53cc4d505bf. This report supersedes older-head validation.

Test fix 46f2f1844 repairs three typed model-fetch mocks, waits for task initialization before observing transport errors, restores the console spy on failure, and verifies no append is posted through an unavailable provider. Subsequent upstream conflicts were resolved preserving both the lifecycle gap audit/fan-out checks and transcript work.

Validation: 205 focused tests and 3,246 core-coverage tests passed (9 skipped). After the upstream update, 617 focused regressions and a fresh extension suite of 9,004 passed /37 skipped passed. Workspace tests/types/lint, lifecycle/transcript models, and standalone fan-out passed; suppression counts unchanged. Local Node 22.22.2 is below requested 22.23.1; no local E2E/visual run claimed.

Current-head remote compile, Linux and Windows tests, patch coverage, mocked E2E, visuals, packaging, security, and translations pass. Automated review has completed; the review gate awaits fresh human maintainer or CODEOWNER approval. GitHub reports MERGEABLE.

Prerequisite #1703 is open with the unchanged transport foundation, tests/model and additive types. Its scope is 262 extension executable lines. After it actually merges unchanged and this branch is updated, this PR should fall from 567 to 305. Opening the prerequisite alone does not satisfy the unchanged 500-line limit; this PR mutation execution is still blocked before it starts.

The prerequisite actual mutation run is advisory/inconclusive: 303 generated, 283 killed, 5 survived, 2 uncovered, 11 timeouts, 2 runtime errors. Details: https://github.com/Zoo-Code-Org/Zoo-Code/pull/1703#issuecomment-5745959486 . No limits, exclusions, workflows, or review labels were relaxed. Neither PR has been merged.

src/core/task/Task.ts
```diff
@@ -1323,12 +1344,60 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
...
+					this.clineMessages === replacement &&
+					this.todoList === replacementTodos &&
+					isDeepStrictEqual(this.clineMessages, snapshot.messages) &&
+					isDeepStrictEqual(this.todoList, snapshot.todos) &&
+					isDeepStrictEqual(this.cloudSyncedMessageTimestamps, snapshot.cloudSyncedTimestamps)
+				) {
+					this.clineMessages = previousMessages
+					this.todoList = previousTodos
+					this.cloudSyncedMessageTimestamps = previousCloudSyncedTimestamps
+				}
```
edelauna, line 1388: What signals to an operator that in-memory state was rolled back here, vs. a plain persistence failure? The caller's log ("Error processing pending edit") doesn't distinguish the two — a `console.warn` with `taskId`, `instanceId`, and the array length delta before re-throw would make reversion diagnosable from extension logs alone.

src/core/webview/ClineProvider.ts
```diff
@@ -1474,16 +1523,143 @@ export class ClineProvider
...
+		const currentTask = this.getCurrentTask()
+		// Every producer, including the legacy CLI path, must pass the same identity
+		// check before reading or cloning payloads from the focused task.
+		if (currentTask?.taskId !== request.taskId || currentTask?.instanceId !== request.taskInstanceId) {
+			return Promise.resolve()
+		}
+		if (process.env.ROO_CLI_RUNTIME === "1") {
+			return request.kind === "update"
+				? this.postMessageToWebview({ type: "messageUpdated", clineMessage: structuredClone(request.message) })
+				: this.postStateToWebviewWithoutTaskHistory()
```
edelauna, line 1606: Could this path be reached with a browser webview attached (e.g. in development or testing)? If so, `ExtensionStateContext.tsx` strips `clineMessages` from state messages unconditionally — the webview's transcript would never advance. The `messageUpdated` case also triggers a resync loop that resolves only by timeout (`CLINE_MESSAGES_RESYNC_TIMEOUT_MS`), since the resync response is again `postStateToWebviewWithoutTaskHistory()`.

src/core/webview/ClineProvider.ts
```diff
@@ -1020,8 +1049,35 @@ export class ClineProvider
 	}
 
 	async resolveWebviewView(webviewView: vscode.WebviewView | vscode.WebviewPanel) {
+		if (this._disposed) return
+		this.clearWebviewResources()
```
edelauna, line 1053: Is there a log entry that captures when a new renderer replaces the prior transport? The sidebar-dispose path at line 1076 logs before calling `clearWebviewResources()`, but this renderer-attach path doesn't — transport generation history would be invisible when debugging renderer churn or gray screens.

src/core/webview/transcriptTransport.ts
```diff
@@ -0,0 +1,412 @@
...
+	}
+
+	private send(frame: TranscriptFrame): void {
+		const completion: SendCompletion = {
+			finish: (success, error) => {
+				completion.finish = undefined
+				if (this.pendingSend !== completion) return
+				this.pendingSend = undefined
+				this.apply({ type: "settle", frame, success }, error)
+				if (!success) this.callbacks?.onError(error)
```
edelauna, line 398: Could `frame` be passed to `onError` here? It's in scope and carries `frame.job.taskId`, `frame.job.seq`, and `frame.phase` — right now it's not possible to tell from extension logs which task's transcript delivery failed or which chunk was in flight.

webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -336,13 +346,218 @@ export const ExtensionStateContextProvider: React.FC<{
...
+				nextMessages = [...clineMessagesRef.current, clineMessage]
+				clineMessagesIndex.set(clineMessage.ts, nextMessages.length - 1)
+			} else {
+				const index = clineMessagesIndex.get(clineMessage.ts)
+				if (index === undefined) {
+					requestClineMessagesResync(seq)
+					return
+				}
+				// Timestamp lookup is O(1) on average; the immutable array copy is still O(N).
+				nextMessages = [...clineMessagesRef.current]
```
edelauna, line 471: The comment acknowledges this is still O(N) — has `requestAnimationFrame` batching been considered as a follow-up? At high streaming rates with long transcripts (`leading:true, maxWait:500ms`) this copies the full array on every debounced update tick.

webview-ui/src/context/ExtensionStateContext.tsx
```diff
@@ -336,13 +346,218 @@ export const ExtensionStateContextProvider: React.FC<{
...
+			resyncTimeoutRef.current = window.setTimeout(() => {
+				resyncPendingRef.current = false
+				resyncTimeoutRef.current = undefined
+			}, CLINE_MESSAGES_RESYNC_TIMEOUT_MS)
+			vscode.postMessage({
+				type: "requestClineMessagesResync",
+				taskId: activeTaskIdRef.current,
+				expectedSeq: clineMessagesSeqRef.current + 1,
+				receivedSeq,
+			})
```
edelauna, line 388: Would a `reason` field here help with diagnosis? At least six paths call this (`gap`, `malformed`, `snapshot_timeout`, `snapshot_interrupted`, `missing_index`, `legacy_unsequenced`) — the server-side log at `ClineProvider.ts:1632` can't distinguish them without correlating browser console output separately.

edelauna (changes requested):
Nice! Thanks so much for continuing work on this. Had a couple comments, but agree that this is a good approach to take.

PR #1368, opened 2026-08-25
fix(settings): preserve configured LiteLLM model ID in model picker

### LINKED ISSUE #1367: LiteLLM settings: configured model ID silently reverts to default and cannot be changed
**Description**

With the LiteLLM provider, the model ID on the settings screen cannot be changed: after picking a model (e.g. via "Use custom model..."), the picker silently reverts to the hardcoded default (`claude-3-7-sonnet-20250219`). The user's selection is not displayed afterwards, so the setting appears broken.

**Steps to reproduce**

1. Open Settings → Providers → LiteLLM.
2. Set the base URL and API key, then refresh the model list.
3. Pick a model ID that is not present in the fetched `/models` list (e.g. a custom alias) via "Use custom model...".
4. The picker immediately reverts to the default model ID.
5. Save, close and reopen Settings — the configured model ID is not shown; the default is shown instead.

**Root cause**

The LiteLLM case in `webview-ui/src/components/ui/hooks/useSelectedModel.ts` validates the configured `litellmModelId` against the fetched router-model list and, when it is absent, substitutes the hardcoded default via `getValidatedModelId()`. LiteLLM is a proxy that fronts arbitrary models and aliases, so a configured ID is the user's explicit selection even when it is absent from the fetched list (custom aliases, incomplete or stale listings, renamed deployments).

**Expected behavior**

- A configured `litellmModelId` is displayed and preserved, regardless of whether it appears in the fetched list.
- The hardcoded default is only used when nothing is configured and a populated list exists.

easonLiangWorldedtech:
Fixes #1367

## Problem

On the LiteLLM settings screen the user could not change the model ID: selecting a model that is absent from the fetched `/models` list (custom aliases, incomplete/stale listings, renamed deployments) silently reverted the picker to the hardcoded default (`claude-3-7-sonnet-20250219`), and the saved custom ID was not displayed after reopening Settings.

## Root cause

The LiteLLM case in `useSelectedModel` validated the configured `litellmModelId` against the fetched list with `getValidatedModelId()` and substituted the hardcoded default whenever the configured ID was missing from that list. LiteLLM is a proxy that fronts arbitrary models, so a configured ID is the user's explicit selection and must never be replaced by a hardcoded default.

## Fix

- `webview-ui/src/components/ui/hooks/useSelectedModel.ts` — the LiteLLM case now preserves the configured `litellmModelId` even when it is absent from the fetched list. The hardcoded default is only used when nothing is configured and a populated list exists; the empty-ID behavior for the empty-list case is unchanged.
- `webview-ui/src/components/ui/hooks/__tests__/useSelectedModel.spec.ts` — the previous test codified the buggy behavior (asserting a configured ID reverts to the default); replaced with regression tests: a configured ID absent from a populated list is preserved, and an unconfigured ID with a populated list falls back to the default.
- `webview-ui/src/components/settings/__tests__/ModelPicker.spec.tsx` — new component test covering the full user flow: open the picker, type a custom ID, choose "Use custom model", re-render with the updated configuration, and assert the button shows the custom ID instead of the default.

## Verification

- Both new tests fail on the unfixed code (verified by stashing the fix) and pass with it.
- `vitest run src/components/ui/hooks src/components/settings` — 54 files / 626 tests passed.
- `vitest run src/components/chat` (other `useSelectedModel` consumers) — 37 files / 386 tests passed.
- `eslint --max-warnings=0` on touched files — clean; `tsc` typecheck — clean.


webview-ui/src/components/ui/hooks/useSelectedModel.ts
```diff
@@ -84,12 +95,19 @@ export const useSelectedModel = (apiConfiguration?: ProviderSettings) => {
...
+	// dynamic providers require a populated provider entry before the
+	// selection is resolved.
 	const hasValidRouterData =
 		needRouterModels && dynamicProvider
-			? routerModels.data &&
-				routerModels.data[dynamicProvider] !== undefined &&
-				typeof routerModels.data[dynamicProvider] === "object" &&
-				!routerModels.isLoading
+			? dynamicProvider === providerIdentifiers.litellm
+				? !routerModels.isLoading
```
edelauna: When `isError=true` and `isLoading=false`, this now resolves for LiteLLM and returns the configured model ID — the old guard (`data[litellm] !== undefined`) would have kept it unresolved. Is that path covered in the litellm describe block?

edelauna (approved):

Issue #1375, opened 2026-08-26
[EPIC] File Write Safety Prevent Concurrent Write Races Data Corruption


easonLiangWorldedtech:
## Problem

The agent's file-write path has no version guard, no atomic publish, and no per-step
visibility:

- Concurrent writers silently overwrite each other's work: parallel subtasks share one
  workspace, multiple VS Code windows are separate extension-host processes, and the
  JetBrains plugin has confirmed real corruption under two IDE instances (#1231).
- Model writes land directly in the shared workspace with no checkpoint to roll back to;
  task start creates no baseline at all.
- No staleness detection: a file that changed after the model read it is silently
  overwritten on the next write.
- The write path carries artificial latency (1 s delay per write) on top of the missing
  safety.

## Known issues

| # | Title | Status | Fix in this epic |
|---|-------|--------|------------------|
| [#1371](https://github.com/Zoo-Code-Org/Zoo-Code/issues/1371) | MCP settings wiped by concurrent windows | Open | **A5** — root cause is the check-then-create TOCTOU in McpHub.getMcpSettingsFilePath() (L496-517), not initializeMcpServers() |
| [#1021](https://github.com/Zoo-Code-Org/Zoo-Code/issues/1021) | Race in abandonSubtask for saveClineMessages | Open | **A5** — fire-and-forget saveClineMessages() re-attaches abandoned subtasks |
| [#920](https://github.com/Zoo-Code-Org/Zoo-Code/issues/920) | Concurrent task history updates cause lost entries | Open | **A5** — cross-instance (parallel tabs) regression test on the locked-merge write |
| [#1231](https://github.com/Zoo-Code-Org/Zoo-Code/issues/1231) | Global _index.json full rewrite unsafe (corruption confirmed) | Closed | Already fixed — per-task history_item.json + merge-under-lock; that is now the reference storage pattern |
| [#1221](https://github.com/Zoo-Code-Org/Zoo-Code/issues/1221) | Truncated tool-call args silently written to disk | Open | **A6 (separate story)** — parser/finalization defect, not a filesystem-safety problem |
| [#376](https://github.com/Zoo-Code-Org/Zoo-Code/issues/376) | File-level write serialization | Open | Superseded by **A3** |
| [#370](https://github.com/Zoo-Code-Org/Zoo-Code/issues/370) | Extension-side task-scoping guard | Open | Related; stays open |

### Agent write path

[...]

PR #1380, opened 2026-08-27
fix(mcp): preserve concurrent MCP settings during initial creation (fixes #1371)

### LINKED ISSUE #1371: MCP settings wiped when multiple windows open — race in McpHub.getMcpSettingsFilePath() direct fs.writeFile
# MCP settings wiped when multiple windows open — race in McpHub.getMcpSettingsFilePath() direct fs.writeFile

## Summary

Opening multiple VS Code windows concurrently can wipe `mcp_settings.json` to a 122-byte stub `{"mcpServers": {}}`, losing all MCP server configurations. Root cause is a race in `McpHub.getMcpSettingsFilePath()` which uses direct `fs.writeFile` without advisory lock, bypassing the existing `safeWriteJson` utility.

## Environment

- Extension: `zoocodeorganization.zoo-code` v3.80.0
- OS: Linux (generic)
- VS Code: multiple windows / workspaces open simultaneously

## Steps to Reproduce

1. Have a populated `mcp_settings.json` at `~/.config/Code/User/globalStorage/zoocodeorganization.zoo-code/settings/mcp_settings.json` (or equivalent globalStorage path) with several `mcpServers` entries.
2. Open 2+ VS Code windows at roughly the same time (e.g., `code .` in two workspaces, or restoring session with multiple windows).
3. Each window constructs `McpHub`, which calls `getMcpSettingsFilePath()` → `ensureSettingsDirectoryExists()` → `fileExistsAtPath()` → `fs.writeFile` if not exists.
4. Observe file truncated to 122 bytes.

## Observed Behavior

- File `mcp_settings.json` becomes:
  ```json
  {
    "mcpServers": {

    }
  }
  ```
  (122 bytes, pretty-printed empty stub)
- `stat` shows `Birth == Modify == 07:54` (file recreated at that time), indicating overwrite rather than edit.
- `FileSystemWatcher` (`watchMcpSettingsFile`) silently accepts the stub as valid config and propagates empty server list to all windows via `debounceConfigChange` → `updateServerConnections({}, "global")`.
- All MCPs disappear from UI; no error shown (empty object passes `McpSettingsSchema` validation).

## Expected Behavior

- Concurrent initializations should not clobber existing config.
- File creation should be atomic and locked, preserving existing servers.

## Root Cause

**File:** `src/services/mcp/McpHub.ts:496-517` (`getMcpSettingsFilePath()`)

```ts
const fileExists = await fileExistsAtPath(mcpSettingsFilePath)
if (!fileExists) {
  await fs.writeFile(
    mcpSettingsFilePath,
    `{
  "mcpServers": {

  }
}`,
  )
}
```

- Uses direct `fs.writeFile` without lock, while `src/utils/safeWriteJson.ts:45` exists and is already used elsewhere in same file (lines 2073, 2158, 2367) for other config writes.
- `safeWriteJson` provides `proper-lockfile` advisory lock (stale 31s, 5 retries, exponential backoff), temp file + atomic rename, and backup/rollback. This path bypasses it.
- TOCTOU: two processes both see `fileExists==false` (or one checks while other is mid-write), second blind write wins and truncates.
- No merge guard: even if file was just created by another window with content, it is overwritten with empty stub.

**Watcher:** `watchMcpSettingsFile()` at `McpHub.ts:519` debounces and calls `handleConfigFileChange` which does `JSON.parse` + `McpSettingsSchema.safeParse` — empty `mcpServers` is valid, so no error is surfaced.

## Impact

- **Data loss:** All MCP server definitions wiped (API keys, commands, URLs, env). User must restore from backup.
- **Silent:** No error toast; user discovers only when MCPs missing.
- **Reproducible:** Race window is small but reliably hit when opening multiple windows on session restore or via CLI.

## Proposed Fix

Replace direct write with `safeWriteJson`:

```diff
- await fs.writeFile(
-   mcpSettingsFilePath,
-   `{
-   "mcpServers": {
-
-   }
- }`,
- )
+ await safeWriteJson(mcpSettingsFilePath, { mcpServers: {} }, { prettyPrint: true })
```

- Import already exists: `import { safeWriteJson } from "../../utils/safeWriteJson"` at `McpHub.ts:44`.
- `safeWriteJson` handles lock, atomic write, dir creation.

**Optional stronger guard** (handles TOCTOU between `fileExistsAtPath` and lock):

```ts
await safeWriteJson(mcpSettingsFilePath, { mcpServers: {} }, {
  prettyPrint: true,
  merge: (existing, incoming) => {
    if (existing && typeof existing === "object" && "mcpServers" in existing) {
      const ex = existing as { mcpServers?: Record<string, unknown> }
      if (ex.mcpServers && Object.keys(ex.mcpServers).length > 0) return existing
    }
    return incoming
  }
})
```

This makes it atomic read-modify-write under lock.

Patch prepared against `v3.80.0` tag; source-only (installed `dist/extension.js` is minified, not patched). Patch file available on request.

## Workaround Until Fixed

- Open single window at a time, or stagger window opens by a few seconds.
- Keep manual backup: e.g., `~/Desktop/CODE/ZooMCPBakup/mcp_settings.json` (or any versioned backup outside globalStorage).
- Restore via `cp ~/Desktop/CODE/ZooMCPBakup/mcp_settings.json ~/.config/Code/User/globalStorage/zoocodeorganization.zoo-code/settings/mcp_settings.json` and reload window.

## Additional Context

- No duplicate found for `mcp_settings.json race OR McpHub race OR mcp wipe` (searched via `gh issue list`).
- File references: `McpHub.ts:496`, `safeWriteJson.ts:45`, `GlobalFileNames.mcpSettings`.
- Willing to submit PR with above patch.

## Checklist

- [x] Searched existing issues — no duplicate
- [x] Anonymized paths (no personal info)
- [x] Includes code refs and proposed fix

easonLiangWorldedtech:
> Tracking issue: [#1385](https://github.com/Zoo-Code-Org/Zoo-Code/issues/1385)

## Summary

Fixes #1371 (part of epic #1375).

MCP settings initialization has a check-then-create race: `getMcpSettingsFilePath()` did `fileExistsAtPath()` followed by an **unconditional** `fs.writeFile` of the empty stub. Two VS Code windows (separate extension-host processes) racing at startup both see the file as absent, and the second blind write truncates the first window's config to the 122-byte stub — silent data loss of all configured MCP servers.

## Changes

- `src/services/mcp/McpHub.ts` — the stub creation in `getMcpSettingsFilePath()` now goes through `safeWriteJson` with a `merge` callback:
  - the merge read happens **under the existing advisory lock**, so any config written by a concurrent process after the existence check is visible to it;
  - if the on-disk content has a proper `mcpServers` object, it is preserved; otherwise the default stub is written;
  - the fast path is unchanged (file exists → no write at all), so no new writes, no mtime churn, no watcher-triggered MCP reloads.
- `src/services/mcp/__tests__/McpHub.spec.ts`:
  - regression test reproduces the interleaving: the existence check sees an absent file, but the locked read sees a concurrent window's config; asserts the creation write carries that config, not the stub. (This test fails against the old raw `fs.writeFile` implementation.)
  - the `safeWriteJson` spec mock now honors `options.merge` (it previously ignored options, which masked merge semantics for every test in this file).

## Notes

- `safeWriteJson` was already the established pattern for this file (used by the add/remove-server flows); this change removes the one remaining raw `fs.writeFile` in `McpHub`.
- No suppression count changes; `eslint --max-warnings=0` clean on both touched files.
- `pnpm --dir src exec vitest run services/mcp/__tests__/McpHub.spec.ts` → 64/64 passing.

## Summary by CodeRabbit
* **Bug Fixes**
  * Improved reliability when creating MCP settings.
  * Preserves settings added concurrently instead of overwriting them.

---

**Update (CodeRabbit-sync from trial #1413):** head `9dd9825c4` — mcp_settings merge array guard, spec-mock production error parity, unknown-safe error.code read (trial addenda 178e6f405 + 4fc14c48d). Review context: [trial PR #1413](https://github.com/Zoo-Code-Org/Zoo-Code/pull/1413).
---

**Review-gate re-trigger (2026-08-30):** empty commit 046b78cad (no code change) re-runs CI and CodeRabbit current-head review under the org new PR review gate; the code head remains 9dd9825c4.


easonLiangWorldedtech:
Patch coverage gap (Codecov: 2 lines, 1 missing + 1 partial in the merge callback) addressed in 332253af7: added the three remaining fallback cases of getMcpSettingsFilePath (absent file, existing content without an mcpServers object, and mcpServers present but not an object). All changed lines and all merge-callback branches are now covered locally (v8 lcov: lines 519/520/521/523 all hit; BRDA outcomes all > 0); 67/67 McpHub spec tests pass.

easonLiangWorldedtech:
@coderabbitai approve

easonLiangWorldedtech:
## CI status update (2026-09-14)

The `mutation-diff` gate was failing before Stryker started:

> Mutation gate failed: extension has 1009 changed executable lines (limit 500).

Root cause: the gate diffs from the PR's merge-base, which had drifted ~1 week of main commits, so main's own churn was counted as changed code. This commit merges the latest main (`ba46d1f34`) into the branch — the merge-base advances and the diff now contains only this PR's own changes (`McpHub.ts` +25/-6 plus two spec files; specs are excluded from the gate).

Local verification at the merge commit:
- `McpHub.spec.ts` + `McpHub.settingsCreation.integration.spec.ts`: 70/70 pass
- Post-merge delta vs main: 3 files (McpHub.ts 25+/6-, integration spec 176+, McpHub.spec 155+/6-)

Expecting a re-run of the full check suite on the new head.

src/services/mcp/McpHub.ts
```diff
@@ -506,13 +506,32 @@ export class McpHub {
...
+					prettyPrint: true,
+					merge: (existing) => {
+						const parsed = existing as { mcpServers?: unknown } | null
+						// Arrays satisfy `typeof === "object"` but are not a valid
+						// mcpServers map; preserve only a plain object, otherwise the
+						// file would be rewritten with a value McpSettingsSchema
+						// rejects on the next load.
+						if (
+							parsed &&
+							parsed.mcpServers &&
```
edelauna, line 526: Is `{ mcpServers: null }` covered by the unit tests? `typeof null === 'object'` and `!Array.isArray(null)` are both `true`, so removing this truthy check would pass `null` through to `return existing` and write `{ mcpServers: null }` to disk — which `McpSettingsSchema` would then reject on the next load.

src/services/mcp/__tests__/McpHub.spec.ts
```diff
@@ -212,6 +245,122 @@ describe("McpHub", () => {
...
+			// safeWriteJson mock mirrors the production contract — reject before
+			// the merge callback runs instead of treating the file as absent.
+			vi.mocked(fs.readFile).mockRejectedValueOnce(
+				Object.assign(new Error("EACCES: permission denied"), { code: "EACCES" }),
+			)
+
+			await expect(mcpHub.getMcpSettingsFilePath()).rejects.toThrow("EACCES: permission denied")
+			expect(fs.writeFile).not.toHaveBeenCalled()
+		})
+	})
```
edelauna, line 402: Is there a test case for malformed JSON in the locked read? `safeWriteJson` recovers from `SyntaxError` (→ `existing = null` → stub), and the mock already mirrors that contract at line 65, but I don't see a test that exercises this path.

edelauna (changes requested):
Thanks for making this fix - would like to harden it a bit first before merge.

src/services/mcp/McpHub.ts
easonLiangWorldedtech: Covered in fe0d85e0d — new test `writes the default stub when the existing mcpServers value is null`: the locked read returns `{"mcpServers": null}` and the test asserts the written file is the `{ mcpServers: {} }` stub, not the null map.

Since `typeof null === "object"` and `!Array.isArray(null)` both hold, this pins the truthy check on `parsed.mcpServers`: if it were removed, the written data would be `{"mcpServers":null}` and the `toEqual({ mcpServers: {} })` assertion fails. The production guard was already in place (the truthy check at this site) — the gap was the missing coverage.

src/services/mcp/__tests__/McpHub.spec.ts
easonLiangWorldedtech: Added in fe0d85e0d — new test `writes the default stub when the locked read encounters malformed JSON`: the locked read resolves with an unparseable body, `JSON.parse` throws `SyntaxError`, the `safeWriteJson` mock (mirroring production) maps that to `existing = null`, and the test asserts creation resolves with the `{ mcpServers: {} }` stub written. It is the recoverable-absence branch — the existing EACCES test covers the rejecting counterpart.

edelauna (approved):
Thanks!

PR #1442, opened 2026-08-29
refactor(api): make Gemini CLI handler routing explicit

### LINKED ISSUE #1029: [ENHANCEMENT] Make Gemini CLI handler routing explicit

`geminiCli` is the only provider name without an explicit routing case in `buildApiHandler`, so it silently falls through to the default Anthropic handler. This makes the intended routing behavior unclear and easier to change accidentally.

Contributors working on provider registration and API handler routing encounter this when reviewing or modifying `src/api/index.ts`. The gap was identified in review feedback on PR #1012: https://github.com/Zoo-Code-Org/Zoo-Code/pull/1012#discussion_r3652458986

The Gemini CLI provider's current fallback behavior should be represented explicitly so readers can tell that using the Anthropic handler is intentional until a dedicated Gemini CLI handler exists.

Keep the existing runtime behavior unchanged. The change should improve clarity without introducing a dedicated handler as part of this issue.

### Request checklist

N/A

- Given the provider routing switch in `buildApiHandler`
- When a contributor reviews the supported provider names
- Then `geminiCli` has an explicit case
- And the case documents that it currently uses the Anthropic handler pending a dedicated implementation
- And runtime routing behavior remains unchanged

Add an explicit `case providerIdentifiers.geminiCli` immediately before the existing default branch, with a short comment explaining the intentional fallback to the Anthropic handler.

The explicit case still shares behavior with the default branch, but this small duplication of routing intent is preferable to an undocumented implicit fallback.

enestunc10:
Fixes #1029

  Makes the routing for geminiCli explicit within buildApiHandler in src/api/index.ts. Previously, geminiCli was the
  only provider without an explicit case statement, causing it to silently fall through to the default Anthropic
  handler.

  This adds an explicit case providerIdentifiers.geminiCli: with an explanatory comment documenting that falling
  through to the Anthropic handler is intentional pending a dedicated Gemini CLI handler implementation.

  ## Checklist

  [✓] Tested changes locally
  [✓] Followed repository contributing guidelines
  [✓] No breaking runtime changes

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -235,6 +235,8 @@ export function buildApiHandler(configuration: ProviderSettings): ApiHandler {
 			return new BasetenHandler(options)
 		case providerIdentifiers.poe:
 			return new PoeHandler(options)
+		case providerIdentifiers.geminiCli:
+			// Intentionally falls through to the Anthropic handler pending a dedicated Gemini CLI handler implementation.
 		default:
 			return new AnthropicHandler(options)
 	}


edelauna (approved):
Thanks for fixing this

PR #1459, opened 2026-08-31
feat(zoo-gateway): cache model catalog per session with ETag revalida…

JamesRobert20:
## Summary

- Cache zoo-gateway (and kimi-code) model catalogs in a 5-minute in-memory session store keyed by provider, base URL, and session token hash
- Send `If-None-Match` on `/models` when an ETag is known; reuse the prior catalog on `304`
- Clear zoo-gateway session cache entries in `clearZooCodeToken()` on sign-out
- Prune stale `eslint-suppressions.json` entries uncovered while fixing zoo-gateway fetcher tests (required for pre-commit lint)
- **Fix: surface reasoning traces from delta stream**: add `extractReasoningFromDelta` call in the Zoo Gateway streaming loop so thinking traces (`delta.reasoning_content` / `delta.reasoning`) are yielded as reasoning chunks, matching every other OpenAI-compatible provider. Pairs with the server-side fix that now emits `delta.reasoning` in SSE chunks.

## Why

Auth-scoped providers intentionally skip disk and the shared `modelCache` memory store. Without a session cache, every model discovery call hit the gateway. That drove unnecessary request invocations and repeated large catalog downloads.

The reasoning traces fix closes a gap where thinking models (e.g. `qwen3.8-27b`) running through the Zoo Gateway produced empty reasoning blocks in the extension, the delta was correctly emitted server-side by the AI SDK but neither the server SSE handler nor the extension stream loop forwarded it.

## Safety considerations

- Empty API responses are never written to the session cache
- Failed or empty refreshes return the last non-empty session catalog instead of `{}`
- Session cache stays separate from disk/shared cache (no cross-user leakage)
- `useZooGatewayRouterModelsSync` behavior is unchanged: empty zoo-gateway fetches do not overwrite routerModels in React Query

## Test plan

- [x] `pnpm exec vitest run api/providers/fetchers/__tests__/modelCache.spec.ts api/providers/fetchers/__tests__/zoo-gateway.spec.ts` (78 tests - 17 new)
- [ ] Sign in, confirm model picker populates
- [ ] Within 5 minutes, confirm repeated opens do not refetch `/models` (network tab)
- [ ] After TTL, confirm catalog still loads (ETag `304` path when unchanged)
- [ ] Sign out and sign in as another account; confirm catalog refetches and session cache is cleared
- [ ] On a thinking model (e.g. qwen3.8-27b), confirm reasoning traces appear in the extension UI


src/api/providers/fetchers/modelCache.ts
```diff
@@ -267,9 +432,18 @@ async function fetchModelsFromProvider(options: GetModelsOptions): Promise<Model
...
+			if (result.kind === "not_modified") {
+				models = {}
+				break
+			}
+			models = result.models
 			break
+		}
 		case providerIdentifiers.kimiCode:
 			models = await getKimiCodeModels(options.apiKey)
 			break
```
edelauna: Is this branch reachable? Both `getModels` (line 475) and `refreshModels` (line 570) return early via `isAuthScopedProvider` before `fetchModelsFromProvider` is ever called, so these `zooGateway` and `kimiCode` arms never execute. The 304 path here also can't fire since no `If-None-Match` is sent — but a reader might think it's live.

Worth removing both cases and adding an explicit guard at the top of `fetchModelsFromProvider` to catch any future accidental routing of auth-scoped providers here?

src/api/providers/fetchers/modelCache.ts
```diff
@@ -108,10 +108,175 @@ const AUTH_SCOPED_PROVIDERS: ReadonlySet<RouterName> = new Set([
...
+		if (existing && Object.keys(existing.models).length > 0) {
+			return existing.models
+		}
+
+		return fetched.models
+	} catch (error) {
+		if (existing && Object.keys(existing.models).length > 0) {
+			return existing.models
+		}
+		throw error
```
edelauna: Is the cold-start error path covered by a test? The existing spec ("returns the prior session catalog when a fetch throws") requires a prior successful fetch to populate `existing` — when `existing` is `undefined` and this is the very first call, the error propagates uncaught to the `getModels` caller. A test asserting `await expect(freshGetModels(options)).rejects.toThrow(...)` on a first-call network failure would pin this contract.

src/api/providers/fetchers/__tests__/modelCache.spec.ts
```diff
@@ -1204,3 +1210,309 @@ describe("compound cache key derivation across scoping dimensions", () => {
...
+
+			const options = { provider: providerIdentifiers.zooGateway, apiKey: "session-token" }
+			await freshGetModels(options)
+
+			vi.advanceTimersByTime(5 * 60 * 1000 + 1)
+
+			const second = await freshGetModels(options)
+
+			expect(second).toEqual(zooModels)
+			expect(freshMockGetZooGatewayModels).toHaveBeenCalledTimes(2)
```
edelauna: Does this verify that the TTL was actually reset? If `touchAuthSessionEntry` kept the original `fetchedAt` instead of updating it, `second.toEqual(zooModels)` and `toHaveBeenCalledTimes(2)` would still both pass — but the entry would expire again immediately after. What about making a third `freshGetModels` call here (within the new TTL window, before advancing time) and asserting call count remains 2?

src/api/providers/fetchers/modelCache.ts
```diff
@@ -108,10 +108,175 @@ const AUTH_SCOPED_PROVIDERS: ReadonlySet<RouterName> = new Set([
...
+
+	try {
+		const fetched = await fetchAuthScopedModelsFromProvider(
+			assertAuthScopedGetModelsOptions(options),
+			existing?.etag,
+		)
+
+		if (fetched.notModified) {
+			if (existing && Object.keys(existing.models).length > 0) {
+				touchAuthSessionEntry(cacheKey, existing)
```
edelauna: If `clearAuthSessionModelsForProvider` runs during the `await` above — e.g. a concurrent sign-out — `existing` is a pre-clear snapshot. A 304 response here would then call `touchAuthSessionEntry(cacheKey, existing)` and re-insert the cleared entry. Would re-reading from the Map be safer?

```suggestion
			const current = getAuthSessionEntry(cacheKey)
			if (current && Object.keys(current.models).length > 0) {
				touchAuthSessionEntry(cacheKey, current)
```

src/api/providers/fetchers/modelCache.ts
```diff
@@ -108,10 +108,175 @@ const AUTH_SCOPED_PROVIDERS: ReadonlySet<RouterName> = new Set([
...
+			authSessionCache.delete(key)
+		}
+	}
+}
+
+type AuthScopedFetchResult = {
+	models: ModelRecord
+	etag?: string
+	notModified?: boolean
+}
```
edelauna, line 247: `ZooGatewayModelsFetchResult` (the upstream type this wraps) uses a `kind` discriminant. Would aligning `AuthScopedFetchResult` to the same shape be cleaner? When `notModified` is `true`, `models` is a meaningless `{}` — the type doesn't express that.

```suggestion
type AuthScopedFetchResult = { kind: "ok"; models: ModelRecord; etag?: string } | { kind: "not_modified" }
```

src/api/providers/fetchers/zoo-gateway.ts
```diff
@@ -20,29 +20,51 @@ const MODEL_DISCOVERY_TIMEOUT_MS = 15_000
...
+		if (response.status === 304) {
+			return { kind: "not_modified" }
+		}
+
+		const etag =
+			typeof response.headers.etag === "string"
+				? response.headers.etag
+				: typeof response.headers.ETag === "string"
+					? response.headers.ETag
+					: undefined
```
edelauna: Node.js lowercases all HTTP response header names before they reach userland, so `response.headers.ETag` will always be `undefined` here. Only `response.headers.etag` can hold a value.

```suggestion
		const etag = typeof response.headers.etag === "string" ? response.headers.etag : undefined
```

src/api/providers/fetchers/modelCache.ts
```diff
@@ -108,10 +108,175 @@ const AUTH_SCOPED_PROVIDERS: ReadonlySet<RouterName> = new Set([
...
+	const { provider } = options
+	const cacheKey = getCacheKey(options)
+	const existing = getAuthSessionEntry(cacheKey)
+
+	if (!forceRefresh && existing && isAuthSessionFresh(existing) && Object.keys(existing.models).length > 0) {
+		return existing.models
+	}
+
+	try {
+		const fetched = await fetchAuthScopedModelsFromProvider(
```
edelauna: What happens if two `getModels` calls race here — both miss the session cache (e.g. two panels opening on startup) and both enter `fetchAuthScopedModelsFromProvider`? The non-auth-scoped path coalesces concurrent fetches via `dedupedFetch`. Is the doubled API call intentional for auth-scoped providers, or should we gate the fetch behind an in-flight map keyed on `cacheKey`?

edelauna (changes requested):
Nice! Had some questions regarding the implementation.

src/api/providers/fetchers/__tests__/modelCache.spec.ts
```diff
@@ -1204,3 +1238,926 @@ describe("compound cache key derivation across scoping dimensions", () => {
...
+	it("treats an entry as stale at exactly the TTL boundary", async () => {
+		// Kills: < vs <= equality-operator mutant on isAuthSessionFresh
+		vi.useFakeTimers()
+		try {
+			mockGetZooGatewayModels.mockResolvedValue(zooGatewayOk(zooModels))
+			const options = { provider: providerIdentifiers.zooGateway, apiKey: "session-token" }
+
+			await getModels(options)
+			// Advance to exactly TTL — entry is now stale (< not <=)
+			vi.advanceTimersByTime(5 * 60 * 1000)
```
edelauna, line 1447: This test confirms the entry is stale at exactly `AUTH_SESSION_TTL_MS`. Is there a companion assertion that advances to `TTL - 1` ms and confirms the entry is still served from cache? Without it, a mutant that tweaks the staleness comparison (e.g. `<` → `<=`) survives the suite.

src/api/providers/fetchers/zoo-gateway.ts
```diff
@@ -20,29 +20,48 @@ const MODEL_DISCOVERY_TIMEOUT_MS = 15_000
  * Fetches models from the Zoo Gateway API. Requires authentication via the zoo_ext_ token.
  */
 
-export async function getZooGatewayModels(options?: ApiHandlerOptions): Promise<Record<string, ModelInfo>> {
+export type ZooGatewayModelsFetchResult =
+	| { kind: "ok"; models: Record<string, ModelInfo>; etag?: string }
+	| { kind: "not_modified" }
+
+export async function getZooGatewayModels(
```
edelauna, line 27: The JSDoc block (`getZooGatewayModels` / `Fetches models from the Zoo Gateway API…`) sits above the new `ZooGatewayModelsFetchResult` type, so TypeScript attaches it to the type instead of the function. Would it make sense to move the block below the type, and document `ifNoneMatch` and the `not_modified` return path while it's there?

src/api/providers/fetchers/modelCache.ts
```diff
@@ -108,10 +119,243 @@ const AUTH_SCOPED_PROVIDERS: ReadonlySet<RouterName> = new Set([
...
+				})
+			}
+
+			if (existing && authSessionHasModels(existing.models)) {
+				return existing.models
+			}
+
+			return fetched.models
+		} catch (error) {
+			if (existing && authSessionHasModels(existing.models)) {
```
edelauna, line 341: When the fetch fails and a stale entry is available, the stale models are returned silently. Since kimi-code has no internal logging, would it help to add a `console.error` here so the failure is observable?

src/api/providers/fetchers/modelCache.ts
```diff
@@ -108,10 +119,243 @@ const AUTH_SCOPED_PROVIDERS: ReadonlySet<RouterName> = new Set([
...
+
+			if (fetched.notModified) {
+				// Re-read from the Map: a concurrent sign-out could have cleared
+				// the entry between when we captured `existing` and now.
+				const current = getAuthSessionEntry(cacheKey)
+				if (current && authSessionHasModels(current.models)) {
+					touchAuthSessionEntry(cacheKey, current)
+					reportedEmptyModelResponse.delete(cacheKey)
+					return current.models
+				}
```
edelauna, line 310: The comment on line 303 correctly flags the sign-out race for the `not_modified` path — but if `current` is `undefined` (sign-out fired before the 304 arrived), the code falls through to line 333 and returns `existing.models`, the now-invalidated catalog. Should this branch return early?

```suggestion
				if (current && authSessionHasModels(current.models)) {
					touchAuthSessionEntry(cacheKey, current)
					reportedEmptyModelResponse.delete(cacheKey)
					return current.models
				}
				// Sign-out cleared the entry while the 304 was in-flight.
				return {}
```

src/api/providers/fetchers/modelCache.ts
```diff
@@ -108,10 +119,243 @@ const AUTH_SCOPED_PROVIDERS: ReadonlySet<RouterName> = new Set([
...
+				assertAuthScopedGetModelsOptions(options),
+				existing?.etag,
+			)
+
+			if (fetched.notModified) {
+				// Re-read from the Map: a concurrent sign-out could have cleared
+				// the entry between when we captured `existing` and now.
+				const current = getAuthSessionEntry(cacheKey)
+				if (current && authSessionHasModels(current.models)) {
+					touchAuthSessionEntry(cacheKey, current)
```
edelauna, line 307: A mutation from `current` to `existing` here (passing the pre-fetch snapshot to `touchAuthSessionEntry` instead of the re-read value) would survive the suite — no test fires sign-out between the 304 arriving and this touch call. Worth adding one?

src/services/zoo-code-auth.ts
```diff
@@ -154,6 +154,9 @@ export async function clearZooCodeToken(): Promise<void> {
 	await secretStorage.delete(ZOO_CODE_TOKEN_KEY)
 	_cachedToken = undefined
 	_sessionCleared = true
+	const { clearAuthSessionModelsForProvider } = await import("../api/providers/fetchers/modelCache")
+	const { providerIdentifiers } = await import("@roo-code/types")
```
edelauna, line 158: `@roo-code/types` is a leaf package with no circular-import risk — only the `modelCache` import needs to be dynamic here. Could `providerIdentifiers` move to the static imports at the top of the file? Same pattern at `src/integrations/kimi-code/oauth.ts:236`.

edelauna (commented):
Approving, but had a couple comments related to the implementation, main one being the edge case for concurrent signout and 304 arriving concurrently, will try to implement a fix.

edelauna (approved):

PR #1481, opened 2026-09-01
fix(terminal): stop forcing en_US.UTF-8 in execa command environment

### LINKED ISSUE #1084: Zoo Code forces en_US.UTF-8 locale on every command it runs, causing errors on systems that don't have that locale installed

Every time Zoo Code runs a command, it injects LANG=en_US.UTF-8 and LC_ALL=en_US.UTF-8 into the shell environment. On computers where that specific locale isn't installed, every command produces a setlocale warning, and locale-sensitive tools behave as if the system is US English even when it isn't.

Anyone using Zoo Code on a machine whose locale is not en_US.UTF-8 — for example Australian users (en_AU), British users (en_GB), or anyone in a non-US locale that only has their own locale generated. They see the warning on every single command Zoo Code executes, regardless of what the command is. The warning doesn't appear in a normal terminal — only in commands run by Zoo Code.

### Reproduction steps

On a Linux system (tested on WSL2/Arch Linux) generate only your local locale, e.g. en_AU.UTF-8, and confirm en_US.UTF-8 is NOT listed by locale -a
Make sure the system locale is set correctly (LANG=en_AU.UTF-8 in /etc/locale.conf)
Open a project in Zoo Code (VSCode Remote-WSL) and ask it to run any simple command, e.g. echo hello
Observe the command output

### Expected result

The command runs cleanly with the system's locale (en_AU.UTF-8), with no locale warnings.

### Actual result

Every command output is prefixed with:  /bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8): No such file or directory  The warning appears only in Zoo Code-spawned shells. A normal interactive terminal on the same machine shows no warning, confirming the system config is correct and Zoo Code is overriding it.

Switching Zoo Code's UI language (e.g., to another locale) — the warning and the injected en_US.UTF-8 did not change at all
Setting the locale via ~/.vscode-server/server-env-setup (which correctly sets LANG=en_AU.UTF-8 on the VSCode server process) — the server itself then runs with the correct locale, but Zoo Code-spawned command shells still get en_US.UTF-8
Generating the en_US.UTF-8 locale as a workaround — the warning disappears (but this is a workaround, not a fix)

### App Version

v3.74.0

### API Provider (optional)

None

### Model Used (optional)

_No response_

_No response_

```shell
Root cause found in the extension bundle (dist/extension.js). The locale is a hardcoded literal (appears exactly once in the whole bundle, unrelated to any language/locale setting):

env: {...RKe.default.env, LANG: "en_US.UTF-8", LC_ALL: "en_US.UTF-8"}

Shell environment of a command spawned by Zoo Code (parent processes all have the correct en_AU.UTF-8):

LANG=en_US.UTF-8
LC_ALL=en_US.UTF-8

Suggested fix: inherit the server/host locale instead of hardcoding it (e.g., only set LANG/LC_ALL if not already present, or make it configurable like VSCode's terminal.integrated.env.linux).
```

mmskazak:
### Related GitHub Issue

Closes: #1084

### Description

`ExecaTerminalProcess` unconditionally overwrote `LANG`/`LC_ALL` with `en_US.UTF-8` for every command it runs, even when the host already had a correctly configured non-US UTF-8 locale (e.g. `en_AU.UTF-8`). This produced a `setlocale: LC_ALL: cannot change locale (en_US.UTF-8): No such file or directory` warning on every single command for anyone whose system locale isn't `en_US.UTF-8` and doesn't have that specific locale generated (confirmed root cause in the issue via the bundled `dist/extension.js`).

- `src/integrations/terminal/ExecaTerminalProcess.ts`: added `ensureUtf8Locale(value)`, which:
  - Preserves the existing `LANG`/`LC_ALL` value as-is if it already specifies a UTF-8 encoding (e.g. `en_AU.UTF-8` stays `en_AU.UTF-8`).
  - Upgrades the encoding portion of a locale that specifies a non-UTF-8 encoding while keeping its language/territory (e.g. `de_DE.ISO-8859-1` becomes `de_DE.UTF-8`).
  - Falls back to `en_US.UTF-8` only when the value is unset, or is one of the encoding-less POSIX defaults (`C`/`POSIX`) — matching the original intent of the code (the comment says "Ensure UTF-8 encoding for Ruby, CocoaPods, etc.") without clobbering a locale the system already had correctly configured.
- `src/integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts`: updated the existing "should set LANG and LC_ALL to en_US.UTF-8" test to explicitly unset `LANG`/`LC_ALL` first (previously it implicitly relied on the old code ignoring the ambient environment entirely, which made the assertion depend on whatever locale happened to be set on the machine running the test — it silently passed in CI only because CI runners don't set a non-US UTF-8 locale). Added two new cases covering the actual bug: preserving an already-UTF-8 non-US locale, and upgrading a non-UTF-8 encoding while keeping the language/territory.

### Test Procedure

- `npx vitest run integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts` — 15/15 passed, including the 2 new cases reproducing the reported bug (`en_AU.UTF-8` preserved) and the encoding-upgrade case (`de_DE.ISO-8859-1` → `de_DE.UTF-8`).
- `npx vitest run integrations/terminal` — 189 passed / 23 skipped, no regressions in the wider terminal integration suite.
- `tsc --noEmit` — clean.
- `eslint . --ext=ts --max-warnings=0` (via `pnpm lint`, all 13 packages) — clean, no new `@typescript-eslint/no-explicit-any` suppressions added (used a `unknown` double-cast for the two new test assertions instead of `any`, matching the file's existing suppression budget in `eslint-suppressions.json`).
- Manually verified the reported scenario at the unit level: with `LANG=en_AU.UTF-8` / `LC_ALL=en_AU.UTF-8` set, the spawned command environment now keeps `en_AU.UTF-8` instead of being forced to `en_US.UTF-8`.

### Pre-Submission Checklist

- [x] **Issue Linked**: This PR is linked to #1084.
- [x] **Scope**: Limited to the locale-forcing fix and its test coverage.
- [x] **Self-Review**: Reviewed the diff myself.
- [x] **Testing**: New tests added covering the reported bug and the encoding-upgrade case; full terminal suite passes.
- [ ] **Visual Snapshot** (UI changes only): Not applicable — no UI change.
- [x] **Documentation Impact**: No user-facing documentation update needed; added a changeset describing the fix for the release notes.

### Documentation Updates

### Additional Notes

I found this issue unassigned and unclaimed while looking for a well-scoped bug to fix. I wasn't able to complete the Discord assignment step described in CONTRIBUTING.md before opening this — happy to withdraw or wait if a maintainer would rather this go through that process first, but the fix itself is small, fully tested, and root-caused directly from the issue's own diagnosis.

### Get in Touch

GitHub: @mmskazak

### DIFF
--- a/.changeset/fix-execa-terminal-locale-override.md
+++ b/.changeset/fix-execa-terminal-locale-override.md
@@ -1,0 +1,5 @@
+---
+"zoo-code": patch
+---
+
+Fix commands run by Zoo Code forcing `LANG`/`LC_ALL` to `en_US.UTF-8` even when the system already has a correctly configured non-US UTF-8 locale (e.g. `en_AU.UTF-8`), which caused a `setlocale: LC_ALL: cannot change locale` warning on every command for anyone whose system locale isn't `en_US.UTF-8`. The existing locale is now preserved when it already specifies a UTF-8 encoding; only an unset locale or an encoding-less POSIX default (`C`/`POSIX`) falls back to `en_US.UTF-8`, and a locale with a non-UTF-8 encoding has its encoding upgraded while its language/territory is kept.
--- a/src/eslint-suppressions.json
+++ b/src/eslint-suppressions.json
@@ -1206,7 +1206,7 @@
 	},
 	"integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts": {
 		"@typescript-eslint/no-explicit-any": {
-			"count": 10
+			"count": 8
 		}
 	},
 	"integrations/terminal/__tests__/OutputInterceptor.test.ts": {
--- a/src/integrations/terminal/ExecaTerminalProcess.ts
+++ b/src/integrations/terminal/ExecaTerminalProcess.ts
@@ -8,21 +8,26 @@ import { BaseTerminalProcess } from "./BaseTerminalProcess"
 
 /**
  * Returns a UTF-8 locale string derived from `value`, preserving the
- * language/territory the system already has configured instead of
- * forcing en_US. Falls back to en_US.UTF-8 when `value` is unset or is
- * one of the encoding-less POSIX defaults ("C"/"POSIX").
+ * language/territory (and any @modifier, e.g. "de_DE@euro") the system
+ * already has configured instead of forcing en_US. Falls back to
+ * en_US.UTF-8 when `value` is unset or is one of the encoding-less POSIX
+ * defaults ("C"/"POSIX").
  */
 export function ensureUtf8Locale(value: string | undefined): string {
 	if (!value || value === "C" || value === "POSIX") {
 		return "en_US.UTF-8"
 	}
 
-	if (/utf-?8$/i.test(value)) {
+	const atIndex = value.indexOf("@")
+	const modifier = atIndex === -1 ? "" : value.slice(atIndex)
+	const localeAndEncoding = atIndex === -1 ? value : value.slice(0, atIndex)
+
+	if (/utf-?8$/i.test(localeAndEncoding)) {
 		return value
 	}
 
-	const [base] = value.split(".")
-	return `${base}.UTF-8`
+	const [base] = localeAndEncoding.split(".")
+	return `${base}.UTF-8${modifier}`
 }
 
 export class ExecaTerminalProcess extends BaseTerminalProcess {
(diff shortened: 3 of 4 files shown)


mmskazak:
Thanks for the review — both actionable findings were real bugs, fixed in the latest commit:

- \`LC_ALL\` was being fabricated to \`en_US.UTF-8\` whenever it was unset, even if \`LANG\` was already correctly configured. Since \`LC_ALL\` overrides \`LANG\`, this silently re-forced en_US for the common case where only \`LANG\` is set — recreating the exact bug from #1084. \`LC_ALL\` is now only normalized (and only ever set) when the environment already had it.
- \`ensureUtf8Locale()\` now parses a trailing \`@modifier\` (e.g. \`de_DE@euro\`) separately, so it survives an encoding upgrade instead of being misplaced or dropped.

Added regression tests for both, plus the isolated C/POSIX cases you asked for, and replaced the remaining \`as any\` test casts with a typed helper (pruned the suppression count for the test file accordingly).

Left the changeset in place — this repo's other routine fix PRs (e.g. fix-diff-scroll-position.md) add one the same way, so it looks like standard convention here rather than release-prep-only.

PR #1484, opened 2026-09-01
fix: _isGrokXAI() false-positive substring match breaks token usage for domains containing "x.ai"

### LINKED ISSUE #1483: [BUG] _isGrokXAI() false-positive: substring match on "x.ai" breaks token usage for any OpenAI-compatible provider whose domain contains "x.ai" (e.g. box.ai, fox.ai, max.ai)
### Description

The `_isGrokXAI()` method in `src/api/providers/openai.ts` uses `urlHost.includes("x.ai")` to detect whether the base URL points to the xAI/Grok API. However, `.includes()` is a **substring match**, meaning any domain that contains "x.ai" anywhere in its host will be falsely identified as a Grok/xAI endpoint.

This includes domains like:
- `box.ai`
- `fox.ai`
- `max.ai`
- Any domain ending in `x.ai` that is not actually `api.x.ai` or a subdomain of `x.ai`

When this false-positive triggers, `stream_options: { include_usage: true }` is **omitted** from the API request. As a result, the API never returns usage data, and the **token bar shows 0** — a silent failure with no error message.

### Root Cause

```typescript
// src/api/providers/openai.ts, line 519-522
private _isGrokXAI(baseUrl?: string): boolean {
    const urlHost = this._getUrlHost(baseUrl)
    return urlHost.includes("x.ai")  // <-- substring match, false-positive
}
```

The bug affects **two** code paths:
1. `createMessage()` — main message streaming (line 153)
2. `handleO3FamilyMessage()` — O3 family model streaming (line 351)

Both conditionally omit `stream_options` when `_isGrokXAI()` returns true:
```typescript
...(isGrokXAI ? {} : { stream_options: { include_usage: true } })
```

### Reproduction

1. Configure an OpenAI-compatible provider with base URL `https://box.ai/v1` (or any URL whose host contains "x.ai" as a substring but is not `api.x.ai`)
2. Send a message in Zoo Code
3. Observe: token bar shows 0, no usage data is displayed

### Expected Behavior

`_isGrokXAI()` should only return `true` for:
- `api.x.ai` (exact match)
- `*.x.ai` (subdomains of x.ai, e.g. `custom.x.ai`)

It should **not** match domains like `box.ai`, `fox.ai`, `max.ai` where "x.ai" appears as a substring of a different domain.

### Proposed Fix

```typescript
private _isGrokXAI(baseUrl?: string): boolean {
    const urlHost = this._getUrlHost(baseUrl)
    return urlHost === "api.x.ai" || urlHost.endsWith(".x.ai")
}
```

### Environment

- Zoo Code version: 3.80.0 (VS Code), 3.80.1 (IntelliJ)
- Affected platforms: VS Code, IntelliJ IDEA
- File: `src/api/providers/openai.ts`

### Severity

Medium-High — silently breaks a core feature (token tracking) for a broad class of users (anyone using a proxy/provider whose domain contains "x.ai" as a substring), with no error message or visible indication of the problem.

BambinoSK:
### Summary

The `_isGrokXAI()` method in `src/api/providers/openai.ts` used `urlHost.includes("x.ai")` which is a substring match. Any domain containing "x.ai" as a substring (e.g. `box.ai`, `fox.ai`, `max.ai`) was falsely identified as a Grok/xAI endpoint. This caused `stream_options: { include_usage: true }` to be omitted from API requests, so the API never returned usage data and the token bar showed 0 — a silent failure with no error message.

### Root Cause

```typescript
// Before (buggy)
private _isGrokXAI(baseUrl?: string): boolean {
    const urlHost = this._getUrlHost(baseUrl)
    return urlHost.includes("x.ai")  // substring match — false positive for box.ai, fox.ai, etc.
}
```

The bug affects two code paths:
1. `createMessage()` — main message streaming (line 153)
2. `handleO3FamilyMessage()` — O3 family model streaming (line 351)

Both conditionally omit `stream_options` when `_isGrokXAI()` returns true:
```typescript
...(isGrokXAI ? {} : { stream_options: { include_usage: true } })
```

### Fix

```typescript
// After (fixed)
private _isGrokXAI(baseUrl?: string): boolean {
    const urlHost = this._getUrlHost(baseUrl)
    return urlHost === "api.x.ai" || urlHost.endsWith(".x.ai")
}
```

This ensures only `api.x.ai` and subdomains of `x.ai` (e.g. `custom.x.ai`) are detected as Grok/xAI endpoints.

### Changes

- `src/api/providers/openai.ts`: Changed `_isGrokXAI()` to use exact host match or subdomain suffix check instead of substring `includes()`
- `src/api/providers/__tests__/openai.spec.ts`: Added test suite "Grok xAI false-positive prevention" with 5 test cases:
  - `box.ai` should NOT be detected as Grok xAI
  - `fox.ai` and `max.ai` should NOT be detected as Grok xAI
  - `api.x.ai` SHOULD be detected as Grok xAI
  - `custom.x.ai` (subdomain) SHOULD be detected as Grok xAI
  - `stream_options` should be included when using a non-Grok provider whose URL contains "x.ai" substring

### Testing

All existing Grok xAI tests continue to pass. New tests verify the false-positive scenarios are fixed.

### Related Issue

Fixes #1483

### AI Assistance Disclosure

This PR was developed with AI assistance (Roo Code / Zoo Code with GLM-5.2). The contributor has reviewed and understands every meaningful change, can explain the implementation and tradeoffs, and has verified the fix against the actual installed plugin (both VS Code and IntelliJ). The fix is a one-line change to the `_isGrokXAI()` method plus corresponding test cases.

### DIFF
--- a/src/api/providers/__tests__/openai.spec.ts
+++ b/src/api/providers/__tests__/openai.spec.ts
@@ -1060,29 +1060,54 @@ describe("OpenAiHandler", () => {
 				openAiModelId: "gpt-4o",
 			}
 			const handler = new OpenAiHandler(nonGrokOptions)
-			// @ts-expect-error - accessing private method for testing
-			expect(handler._isGrokXAI(nonGrokOptions.openAiBaseUrl)).toBe(false)
+			expect(handler["_isGrokXAI"](nonGrokOptions.openAiBaseUrl)).toBe(false)
 		})
 
 		it("should NOT detect as Grok xAI for other domains containing 'x.ai' substring (e.g. fox.ai, max.ai)", () => {
 			const handler = new OpenAiHandler({ ...mockOptions, openAiBaseUrl: "https://fox.ai/v1" })
-			// @ts-expect-error - accessing private method for testing
-			expect(handler._isGrokXAI("https://fox.ai/v1")).toBe(false)
-
-			// @ts-expect-error - accessing private method for testing
-			expect(handler._isGrokXAI("https://max.ai/v1")).toBe(false)
+			expect(handler["_isGrokXAI"]("https://fox.ai/v1")).toBe(false)
+			expect(handler["_isGrokXAI"]("https://max.ai/v1")).toBe(false)
 		})
 
 		it("should detect as Grok xAI for api.x.ai", () => {
 			const handler = new OpenAiHandler({ ...mockOptions, openAiBaseUrl: "https://api.x.ai/v1" })
-			// @ts-expect-error - accessing private method for testing
-			expect(handler._isGrokXAI("https://api.x.ai/v1")).toBe(true)
+			expect(handler["_isGrokXAI"]("https://api.x.ai/v1")).toBe(true)
 		})
 
 		it("should detect as Grok xAI for subdomains of x.ai (e.g. custom.x.ai)", () => {
 			const handler = new OpenAiHandler({ ...mockOptions, openAiBaseUrl: "https://custom.x.ai/v1" })
-			// @ts-expect-error - accessing private method for testing
-			expect(handler._isGrokXAI("https://custom.x.ai/v1")).toBe(true)
+			expect(handler["_isGrokXAI"]("https://custom.x.ai/v1")).toBe(true)
+		})
+
+		it("should detect as Grok xAI when api.x.ai uses a non-default port", () => {
+			const handler = new OpenAiHandler({ ...mockOptions, openAiBaseUrl: "https://api.x.ai:8443/v1" })
+			expect(handler["_isGrokXAI"]("https://api.x.ai:8443/v1")).toBe(true)
+		})
+
+		it("should exclude stream_options when streaming with api.x.ai on a non-default port", async () => {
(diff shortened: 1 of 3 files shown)


src/api/providers/__tests__/openai.spec.ts
```diff
@@ -1052,6 +1052,92 @@ describe("OpenAiHandler", () => {
...
+				}),
+				{},
+			)
+
+			const mockCalls = mockCreate.mock.calls
+			const lastCall = mockCalls[mockCalls.length - 1]
+			expect(lastCall[0]).toHaveProperty("stream_options")
+			expect(lastCall[0].stream_options).toEqual({ include_usage: true })
+		})
+	})
```
edelauna: These tests cover the `createMessage` path. Does `handleO3FamilyMessage` (which also calls `_isGrokXAI` at line 351 and gates `stream_options` at line 363) need coverage here too — e.g. an O3 model with an `api.x.ai` base URL?

edelauna (approved):

PR #1369, opened 2026-08-25
fix(openrouter): support extended reasoning efforts and add refresh models button

myk1yt:
### Description

This PR resolves two related issues with the OpenRouter provider:

1. **Extended Reasoning Effort Support (`xhigh`, `max`)**:
   - **Problem**: When fetching models dynamically from OpenRouter (`/api/v1/models`), reasoning capability is advertised as `"reasoning"` in `supported_parameters`. The parser set `supportsReasoningEffort: true` (boolean). In `ThinkingBudget`, boolean `true` fell back to the basic 3-tier enum (`["low", "medium", "high"]`), preventing users from selecting higher reasoning tiers (`xhigh` / "Extra High" and `max` / "Max") that OpenRouter models (like OpenAI o1, o3-mini, etc.) actually support.
   - **Fix**: In `src/api/providers/fetchers/openrouter.ts`, `supportsReasoningEffort` is now explicitly set to `["low", "medium", "high", "xhigh", "max"]` when `"reasoning"` is present in `supported_parameters`.

2. **"Refresh Models" Button in OpenRouter Settings**:
   - **Problem**: Model lists fetched from OpenRouter are persisted in disk cache without a manual invalidation trigger in the UI, unlike LiteLLM, Moonshot, Poe, etc.
   - **Fix**: Added the standard "Refresh Models" button and IPC message listener in `webview-ui/src/components/settings/providers/OpenRouter.tsx` (matching the pattern in `LiteLLM.tsx` / `Moonshot.tsx`), allowing users to flush the cache and fetch fresh model metadata on demand.

**Scope & Blast Radius**:
- Changes are strictly isolated to the OpenRouter provider (`fetchers/openrouter.ts` and `providers/OpenRouter.tsx`). No shared core components were modified.

### Test Procedure

1. **Verify "Refresh Models" Button**:
   - Open Settings → API Provider → Select **OpenRouter**.
   - Verify the "Refresh Models" button appears above the model selector.
   - Click "Refresh Models" → verify the loading spinner displays, cache is flushed, and success feedback is shown.
2. **Verify Reasoning Effort Options**:
   - Select a reasoning-capable model (e.g. `openai/o3-mini`, `openai/o1`, `deepseek/deepseek-r1`).
   - Open the **Model Reasoning Effort** dropdown.
   - Confirm all 5 options are present: `None`, `Low`, `Medium`, `High`, `Extra High`, `Max`.
[...]

PR #1495, opened 2026-09-03
fix(delegation): preserve live child delegation links across extension host startup (multi-window subtask return)

myk1yt:
## Summary
- Preserves live child delegation links across extension host startup when users work in multiple VS Code windows. Previously, when the extension host started in another window, a child task actively running there was misjudged as a crash orphan and repaired, breaking the parent's `delegatedToId`/`awaitingChildId` links so completing the subtask failed to return to the parent.

## Root Cause
- `TaskHistoryStore.reconcileDelegationState()` treated any persisted "active" child without a live session as a crash orphan, regardless of whether another extension host was still actively writing its history file.

## Fix (2 files)
1. `src/core/task-persistence/TaskHistoryStore.ts` (+35)
   - Add `LIVE_CHILD_MTIME_THRESHOLD_MS = 5 * 60 * 1000` (≥ reconcile interval so sparse writers are safe).
   - In reconcile: before repairing a persisted-active child, check its `history_item.json` mtime. If written within the threshold, the child is live in another window → skip repair and log.
   - New helper `getChildFileMtimeMs(childId)` returns mtime (undefined → conservatively proceed with repair).
2. `src/core/task-persistence/__tests__/TaskHistoryStore.reconciliation.spec.ts` (+98)
   - New tests: "skips repair for active child with recent mtime (live in another window)" and "repairs active child with stale mtime (crash orphan)".
   - Existing multi-window-sensitive tests updated with `markStaleMtime()` to represent real crash orphans.

## Verification
- Local: `TaskHistoryStore.reconciliation.spec.ts` 50/50 pass; `attemptCompletionTool.spec.ts` 22/22; delegation regression specs (`history-resume-delegation` / `nested-delegation-resume`) 23/23; `tsc --noEmit` 0 errors; full lint (13 packages) PASS.
- Merged into integration build `feat/combined-vsix-260903` and shipped in VSIX `3.80.1-combined-260903` — manual multi-window provider-switch → subtask complete → parent return verified.

## Notes
- Fork branch `fix/returntoparent` is purely this fix (1 commit, 2 files) rebased on latest main.
- Backups: `backup/fix-returntoparent-260903` (originals preserved separately).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,14 @@ qdrant_storage/
 plans/
 
 roo-cli-*.tar.gz*
+
+# Local worktrees (multi-branch development)
+.wt-*/
+
+# Temporary scratch files
+.tmp-*
+.zoo-status.txt
+untracked-*.txt
+class_*.txt
+check-dup2-result.txt
+*.tsbuildinfo
--- a/.gitignore
+++ b/.gitignore
@@ -59,14 +59,3 @@ qdrant_storage/
 plans/
 
 roo-cli-*.tar.gz*
-
-# Local worktrees (multi-branch development)
-.wt-*/
-
-# Temporary scratch files
-.tmp-*
-.zoo-status.txt
-untracked-*.txt
-class_*.txt
-check-dup2-result.txt
-*.tsbuildinfo
--- a/src/core/task-persistence/TaskHistoryStore.ts
+++ b/src/core/task-persistence/TaskHistoryStore.ts
@@ -97,6 +97,13 @@ export class TaskHistoryStore {
 	/** Periodic reconciliation interval in milliseconds. */
 	private static readonly RECONCILE_INTERVAL_MS = 5 * 60 * 1000
 
+	/**
+	 * Maximum age (in ms) of a child's history file mtime for the child to be
+	 * considered live in another window. Kept at least as long as the reconcile
+	 * interval so live tasks with sparse writes are not misjudged as orphans.
+	 */
+	private static readonly LIVE_CHILD_MTIME_THRESHOLD_MS = 5 * 60 * 1000 // 5 minutes
+
 	constructor(globalStoragePath: string, options?: TaskHistoryStoreOptions) {
 		this.globalStoragePath = globalStoragePath
 		this.onWrite = options?.onWrite
@@ -466,6 +473,19 @@ export class TaskHistoryStore {
 						)
 						repairsInThisPass++
 					} else if ((child.status ?? "active") === "active" && persistedActiveIds.has(child.id)) {
+						// Cross-instance liveness guard: a child whose history file was written
+						// recently is owned by another live window, not a crash orphan.
+						const mtimeMs = await this.getChildFileMtimeMs(child.id)
+						const isLiveElsewhere =
+							mtimeMs !== undefined &&
+							Date.now() - mtimeMs < TaskHistoryStore.LIVE_CHILD_MTIME_THRESHOLD_MS
+						if (isLiveElsewhere) {
+							console.log(
+								`[TaskHistoryStore] Skipping repair for live child ${child.id} ` +
+									`(mtime ${Math.round((Date.now() - mtimeMs) / 1000)}s ago) — owned by another window`,
+							)
+							continue
+						}
 						// An active child persisted across startup cannot have a live task session
 						// behind it. Mark it interrupted before releasing the parent's delegation
 						// link so the normal resume/re-delegate flow can take over. This is an
@@ -1091,5 +1111,20 @@ export class TaskHistoryStore {
 	private async getTaskFilePath(taskId: string): Promise<string> {
 		const tasksDir = await this.getTasksDir()
 		return path.join(tasksDir, taskId, GlobalFileNames.historyItem)
+	}
+
+	/**
+	 * Returns the mtime (ms epoch) of the child's history_item.json, or undefined
+	 * when unreadable. A recent mtime means another live extension host is actively
+	 * persisting this child, so startup repair must not treat it as a crash orphan.
+	 */
+	private async getChildFileMtimeMs(childId: string): Promise<number | undefined> {
+		try {
+			const filePath = await this.getTaskFilePath(childId)
+			const stat = await fs.stat(filePath)
+			return stat.mtimeMs
+		} catch {
+			return undefined // File missing/unreadable → conservatively proceed with repair
+		}
 	}
 }
--- a/src/core/task-persistence/TaskHistoryStore.ts
+++ b/src/core/task-persistence/TaskHistoryStore.ts
@@ -534,7 +534,10 @@ export class TaskHistoryStore {
 	 * Replay the durable active-child repair intent, if one was left by a crash.
 	 * The expected fields are guards: an intent may update only the missing side
 	 * when the other side is already at its target, or when both records still
-	 * describe the original delegated handoff.
+	 * describe the original delegated handoff. Before writing the child, the same
+	 * cross-window liveness guard as `reconcileDelegationStateCore` applies: a
+	 * child whose history file was touched recently belongs to another live
+	 * window, so the stale intent is quarantined instead of replayed.
 	 *
 	 * This method acquires the store's non-reentrant promise-chain lock. It must be
 	 * called outside an existing `withLock` callback; locked callers must use the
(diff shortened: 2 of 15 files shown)


src/core/task-persistence/TaskHistoryStore.ts
```diff
@@ -466,6 +473,20 @@ export class TaskHistoryStore {
...
+						const isLiveElsewhere =
+							// Stryker disable next-line ConditionalExpression: replacing `mtimeMs !== undefined` with `true` is mutation-equivalent; with a defined mtimeMs `true && X === X`, and with undefined the right operand is `NaN < threshold === false`, identical to the short-circuit result.
+							mtimeMs !== undefined &&
+							Date.now() - mtimeMs < TaskHistoryStore.LIVE_CHILD_MTIME_THRESHOLD_MS
+						if (isLiveElsewhere) {
+							console.log(
+								`[TaskHistoryStore] Skipping repair for live child ${child.id} ` +
+									`(mtime ${Math.round((Date.now() - mtimeMs) / 1000)}s ago) — owned by another window`,
+							)
+							continue
```
edelauna, line 512: Does `replayDelegationRepairIntent` also get this guard? If this window crashed mid-repair and Window B then started the same child task, the replay path checks `matchesDelegationRepairChildPreconditions` (status / `parentTaskId` / `rootTaskId`) but not the mtime — so a Window-B-live child could pass those checks and end up written as `"interrupted"` on restart.

src/core/task-persistence/TaskHistoryStore.ts
```diff
@@ -97,6 +97,13 @@ export class TaskHistoryStore {
 	/** Periodic reconciliation interval in milliseconds. */
 	private static readonly RECONCILE_INTERVAL_MS = 5 * 60 * 1000
 
+	/**
+	 * Maximum age (in ms) of a child's history file mtime for the child to be
+	 * considered live in another window. Kept at least as long as the reconcile
+	 * interval so live tasks with sparse writes are not misjudged as orphans.
+	 */
+	private static readonly LIVE_CHILD_MTIME_THRESHOLD_MS = 5 * 60 * 1000 // 5 minutes
```
edelauna, line 120: Does `startPeriodicReconciliation` also call `reconcileDelegationState` on each tick? From what I can see it only calls `reconcile()` — so if a child passes this threshold at startup but then crashes, would it stay orphaned until the next VS Code restart rather than being caught within the next 5-minute interval?

src/core/task-persistence/TaskHistoryStore.ts
```diff
@@ -466,6 +473,20 @@ export class TaskHistoryStore {
...
 					} else if ((child.status ?? "active") === "active" && persistedActiveIds.has(child.id)) {
+						// Cross-instance liveness guard: a child whose history file was written
+						// recently is owned by another live window, not a crash orphan.
+						const mtimeMs = await this.getChildFileMtimeMs(child.id)
+						const isLiveElsewhere =
+							// Stryker disable next-line ConditionalExpression: replacing `mtimeMs !== undefined` with `true` is mutation-equivalent; with a defined mtimeMs `true && X === X`, and with undefined the right operand is `NaN < threshold === false`, identical to the short-circuit result.
+							mtimeMs !== undefined &&
+							Date.now() - mtimeMs < TaskHistoryStore.LIVE_CHILD_MTIME_THRESHOLD_MS
+						if (isLiveElsewhere) {
+							console.log(
```
edelauna: All the surrounding reconciliation logs in this block use `console.warn` (lines 452, 471, 495, 512). Should this be `warn` too, so it shows up at the same log level?

```suggestion
						console.warn(
```

src/core/task-persistence/__tests__/TaskHistoryStore.reconciliation.spec.ts
```diff
@@ -161,12 +171,90 @@ describe("TaskHistoryStore reconcileDelegationState", () => {
...
+			getChildFileMtimeMs: (childId: string) => Promise<number | undefined>
+		}
+
+		expect(await internals.getChildFileMtimeMs("missing-mtime-child")).toBeUndefined()
+
+		const child = makeItem({ id: "present-mtime-child", status: "active" })
+		await seedItems([child])
+		const mtimeMs = await internals.getChildFileMtimeMs("present-mtime-child")
+		expect(typeof mtimeMs).toBe("number")
+		expect(mtimeMs).toBeGreaterThan(0)
```
edelauna, line 325: This covers `getChildFileMtimeMs` returning `undefined` in isolation. Is there also a test that routes the `undefined` return through `reconcileDelegationStateCore` end-to-end — e.g. seed an active child, spy `getChildFileMtimeMs` to return `undefined`, call `store.initialize()`, and assert that repair ran?

.gitignore
edelauna, line 1: I dont think we need this

edelauna (changes requested):
Thanks for addressing this, could you also add this to the existing `lifecycle:model-check` script so that the fix (and bug) could be formally verified

.gitignore
myk1yt: Removed!

src/core/task-persistence/TaskHistoryStore.ts
myk1yt: Yes — the replay path now checks the child's history-file mtime before applying the repair. A Window-B-live child (recent mtime) quarantines the repair intent instead of being written as interrupted; an unreadable mtime conservatively proceeds, matching the reconcile-path convention. Covered by an E2E test reproducing exactly your scenario: crash mid-repair → Window B starts the child → restart.

src/core/task-persistence/__tests__/TaskHistoryStore.reconciliation.spec.ts
myk1yt: Yes — there's now a test doing exactly that: seeds a delegated parent + active child, spies getChildFileMtimeMs to resolve undefined, calls store.initialize(), and asserts the child is repaired to interrupted, the parent released to active, and the delegation links cleared (verified in cache and on disk).

src/core/task-persistence/TaskHistoryStore.ts
myk1yt: It did only call reconcile() — good catch. Ticks now also run delegation reconciliation, so a child that skipped startup repair but crashes afterwards is caught within the next 5-minute interval instead of the next restart. One nuance: the tick excludes children this window itself runs, since a live local child can legitimately go quiet for over 5 minutes (long model turn, user deliberating an ask) without being dead. Cross-window children remain protected by the mtime guard.

myk1yt:
Status update on the outstanding review items:

**edelauna's request (2026-09-03): add this fix to `lifecycle:model-check` for formal verification** — this landed in the 260904 session and is included in this branch: `scripts/check-task-lifecycle.ts` now models the `liveElsewhere` boolean with the `markLiveElsewhere`/`expireLiveElsewhere`/`reconcileStartup` actions, the preserve-vs-repair transition invariant, and the corresponding semantic landmarks (verified: 91 states / 7 actions / 4 landmarks, plus the shared-store submodel at 625 states — `pnpm lifecycle:model-check` passes on the rebased branch).

**CodeRabbit (2026-09-12) items** — both addressed in the new commits:
- `docs/architecture/task-lifecycle-model.md` invariant 8 now describes only the boolean model behavior; mtime/stat-failure/crash-orphan classification was removed from the model-level statement (it lives in the production-mapping table and TaskHistoryStore tests).
- `ClineProvider` continuation scheduling now releases local ownership when `TaskScheduler.schedule()` rejects or resolves without admitting the task, and when the admitted continuation returns no `runPromise` — with 5 new regression tests (rejection, cancellation, no-runPromise, plus the two keep-claim paths; revert-check verified).

The branch is rebased onto current main (23 commits, conflict-free except one faithfully-resolved Task.ts hunk) and all checks are green. Re-requesting review — thanks for the patience.

myk1yt:
Round review disposition (all open threads now resolved):

- **locallyActiveTaskIds eviction leak** (review 657db15e) → fixed in 6d057814f: reconcile() eviction now drops the id from the ownership set, with the requested peer-file-removal regression test (fails without the fix).
- **OpenRouter threads (4)** → not applicable to this PR's current 14-file diff (no webview/OpenRouter changes remain here); dispositioned and tracked on PR #1369.
- **Invariant 8 narrowing** → addressed by 4fc47fc40.
- **Linked-issue / out-of-scope / description findings** → PR body rewritten: linked to #1469 (Closes) and #1624 (Related), every changed file now grouped by theme with justification, checklist completed.

myk1yt:
@coderabbitai review — head is now 6d057814f. Since the last review: (1) the locallyActiveTaskIds eviction leak is fixed with the requested peer-file-removal regression test; (2) the PR body was rewritten — it now links Closes: #1469, groups all 14 changed files by theme with justification (the Task.ts/ClineProvider/lifecycle changes are review-round hardening of the delegation semantics, not unrelated), and completes the checklist. Please re-run the pre-merge checks against the new head and description.

PR #1295, opened 2026-08-19
feat(api): abort signal support for opencode-go, unbound, vercel-ai-gateway, zoo-gateway

easonLiangWorldedtech:
Wires external abort signals and per-request timeouts into the non-streaming `completePrompt` paths and the `createMessage` streaming paths of the Opencode Go, Unbound, Vercel AI Gateway, and Zoo Gateway providers.

- `opencode-go.ts`: forwards `options?.abortSignal` / `options?.timeoutMs` to both the Anthropic (`/v1/messages`) and OpenAI (`chat.completions`) `completePrompt` paths; bridges `metadata?.abortSignal` (Bedrock pattern: pre-aborted guard + `{ once: true }`) into a per-request `AbortController` shared by both streaming wire formats.
- `unbound.ts`: forwards `completePrompt` options to the OpenAI SDK; bridges `metadata?.abortSignal` into a per-request controller for `createMessage`.
- `vercel-ai-gateway.ts`: forwards `completePrompt` options to the OpenAI SDK; bridges `metadata?.abortSignal` into a per-request controller for `createMessage`.
- `zoo-gateway.ts`: forwards `completePrompt` options to the OpenAI SDK; bridges `metadata?.abortSignal` into the existing per-request options (headers + signal) for `createMessage`.

Tests:
- Ported the reference abort/timeout `completePrompt` pass-through tests for all four providers (signal, timeoutMs (incl. 0), and no-options backward compatibility), plus second-argument expectations on existing SDK-mock assertions.
- Added new `createMessage` bridging tests per provider: pre-aborted signal -> request rejects with an error whose `name === "AbortError"` (unbound asserts the SDK-level rejection since its error wrapper preserves main's behavior); abort mid-flight -> in-flight request/stream aborts and the bridged signal is observed aborted.

Part of the abort-signal series (round 1). Builds on #674, #901, #1008. Addresses #404.

PR #1545, opened 2026-09-06
feat(api): abort signal support for openrouter (completePrompt + createMessage)

easonLiangWorldedtech:
Adds abort-signal support to the OpenRouter provider for both `completePrompt` and `createMessage` (round 1 of the abort-signal series).

**Supersedes #1301 (split A of 4).** #1301's combined gateway-a diff (openrouter + requesty + poe, 10 files) generated 518 mutation-diff preflight mutants against the 400 cap, so the series splits by provider. This PR carries the OpenRouter portion only: **+2483/−419** across 5 files. Mutation-diff preflight (base `main` @ `8d296deef`): 283 changed executable lines (< 500) and 319 valid mutants (< 400): 318 killed, 1 timeout (0.3%), plus 30 mutants on 11 `Stryker disable next-line` directive lines reported as ignored, each with a documented unobservability rationale in the directive comment. The a+d total (2902) exceeds the generic 400-line soft design budget and the largest comparable series precedent (#1295 at 2554 a+d) because `createMessage` bridging and its 86-test suite are inseparable for this provider; the a+d figure is design guidance, not a CI hard gate — the two CI hard gates (500 changed executable lines, 400 valid mutants) are both satisfied. Sibling series precedents at comparable sizes: B1 #1537 (1915 a+d), #1311 (1579 a+d), #1295 (2554 a+d) — all CI-green.

### completePrompt

- Accepts `CompletePromptOptions` (`abortSignal` and/or `timeoutMs`) and forwards them to the OpenAI SDK client: `RequestOptions.signal` / `RequestOptions.timeout` are included only when actually set; `timeoutMs <= 0` never passes `0` to the SDK (the SDK treats `0` as an immediate abort). The client-level timeout remains the default safety net.
- If the caller's signal aborts (or the per-request timeout fires) while the request is in flight, the provider rejects with a DOM-standard `AbortError` (`error.name === "AbortError"`).
- Pre-aborted guard: a call whose signal is already aborted rejects immediately with the canonical abort error (the same `AbortError` message every other abort path produces).
- If the request resolves *after* the abort, the late result is discarded and `AbortError` is thrown instead.
- Model discovery (`fetchModel`) now settles promptly on cancellation via the new shared `rejectOnAbort()` helper (see below).

### createMessage (new bridging)

Bridges the caller's `metadata.abortSignal` into a per-request `AbortController` (Bedrock pattern):

- The request-local controller is captured by closure (not a mutable field), so concurrent requests do not interfere.
- Pre-aborted guard: if the signal is already aborted, the stream rejects with `AbortError` immediately without calling the API.
- The external listener is stored in a named const and removed in `finally`, so listeners never outlive the request.
- The SDK request is driven by the controller's signal, and abort-driven stream failures are normalized to `AbortError`.
- Buffered-chunk guard: the stream loop re-checks `controller.signal.aborted` before processing each chunk (openai@5.23.2 can swallow a mid-stream AbortError and keep delivering buffered chunks), and the post-loop check rejects with `AbortError` instead of completing silently after partial output.
- Abandoned-generator cleanup: the `finally` block aborts the per-request controller, so a consumer that stops iterating early (early `break`, downstream error) cancels the in-flight stream instead of leaving it open until the client-level timeout. On the completed and already-aborted paths the abort is a no-op.

### Shared helper

- `src/api/providers/utils/abort-signal.ts` — new `rejectOnAbort(pending, signal, providerName)`: awaits `pending` but rejects with the provider's abort error when `signal` aborts first. For async phases with no native signal support (model discovery) that must still settle promptly on cancellation; the abort listener is detached once `pending` settles (success or failure).
- `abort-signal.spec.ts` — tests for `rejectOnAbort`: same-reference listener identity (the exact registered reference is the one removed on success and on failure), `{ once: true }` registration (asserted by identity against the captured listener), and a settle guard (the shared `settlesWithin` helper below) that converts structural hangs into fast failures so mutation testing reports Kills instead of Timeouts.
- `src/test-utils/promise.ts` — new shared `settlesWithin(promise, ms)` settle guard used by both `openrouter.spec.ts` and `abort-signal.spec.ts` (replaces their duplicated local copies).

### Tests

- `completePrompt`: signal/timeout pass-through, `timeoutMs <= 0` handling, backward compatibility without options, pre-aborted reject, mid-flight abort reject, late-result discard, and mapping of an abort-named model-discovery failure to the canonical `AbortError` without calling the API.
- `createMessage` bridging: pre-aborted signal rejects with the provider `AbortError` message without calling the API or starting a second model-discovery lookup; deferred-discovery and mid-flight aborts reject the stream with `name === "AbortError"` (awaited through a file-local settle guard so a broken abort path fails fast instead of timing out); the external listener is registered with `{ once: true }` and the exact registered reference is removed after settlement; the `reasoning_details` accumulator resets between requests on the same handler; Gemini 2.5 Pro models (preview and non-preview) get `reasoning: { exclude: true }` by default, non-Gemini models without configured reasoning get no `reasoning` param, and user-configured reasoning is preserved for Gemini 2.5 Pro; buffered chunks delivered by the iterator after a swallowed mid-stream AbortError are not emitted ("does not emit buffered chunks after a mid-stream abort (iterator keeps delivering)"); abandoning the generator mid-stream aborts the per-request controller so the in-flight stream is cancelled ("cancels the in-flight stream when the consumer abandons the generator"); the pre-aborted `completePrompt` guard rejects with the canonical `AbortError` message.
- Mutation coverage: the new and extended assertions kill every surviving changed-code mutant (0 survived, 0 no-coverage); the `openrouter.spec.ts` suite grows 26 → 86 tests (60 new tests across `completePrompt` and `createMessage`), all passing.
- 11 `// Stryker disable next-line` directive lines (30 mutants) cover lines whose variants are provably unobservable: the `sanitizeGeminiMessages`/message-conversion invariants (assistant-only `tool_calls`, non-empty `tool_calls`, a defined `reasoning_details` array, and the `?? []` fallback that never runs), no-op optional chains behind the `parseResult.data.error` definedness guard, a rethrow whose message is masked by the canonical `createAbortError`, the empty-accumulator consolidation that `getReasoningDetails()` maps to `undefined`, and the positive-`maxTokens` model-info guarantee. Each directive states its rationale in-line.

Part of the abort-signal series (round 1). Builds on #674, #901, #1008. Addresses #404. Supersedes #1301 (split A of 4; siblings: #1537 + #1538 (requesty, stacked), #1535 (poe)).

### DIFF
--- a/src/api/providers/__tests__/openrouter.spec.ts
+++ b/src/api/providers/__tests__/openrouter.spec.ts
@@ -17,11 +17,11 @@ const MOCK_TIMEOUT_MS = 300_000
 import { Anthropic } from "@anthropic-ai/sdk"
 import OpenAI from "openai"
 
-import { providerIdentifiers } from "@roo-code/types"
+import { providerIdentifiers, type ModelRecord } from "@roo-code/types"
 
 import { OpenRouterHandler } from "../openrouter"
 import { Package } from "../../../shared/package"
-import { makeApiHandlerOptions } from "../../../test-utils/api"
+import { makeApiHandlerOptions, makeCreateMessageMetadata } from "../../../test-utils/api"
 import { asyncStreamFrom, collectStream } from "../../../test-utils/stream"
 import { collectStreamAndParseToolCalls } from "../../../test-utils/native-tool-call-stream"
 import { clearAllMocks } from "../../../test-utils/reset"
@@ -108,6 +108,29 @@ vitest.mock("../fetchers/modelCache", () => ({
 		return getModels(options)
 	}),
 }))
+
+/**
+ * Fail fast when an awaited operation never settles. Mutations that break abort
+ * propagation would otherwise hang the test until Stryker's per-mutant timeout,
+ * marking the mutant "Timeout" instead of "Killed".
+ */
+function settlesWithin<T>(promise: Promise<T>, ms: number): Promise<T> {
+	return new Promise<T>((resolve, reject) => {
+		const timer = setTimeout(() => reject(new Error("operation did not settle within " + ms + "ms")), ms)
+		void promise.then(
+			(value) => {
+				clearTimeout(timer)
+				resolve(value)
+			},
+			(error) => {
+				clearTimeout(timer)
+				reject(error)
+			},
+		)
+	})
+}
+
+const ABORT_SETTLE_MS = 150
 
 describe("OpenRouterHandler", () => {
 	const mockOptions = makeApiHandlerOptions({
@@ -294,7 +317,10 @@ describe("OpenRouterHandler", () => {
 					temperature: 0,
 					top_p: undefined,
 				}),
-				{ headers: { "x-anthropic-beta": "fine-grained-tool-streaming-2025-05-14" } },
+				{
+					headers: { "x-anthropic-beta": "fine-grained-tool-streaming-2025-05-14" },
+					signal: expect.any(AbortSignal),
+				},
 			)
 		})
 
@@ -337,7 +363,10 @@ describe("OpenRouterHandler", () => {
 						}),
 					]),
 				}),
-				{ headers: { "x-anthropic-beta": "fine-grained-tool-streaming-2025-05-14" } },
+				{
+					headers: { "x-anthropic-beta": "fine-grained-tool-streaming-2025-05-14" },
+					signal: expect.any(AbortSignal),
+				},
 			)
 		})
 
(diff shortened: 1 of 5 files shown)


easonLiangWorldedtech:
**Review-gate status is stale (mergeability already settled) — one maintainer action would clear it; no push needed**

The `Zoo Code / PR review gate` commit status on head `9bd5d6f81` reads:

> Wait for GitHub to finish calculating mergeability.

That status was set at 2026-09-07 02:36:29Z by the reconcile triggered by the #1425 main push, while GitHub was still recomputing mergeability. It has since settled — the PR API now reports `mergeable: true` (mergeable_state `blocked` = branch-protection review requirements only). Everything this PR needs is already in place:

- required CI green on head `9bd5d6f81`
- CodeRabbit **APPROVED** `9bd5d6f81` (2026-09-06 07:44:43Z, no actionable findings)
- all 5 CODEOWNER review requests in place

The stale status clears on the next reconcile of this PR. The hourly `schedule` sweep is the fallback reconcile, but it is running late (the 04:00Z and 05:00Z slots had not fired as of 05:18Z — zero `schedule` runs in the 03:18–05:18Z window; GitHub delays schedule events under platform load). Fork-side events cannot force a reconcile here (verified: no label write permission, `workflow_dispatch` 403 with the pull-only token, reviewer management 404, API draft toggle is a silent no-op), so the heal currently depends on an external event.

Any one of these would re-reconcile the PR immediately:

1. add or remove any label on this PR (fires a trusted `labeled`/`unlabeled` event), or
2. run the "Label PR review state" workflow dispatch with `pull_request_number: 1545`, or
3. push to `main` (auto-reconciles).

**Please do not push a new commit to this branch** — head `9bd5d6f81` is intentionally unchanged so the CodeRabbit APPROVED on it remains valid.

Full series state: tracking issue easonLiangWorldedtech/Zoo-Code#42 (snapshot 19, item 14).

@taltas @JamesRobert20 @navedmerchant @hannesrudolph @edelauna

PR #1608, opened 2026-09-11
fix(vscode-lm): integrate guarded recovery into streaming responses

simurg79:
## Summary

Activates the leaked tool-call parser introduced in #1188 inside `createMessage`. This is the second half of a two-part split of #1188; it contains all of the streaming integration and its tests.

**Draft, and dependent on #1188.** It must not be merged before #1188.

## Depends on

- https://github.com/Zoo-Code-Org/Zoo-Code/pull/1188 (part A — the parser, guards, schema conversion and their direct tests; inactive in production)

## What is in this PR (part B)

- Streaming salvage state inside `createMessage`.
- Start-marker detection with partial-marker carry across chunk boundaries.
- Buffering until the invoke block completes.
- An overflow fallback that releases unclosed markup as plain text when limits are exceeded.
- An ordered flush that emits prose before any recovered call, and runs before native tool calls.
- The streaming integration tests.

Diff versus part A: 2 files changed, 361 insertions, 3 deletions. Combined with part A versus `main`: 4 files changed, 1357 insertions, 6 deletions.

The combined result is byte-for-byte identical to the previously reviewed head of #1188 (`34e16a49d01a16525d09f0d8250aa696143207e6`): the tree of this branch equals that commit's tree exactly. No tests were dropped, no guard was weakened, and nothing was refactored during the split.

## Why the split

The changed-code mutation gate caps a run at 400 selected mutants. Measured with instrumentation-only Stryker 10.0.0 runs:

| Revision pair | Selected candidates | Cap |
| --- | --- | --- |
| Part A vs `main` | 320 | 400 |
| This PR vs part A (incremental) | 110 | 400 |
| Combined vs `main` | 430 | 400 |

**Important:** until #1188 is merged, CI on this branch measures the combined **430** against `main`, not the incremental **110**. So this branch's own cap compliance cannot be demonstrated by CI until part A is in the base. Please do not read a red combined run here as evidence the incremental change is over cap — and equally, the incremental figure is not a passing CI result.

## Tests

- 114/114 passing (provider suite, at this branch's exact source tree).
- Lint and type-check pass; no increase in ESLint suppression counts.

## Mutation-testing status — known failing, disclosed

This PR does **not** pass the changed-code mutation gate. Locally measured, incremental against part A, over the selected changed-code range:

- 79 killed, **30 survived, 1 uncovered** → 31 blocking, gate result **FAIL**.

For reference, part A measures 223 killed, 1 timeout, 93 survived, 3 uncovered → 96 blocking, also **FAIL**.

These are observed failures as run here. I am **not** claiming the surviving mutants are inherited or pre-existing, and no threshold was weakened or waived. Remediating them is out of scope for this structural split.

Caveat: a Windows extensionless-Vitest shim `ENOENT` prevented an end-to-end run of the gate script locally, so a pinned JS invocation and harness were used, with source hashes verified against the pushed trees. **CI remains authoritative.**

## Merge order

1. #1188 (part A)
2. This PR

After part A merges, this PR's base and CI should be refreshed and the diff re-inspected. No rebase is being asserted as necessary in advance; if history requires it later, that would be handled separately and explicitly.


simurg79:
### CI-fix investigation status: blocked, no fixes published

Issue #648, opened 2026-06-17
[BUG] Un-controllable Token Exceed with MCP results


sihangchen97:
### Problem (one or two sentences)

The conversation will get stuck when MCP tools return a large amount of context.
- with “Compression API failed: OpenAI completion error: 400 request (359746 tokens) exceeds the available context size (128000 tokens), try increasing it”

### Context (who is affected and when)

N/A

### Reproduction steps

1) Clone a **rather large** git repo
2) Activate "Git Tools" from "Zoo Marketplace - MCP"
3) Ask Zoo to check the diff and status of this git repo
4) Zoo will use the MCP tool for the task
5) The conversation gets stuck with a compression error

### Expected result

Run normally with either trunc or compressed MCP result

### Actual result

“Compression API failed: OpenAI completion error: 400 request (359746 tokens) exceeds the available context size (128000 tokens), try increasing it”

### App Version

3.60.0

### API Provider (optional)

OpenAI Compatible

### Model Used (optional)

Qwen3.6-27B

### Relevant logs or errors (optional)

```shell

```

pajitosingh:
I can confirm.  This happens with html_get use with a browser mcp (playwright) as well.  A large html dump will take the context over 1M sometimes and then the session is borked.

Awilen-Bernkastel:
Adding to this: my own workflow first compacts fine, then the subsequent compaction runs try to reuse the full context from before compaction. This leads to a silly situation in which a session tries to compact a million-tokens worth of conversation from a 128k window: `Condensing API call failed: OpenAI completion error: 400 request (1009019 tokens) exceeds the available context size (128000 tokens), try increasing it` (Yes, somehow this is still working out fine for my purpose, despite the context window crops.)

This didn't come from a massive MCP call result filling the context window, just normal usage. This may be another bug I could describe as "the compaction algorithm keeps reusing the full conversation instead of previous compaction results."

PR #1617, opened 2026-09-12
fix(context): make zero-progress truncation recovery degrade tool results instead of reporting success

### LINKED ISSUE #1254: [BUG] Ensure context truncation always makes measurable progress
## Problem

Fallback context truncation can report success without reducing context. With a short history such as an assistant `tool_use` followed by one oversized user `tool_result`, the current 50% message calculation rounds down to zero removable messages. Zoo Code still emits a truncation event, leaves the oversized history unchanged, and retries into the same failure indefinitely.

The reproduced signal is `Context Truncated` with `0 messages removed` while the recalculated context remains above the allowed input budget.

## Scope

Make context recovery monotonic and bounded.

- Treat recovery as successful only when the recalculated model-facing token count decreases.
- When message-level truncation cannot remove a valid turn/tool pair, shrink the largest eligible textual tool-result block while preserving its tool identifier and result shape.
- Recount after degradation and verify the result fits the target or makes measurable progress.
- If protected content makes the target impossible, return a distinct actionable error instead of emitting another successful truncation event or retrying indefinitely.
- Keep retries bounded.

## Acceptance criteria

- [ ] One-to-four-message histories cannot emit a successful zero-progress truncation result.
- [ ] An oversized `tool_use`/`tool_result` pair is reduced without orphaning either side.
- [ ] `newContextTokensAfterTruncation` is lower than the pre-recovery count whenever recovery reports success.
- [ ] An impossible budget produces a controlled error and does not loop.
- [ ] Focused tests reproduce the 4 MB tool-result case and verify bounded recovery.

## Out of scope

- Preventing individual tools from producing oversized results
- Condensation-input sanitization
- Full-output spill files or resource references
- Aggregate budgeting across parallel tool calls

## Estimate

1 story point. Independently mergeable and non-stacked.

Parent tracking issue: #648

yetuge:
### Related GitHub Issue

Closes: #1254 (commented "Claiming" per the contribution guide — happy to be reassigned if that flow needs it)

### Description

**Root cause** — the fallback branch in `manageContext` (`src/core/context-management/index.ts`) calls `truncateConversation(messages, 0.5)`. For short histories — e.g. an assistant `tool_use` followed by one oversized user `tool_result` — `Math.floor((visibleCount - 1) * 0.5)` rounds down to 1 and the even-rounding step takes it to **0 removable messages**. `truncateConversation` then returns the original messages with a fresh `truncationId` and `messagesRemoved: 0`, and `manageContext` passed that through as a **success-shaped** result. `Task.ts` emitted the `sliding_window_truncation` event with `messagesRemoved: 0`, the oversized history was never overwritten, and every retry re-entered `manageContext` over budget — the zero-progress loop described in the issue (parent: #648).

**Fix** — make recovery monotonic and bounded, per the issue's acceptance criteria:

1. `manageContext` now treats `messagesRemoved === 0` as zero progress. Instead of reporting a successful truncation, it degrades in place: `shrinkOversizedToolResults` shrinks the **largest eligible textual `tool_result` blocks first** (largest-first so each round frees the most tokens while losing the least information), preserving each block's `tool_use_id` and shape so the `tool_use`/`tool_result` pair is never orphaned. The keep-size targets the over-budget amount using a chars-per-token ratio measured on the block itself, with a 200-char floor below which a block stops being eligible.
2. After degradation the context is recounted, and success is reported **only when the recalculated model-facing token count actually decreased** (message-level truncation success is now verified the same way). This guarantees `newContextTokensAfterTruncation < prevContextTokens` on every reported success.
3. When protected content leaves nothing to remove or shrink, `manageContext` returns a controlled `error`/`errorDetails` result (an actionable message naming the budget numbers) instead of emitting another fake truncation event — the existing `condense_context_error` path in `Task.ts` surfaces it, and no `truncationId` is set, so no misleading truncation UI event is produced.

**Trade-offs / notes for reviewers:**

- The in-place degradation returns a **new** message array via a functional edit applier, so the caller's `truncateResult.messages !== this.apiConversationHistory` reference check keeps working and the degraded history is persisted.
- One visible behavior choice: a degraded-but-successful round reuses the recovery round's `truncationId` and reports `messagesRemoved: 0` with the lowered token counts, so the existing `sliding_window_truncation` UI channel reflects real progress (tokens reduced) rather than staying silent.
- Repeated recovery rounds stay bounded: each round either fits the target, makes measurable progress, or ends in the controlled error — the shrinking floor ensures strictly decreasing block sizes across rounds.
- Out of scope, observed while testing: `src/utils/tiktoken.ts` encodes a whole block in a single WASM call; a pathological multi-MB single-token-run block can make `encoder.encode` throw (`RuntimeError: unreachable`). The fix here does not depend on that path (the 4 MB test below uses realistic mixed content), but it may deserve its own issue.

### Test Procedure

Focused vitest coverage added to `src/core/context-management/__tests__/context-management.spec.ts` (`manageContext fallback recovery for zero-progress truncation`):

- **4 MB tool-result case**: a 3-message history (initial task + assistant `tool_use` + user `tool_result` with a ~4 MB mixed-text payload) over a 100k window with a 30k reserve. On the pre-fix code the new assertions fail (see below); after the fix, the tool_result is shrunk, the pair stays intact (`tool_use_id` preserved, block shape preserved), `messagesRemoved` stays 0, and `newContextTokensAfterTruncation < prevContextTokens`.
- **Impossible budget**: a 3-message plain-text history over budget produces a controlled `error`/`errorDetails` with **no** `truncationId` and an unchanged history.

Red→green evidence (Windows, Node v24, `pnpm exec vitest run core/context-management`):

1. RED — tests added first, run on unmodified `main` (`745656a`): the recovery test fails on `expect(result.messages).not.toBe(messages)` (same reference returned) and `newContextTokensAfterTruncation < prevContextTokens`; the impossible-budget test fails on `expect(result.error).toBeDefined()` — i.e. the current code reports a successful zero-progress truncation. `2 failed | 53 skipped`.
2. GREEN — after the fix: `2 passed`; the full module + neighbors run `73 + 246` passed (context-management, condense, message-manager, checkpoints), and `core/task` 537 passed.
3. Zero regression — full `src` suite (8414 tests) run on the fix branch vs. unmodified `main` on the same machine, compared by test name: **150 pre-existing environment failures on both sides (dist-assets requiring a build, tree-sitter native modules on Windows), 0 new, 0 disappeared**.
4. `tsc --noEmit`, `eslint --max-warnings=0`, and `prettier --check` clean; changeset included.

### Pre-Submission Checklist

- [ ] **Visual Snapshot** (UI changes only): If a user would notice this change at a glance (layout, theme tokens, brand elements, empty/error states), I've added or updated a `*.visual.tsx` snapshot in `webview-ui/`. See `webview-ui/AGENTS.md` → "When a UI change needs a snapshot".

### Visual Snapshots

Not a UI change (no rendered state touched; the existing `sliding_window_truncation` event payload semantics are unchanged).

Not applicable.

### Documentation Updates

### Additional Notes

- AI assistance was used to prepare this change; the implementation, tests, and trade-offs above were reviewed and verified locally as the contribution guide requires.
- The tiktoken WASM observation above is pre-existing and independent of this fix; happy to file it separately if preferred.

### Get in Touch

GitHub handle only for now: @yetuge


yetuge:
Thanks for the review — 5 of the 6 findings are addressed in `cdc509c`; one is deliberately left out and explained below.

**1. `changeset` — removed.** You're right and my mistake: `AGENTS.md` is explicit ("do NOT create `.changeset` files for each commit or code change. Changesets are managed separately by maintainers"). `.changeset/fix-context-truncation-zero-progress.md` is deleted in this commit; nothing else referenced it. (That also clears the markdownlint MD041 warning on it.)

**2. `tests` — array-form `tool_result.content` coverage.** Added a case where `content` is `[text, image, oversized text, text]`. It asserts the item order and all ineligible items (`toEqual` against the originals, including the image block), that only the oversized text item is degraded, and that the caller's history is not mutated. This one is coverage-only — no production change was needed — so I verified it from the other side: mutating `applyToolResultEdits` to write `blockContent[0]` instead of `blockContent[edit.textIndex]` makes the new test fail, reverting makes it pass.

**3. `tests` — controlled error contract.** Adopted your assertions verbatim: `result.error` must contain `"Context window recovery failed"` and `result.errorDetails` must contain `"removed 0 messages and no eligible textual tool_result"`. Same verification: rewording either string makes the test fail, so the assertions are load-bearing rather than passing on a stale/unrelated error.

**4. `correctness` — condensed messages were shrink candidates (`index.ts:193`).** Confirmed, and this was a real bug of the same family as #1254. `findShrinkableToolResults` now builds `new Set(getEffectiveApiHistory(messages))` and skips any message the API never receives, instead of only skipping `truncationParent` / `isTruncationMarker`. Returned indexes still point into the persisted history, so the edits and the caller's persistence path are unchanged.

New test: a fresh-start condense history (every message tagged `condenseParent`, summary last) whose oversized `tool_result` is hidden by that summary. Before the fix it returned a **fake success** — the estimate dropped while the effective request stayed identical; now it returns the controlled error. Red confirmed by stashing the source change only: `expect(result.error).toContain(...)` fails on the pre-fix code.

**5. `correctness` — the notice could grow the block (`index.ts:309`).** Confirmed and fixed. `keepChars` now reserves `buildTruncationNotice(...).length` before slicing, and a candidate whose rewritten body would not be strictly shorter is skipped. Token bookkeeping now uses the actual `text.length - newText.length` instead of the pre-notice figure.

New test: one oversized `tool_result` next to a second one sitting just above the 200-character floor. Before the fix, the short block was rewritten to `200 chars + notice` = 273 characters — **18 characters longer** than the 255 it started with, while the notice claimed "70 characters removed to fit the context budget". It is now left byte-identical (`toBe` on the original string) while the oversized block still degrades and recovery still succeeds. Red confirmed the same way.

**6. `stability` — reject an over-budget degraded result before retrying (`index.ts:592`). Not in this commit, on purpose.** I agree with the finding: `manageContext` currently treats "tokens decreased" as success even when the result is still above `allowedTokens`. I left it out because the fix is not local — it redefines what *success* means for a partially degraded history, and your suggested follow-through (treating it as terminal in both the normal caller and `handleContextWindowExceededError`) changes retry behaviour in `Task.ts` for every provider. There is also a design choice I don't want to make silently: once the floor is reached and we are still over budget, should we return the partially degraded history together with the error (persisting the progress we did make) or the original history untouched?

If you want it in this PR, this is the plan I'd push as a follow-up commit: accept degradation only when `newContextTokensAfterDegradation <= allowedTokens`, otherwise return the degraded messages **plus** a distinct controlled error and no `truncationId` (so the caller persists the smaller history but emits no truncation event), then make both callers treat that error as terminal. Happy to add it — just confirm the history-return question above.

**Verification (local, Windows)**
- `vitest run core/context-management/__tests__/context-management.spec.ts` → **58 passed** (3 new; 55 before).
- `vitest run core/context-management core/condense core/message-manager core/checkpoints core/task` → **786 passed**.
- Full `src` suite against a clean `main` baseline, compared test-by-test: **0 new failures**; the 149 remaining failures are the pre-existing platform ones (dist assets not built, tree-sitter native module) that fail identically on `main`.
- `tsc --noEmit`, `eslint --max-warnings=0` (with `--prune-suppressions`) and prettier on both changed files: clean.

The two added tests should also lift the patch coverage Codecov reported (75.6%).

yetuge:
Thanks for the re-review — confirmed and fixed in `ffe1778`.

**`correctness` — the orphan filter's clone broke identity matching (`index.ts:212`).** Real gap in yesterday's finding-#4 fix: `getEffectiveApiHistory` returns a **clone** of a kept user message whenever it filters an orphan `tool_result` out of it, so that message can never match a persisted message by reference and `findShrinkableToolResults` skipped the whole message — including its other, API-visible oversized `tool_result`. Recovery then reported a controlled error while the over-budget request still carried the unshrunk result.

**Fix — match by block identity, not message identity.** The orphan filter keeps the surviving block objects (it only rebuilds the message and its `content` array), so collecting the effective history's blocks into a reference `Set` maps API-visible content back to the persisted history exactly: the filtered orphan block stays invisible (shrinking it would still be token-estimate-only progress), and the surviving blocks in the same message stay shrinkable. Returned indexes still point into the persisted history, so `applyToolResultEdits` and the caller's persistence path are unchanged. Per your note, I did not track message indexes only — that would have exposed the filtered orphan block to `applyToolResultEdits`.

**Regression test**: one kept message carrying an orphan `tool_result` (`toolu_old`, condensed away) next to an oversized valid `tool_result` (`toolu_big`), after a summary. On the pre-fix code the recovery failed with `Context window recovery failed ... no eligible textual tool_result` even though the API-visible history contains the oversized block; after the fix the oversized block degrades, recovery succeeds, and the orphan block survives byte-identical (`toBe` on the original string). Red confirmed by checking out the pre-fix source while keeping the new test.

**Verification (local, Windows)**
- `vitest run core/context-management/__tests__/context-management.spec.ts` → **59 passed** (1 new; 58 before); the "hidden by a condensation summary" case still returns the controlled error.
- `vitest run core/context-management core/condense core/message-manager core/checkpoints core/task` → **787 passed**.
- Full `src` suite against the previous head (`cdc509c`) baseline, compared test-by-test: **0 new failures** (8409 vs 8408 named tests; the 150 remaining failures are the pre-existing platform ones — dist assets not built, tree-sitter native module — failing identically on both sides; the known flaky `CustomModesManager` YAML case failed on both sides this session).
- `tsc --noEmit`, `eslint --max-warnings=0` and prettier on both changed files: clean; pre-commit hook (`lint-staged` + full turbo lint 11/11) passed.

yetuge:
Confirmed and fixed — thanks for the detailed outside-diff note. Both failure modes were real on `4bdebd6b`; the fix is in 63bc873.

**What was wrong**

1. `countModelFacingTokens` walked the persisted history and skipped messages by `truncationParent`/`isTruncationMarker` metadata only. With a condense summary present, hidden pre-summary content was still charged to the recovery result, and orphan `tool_result` blocks filtered out of `getEffectiveApiHistory` (message clones) were counted too. A degradation that only moved API-visible tokens could therefore fail `newContextTokensAfterDegradation < prevContextTokens` and be rejected as zero progress, ending in the controlled error even though the next request would have been smaller.
2. `findShrinkableToolResults` vetoed whole messages on the same metadata. A message whose `truncationParent` was rewound away (marker deleted) is API-visible again per `getEffectiveApiHistory`'s no-summary path, but its oversized `tool_result` had no shrink candidate — contradicting the `apiVisibleBlocks` membership check on the very next line.

**The fix (63bc873)**

- `countModelFacingTokens` now iterates `getEffectiveApiHistory(msgs)`: the fresh-start summary slice, truncation-tagged messages (only while their marker exists), orphan-block filtering, and the markers themselves are all reflected — the recount now matches what the next request actually carries.
- `findShrinkableToolResults` relies on `apiVisibleBlocks` alone; the unconditional message-level guard is removed.

**Verification**

- Two new regressions, both failing on `4bdebd6b` and passing now: (a) an oversized `tool_result` in a message carrying an orphaned `truncationParent` (previously "no eligible textual tool_result" → controlled error); (b) a persisted orphan `tool_result` block (≈57k tokens) masking a successful degradation (previously rejected at 89k recounted vs 65k budget baseline).
- Module: 61 passed (59 + 2). Affected surface (`context-management`, `condense`, `message-manager`, `checkpoints`, `task`): 789 passed.
- Full suite vs clean `4bdebd6b`: 8422 → 8424 named tests, the same 150 pre-existing Windows platform failures on both sides, 0 new / 0 disappeared by (classname, test) comparison.
- `tsc --noEmit`, `eslint --max-warnings=0` (both files), prettier clean; pre-commit (lint-staged + full turbo lint) passed.

The two mutation-survival advisories on the old guard (lines 219/222) are moot now that the guard is gone.

One adjacent observation, deliberately out of scope here: `truncateConversation` itself is not summary-aware — it counts pre-summary messages as visible when choosing what to tag, so with a summary present it can tag already-hidden content (`messagesRemoved > 0`) while the API-visible history only shrinks by whatever post-summary messages fell in range. The effective-history recount now reports such rounds accurately, but if that tagging behavior is worth changing, it looks like its own (larger) change.

yetuge:
Both rounds addressed in `172b3e66a`; the branch is mergeable again.

**`index.ts:611` — budget fit (both rounds).** Confirmed, and it was a real hole in the guard I added last round. `manageContext` accepted the fallback truncation on `messagesRemoved > 0 && newContextTokensAfterTruncation < prevContextTokens` alone, so removing cheap messages while protected content stayed oversized returned a successful over-budget history. `Task.ts` then persists and sends it, and the next context-window error retries the same recovery — exactly the loop this recovery path exists to break. I did *not* add the `<= allowedTokens` requirement to the degradation path only; that path already had it from the previous commit, which is what made the asymmetry visible.

I kept `< prevContextTokens` rather than switching the comparison to `modelFacingTokensBeforeRecovery`, per the second round's wording. Switching the baseline is not equivalent: it fails 13 pre-existing tests that assert a successful truncation, because those fixtures count `prevContextTokens` from a persisted total that intentionally exceeds the model-facing recount. The budget check is what actually closes the hole; the decrease baseline is a separate question from the undercount discussion below.

**Failure detail now distinguishes three outcomes** instead of collapsing two of them into the zero-progress message: degradation that reduced tokens but stayed over budget, truncation that reduced tokens but stayed over budget, and truncation that never reduced the model-facing context. The middle one previously would have reported "selected N messages but did not reduce the model-facing context" for a run that *did* reduce it.

**`spec.ts:2518` fixture — accepted.** You were right that the case could pass without exercising the guard. With `totalTokens: 0`, `prevContextTokens` is only the summary count while the recount also includes the system prompt, so the pre-fix source already failed the decrease check for an unrelated reason. It now uses a nonzero persisted `totalTokens: 150000` that exceeds the model-facing count, so the decrease check passes on the removal alone and only the budget fit can reject the result — the false-success path itself. The undercounting variant you described is kept as a separate case with `totalTokens: 0`, asserting the marker path still reports no effective reduction.

**Verification (local, Windows)**
- Both new cases fail on the pre-fix source (`expected undefined to be true`) and pass after: file 65/65.
- `tsc --noEmit` clean; `eslint` clean on both files; repo `lint` via the pre-commit hook 11/11.
- Merge of the 26 upstream commits that had landed since `bd6914d9a` resolved by taking `origin/main`'s `verify-coverage-contract.mjs` wholesale: upstream replaced the hardcoded source/line baseline with the structural `parseCoverageSourceLines` + `mergeCoverageSources` checks, so the baseline bump in my commit is obsolete. The file no longer appears in this PR's diff.

yetuge:
Addressed in `f00dc8026`, though not by switching to `modelFacingTokensBeforeRecovery` — that comparison rejects results that fit. The finding that the two sides of the condition measure different things is correct, and taking it seriously is what led to the fix.

**You were right about the mixed accounting.** The condition compared the recount against `prevContextTokens` and against `allowedTokens`. `allowedTokens` is a budget derived from the model's window; `prevContextTokens` is a locally composed figure (persisted total plus the last message) supplied by the caller. Recovery is only entered when `prevContextTokens > allowedTokens`, so landing under the budget already implies the result is below that figure — the comparison could never change the outcome. I removed it rather than re-pointed it. The condition is now `messagesRemoved > 0 && newContextTokensAfterTruncation <= allowedTokens`, which is also what the degradation path already required.

**Why not `modelFacingTokensBeforeRecovery`.** I instrumented the branch to get both values on a real run and they disagree in the direction that matters. On the existing `should truncate if tokens are above max tokens threshold` case:

```
prevContextTokens = 70001   modelFacingTokensBeforeRecovery = 15
newContextTokensAfterTruncation = 30   allowedTokens = 60000
```

`new` is *larger* than `modelFacingTokensBeforeRecovery`, because truncation is non-destructive — it tags messages and appends a marker rather than deleting them, so on a short history the marker costs more than the content that stops being sent. Comparing against that baseline makes `30 < 15` false and returns a terminal recovery error for a request that occupies 30 tokens against a 60,000-token budget. That is the 13 failures the previous revision reported, and they are the check being wrong rather than the fixtures being stale: the fixtures are short on purpose, and short histories are exactly where the marker outweighs what it hides.

I confirmed the direction rather than asserting it: with the comparison set to `modelFacingTokensBeforeRecovery`, the regression I added for the budget-fitting path fails with `recoveryFailed: true`; with the condition above, it passes. A failed recovery means the caller reports an unrecoverable over-budget conversation and stops, so the cost of being wrong here is refusing to continue a session that was safe to send.

**On the scenario in the comment** (`prevContextTokens` 70,000, budget 90,000, recount 100,000 → 80,000): that state does not reach this branch, because recovery is entered on `prevContextTokens > allowedTokens` and 70,000 is under 90,000. The disagreement you describe is real, but it lives in the caller's `totalTokens` being below `allowedTokens` while the effective history is not — a state this function cannot observe from where it is called.

**Regression added** (`accepts truncation that fits the budget even when the recount is not below the persisted count`): a persisted total well above the real content, asserting `recoveryFailed` is unset, no error, a truncation id, 2 messages removed, and `newContextTokensAfterTruncation <= 60000`. It encodes the budget as the contract and fails if the condition is re-pointed at either baseline.

**Verification**
- `core/context-management/__tests__/context-management.spec.ts` — 66 passed (65 before, +1 new).
- `context-management` + `task/__tests__` — 549 passed, 31 files.
- `tsc --noEmit` clean, `eslint` clean on both files, repo lint via pre-commit hook 11/11.
- The temporary instrumentation used to obtain the three values above is not part of the commit.

PR #1634, opened 2026-09-14
fix: clear nativeArgs when tool-call finalize fails (#1221)

### LINKED ISSUE #1221: Truncated tool-call arguments can be silently written to disk (stale partial-parse nativeArgs reused on finalization failure)
### Summary

When a streamed tool call's arguments are cut off mid-value (e.g. the model hits `max_tokens` while still writing a string argument), the value that was only ever meant for live streaming-progress display can end up used as the final, executed value with no error or diagnostic indicating it was incomplete.

### Where

Observed at commit `abaf732bb7a96848a7c6f5dd2ce485b86bff086d`.

- `src/core/assistant-message/NativeToolCallParser.ts`: `processStreamingChunk()` uses the `partial-json` package (`^0.1.7` in `src/package.json`) to extract a best-effort value from the JSON accumulated so far, purely to show progress during streaming - `createPartialToolUse()`'s `write_to_file` case builds `nativeArgs.content` directly from that lenient parse.
- When the stream ends, `finalizeStreamingToolCall()` calls the strict `parseToolCall()` (`JSON.parse` on the full accumulated text). A genuinely truncated payload correctly throws and the function returns `null`.
- `src/core/task/Task.ts` (search for `finalizeStreamingToolCall returned null`): when finalization returns `null`, the code reuses the *existing* tool-use object - the one built during streaming from the `partial-json` output - and just sets `.partial = false`. It does not clear `nativeArgs`.
- `src/core/assistant-message/presentAssistantMessage.ts`: the guard here is `!block.nativeArgs`. Since `nativeArgs` was already populated during streaming, this guard does not trip.
- `src/core/tools/WriteToFileTool.ts`: the only content check is `newContent === undefined`. A truncated-but-present string passes this too.

### Repro / PoC

Given a `write_to_file` call whose `content` argument is cut off mid-value:

```
{"path":"config/database.yml","content":"production:\n  host: db.prod.internal\n  password: correct-horse-battery-sta
```

Running the actual `partial-json` package (v0.1.7, the version pinned in `src/package.json`) on this string:

```
parse(text) -> { path: "config/database.yml",
                  content: "production:\n  host: db.prod.internal\n  password: correct-horse-battery-sta" }
```

No error. `content` is present and looks like a normal, complete string - exactly what `nativeArgs.content` would still hold when finalization fails and the stale streaming-phase object is reused, so `WriteToFileTool` would write this content to disk exactly as shown, silently.

For contrast, running the same input through [prefix-safe-json](https://github.com/canblmz1/prefix-safe-json) (an incremental JSON parser I maintain, built specifically around never exposing a value until it's unambiguously complete):

```
outcome: "truncated", executable: false
stableValue: { path: "config/database.yml" }   // content is absent, not guessed
```

### Impact

Any tool whose schema includes a free-form string argument (file content, shell command, etc.) can have that argument silently truncated-but-executed when the model's response is cut off mid-argument - a data-integrity issue for any tool with a real side effect.

### Context

This codebase shares this exact code path (including comments) with `RooCodeInc/Roo-Code`, which this appears to be derived from - I reported the same issue there as well, in case it's useful for cross-referencing a fix.

### Suggested direction (not prescriptive)

When `finalizeStreamingToolCall()` returns `null`, clear/discard the reused tool-use object's `nativeArgs` (and ideally `params`) rather than only marking it non-partial - falling back to the existing `!block.nativeArgs` guard at that point instead of leaving stale partial-parse data in place.

I found this by cloning the repository and reading the real source (paths/lines above), then verified the specific claim about `partial-json`'s output by running the actual package version you depend on - not speculation. Happy to answer questions or share the small repro script if useful.

canblmz1:
## Problem

When a streamed native tool call's arguments are truncated mid-value (e.g. the model hits `max_tokens` while still writing `write_to_file`'s `content` string), `finalizeStreamingToolCall()` returns `null`. `Task.ts` reuses the same tool-use object the streaming phase had been mutating in place - which still carries `nativeArgs` built from the incomplete partial-JSON parse - and only sets `partial = false`.

`presentAssistantMessage.ts` already has a guard meant for exactly this case:

```ts
if (isKnownTool && !block.nativeArgs && !customTool) {
  // structured tool_result error, no execution
}
```

But since `nativeArgs` was never actually cleared, `!block.nativeArgs` is never true, and the guard never fires. Truncated arguments (e.g. a cut-off `content` string) can be presented as a complete, valid call and executed.

## Fix

Clear `nativeArgs` alongside `partial = false` at the finalize-null site in `Task.ts`, so the existing guard does what its own comment already said it did. `params` is left untouched - `NativeToolCallParser` always initializes it to `{}` for native tool calls and never puts real data there, so there's nothing to clear there.

## Test

Adds `truncated-native-tool-args.spec.ts`, mirroring the exact `Task.ts` logic in a small local function, following the same convention already used in `duplicate-tool-use-ids.spec.ts` for this kind of internal streaming logic. Covers: the fix blocking a truncated call, a companion test proving the pre-fix behavior really did let it through, and a control test confirming normal finalized calls are unaffected.

Full `core/task` suite passes: 27 test files, 381 tests, no regressions.

Fixes #1221.


canblmz1:
Added an integration-level test alongside the existing simulation-based ones, per the review comment - drives a truncated \`write_to_file\` call through the real \`Task\` streaming + \`presentAssistantMessage\` flow (\`recursivelyMakeClineRequests\` + a mocked \`attemptApiRequest\` stream) instead of mirroring the finalize-null logic in an isolated function.

Two things worth noting from building it:

1. \`writeToFileTool.handle\` legitimately gets called with \`partial: true\` while the call is still streaming (\`BaseTool.handle\` short-circuits to a no-op preview hook in that case, per its own doc comment). The test asserts it's never called with \`partial: false\` specifically - that's the flag that actually reaches \`execute()\`/disk writes.
2. Confirmed the test fails for the right reason by temporarily reverting the \`Task.ts\` fix and re-running it: without the fix, \`handle\` does get called with \`partial: false\`. An earlier version of this using a \`.json\` target path was a false negative - the default test task mode's Architect-only markdown file restriction blocked the write independently of the \`nativeArgs\` guard, masking the actual bug. Switched to a \`.md\` path so the guard under test is what's actually exercised.

Kept the existing \`truncated-native-tool-args.spec.ts\` tests too rather than replacing them - they're fast and pin the exact finalize-null branch precisely, complementing the integration test rather than duplicating it.

Full \`core/task\` suite: 27 files, 382 tests, all passing.

src/core/task/Task.ts
```diff
@@ -3746,12 +3746,18 @@ export class Task extends EventEmitter<TaskEvents> implements TaskLike {
...
+							// place, so it still carries nativeArgs built from the incomplete partial
+							// parse (e.g. a truncated write_to_file `content` string) - that value was
+							// only ever meant for live progress display, never for execution. Mark the
+							// tool as non-partial so it's presented as complete, and clear nativeArgs so
+							// presentAssistantMessage's `!block.nativeArgs` guard actually short-circuits
+							// it with a structured tool_result instead of executing the truncated value.
 							const existingToolUse = this.assistantMessageContent[toolUseIndex]
 							if (existingToolUse && existingToolUse.type === "tool_use") {
 								existingToolUse.partial = false
+								existingToolUse.nativeArgs = undefined
```
taltas, line 3826: Would it also be worth clearing `params` here, since the streaming partial parse still leaves the truncated values in `params` and they get echoed into API history via the `toolUse.nativeArgs || toolUse.params` fallback?

src/core/task/__tests__/truncated-native-tool-args.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+import { isValidToolName } from "../../tools/validateToolUse"
+import type { ToolUse, WriteToFileToolUse } from "../../../shared/tools"
+
+describe("Truncated native tool-call args on finalize failure (issue #1221)", () => {
+	/**
+	 * Simulates the finalize-null branch from Task.ts (~line 3748) as it exists
+	 * after the fix: on finalizeStreamingToolCall() returning null, mark the
+	 * tool non-partial and clear nativeArgs.
+	 */
+	function finalizeNullBranch(existingToolUse: ToolUse): ToolUse {
```
taltas, line 32: Is there a reason to keep these hand-written copies of the finalize-null branch and guard, given they can drift from the real logic that `Task.spec.ts` already exercises end-to-end?

src/core/task/Task.ts
canblmz1: Good catch, and you're actually right about the mechanism - I'd assumed params stayed {} for native calls based on how it's built at finalize time (Task.ts:713 area), but missed that the streaming partial-update path (NativeToolCallParser.ts:391) populates it too, for handlePartial's UI hooks.

Checked whether that made it an execution risk though: BaseTool.handle only ever reads nativeArgs to build execute()'s params, and throws instead of falling back to block.params when nativeArgs is undefined - so it was never actually exploitable. The real effect was just that the truncated content kept ending up in conversation history via the nativeArgs || params fallback, under a different name than before.

Cleared params to {} too in the latest commit and added an assertion that the recorded history entry for a truncated call doesn't contain the leaked content.

src/core/task/__tests__/truncated-native-tool-args.spec.ts
canblmz1: Thought about pulling this file honestly. The drift risk is real, but I kept both - the Task.spec.ts test is the one that actually proves the fix works (drove it through a fix revert to confirm it fails for the right reason), so it'd catch drift in the finalize-null branch even if this file went stale. This one's just cheap and pins the exact guard condition down precisely, which is handy if someone's trying to understand what the bug was without reading through a mocked stream setup.

Not attached to it though - if you'd rather it go, say so and I'll pull it.

taltas: I'm fine with leaving it, I think it's ok to have it here in case the other spec changes

canblmz1: Thanks for taking a look.

canblmz1:
@coderabbitai review

canblmz1:
CodeRabbit won't take review commands from non-members here, so it's still holding the CHANGES_REQUESTED from the first commit. What it asked for is in now (an integration test in Task.spec.ts that drives the real streaming flow), along with the params follow-up from taltas's review. Could a maintainer restart the review when there's a chance?

taltas:
@coderabbitai can you review this please

PR #1635, opened 2026-09-14
feat: add model selector UI to chat

### LINKED ISSUE #1502: [ENHANCEMENT] Add a model selector to the chat input area

The chat input area only had an API configuration (profile) selector, so users had to leave the chat and go into Settings to change the actual model — even though switching models is a very common action.

This affects anyone chatting who wants to quickly try a different model mid-conversation. Currently they have to interrupt their workflow, open Settings, find the relevant provider, and change the model there before returning to chat.

Add a model selector directly in the chat input toolbar, next to the existing mode/API-config selectors, so users can switch models inline without leaving the chat.

- Should support both dynamic-model providers (e.g., OpenRouter) and static-model providers.
- Should be searchable when the model list is long.
- Should degrade gracefully for providers where inline model selection isn't supported (e.g., show a tooltip pointing back to Settings) instead of breaking or hiding the control entirely.
- Needs i18n coverage across all supported locales.

### Request checklist

_No response_

_No response_

Add a `ModelSelector` component reusing `useRouterModels`/`getStaticModelsForProvider` to source the model list per provider, and mount it in `ChatTextArea` alongside the other selectors. Add `selectModel` / `selectModelUnsupported` strings to `chat.json` in every locale.

_No response_

daewoongoh:
### Related GitHub Issue

Closes: #1502 

### Description

Adds a `ModelSelector` to the chat input toolbar so users can pick a model directly from chat instead of going through Settings.

- New `ModelSelector` component (`webview-ui/src/components/chat/ModelSelector.tsx`), mounted in `ChatTextArea` next to the existing `ModeSelector` / `ApiConfigSelector`.
- Resolves the model list per provider: dynamic providers (e.g. OpenRouter) via `useRouterModels`, static-model providers via `getStaticModelsForProvider`.
- Falls back gracefully for unsupported/retired providers, showing a `selectModelUnsupported` tooltip that points back to Settings instead of hiding or breaking the control.
- Uses `Fzf` for search once the model list is long enough (`SEARCH_THRESHOLD`).
- Adds `selectModel` / `selectModelUnsupported` i18n strings to `chat.json` for all supported locales.

### Test Procedure

- Added unit tests in `webview-ui/src/components/chat/__tests__/ModelSelector.spec.tsx` covering supported/unsupported providers, dynamic vs. static model lists, and search behavior.
- Manual testing: open chat, switch between API configurations using different providers (e.g. OpenRouter, Anthropic), confirm the model dropdown lists the correct models, selecting one updates the active configuration, and providers without inline model listing show the fallback tooltip.

### Pre-Submission Checklist

- [x] **Visual Snapshot** (UI changes only): If a user would notice this change at a glance (layout, theme tokens, brand elements, empty/error states), I've added or updated a `*.visual.tsx` snapshot in `webview-ui/`. See `webview-ui/AGENTS.md` → "When a UI change needs a snapshot".

### Documentation Updates

### Get in Touch

hehegwk_23849


daewoongoh:
@coderabbitai review

daewoongoh:
Addressed all 3 actionable comments in 6d755f096:

- **aria-pressed on model list items**: Added `aria-pressed={isCurrentModel}` to the model button so assistive tech can announce the current selection. Extended the dynamic-model-list test to assert `aria-pressed` is `true`/`false` for selected/unselected items.
- **Retired-provider fallback label**: The disabled view now falls back to `rawProvider` (the actual configured, retired provider, e.g. `groq`) instead of the generic `openrouter` mapping when there's no selected model label. If a selected model label *is* available, it's still shown as before. Updated the existing test and added a new one covering the "label preserved" case.
- **Portal container assertion**: The `PopoverContent` mock now captures its `container` prop, and the portal test asserts it's the exact instance returned by `useRooPortal` (`document.body`), not just that the hook was called.

Also added a test for the `apiConfiguration={{}}` (unset `apiProvider`) path flagged by the pre-merge Regression Evidence check, verifying the `useRouterModels` fallback args and the resulting disabled state.

All 40 ModelSelector tests, lint, and typecheck pass. Ready for re-review.

webview-ui/src/components/chat/ModelSelector.tsx
```diff
@@ -0,0 +1,282 @@
...
+			if (!modelConfig) {
+				return
+			}
+
+			const updated: ProviderSettings = {
+				...apiConfiguration,
+				reasoningEffort: undefined,
+				modelMaxTokens: undefined,
+				modelMaxThinkingTokens: undefined,
+			}
```
edelauna, line 160: When the provider is Bedrock and the user switches away from `custom-arn`, the shared `handleModelChangeSideEffects` helper (`providerModelConfig.ts:249-252`) also clears `awsCustomArn`. Leaving it set here means the extension reads the stale ARN at `bedrock.ts:231` and ignores the newly selected `apiModelId`.

```suggestion
				const updated: ProviderSettings = {
					...apiConfiguration,
					reasoningEffort: undefined,
					modelMaxTokens: undefined,
					modelMaxThinkingTokens: undefined,
				}
				if (provider === providerIdentifiers.bedrock && modelId !== "custom-arn") {
					;(updated as Record<string, unknown>)["awsCustomArn"] = undefined
				}
```

webview-ui/src/components/chat/ModelSelector.tsx
```diff
@@ -0,0 +1,282 @@
...
+			;(updated as Record<string, unknown>)[modelConfig.field] = modelId
+
+			vscode.postMessage({
+				type: "upsertApiConfiguration",
+				text: currentApiConfigName,
+				apiConfiguration: updated,
+			})
```
edelauna: `ModeSelector` and `ApiConfigSelector` both accept an `onChange` prop and let `ChatTextArea` own the `vscode.postMessage` call. Is there a reason this component posts directly rather than following the same pattern? It would make the component easier to test in isolation and allow the parent to add telemetry.

webview-ui/src/components/chat/ModelSelector.tsx
```diff
@@ -0,0 +1,282 @@
...
+import { vscode } from "@/utils/vscode"
+
+import {
+	getProviderModelConfig,
+	getStaticModelsForProvider,
+	isStaticModelProvider,
+} from "../settings/utils/providerModelConfig"
+import { filterModels } from "../settings/utils/organizationFilters"
+
+const SEARCH_THRESHOLD = 6
```
edelauna: `ModeSelector.tsx` already declares `const SEARCH_THRESHOLD = 6` at line 20. Worth extracting to a shared file in this directory?

webview-ui/src/components/chat/__tests__/ModelSelector.spec.tsx
```diff
@@ -0,0 +1,1189 @@
...
+		await user.tab()
+		expect(screen.getByRole("button", { name: "openrouter/model-0" })).toHaveFocus()
+		await user.keyboard(" ")
+		expect(screen.queryByTestId("popover-content")).not.toBeInTheDocument()
+		expect(vscode.postMessage).toHaveBeenCalledWith(
+			expect.objectContaining({
+				type: "upsertApiConfiguration",
+				apiConfiguration: expect.objectContaining({ openRouterModelId: "openrouter/model-0" }),
+			}),
+		)
```
edelauna: This test renders without `currentApiConfigName`, so `text` in the posted message will be `undefined`. Could we assert that here to catch a mutation that defaulted it to an empty string or a fixed name?

```suggestion
		expect(vscode.postMessage).toHaveBeenCalledWith(
			expect.objectContaining({
				type: "upsertApiConfiguration",
				text: undefined,
				apiConfiguration: expect.objectContaining({ openRouterModelId: "openrouter/model-0" }),
			}),
		)
```

webview-ui/src/components/chat/__tests__/ModelSelector.spec.tsx
```diff
@@ -0,0 +1,1189 @@
...
+		fireEvent.click(screen.getByTestId("model-selector-trigger"))
+
+		const list = within(screen.getByTestId("popover-content"))
+
+		// Only the allowed model is present.
+		expect(list.getByText("claude-3-5-haiku-20241022")).toBeInTheDocument()
+
+		// Other models are filtered out by the allow list.
+		expect(list.queryByText("claude-sonnet-4-5")).not.toBeInTheDocument()
+	})
```
edelauna, line 1067: The three allow-list tests all include the current provider in `allowList.providers`. What happens when `allowAll: false` but the provider isn't listed at all? `filterModels` returns `{}` in that case, which should render the disabled view — worth adding a test case for that path?

edelauna (changes requested):
Nice! This would be a great feature. Thanks for adding it - had some comments.

webview-ui/src/components/chat/ModelSelector.tsx
daewoongoh: fixed exactly as suggested in 54155ee06: handleSelect now clears awsCustomArn when the provider is Bedrock and the selected model isn't custom-arn. Also added a regression test that seeds a stale awsCustomArn and asserts it's cleared on the onChange payload.

webview-ui/src/components/chat/ModelSelector.tsx
daewoongoh: Refactored in 54155ee06 to match ModeSelector/ApiConfigSelector: ModelSelector now takes an onChange(apiConfiguration) prop instead of calling vscode.postMessage directly, and ChatTextArea owns the upsertApiConfiguration post via a new handleModelChange callback. Updated the spec file to assert on the onChange mock instead of vscode.postMessage.

webview-ui/src/components/chat/ModelSelector.tsx
daewoongoh: Extracted to webview-ui/src/components/chat/selectorConstants.ts and imported from both ModeSelector.tsx and ModelSelector.tsx.

webview-ui/src/components/chat/__tests__/ModelSelector.spec.tsx
daewoongoh: Added "shows the disabled view when the provider is entirely absent from the organization allow list" in 54155ee06, covering the filterModels → {} → disabled-view path.

webview-ui/src/components/chat/ModelSelector.tsx
```diff
@@ -0,0 +1,287 @@
...
+					provider,
+					organizationAllowList,
+				) ?? {}
+			)
+		}
+
+		return filterModels(modelsWithoutCustomArn, provider, organizationAllowList) ?? {}
+	}, [modelConfig, dynamicProvider, routerModels.data, provider, apiConfiguration, organizationAllowList])
+
+	const modelIds = useMemo(() => Object.keys(models), [models])
```
edelauna, line 88: Could this list exclude models whose `ModelInfo.deprecated` flag is set, while retaining the currently selected deprecated model? Otherwise, a dynamic provider can expose a deprecated model here and `handleSelect()` can save it, unlike the settings model picker.

edelauna (changes requested):
Thanks! This is a cool feature. Had 1 additional request for models that get displayed

PR #1637, opened 2026-09-14
fix(task): keep delegated child mode isolated

### LINKED ISSUE #1632: [BUG] Continuous re-delegation
**Describe the bug**

I'm having an issue where I start in orchestrator mode and every spawned subtask thinks it's still in orchestrator mode and re-delegates.  I've seen this go 30+ deep.

**To Reproduce**

**Expected behavior**

Roo to recognize the current mode and act accordingly

**Screenshots**

Image

Image

**Video**
Add a video of the incident happening, it will help us debug

v3.82.1

**Additional context**

PierrunoYT:
## Summary
- use each task's local mode when generating environment details and validating tool execution
- use task-local mode for mode switching, MCP restrictions, skills, and slash-command fallback
- prevent Code-mode children of Orchestrator tasks from continuously re-delegating

Fixes #1632

## Verification
- focused Vitest suites: 82 tests passed
- TypeScript typecheck passed
- ESLint passed with suppression pruning and zero warnings


src/core/tools/__tests__/mcpServerRestriction.spec.ts
```diff
@@ -77,7 +83,7 @@ describe("getAllowedMcpServersForTask", () => {
 
 	it("returns undefined when the mode cannot be resolved", async () => {
 		vi.mocked(getModeBySlug).mockReturnValue(undefined as any)
-		const task = makeTask({ mode: "missing", customModes: [] })
+		const task = makeTask({ mode: "code", customModes: [] }, "missing")
 		await expect(getAllowedMcpServersForTask(task)).resolves.toBeUndefined()
 	})
```
edelauna, line 88: The modified test above adds `expect(getModeBySlug).toHaveBeenCalledWith("code", [])` to catch a mode-source mutation, but this test does not have the equivalent. `getModeBySlug` is mocked to return `undefined` for any input, so passing `state.mode` ("code") instead of `taskMode` ("missing") would still produce `toBeUndefined()` passing. Worth adding `expect(getModeBySlug).toHaveBeenCalledWith("missing", [])` here?

src/core/tools/__tests__/switchModeTool.spec.ts
```diff
@@ -41,6 +41,7 @@ describe("SwitchModeTool", () => {
 		mockGetState = vi.fn().mockResolvedValue({ mode: "code", customModes: [] })
 
 		mockTask = {
+			getTaskMode: vi.fn().mockResolvedValue("code"),
 			consecutiveMistakeCount: 0,
```
edelauna, line 45: The three happy-path tests (lines 155, 179, 297) use this `getTaskMode` default alongside `mockGetState → { mode: "code" }`, so both sources return the same value. A mutation that reads `state.mode` instead of `task.getTaskMode()` for the "from" label would still produce `"Successfully switched from Code mode…"` and pass — only the test at line 330 (architect case) catches it. Would it help to set `mockGetState` to `{ mode: "orchestrator" }` in those three tests?

edelauna (approved):
Thanks for this! Will merge - noticed some test strengthening, but will leave it to you if you'd like to close this in a later ticket.

PR #1663, opened 2026-09-16
test(e2e): poll restart conversation history

### LINKED ISSUE #1641: [BUG] Make the restart-persistence e2e tolerate the atomic-rename swap window when reading API conversation history
## Problem

The restart-persistence e2e asserts on single, unguarded reads of a task's persisted API conversation history: `api.hasTaskApiConversationHistorySequence` at `apps/vscode-e2e/src/suite/restart-persistence.test.ts:89`, and again at `restart-persistence.test.ts:109` after the task is resumed. Nothing polls.

Observed in version v3.82.1

Those reads race the writes that produce the file. History is persisted through `safeWriteJson`, which renames the existing target to a backup (`src/utils/safeWriteJson.ts:130`) and only then renames the new temp file into place (`src/utils/safeWriteJson.ts:142`); between the two renames the target path does not exist. The read path treats that miss as absent history rather than as a transient state: `hasTaskApiConversationHistorySequence` (`src/extension/api.ts:256`) receives an empty history from the task loader when the file is missing, and reports `false` when the read fails at all. A read that lands in that swap window therefore reports the turn sequence as missing even though the write is in flight and will land.

## Consequence

A swap-window miss fails the restart assertion while the same read succeeds moments later: an intermittent e2e failure with no production bug behind it. Completion-time ordering needs no new test: `src/core/task/__tests__/Task.persistence.spec.ts:509` already pins that the durable history write settles before completion is emitted, with siblings at `Task.persistence.spec.ts:601` and `Task.persistence.spec.ts:670` covering the retry paths. What is missing is test-side tolerance on the reads above.

## Acceptance criteria

1. The restart-persistence e2e waits for the conversation history to become visible before asserting on it, through a bounded poll-until-visible with a deadline; no fixed sleeps.
2. The wait uses the same read that the assertions themselves use, so a swap-window miss can no longer surface as a false negative.

PierrunoYT:
### Related GitHub Issue

Closes: #1641

### Description

The restart-persistence E2E previously checked persisted API conversation history with one immediate read after extension-host restart and another after reopening the task. Those reads could land in `safeWriteJson`'s atomic-rename swap window, where the destination path is briefly absent, and report a false negative.

This change routes both checks through a shared `waitForMarkedCompletion` helper. It repeatedly calls the same `api.hasTaskApiConversationHistorySequence` read used by the original assertions until the sequence becomes visible or the existing 30-second `waitFor` deadline expires. There are no fixed sleeps and no production persistence behavior changes.

### Test Procedure

```shell
pnpm --filter @roo-code/vscode-e2e lint
pnpm --filter @roo-code/vscode-e2e check-types
pnpm lint
pnpm check-types
pnpm exec prettier --check apps/vscode-e2e/src/suite/restart-persistence.test.ts
git diff --check
```

All commands pass. The commit and push hooks also completed the full workspace lint/typecheck gates (11/11 packages).

The targeted command below built the extension and webview and downloaded the pinned VS Code binary, but could not launch native VS Code in the local orb because no X display or `xvfb-run` is available. CI's mocked E2E runner should execute this exact scenario:

```shell
TEST_FILE=restart-persistence.test pnpm --filter @roo-code/vscode-e2e test:ci:mock
```

### Pre-Submission Checklist

- [x] **Scope**: The change is limited to the restart-persistence E2E read race.
- [x] **Self-Review**: I have performed a self-review.
- [x] **Testing**: The affected package and workspace lint/typecheck gates pass; CI will run the graphical E2E.
- [x] **Visual Snapshot**: N/A — test synchronization only; no UI changes.
- [x] **Contribution Guidelines**: I have read and agree to the contributor guidelines.

### Visual Snapshots

N/A — no UI changes.

N/A.

### Documentation Updates

### Additional Notes

The production write ordering and history loader are intentionally unchanged. This only makes the E2E tolerate the already-documented atomic publication window.

### Get in Touch

GitHub: @PierrunoYT


edelauna (approved):
Thanks for addressing this.

Issue #1541, opened 2026-09-06
perf: globalState write-through causes extension host freezes and OOM pause on large task histories


edelauna:
## Summary

The `TaskHistoryStore` correctly persists history items as individual JSON files on disk. However, `ClineProvider` also maintains a **5-second debounced write-through** of the full `getAll()` array back into VS Code's `globalState("taskHistory")`. On any session with many or large tasks, this blob exceeds VS Code's recommended size, causing repeated extension host freezes and a "Pausing before OOM" self-protective pause.

## Reproduction

Run a multi-child orchestrator session with 10+ tasks (or tasks with large `completionResultSummary` fields). Observe in the browser devtools console:

```
WARN [mainThreadStorage] large extension state detected
(ZooCodeOrganization.zoo-code, global: true): 2885 KB
Consider to use 'storageUri' or 'globalStorageUri' to store this data on disk instead.
```

This warning fires **on every debounced write** (every 5 s while a task is active), and on `flushGlobalStateWriteThrough` it fires synchronously. The extension host logs `unresponsive` / `responsive` pairs around each flush as it blocks serializing the blob over the RPC bridge. The webview shows "Pausing before OOM" when the blob approaches the size limit.

## Root cause

`ClineProvider.ts:249`
```ts
private static readonly GLOBAL_STATE_WRITE_THROUGH_DEBOUNCE_MS = 5000
```

`ClineProvider.ts:3440-3470` — `scheduleGlobalStateWriteThrough` / `flushGlobalStateWriteThrough`

Every task update calls `taskHistoryStore.atomicReadAndUpdate()`, which calls the `onWrite` callback, which schedules the write-through. Each write serializes **all** history items (including full `task`, `completionResultSummary`, `childIds`, etc.) into a single VS Code key-value entry. A 13-task session with rich summaries produces a ~2.9 MB blob.

The write-through exists because some code paths still fall back to reading `globalState("taskHistory")` (e.g. `ClineProvider.ts:1988, 2251, 2521`) and webview hydration reads it from state.

## Impact

- Extension host becomes unresponsive for several seconds on every flush
- Triggers the extension's own OOM guard ("Pausing before OOM") on large sessions
- Grows unboundedly — each new task or updated `completionResultSummary` increases the blob permanently
- Worst for orchestrator / multi-child sessions, which produce the largest histories

## Proposed fix

[...]

PR #1664, opened 2026-09-16
perf(webview): stop task history globalState writes

### LINKED ISSUE #1542: perf: remove globalState write-through for taskHistory
Part of #1541.

Remove the 5-second debounced write-through that flushes the full `TaskHistoryStore.getAll()` array into VS Code's `globalState("taskHistory")` on every task update. This is the change that immediately stops extension host freezes and the OOM pause.

## Changes

`ClineProvider.ts` only:
- Delete `GLOBAL_STATE_WRITE_THROUGH_DEBOUNCE_MS` constant (line 249)
- Delete `globalStateWriteThroughTimer` field (line 248)
- Delete `scheduleGlobalStateWriteThrough()` and `flushGlobalStateWriteThrough()` methods (lines ~3440–3470)
- Remove the `onWrite` callback hookup that calls `scheduleGlobalStateWriteThrough`

## Safety

Safe to merge standalone. The three `getGlobalState("taskHistory")` fallback reads (lines 1988, 2251, 2521) already call `taskHistoryStore.get()` first via `??` — the globalState fallback silently returns nothing when the key is stale or absent.

PierrunoYT:
## Summary

- stop registering the task-history store write callback that queued full-history `globalState` updates
- remove the debounced write-through timer and disposal-time flush
- keep file-backed `TaskHistoryStore` persistence and legacy migration/fallback reads unchanged
- add a regression assertion that the provider does not register a write-through callback

## Why

Each task mutation could schedule serialization of the entire task history into VS Code `globalState`. Large orchestrator sessions repeatedly copied multi-megabyte histories across the extension-host storage boundary, causing freezes and OOM pressure even though per-task files are already authoritative.

## Validation

- `pnpm --dir src test core/webview/__tests__/ClineProvider.taskHistory.spec.ts` (24 passed)
- `pnpm --dir src test` (8,413 passed, 39 skipped)
- `pnpm lint` (11/11 packages)
- `pnpm check-types` (11/11 packages)
- `pnpm lifecycle:model-check`
- touched-file ESLint with suppression pruning
- Prettier and `git diff --check`

Closes #1542
Related to #1541


edelauna:
@coderabbit review

src/core/webview/ClineProvider.ts
```diff
@@ -344,14 +342,8 @@ export class ClineProvider
...
-		// Initialize the per-task file-based history store.
-		// The globalState write-through is debounced separately (not on every mutation)
-		// since per-task files are authoritative and globalState is only for downgrade compat.
-		this.taskHistoryStore = new TaskHistoryStore(this.contextProxy.globalStorageUri.fsPath, {
-			onWrite: async () => {
-				this.scheduleGlobalStateWriteThrough()
-			},
-		})
+		// Initialize the authoritative per-task file-based history store.
+		this.taskHistoryStore = new TaskHistoryStore(this.contextProxy.globalStorageUri.fsPath)
```
edelauna, line 346: Stops new writes, but the old `"taskHistory"` blob is still sitting in globalState from before migration. VS Code fires the large-state warning whenever it reads or writes the total extension state, so the warning will keep appearing even after this PR lands. Should `initializeTaskHistoryStore` (or the migration block in it) also call `this.context.globalState.update("taskHistory", undefined)` after marking the migration complete?

edelauna (changes requested):
Nice! Thanks you for addressing this, had one question if we could also address existing extension state.

src/core/webview/ClineProvider.ts
edelauna: @PierrunoYT - after reviewing the recent changes, lets revert this request, and instead file a separate ticket for it, it might not be a simple addition for handling old blobs.

PR #1673, opened 2026-09-18
fix(terminal): prevent inline terminal cmd.exe fallback on Windows

### LINKED ISSUE #705: [BUG] PowerShell cmdlets fail when Inline Terminal is ON — execa uses cmd.exe instead of user shell
## A Note from the User

This is a message from me, a human user. I asked AI to handle the rest of the issue report.

I tried to patch the problems I encountered while using Zoo Code myself. The patches may be incomplete or may not work correctly. Please, the bug fix team, judge them directly. I hope Zoo Code continues to grow and mature.

---

## Problem

When "Use Inline Terminal (recommended)" setting is ON, PowerShell cmdlets like `Test-Path`, `Get-ChildItem` return `'Test-Path' is not recognized as an internal or external command`. The inline terminal uses execa backend which defaults to cmd.exe on Windows, ignoring the user's shell preference.

## Context

- **OS**: Windows 11, PowerShell 5.1
- **Zoo Code Version**: v3.63.100188
- **Model**: mimo-v2.5-pro
- **Setting**: "Use Inline Terminal (recommended)" = ON

## Reproduction Steps

1. Zoo Code Settings → Terminal → "Use Inline Terminal (recommended)" ON
2. Execute `Test-Path "README.md"` via `execute_command`
3. Runs in cmd.exe, PowerShell cmdlet fails

## Expected Result

`Test-Path "README.md"` should return `True`.

## Actual Result

```
'Test-Path' is not recognized as an internal or external command, operable program or batch file.
```

## Root Cause

From Zoo Code documentation:
> **ON (Use Inline Terminal)**: Runs commands with the Inline Terminal (in chat), **bypassing shell profiles and VS Code shell integration**. Mechanism: uses a background **"execa" provider** and streams output directly.

When Inline Terminal is ON:
1. Does **NOT** use VS Code's PowerShell terminal
2. Creates **background process** via Node.js `execa` library
3. execa uses **cmd.exe by default** on Windows

The `jJe` (execaTerminal) class in `extension.js` does not respect the user's shell settings when calling execa.

## Applied Patch

We resolved this by turning off Inline Terminal and adding a PowerShell shellPath fallback in the `Pn` constructor (v6 patch):

```javascript
// When getProfileShell() returns undefined in Pn constructor
let explicitShell = process.platform === "win32"
  ? (process.env.SystemRoot || "C:\\Windows") + "\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"
  : (process.env.SHELL || "/bin/bash");
n.shellPath = explicitShell;
```

## Recommended Official Fix

**Option 1**: Respect user shell settings in execa backend:
```javascript
// In jJe (execaTerminal) class
const shellPath = process.platform === "win32"
  ? (Pn.getProfileShell()?.shellPath || process.env.ComSpec || "cmd.exe")
  : (process.env.SHELL || "/bin/bash");
```

**Option 2**: Add shell option for inline terminal in settings:
```json
{
  "roo-cline.inlineTerminal.shell": "powershell"
}
```

**Option 3**: Change default to Inline Terminal OFF (currently ON by default)

xcloudx01:
### Related GitHub Issue

Closes #705

### Description

Fixes a Windows shell mismatch in the Inline Terminal Execa execution path.

When no explicit `execaShellPath` was configured, Execa used `shell: true`, which lets Windows fall back to `%ComSpec%` (`cmd.exe`). This could cause commands to execute under Command Prompt even when Zoo had resolved PowerShell as the intended shell.

The fix preserves an explicit `execaShellPath` when configured and otherwise uses Zoo's existing `getShell()` resolver.

The change is intentionally limited to the Execa execution path and its regression tests.

### Test Procedure

Verified the following:

* Explicit `execaShellPath` overrides are preserved unchanged.
* With no explicit Execa shell, the resolved Zoo shell is used instead of `shell: true`.
* PowerShell resolves correctly through the normal profile configuration.
* Deliberately selecting Command Prompt remains supported.
* `ExecaTerminalProcess.spec.ts`: 18/18 tests passing.
* ESLint passes for the changed files.
* `tsc --noEmit` passes.
* `git diff --check` passes.
* Live-tested on Windows with `%ComSpec%` still set to `cmd.exe`; Inline Terminal successfully executed under PowerShell 7.6.6.

### Pre-Submission Checklist

* [x] **Issue Linked**: This PR is linked to #705.
* [x] **Scope**: Changes are limited to the reported shell fallback issue.
* [x] **Self-Review**: I have reviewed the final diff.
* [x] **Testing**: Regression tests cover the changed behavior.
* [x] **Visual Snapshot**: Not applicable; no UI changes.
* [x] **Documentation Impact**: No documentation updates are required.
* [x] **Contribution Guidelines**: I have reviewed the contributor guidelines.

### Documentation Updates

* [x] No documentation updates are required.

### Additional Notes

The production change is intentionally small: one shell-resolution import and one fallback change in `ExecaTerminalProcess.ts`. No generated bundles, VSIX files, or unrelated changes are included.


src/integrations/terminal/ExecaTerminalProcess.ts
```diff
@@ -40,7 +41,7 @@ export class ExecaTerminalProcess extends BaseTerminalProcess {
 			this.isHot = true
 
 			this.subprocess = execa({
-				shell: BaseTerminal.getExecaShellPath() || true,
+				shell: BaseTerminal.getExecaShellPath() ?? getShell(),
```
taltas, line 44: Can we use a truthy fallback here so an empty persisted shell path still resolves through `getShell()` instead of disabling shell parsing?

taltas (changes requested):
Requesting changes for the inline terminal shell fallback regression described in the inline comment.

src/integrations/terminal/ExecaTerminalProcess.ts
xcloudx01: Done, switched this to a truthy fallback so an empty persisted shell path resolves through `getShell()`.

src/integrations/terminal/__tests__/ExecaTerminalProcess.spec.ts
```diff
@@ -191,4 +198,52 @@ describe("ExecaTerminalProcess", () => {
...
+			expect(getShellSpy).toHaveBeenCalledTimes(1)
+			expect(capturedShellOption().shell).toBe("/bin/zsh")
+		})
+
+		it("keeps the Execa shell equal to getShell() when a Zoo profile override is set", async () => {
+			BaseTerminal.setExecaShellPath(undefined)
+			const getShellSpy = vi.spyOn(shellUtils, "getShell").mockReturnValue("C:\\Windows\\System32\\pwsh.exe")
+			await terminalProcess.run("echo test")
+			expect(getShellSpy).toHaveBeenCalledTimes(1)
+			expect(capturedShellOption().shell).toContain("pwsh.exe")
```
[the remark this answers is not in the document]
xcloudx01: Updated the assertion to require the exact mocked PowerShell path instead of only checking for `pwsh.exe`.

taltas (approved):

PR #1678, opened 2026-09-18
fix(task): preserve subtask links after repeated Stop

### LINKED ISSUE #1676: [BUG] Repeated Stop breaks resume for interrupted subtasks
### Problem

Repeated Stop can run after Zoo Code has already saved a delegated subtask as interrupted.

The second cancellation repeats the same status change. The lifecycle guard rejects it. Recovery then clears the parent and root links.

### Context

This can occur after a delayed response, a repeated click, an extension host interruption, or a window reload.

The parent still waits for the interrupted child. The child must keep its links so the user can resume it.

### Reproduction steps

1. Start a delegated subtask.
2. Save the child with status `interrupted`.
3. Keep the parent in status `delegated` with `awaitingChildId` set to the child.
4. Run `cancelTask()` for the child again.
5. Examine the saved child history.

### Expected result

The second cancellation succeeds without another status change. The child keeps its parent and root links.

### Actual result

Zoo Code treats the repeated cancellation as a persistence error. It clears the child links and blocks the normal resume path.

### App Version

`3.82.1`

### API Provider

`Not Applicable / Other`

### Model Used

`N/A`

### Relevant error

```text
Invalid task status transition: interrupted → interrupted
```

jaszhix:
### Related GitHub Issue

Closes #1676

### Description

Repeated Stop can run after Zoo Code has already saved a delegated subtask as interrupted.

The second cancellation repeats the same status change. The lifecycle guard rejects it. Recovery then clears the parent and root links.

This change treats the saved interrupted state as a successful replay. The child keeps its parent and root links.

The existing code still handles normal cancellation and persistence errors. The strict lifecycle guard remains unchanged.

### Test Procedure

Run the focused cancellation suite:

```bash
pnpm --dir src exec vitest run --maxWorkers=1 core/webview/__tests__/ClineProvider.flicker-free-cancel.spec.ts
```

Expected result: 11 tests pass.

Run ESLint:

```bash
pnpm --dir src exec eslint --prune-suppressions --max-warnings=0 \
  core/webview/ClineProvider.ts \
  core/webview/__tests__/ClineProvider.flicker-free-cancel.spec.ts
```

Expected result: ESLint exits with code 0.

Run the package type check:

```bash
pnpm --dir src check-types
```

Expected result: The command exits with code 0.

### Pre-Submission Checklist

- [x] **Issue Linked**: Closes #1676.
- [x] **Scope**: The change fixes one cancellation error.
- [x] **Self-Review**: The change was compared with current `main` and related PRs.
- [x] **Testing**: The focused tests, ESLint, and package type check passed.
- [x] **Visual Snapshot**: Not applicable.
- [x] **Documentation Impact**: No documentation update is required.
- [x] **Contribution Guidelines**: Reviewed.

### Visual Snapshots

N/A

### Videos

N/A

### Documentation Updates

- [x] No documentation update is required.


src/core/webview/ClineProvider.ts
```diff
@@ -3639,14 +3639,23 @@ export class ClineProvider
 					if (parentHistory?.status === "delegated" && parentHistory?.awaitingChildId === task.taskId) {
 						// Mark the child interrupted and leave parent delegated with awaitingChildId
 						// intact — the user can resume this child later and it will report back.
-						historyItem = interruptDelegatedChild(parentHistory, historyItem!)
-						await this.updateTaskHistory(historyItem)
+						// A previous cancellation may already have persisted the interrupted status
+						// before its caller lost the response. Treat that replay as success without
+						// weakening the lifecycle state machine's self-loop rejection.
+						if (historyItem!.status !== "interrupted") {
```
edelauna, line 3650: The sibling `markDelegatedChildInterrupted` re-fetches child history inside this same lock (around line 712) with a note about closing a TOCTOU window with `cancelTask`. Should `historyItem` be re-fetched here too before the status check, for consistency?

src/core/webview/__tests__/ClineProvider.flicker-free-cancel.spec.ts
```diff
@@ -663,6 +663,80 @@ describe("ClineProvider flicker-free cancel", () => {
...
+				parentTaskId: "parent-1",
+				rootTaskId: "root-1",
+				parentTask: expect.objectContaining({ taskId: "parent-1" }),
+				rootTask: expect.objectContaining({ taskId: "root-1" }),
+			}),
+		)
+		expect(provider["cancelledDelegationChildIds"].has("child-1")).toBe(false)
+		expect(mockOutputChannel.appendLine).not.toHaveBeenCalledWith(
+			expect.stringContaining("Invalid task status transition: interrupted → interrupted"),
+		)
```
edelauna: `updateTaskHistory` is mocked above, so the real transition validator never runs and this log string is never produced regardless of which branch executes. Does the `not.toHaveBeenCalled()` check on line 723 already cover the mutation this targets?

src/core/webview/__tests__/ClineProvider.flicker-free-cancel.spec.ts
```diff
@@ -663,6 +663,80 @@ describe("ClineProvider flicker-free cancel", () => {
 		)
 	})
 
+	it("preserves delegated lineage when cancelling an already-interrupted child", async () => {
```
edelauna: Is there also a variant for the path where the first cancel completed cleanly (so `cancelledDelegationChildIds` is empty before the second cancel)? The current test always pre-seeds the set, but the `delete` no-op path isn't explicitly exercised.

edelauna (changes requested):
Thanks for this change!

Could you also add the following:

1. **Re-fetch child history inside the lock**  — matching `markDelegatedChildInterrupted` at line 712. This closes the inconsistency without needing to reason about the concurrent case at all.

2. **Add a second test variant for the no-stale-entry path** — the current test always pre-seeds `cancelledDelegationChildIds` to simulate a prior failed cancel. The more common second-Stop scenario (first cancel fully succeeded, set is empty) is not covered.

src/core/webview/ClineProvider.ts
jaszhix: Good catch. `cancelTaskInternal` read the child before abort cleanup, so that snapshot could be stale when the parent transition lock was acquired. I now refresh the child inside the lock before checking its status, matching `markDelegatedChildInterrupted`. I also added a regression where the first read is `active` and the in-lock read is `interrupted`.

src/core/webview/__tests__/ClineProvider.flicker-free-cancel.spec.ts
jaszhix: Yes. `not.toHaveBeenCalled()` already covers the mutation path this test needs. I removed the redundant log assertion.

src/core/webview/__tests__/ClineProvider.flicker-free-cancel.spec.ts
jaszhix: Agreed. I parameterized the test to cover both a pre-seeded stale guard and an already-empty guard. Both cases preserve lineage and leave the guard empty.

src/core/webview/__tests__/ClineProvider.flicker-free-cancel.spec.ts
```diff
@@ -663,6 +663,145 @@ describe("ClineProvider flicker-free cancel", () => {
...
+			if (id === "parent-race") return Promise.resolve({ historyItem: parentHistory })
+			throw new Error(`unexpected task lookup: ${id}`)
+		}) as unknown as ClineProvider["getTaskWithId"]
+
+		const updateTaskHistorySpy = vi.spyOn(provider, "updateTaskHistory").mockResolvedValue([])
+		vi.spyOn(provider, "createTaskWithHistoryItem").mockResolvedValue(undefined as unknown as CreatedHistoryTask)
+
+		await provider.cancelTask()
+
+		expect(childReads).toBe(2)
```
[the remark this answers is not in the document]
jaszhix: Fixed in 721747f8f.

edelauna (approved):
Thanks for addressing this!

PR #1682, opened 2026-09-19
fix(prompts): report the shell that actually runs under Inline Terminal

### LINKED ISSUE #1568: [BUG] System prompt reports wrong shell when Inline Terminal is used
## Problem / Value
When the Inline Terminal is enabled, Zoo tells the model that its commands will run in the shell configured in VS Code, but the commands actually run in a different shell - on Windows, the plain Command Prompt. The model then writes commands in a dialect the terminal cannot understand, and has to guess and retry before anything works. Making the reported shell match the real one would remove failed commands, wasted turns, and needless confusion for both the model and the user.

## Context
Affects Windows users whose VS Code default terminal is something other than Command Prompt (PowerShell, PowerShell 7, Git Bash, etc.) while the recommended "Use Inline Terminal (recommended)" setting is left on - which is the out-of-the-box configuration.

What happens today: the environment summary handed to the model names the VS Code shell (PowerShell in my case), regardless of the fact that, with the Inline Terminal active, commands the model runs are executed by Command Prompt. The model only discovers the mismatch by failing.

What should happen: the model should always be told the kind of shell it will actually get, so its very first command is written for the right shell.

## Reproduction
1) On Windows, set your VS Code default terminal to PowerShell
2) In Zoo settings, leave "Use Inline Terminal (recommended)" enabled
3) Ask Zoo to do anything that runs a command, and observe the shell Zoo reports to the model versus the shell the command actually runs in
4) Expected result: Zoo tells the model the same shell its commands really run in
5) Actual result: Zoo tells the model the shell is PowerShell, while the commands are executed by Command Prompt, so shell-specific commands fail

## Constraints/Preferences
- Whatever execution mode is active, the shell named to the model must be the shell that actually runs its commands.
- Accompanying advice about shell-specific command syntax and chaining should come from that same, accurate source rather than the VS Code setting alone.

## Related reports
Previously reported and closed: #82 (System prompt on Windows says shell is cmd.exe but PowerShell is used via the VS Code terminal) and #321 (Windows: shell mismatch between AI prompt generation and Inline Terminal execution). Still open, #508 (Inline terminal always uses system shell instead of what shell is configured in VS Code) covers the execution side only, not what the model is told. No open issue was found for the system prompt reporting the wrong shell.

DaubnerF:
## Problem

With "Use Inline Terminal" enabled, which is the default, commands run through execa instead of the VS Code integrated terminal. On Windows execa spawns them in `COMSPEC` (cmd.exe for most users) regardless of the VS Code default profile, but `getShell()` kept resolving the profile. The system prompt therefore told the model "Default Shell: PowerShell" while its commands executed in cmd.exe, and the shell-syntax advice and recommended chaining operator were derived from the same wrong value. The model discovered the mismatch only when its first command failed. #82 and #321 reported adjacent forms of this mismatch.

The headless CLI host hit the same mismatch through a second gap: nothing seeds the inline-terminal flag in `BaseTerminal`, and its static default disagreed with the setting default, so shell resolution fell through to a PowerShell guess with no VS Code present.

## Change

When the Inline Terminal is active, `getShell()` now returns the shell execa actually spawns: `process.env.COMSPEC` (fallback `cmd.exe`) on Windows, `/bin/sh` on POSIX. With the Inline Terminal off, resolution is unchanged. The prompt label stays "Default Shell:"; the fix is the reported value.

The inline-terminal flag in `BaseTerminal` now seeds to the same default as the user setting, which closes the CLI gap. Nothing about which terminal executes commands changes; the flag is consulted for prompt reporting only.

One pre-existing divergence is now documented in `execaDefaultShellForPlatform()`: a `COMSPEC` outside the shell allowlist is gated to the safe cmd.exe default in the reported string, while execa still spawns the configured executable. This PR changes what is reported, not what runs.

## Tests

`src/utils/__tests__/shell.spec.ts` gains 11 cases: win32 COMSPEC reported over a PowerShell profile, no-COMSPEC fallback, explicit execa shell path precedence, allowlist gating and casing, Zoo profile override ignored, POSIX `/bin/sh`, inline-off behavior unchanged, and fresh-module coverage of the unseeded default. The existing #634 divergence suite now pins the inline-off path explicitly. Focused suites are green locally (`shell.spec.ts` 60/60, divergence suite 3/3, lint and typecheck clean); CI runs the full suite.

## Related: intentionally out of scope

- Tracked separately as #1743: a spawn-side hardening gap that exists on main independently of this change: when `execaShellPath` is set, the inline terminal spawns that raw configured path without passing it through the shell allowlist (`src/integrations/terminal/ExecaTerminalProcess.ts`), a file outside this change's set. This change only alters which shell is reported to the model.

Closes #1568

### DIFF
--- a/src/utils/__tests__/shell.spec.ts
+++ b/src/utils/__tests__/shell.spec.ts
@@ -28,6 +28,9 @@ vi.mock("path", async () => {
 	}
 })
 
+// Captured before any test spies patch it, so a suite can restore real profile resolution.
+const getProfileShellOriginal = Terminal.getProfileShell.bind(Terminal)
+
 describe("Shell Detection Tests", () => {
 	let originalPlatform: string
 	let originalEnv: NodeJS.ProcessEnv
@@ -82,6 +85,7 @@ describe("Shell Detection Tests", () => {
 		// Clear Zoo profile override and execa shell path between tests.
 		Terminal.setTerminalProfile(undefined)
 		BaseTerminal.setExecaShellPath(undefined)
+		BaseTerminal.setShellIntegrationDisabled(false)
 	})
 
 	afterEach(() => {
@@ -90,6 +94,7 @@ describe("Shell Detection Tests", () => {
 		vscode.workspace.getConfiguration = originalGetConfig
 		Terminal.setTerminalProfile(undefined)
 		BaseTerminal.setExecaShellPath(undefined)
+		BaseTerminal.setShellIntegrationDisabled(false)
 		vi.clearAllMocks()
 	})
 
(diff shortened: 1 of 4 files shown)


edelauna:
@coderabbit approve

src/utils/shell.ts
```diff
@@ -274,9 +289,10 @@ export function getShell(): string {
 	//    regardless of VS Code profile settings.
 	shell = BaseTerminal.getExecaShellPath() ?? null
 
-	// 2. VS Code profile config (Zoo override first, then default profile).
+	// 2. Inline Terminal active: execa spawns commands in its own default
+	//    shell, so the VS Code terminal profile does not run them (issue #1568).
```
edelauna: The step-2 comment explains the `shellIntegrationDisabled=true` path well. Could it also cover the `false` branch? Before this change the comment read "VS Code profile config (Zoo override first, then default profile)" — without that, it’s not obvious when `getShellFromVSCode()` applies.

src/utils/shell.ts
```diff
@@ -220,6 +220,21 @@ function getShellFromEnv(): string | null {
 	return null
 }
 
+/**
+ * Returns the shell execa actually runs commands in under the Inline Terminal:
+ * Node's spawn-shell default (COMSPEC/cmd.exe on Windows, /bin/sh on POSIX),
+ * independent of the VS Code terminal profile (issue #1568).
+ * Note: a COMSPEC outside the SHELL_ALLOWLIST is reported as the platform-safe
+ * fallback by getShell()'s step-6 gate, so on such machines the reported string
+ * and the spawned executable can differ (allowlist security tradeoff).
```
edelauna: Does this Note correctly describe the divergence? `ExecaTerminalProcess` passes `getShell()` to execa’s `shell:` option, so the step-6 allowlist gate applies to both the reported value and what execa receives — a non-allowlisted COMSPEC would gate to `cmd.exe` in both places. Is the real divergence path actually `setExecaShellPath()` with a non-allowlisted value, since `ExecaTerminalProcess` uses that directly without calling `getShell()`?

src/utils/__tests__/shell.spec.ts
```diff
@@ -90,6 +94,7 @@ describe("Shell Detection Tests", () => {
 		vscode.workspace.getConfiguration = originalGetConfig
 		Terminal.setTerminalProfile(undefined)
 		BaseTerminal.setExecaShellPath(undefined)
+		BaseTerminal.setShellIntegrationDisabled(false)
 		vi.clearAllMocks()
```
edelauna: The `Inline Terminal` suite’s `beforeEach` notes that earlier suites leak a `getProfileShell` spy because this `afterEach` uses `clearAllMocks()` rather than `restoreAllMocks()`. Would switching to `vi.restoreAllMocks()` here fix the root cause and let you drop the `getProfileShellOriginal` capture and the workaround in the new suite’s `beforeEach`?

edelauna (approved):
Will approve and merge - had some minor nits, and documentation improvements if you'd like to tackle in a separate issue.

PR #1704, opened 2026-09-19
fix(anthropic): gate prompt caching on supportsPromptCache instead of model-id list

### LINKED ISSUE #717: [BUG] Anthropic prompt caching gated on hard-coded model-id switch instead of info.supportsPromptCache

In the Anthropic provider, prompt-cache breakpoints (`cache_control`) are attached only inside a hard-coded `switch (modelId)` allowlist of literal model ids. Any cacheable Claude model whose exact id is not pasted into that list falls through to the `default:` branch, which sends the system prompt and messages with **no** `cache_control`, so prompt caching is silently disabled for it.

Affects anyone using a Claude model on the Anthropic provider whose literal id is not in the allowlist — most relevantly, every newly added Anthropic model until someone hand-edits the switch. The model is fully cacheable (`supportsPromptCache: true` in the model registry) but receives full-priced uncached input on every turn because the breakpoint is never emitted. On long agentic loops the system prompt + tool schema + prior turns are re-sent uncached each turn.

### Reproduction steps

1) Environment: Zoo Code with the Anthropic provider, any Claude model that is `supportsPromptCache: true` in `packages/types/src/providers/anthropic.ts` but whose id is absent from the two switch statements in `src/api/providers/anthropic.ts`.
2) Open `src/api/providers/anthropic.ts` at `createMessage`. The outer `switch (modelId)` (~line 91) is the only place that builds the request with `cache_control` on the system block and the last/second-last user messages; the nested `switch (modelId)` (~line 161) is the only place that adds the `prompt-caching` beta header.
3) Select a Claude model not listed in those two switches. The request is built by the `default:` branch (~line 198), which sends `system: [{ text, type }]` (no `cache_control`) and plain messages.
4) Observe in the API usage that `cache_creation_input_tokens` / `cache_read_input_tokens` stay at 0 for that model — caching never engages.

This is a static-analysis finding (a small linter for prompt-cache anti-patterns we call CacheLint) that I then confirmed by reading the live source on `main` at commit `0084cc8`.

### Expected result

Whether caching is applied should be driven by the model's own capability flag, `ModelInfo.supportsPromptCache`, which is already `true` for all entries in `anthropicModels`. Every cacheable Anthropic model should get the `cache_control` breakpoints and the prompt-caching beta header automatically — no hand-maintained id list.

### Actual result

Caching is gated on two parallel hard-coded id allowlists. New/unlisted models fall to the `default:` branch and get zero caching until both switches are manually updated (as PR #555 "Add Fable 5" and the Opus 4.8 PR had to do).

### Suggested fix

Replace the two `switch (modelId)` statements with a single capability check, e.g.:

```ts
if (info.supportsPromptCache) {
  // attach cache_control to the system block + last/second-last user message
  // and push the "prompt-caching-2024-07-31" beta header
} else {
  // existing default: plain system + messages, no cache_control
}
```

`info` is already destructured from `this.getModel()`, so no new plumbing is needed, and the change keeps behavior identical for all currently-listed models while letting future models inherit caching automatically. Happy to verify against a specific model id if helpful.

### App Version

main @ `0084cc8` (latest at time of writing)

### API Provider (optional)

Anthropic

myk1yt:
### Related GitHub Issue

Closes: #717

### Description

Anthropic prompt caching was gated on a hard-coded 19-model-id `switch` (in two places in `src/api/providers/anthropic.ts`), so:

- custom or newly added Anthropic models that support caching silently got NO `cache_control` breakpoints and no `prompt-caching-2024-07-31` beta header (they fell to `default:`), inflating cost/latency;
- the list duplicated knowledge the `ModelInfo` registry already carries.

Both switches are now a single capability check, `if (info.supportsPromptCache)`, where `info` is the resolved model info (`this.getModel()`) already in scope. Where breakpoints are inserted is unchanged — only the condition changed.

Behavior notes (verified against the registry):

- All 20 models in `packages/types/src/providers/anthropic.ts` have `supportsPromptCache: true`, and the old 19-id list covers the registry exactly — so no registry model changes behavior.
- One intended improvement: the virtual id `claude-3-7-sonnet-20250219:thinking` previously fell to `default` (uncached) because the switch matched exact strings; it now resolves to its base model's info and gets caching — exactly the bug class this issue fixes.
- Unknown/custom ids resolve through `guessModelInfoFromId`, which always yields a registry `ModelInfo`; in the worst theoretical case of an undefined flag, `if (undefined)` preserves the old no-caching default. No unsafe path.

### Test Procedure

- `src/api/providers/__tests__/anthropic.spec.ts` (+2 tests):
  - custom id `claude-sonnet-5-bf` (not in the old list; family-guessed info with `supportsPromptCache: true`) now receives system/message `cache_control` and the beta header — fails on `main`.
  - `supportsPromptCache: false` (spy-injected) sends a plain system block, no message `cache_control`, no beta header — fails on `main`.
- `cd src && ./node_modules/.bin/vitest run api/providers/__tests__/anthropic.spec.ts` → 67/67 passed; related suites (anthropic-vertex, anthropic-filter, transform/caching/anthropic) → 64/64 passed.
- `git diff -w` confirms the provider change is condition-only (the visible diff is switch→if re-indent).
- ESLint on both files → clean.

### Pre-Submission Checklist

- [x] **Visual Snapshot** (UI changes only): N/A — no UI changes.

### Visual Snapshots

N/A.

N/A.

### Documentation Updates

### Additional Notes

The nested second check (beta header) is now technically redundant (same `info`, no mutation between), but it mirrors the old structure 1:1 to keep the diff reviewable; happy to simplify in a follow-up if preferred.

### Get in Touch

GitHub: @myk1yt — please tag me here; I monitor notifications.


myk1yt:
@coderabbitai review — the once-per-request supportsPromptCache refactor you asked for landed in 9eb05b2e; requesting the formal review of the current head.

PR #1708, opened 2026-09-20
fix(task): merge stale partial reasoning snapshot instead of duplicating

### LINKED ISSUE #1346: [BUG] Reasoning text rendered twice in Code mode (stale partial snapshot not merged)
Describe the bug

In Code mode, the model's reasoning/thinking text is sometimes rendered twice in the chat panel. The model only infers once — a single API request containing a single reasoning block (verified in api_conversation_history.json, token usage is normal). The duplication is intermittent and always consists of the same reasoning text appearing as a pair: a frozen streaming snapshot followed by the final version, identical except for 1–2 trailing characters (e.g. a missing trailing period).

This is a UI message-store defect, not a double inference and not a provider parsing issue.

To Reproduce

Go to a new task in Code mode with an OpenAI-compatible thinking model configured (e.g. GLM-4.7, Qwen3-thinking, DeepSeek-R1)
Click send on a prompt that triggers multiple tool calls, e.g. "List all .ps1 files on the desktop and inspect each one"
Wait for the model to emit a reasoning segment and then invoke a tool (reasoning block followed by a tool_use block in the same response)
See the thinking area render the same reasoning text twice (frozen streaming snapshot + final write)
Deterministic verification (no visual guesswork):

Open the task's message store:

%APPDATA%\Code\User\globalStorage\zoocodeorganization.zoo-code\tasks\<task-id>\ui_messages.json

The bug is confirmed by an adjacent pair like:

{"type":"say","say":"reasoning","partial":true,  "text":"...(identical for the first N chars)"}
{"type":"say","say":"reasoning","partial":false, "text":"...(1-2 extra trailing chars)"}

With a correct merge, only the single partial:false entry should exist.

Expected behavior

The reasoning text is rendered once. The final partial:false write should merge into (replace) the in-flight partial:true snapshot of the same message instead of appending a new entry.

Screenshots

N/A — the bug is data-observable; the ui_messages.json pair above is more precise than a screenshot. (Happy to provide one on request.)

Video

N/A — see verification steps above.

What version of zoo are you running

3.79.100401 (Windows 10, VS Code)

Additional context

Root cause (traced in the bundled dist/extension.js):

Streaming reasoning calls say("reasoning", text, undefined, /*partial*/ true) on every delta. The dedup logic in say() only inspects the last message:

let d = this.clineMessages.at(-1);
let f = d && d.partial && d.type === "say" && d.say === e;

If any other message (ask / tool / api_req, …) is appended while the reasoning stream is in flight, the final partial:false write no longer sees a same-type partial snapshot as the last message, so it takes the append branch and creates a second entry. The orphaned partial:true snapshot is never removed → both are persisted and both are rendered.

Because it depends on the interleaving of message appends, the bug is a timing race: the same version intermittently duplicates (more frequent with many tool calls, short reasoning segments, and larger stream delta intervals).

Scale of observed data: 655 such duplicate snapshot pairs across 34 historical task directories on this machine.

Distinction from existing issues:

#846 (Ollama dual-channel reasoning: message.thinking + think tags) — different mechanism (provider emits reasoning twice). Here the provider emits it once; the duplication is created entirely client-side in the message store.
#506 (OpenAI Native dual SSE handlers) — different layer (event processing), not message persistence.
Related upstream report with identical symptoms: Kilo-Org/kilocode#5294 (closed as not planned).
Verified fix (local patch, available as a PR):

In both "append new message" branches of say(), scan clineMessages backwards for an existing message with partial === true, same type/say, and prefix-related text (newText.startsWith(oldText) || oldText.startsWith(newText)); if found, merge in place (update text/images/partial flags, keep original ts) instead of appending. Patch applied to 3.79.100401: zero new duplicates since, 655 historical duplicates cleaned. The full patch code and before/after diff are in my writeup — I can submit this as a PR if that is easier than an issue here.

[extension.js.patched.zip](https://github.com/user-attachments/files/31336156/extension.js.patched.zip)

[extension.js.patched.zip](https://github.com/user-attachments/files/31336169/extension.js.patched.zip)

myk1yt:
### Related GitHub Issue

Closes: #1346

### Description

Reasoning text could render twice: `say()`'s partial-dedup only inspected `clineMessages.at(-1)`, so when any other message (an `ask`, a tool result, `api_req_started`, …) was appended while a reasoning stream was in flight, the stranded `partial: true` snapshot was no longer the last message. The final `partial: false` write then took the append branch and persisted a **second** reasoning entry — the duplicate `partial:true`/`partial:false` pair from the issue (identical but for a few trailing chars). The webview renders every stored `say:"reasoning"` message as its own block, so two stored blocks = two rendered blocks; the storage layer is the right fix layer.

Minimal fix in `say()` (following the analysis the reporter posted on the issue): both append branches now scan `clineMessages` backwards for a stranded snapshot — same type/say, `partial === true`, and prefix-related text (`new.startsWith(old) || old.startsWith(new)`) — and merge in place, keeping the original `ts`:

- `partial: true` delta → merges into the stranded snapshot (no forked second partial entry).
- `partial: false` final write → replaces the stranded snapshot in place (`partial: false`, original `ts`, `saveClineMessages()` + `updateClineMessage`).

Safety properties: the scan only runs for say-calls that pass `partial` (reasoning stream, rate-limit/retry notices, completion_result), requires the triple match (same say type + `partial === true` + prefix relation), and returns early for non-string text — unrelated message types and completed messages can never match. Behavior when no snapshot is stranded is byte-identical to before.

### Test Procedure

- `src/core/task/__tests__/Task.spec.ts` (+4 specs, new describe "say() stale partial snapshot dedup"):
  - final write merges into a snapshot stranded by an `api_req_started` interleave → single block, `partial: false`, `saveClineMessages` called
  - partial delta merges into a snapshot stranded by a tool `ask`
  - non-prefix-related text still appends (no over-merge)
  - a completed (`partial: false`) same-type message is not treated as a snapshot
- Red→green verified: the 2 merge specs fail on `main` (duplicate stored); the 2 guard specs pass on both.
- `cd src && ./node_modules/.bin/vitest run core/task/__tests__/Task.spec.ts` → 154/154 passed.
- `cd src && pnpm run check-types` → 0 errors.
- ESLint on both changed files → clean (suppression counts unchanged).

### Pre-Submission Checklist

- [x] **Visual Snapshot** (UI changes only): N/A — no UI changes (the duplicate render is fixed at the storage layer).

### Visual Snapshots

N/A.

N/A.

### Documentation Updates

### Additional Notes

Credit to the issue reporter for the root-cause analysis and patch direction.

### Get in Touch

GitHub: @myk1yt — please tag me here; I monitor notifications.


myk1yt:
@coderabbitai review — the Regression Evidence follow-ups landed in 7f4b23d25: the undefined-text finalize path (production call shape, no-throw + no-match append) and the timestamp/lastMessageTs preservation assertions in both merge tests.

myk1yt:
Persistence Integrity finding addressed at head bcd387990: the stranded-partial delta-merge branch in `Task.say` now awaits `saveClineMessages()` before the webview `updateClineMessage` post (mirroring the sibling complete-merge branches). Regression test pins the ordering with a deferred save — it fails if the awaited save is removed (157/157 Task tests; lifecycle model-check all six submodels pass).

PR #1718, opened 2026-09-20
fix(bedrock): report output truncation and expose model token limits

### LINKED ISSUE #1668: [BUG] Bedrock: responses cut at the output token limit end with an unrelated error, and the limit cannot be raised
**Describe the bug**
When a Bedrock response reaches the output token limit, AWS returns `stopReason: "max_tokens"`, but the Bedrock provider ignores it. The response just stops (often mid-thinking, before any tool call), and the task then reports an unrelated error such as "You did not use a tool…" instead of saying the output limit was reached.

This is more likely on adaptive-thinking models (Claude 4.7+), where Zoo Code sends `effort: "xhigh"` without a thinking budget, so thinking can use most of the limit.

The user also cannot work around it: with reasoning enabled, the "Max Output Tokens" slider stops at 16,384 on Bedrock, while the direct Anthropic provider allows up to 64k/128k for the same models.

**To Reproduce**
1. Select the Amazon Bedrock provider and `anthropic.claude-opus-4-8`, with reasoning enabled.
2. Open the "Max Output Tokens" slider: its maximum is 16,384.
3. Give a complex task (e.g. a large refactor or a long file to write).
4. If thinking and output exceed the limit, the response stops mid-way and the task reports a generic error.

**Expected behavior**
- When a response hits the output limit, the user gets a clear message (e.g. "Output token limit reached, consider increasing Max Output Tokens").
- The slider can go up to the model's documented limit, as with the Anthropic provider. The 16,384 default can stay: a high `max_tokens` reserves Bedrock quota up front and reduces the context available for the conversation.

**Screenshots**

**Video**

3.82.1

**Additional context**
- `messageStop` events are skipped without reading `stopReason` in the Bedrock stream loop (`src/api/providers/bedrock.ts`).
- In `packages/types/src/providers/bedrock.ts`, `maxTokens` is `8192` for Claude Sonnet 4.5/4.6/5 and Opus 4.7/4.8/5, versus `64_000`/`128_000` in `packages/types/src/providers/anthropic.ts`. With reasoning enabled, the slider maximum is `max(model maxTokens, 16_384)`, hence 16,384 on Bedrock. AWS documents 64K max output tokens for Claude Sonnet 4.5: https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-sonnet-4-5.html
- Trade-offs of a higher limit: Bedrock deducts `input + max_tokens` from the tokens-per-minute quota at request start (https://docs.aws.amazon.com/bedrock/latest/userguide/quotas-token-burndown.html), and Zoo Code reserves `max_tokens` from the context window (`src/core/context-management/index.ts`).

PierrunoYT:
## Summary
- Report an explicit output-token-limit error for Bedrock `stopReason: max_tokens` instead of treating a truncated response as a normal completion. Consume trailing usage metadata before reporting the error; use the existing stream-error/retry path.
- Correct documented output ceilings to 64K for Sonnet 4.5/4.6 and 128K for Sonnet 5 and Opus 4.7/4.8/5, preserving the existing 16,384 reasoning-enabled default.
- Expose the max-output control for Sonnet 5 and Opus 5 alongside binary reasoning, and honor its explicit setting with reasoning disabled.
- Add provider, settings, and visual regression coverage, including a baseline generated in the pinned Playwright container.

Fixes #1668.

## Validation
- Provider/shared Vitest suites: 205 passed.
- Settings Vitest suite: 41 passed.
- Pinned-container visual regression: 1 passed after baseline generation.
- Repository lint and type checks passed, including commit/push hooks.
- Browser interaction confirmed the Opus 5 slider starts at 16,384 and reaches 128,000; inspected default and maximum states and the Opus 4.8 controls.
- No live AWS request was made.

## Sources
AWS Bedrock model cards document the limits:
- [Sonnet 4.5](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-sonnet-4-5.html)
- [Sonnet 4.6](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-sonnet-4-6.html)
- [Sonnet 5](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-sonnet-5.html)
- [Opus 4.7](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-7.html)
- [Opus 4.8](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-8.html)
- [Opus 5](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-5.html)


src/api/providers/bedrock.ts
```diff
@@ -487,6 +488,9 @@ export class AwsBedrockHandler extends BaseProvider implements SingleCompletionH
 				modelId: modelConfig.id,
 				thinking: additionalModelRequestFields?.thinking,
 			})
+		} else if (isAdaptiveThinkingModel && modelConfig.info.supportsReasoningBinary) {
+			// Omitting thinking can enable it by default (e.g. Sonnet 5).
+			additionalModelRequestFields = { thinking: { type: "disabled" } }
```
edelauna: isAdaptiveThinkingModel also matches the Fable 5/5.1 ids (both set supportsReasoningBinary), but AWS documents that Fable is adaptive-only and returns a 400 for thinking.type: "disabled". With reasoning off (the default) every request to those models would now fail. Could this gate on a flag only Sonnet 5/Opus 5 set, so Fable keeps the omit behavior?

src/api/providers/bedrock.ts
```diff
@@ -784,9 +789,16 @@ export class AwsBedrockHandler extends BaseProvider implements SingleCompletionH
...
+					outputLimitReached = streamEvent.messageStop.stopReason === "max_tokens"
 					continue
 				}
 			}
+			// Bedrock sends usage metadata after messageStop. Preserve it before reporting truncation.
+			if (outputLimitReached) {
+				throw new Error(
+					"Output token limit reached. Consider increasing Max Output Tokens in the model settings.",
+				)
+			}
```
edelauna: This deterministic error now flows into the mid-stream retry in recursivelyMakeRooRequests, which re-queues identical content with no cap. Under auto-approval a task whose output exceeds maxTokens retries indefinitely, burning input and output tokens on every pass. Should this be typed as non-retryable so it surfaces once, or should the retry raise inferenceConfig.maxTokens?

edelauna (changes requested):

PR #1719, opened 2026-09-20
fix(bedrock): discover regional models and clarify inference routing

### LINKED ISSUE #1667: [BUG] Bedrock: model list ignores the selected region, and "cross-region" / "global" inference checkboxes are misleading
**Describe the bug**
The Bedrock model list is a static, hard-coded list, identical for every region and account. It does not show whether a model is available in the selected region, or whether it can only be used through an inference profile. It also misses models available on Bedrock today (e.g. Llama 4, Mistral Large 3, Nova Premier, Cohere Command R).

On top of that, the two Bedrock routing checkboxes have no description, and their labels suggest the opposite of what they do:

- **"Use cross-region inference"** sounds like requests may leave your area, so privacy-minded users (e.g. EU users under GDPR) leave it off. In reality it uses a geographic inference profile (`eu.`, `us.`, `jp.`…) that keeps processing within that geography. It is also required for recent models: Claude Sonnet 4.5, for example, supports no in-region inference at all, so with the box unchecked every request fails.
- **"Use Global inference (auto-select optimal AWS Region)"** sounds like a harmless performance option. In reality it routes requests to any commercial AWS region worldwide, and when checked it silently takes precedence over the cross-region option.

Nothing in the UI shows the model ID actually sent (`anthropic.…`, `eu.anthropic.…` or `global.anthropic.…`), and the error does not point to these settings.

**To Reproduce**
1. Select the Amazon Bedrock provider, region `eu-west-3`, with valid credentials.
2. Select `anthropic.claude-sonnet-4-5-20250929-v1:0`, leave both "Use cross-region inference" and "Use Global inference" unchecked.
3. Send any message.
4. The request is rejected by AWS, since this model has no in-region support (see the AWS model card below).

**Expected behavior**
- The model list reflects what is available in the selected region, and indicates when a model requires an inference profile.
- Each checkbox has a short description of where requests are processed (geography vs worldwide) and when it is required.
- The error for a model that needs an inference profile suggests enabling cross-region inference.

**Screenshots**

**Video**

3.82.1

**Additional context**
- AWS documentation for Claude Sonnet 4.5 lists no in-region support in any region, only Geo (`us.`, `eu.`, `au.`, `jp.`) and Global inference IDs: https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-sonnet-4-5.html
- AWS recommends Geographic cross-region inference for data residency requirements, and describes Global as routed worldwide: https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference.html
- The model list comes from the static `bedrockModels` table (`packages/types/src/providers/bedrock.ts`), with no region filtering. Models currently offered by Bedrock: https://docs.aws.amazon.com/bedrock/latest/userguide/model-cards.html
- The prefix is added at request time in `getModel()` (`src/api/providers/bedrock.ts`); Global is applied first, cross-region only otherwise.
- Related: #124 / #310, which propose discovering available models and inference profiles through the AWS API (no progress since May).

PierrunoYT:
## Summary
- Add an explicit regional AWS catalogue refresh using ListFoundationModels and paginated ListInferenceProfiles with the current buffered IAM credentials/profile. Entries distinguish foundation models, geographic profiles, global profiles, and application profiles; selecting one fills the custom ARN without saving settings.
- Discard stale catalogue responses when the region or credentials change. Keep the bundled model picker as an explicitly unverified fallback, and hide catalogue selection under restrictive organization model allowlists.
- Explain geographic versus worldwide processing, global precedence, and ARN routing. Preview the effective request model ID using shared routing logic; no routing option is enabled automatically.
- Recognize common on-demand/inference-profile error wording, including straight and curly apostrophes.
- Preserve unfamiliar foundation-model IDs rather than silently invoking the default Claude model, and avoid inheriting Claude reasoning capabilities for those models.

Fixes #1667.

## Verification
- Provider, discovery, routing, and webview-message tests: 148 passed.
- Bedrock settings/catalogue and ApiOptions interaction tests: 77 passed.
- Model-message type tests: 5 passed.
- Pinned Playwright container visual comparisons: 2 passed; container-generated baselines included.
- Browser checks exercised geographic/global ID previews and catalogue ARN selection using mocked discovery responses. Captures were inspected for clipping and readability.
- Repository lint and type checks passed, including commit/push hooks. ESLint suppression count decreased.

## Scope and limitations
Discovery is user-triggered and uses the regional AWS control plane, not a custom runtime endpoint. This implementation requires IAM credentials/profile plus bedrock:ListFoundationModels and bedrock:ListInferenceProfiles; it does not use Bedrock API-key authentication for discovery. Region/credential edits clear the results and require another refresh.

AWS catalogue presence is not proof of invocation permissions, tool support, or direct in-region invocation. Foundation-model entries explicitly warn to check inference-profile requirements; application-profile destinations must be checked in AWS. Unknown model capabilities remain estimated and configurable via the existing custom-ARN settings. The bundled list remains available when discovery is unavailable.

No live AWS discovery or inference request was made.

AWS references: [ListFoundationModels](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_ListFoundationModels.html), [ListInferenceProfiles](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_ListInferenceProfiles.html), [routing geography](https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference.html).


LouisClt:
Reporter of #1667. This covers the report: explicit refresh instead of an automatic call, the
ON_DEMAND filter, the profile kinds, the effective request model ID, and labels that finally
say where processing happens.

Three points:

1. **Discovery failures are not diagnosable.** The bare `catch` in the `requestBedrockModels`
   handler replaces IAM denial, network failure and wrong region with one fixed message, and
   logs nothing. Keeping the original error in the logs would be enough.

2. **Raw `<select>`.** The surrounding Bedrock settings (region, service tier) use the shared
   `Select` component, so the new dropdown does not match the controls directly above it.

3. **`regional` label.** "Foundation model — check inference-profile requirements" contradicts
   `byInferenceType: "ON_DEMAND"`, which already guarantees those entries need no profile.
   Stating that directly would remove the doubt this PR is trying to remove.

Merge order: this PR and #1718 touch `src/api/providers/bedrock.ts`, its spec and
`webview-ui/playwright/gallery/stories.tsx`, and a trial merge of the two branches conflicts.
Whichever lands second needs a rebase.

